PR #2, opened 2020-05-07
components: Move image creation to PF4 wizard

Gundersanne:
Mostly placeholders still.

image

image
image
image
image

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -4,6 +4,7 @@ import React from 'react';
 import asyncComponent from './Utilities/asyncComponent';
 
 const LandingPage = asyncComponent(() => import('./SmartComponents/LandingPage/LandingPage'));
+const CreateImageWizard = asyncComponent(() => import('./SmartComponents/CreateImageWizard/CreateImageWizard'));
 
 const InsightsRoute = ({ component: Component, rootClass, ...rest }) => {
     const root = document.getElementById('root');
@@ -22,6 +23,7 @@ export const Routes = () => {
     return (
         <Switch>
             <InsightsRoute exact path='/landing' component={ LandingPage } rootClass='landingpage'/>
+            <InsightsRoute exact path='/imagewizard' component={ CreateImageWizard } rootClass='imagewizard'/>
             <Redirect to='/landing'/>
         </Switch>
     );
--- a/src/SmartComponents/CreateImageWizard/CreateImageWizard.js
+++ b/src/SmartComponents/CreateImageWizard/CreateImageWizard.js
@@ -1,0 +1,489 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { withRouter } from 'react-router-dom';
+import { connect } from 'react-redux';
+import { actions } from '../redux';
+
+import {
+    Alert,
+    Flex,
+    FlexItem,
+    FlexModifiers,
+    Form,
+    FormGroup,
+    FormSelect,
+    FormSelectOption,
+    Radio,
+    TextContent,
+    TextInput,
+    Title,
+    Wizard,
+} from '@patternfly/react-core';
+
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
+
+import api from './../../api.js';
+
+const ReleaseComponent = (props) => {
+    const options = [
+        { value: 'rhel-8', label: 'Red Hat Enterprise Linux (RHEL) 8.2' },
+    ];
+    return (
+        <Form isHorizontal>
+            <FormGroup label="Release" fieldId="release-select">
+                <FormSelect value={ props.value } onChange={ value => props.setRelease(value) }
+                    aria-label="Select release input" id="release-select">
+                    { options.map(option => <FormSelectOption key={ option.value } value={ option.value } label={ option.label }/>) }
+                </FormSelect>
+            </FormGroup>
+        </Form>
+    );
+};
+
+ReleaseComponent.propTypes = {
+    setRelease: PropTypes.func,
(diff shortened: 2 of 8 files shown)


jgiardino:
Love seeing the progress on this!! 🎉 🎉 🎉 
Here are some things I noticed. I don't see any of these as blockers, but just some things to think about. Let me know if you want to chat about any of this.

cc'ing @andreasn for his awareness on this too. :-)

- Upload step - The label of this seems off to me. I think part of it is that in the flow of the prototype, the user chooses what kind of image to create before being asked for upload details. And the type of image is a really important decision in this flow, and it seems odd that it isn't highlighted as a primary step. But with the combination of options you currently have, it looks like the image type selection and upload details are combined into one step. The best suggestion I have for this combination is something like "Image type and destination" or "Target environment". 🤷 
- Upload step - The sequence of inputs also seems off, and some inputs for the AWS case are missing (at least compared to what's included in cockpit-composer). 
  - Regarding sequence, I conceptually expect anything regarding authentication to come first, before asking for other details like destination, because in an ideal scenario, we'd check that their authentication info is good before letting them go any further. That is, the sooner we can detect a problem and notify the user, the less frustrated they will be with our service. 
  - As far as inputs that are missing, we also ask for the S3 bucket and an image name in cockpit-composer. BUT we also discussed not capturing a value from the user for the image name initially, and instead possibly basing the image name off of the source and possibly include a timestamp. 
- Review step - This is where the user can see all their inputs in one screen to review them one more time before clicking submit. The following are specific things related to this:
  - The step label in the nav vs the review page should match if possible. Currently, the nav is "Upload" and the Review step is "Image upload". 
  - The upload step should include all the inputs that are listed in the form.
  - The Registration step should definitely include activation key. I'm inclined to also include org id, even though it's read-only, because why not. 🤷

jgiardino:
There was some discussion in the team chat about Insights, and it made me think of a few more comments, primarily around how to handle required vs non-required fields and the Registration step.


**required vs non-required fields**
In the prototype for the wizard, I'm using the red asterisk for any field that is required. 

Then on the Review step, if there are required inputs that are missing a value, I do the following (there are examples of these in the prototype that you can reference, except where noted):
- Display an error message at the top. The prototype is using a message that is too specific, but recently in cockpit-composer we implemented this more [generic message](https://github.com/osbuild/cockpit-composer/blob/master/components/Modal/CreateImageUpload.js#L51), so that it can display for any field that requires user attention, not just missing values.
- Display an error message where the user input would display if it had a value. 
  image
- Disable the final submit button if required inputs are missing.

**the Registration step**
Providing an activation key is optional, in that the user can choose to register the system on first boot. But along with activation key, there are other things that are relevant to ask the user for, if the user is choosing to register the system. One is connecting to Insights, and the other is specifying System Purpose. We will definitely want to include Insights. And my understanding of System Purpose is that these values are used along with the activation key in determining which subscription to use for the system. But I don't know how critical System Purpose is for our initial release (cc'ing @andreasn again to weigh in on that one 😺).

Anyway, assuming we want to at least include Insights along with Activation keys, here are examples of how that is handled in the prototype.

1. The first option presented to the user asks if they want to register the system. These are two radio buttons, which allow us to really explain what the alternative options are. These are also required and blank to force the user to consider this option and make a deliberate choice. 
    image
2. If the user chooses to not register the system, then they don't see any additional inputs, like activation key or enabling insights. But if the user chooses to register the system, then additional inputs like activation key and enabling insights display. 
    - Insights should be enabled by default
    - Listing activation keys is a future goal, and for now what you have (i.e. input for key and org id) is good. Here's an example of a [wireframe](https://docs.google.com/presentation/d/1j3vOFW45MW8M_Ec9E8X3Qzs2wp3qHglTOS5mW1GAOn8/edit?usp=sharing) that more closely aligns with that. The order of org ID and activation key are different between your example and this wireframe, but I don't have a strong opinion on that. Maybe @andreasn has some useful insights there?
    image

andreasn:
> But I don't know how critical System Purpose is for our initial release

It will allow the system to make a better guess at what the right subscription is to subscribe to, but I think it's fine to wait until later. The systems will just do a little worse guesses, but they will still make a guess.

> The order of org ID and activation key are different between your example and this wireframe, but I don't have a strong opinion on that.

The wireframe didn't have anything for org ID's if I recall correctly, because the assumption is that (at least initially) customers will only ever have one org ID each.
If I recall correctly, you can only ever have one organization on the Red Hat CDN, and multiple organizations is only a think if you run Satellite. Will need to verify this, but I would say, don't worry too much about it for now.

jgiardino:
@andreasn Thanks for providing feedback! There was one [wireframe (slide 5)](https://docs.google.com/presentation/d/1j3vOFW45MW8M_Ec9E8X3Qzs2wp3qHglTOS5mW1GAOn8/edit#slide=id.g71f80fa48e_3_120) that's closest to what @Gundersanne shows now for specifying activation key, which includes the org ID. I don't remember if it was included in that example because we were questioning whether the user should be able to specify a different org ID, or just because we decided that it's important to show? Also, I recently updated that wireframe to have System Purpose come after activation key, which we had decided was the preferred order.

andreasn:
Ah right. If I recall correctly, that was just a random order. Looking at the[ customer portal documentation](https://access.redhat.com/solutions/3341191) on this topic, it seems org is set before activation key name, so I feel it's probably good to keep that order in this UI as well.
`subscription-manager register --org=ORG ID --activationkey= Key Name`

Gundersanne:
I'm going to merge this as is. It works and will gives us a baseline to work from.

PR #11, opened 2020-09-22
On travis run with NODE_ENV=production

Gundersanne:


major (approved):
I can verify that this change builds the `/beta` routes as it should. Thank you! 🥳

PR #12, opened 2020-09-22
📦 Denote AWS image as raw

major:
We had issues in composer a while back with VHDX in AWS and we have been
using raw images since then.

Signed-off-by: Major Hayden <major@redhat.com>


Gundersanne (approved):

teg:
Could we just drop the file ending altogether? We are not returning the images to the user, they should not care.

Gundersanne:
@teg fine by me too. The original text came from the design slides.

PR #25, opened 2020-10-29
🚀 Deploy to stage-beta and stage-stable

major:
Ensure we deploy to stage-beta (when master is updated) and stage-stable (when master-stable is updated).


teg (approved):
This is the change from how cloudDot used to do it to AppSRE, right?

major:
> This is the change from how cloudDot used to do it to AppSRE, right?

Indeed!

PR #27, opened 2020-10-30
npm: Use npm-update to keep all packages updated

Gundersanne:
Rather than freezing the packages and updating them manually, use
npm-update to have the latest available version.

---

This is similar to what cockpit and cockpit-composer does.


jkozol (approved):
It makes sense to me to use cockpit's bots to keep our packages up to date.

Gundersanne:
Ok i'm re-enabling the workflow. Let's hope it works now

PR #33, opened 2020-11-02
Landing page layout

### LINKED ISSUE #17: Update landing page layout
Update Landing Page layout to match the layout that's shown in the [design document](https://docs.google.com/presentation/d/1YUPsj01oafTfnTqlihf6-stGbuINfQSIN9K9a24-byY/edit?usp=sharing). This includes:
- Removing the extra spacing around the page title
- Adding the basic container for the toolbar, to include only the Create image button at first
- Adding a table to include the details currently provided to the UI for listing images

jgiardino:
This update starts to shift the layout to be closer to the layout used by apps in clouddot, as shown in the example below. Note that I'm using mock data in the following example of the list of images, which might not be an exact match to the data that the API would provide.

image

This PR closes #17


src/SmartComponents/ImagesTable/ImagesTable.js
```diff
@@ -0,0 +1,109 @@
...
+                    <Link to="/imagewizard" className="pf-c-button pf-m-primary" data-testid="create-image-action">
+                        Create image
+                    </Link>
+                </TableToolbar>
+                <Table
+                    aria-label="Images"
+                    rows={ rows }
+                    cells={ this.state.columns }
+                    data-testid="images-table"
+                >
```
Gundersanne, line 83: nitpick: the closing tag on a newline looks weird >.>

src/test/SmartComponents/LandingPage/LandingPage.test.js
```diff
@@ -7,18 +7,14 @@ describe('Landing Page', () => {
...
         // check heading
-        screen.getByRole('heading', { name: /image builder/i });
+        screen.getByRole('heading', { name: /images/i });
     });
-    test('renders CreateImageCard child component', () => {
-        // check CreateImageCard loads
-        screen.getByTestId('create-image-card');
-        // and has correct contents
-        screen.getByText(/create a new image/i);
-        screen.getByRole('button', { name: /create image/i });
```
Gundersanne, line 19: Why was the test that checks the button removed?

Gundersanne (commented):
Thank you! Mostly some questions and nitpicks. I'm happy to move the main tag to the landingpage in a followup.

src/SmartComponents/ImagesTable/ImagesTable.js
jgiardino: I don't have an opinion, because I think it looks weird either way. 🙃

src/test/SmartComponents/LandingPage/LandingPage.test.js
```diff
@@ -7,18 +7,14 @@ describe('Landing Page', () => {
...
-    });
-    test('renders ImagesCard child component', () => {
-        screen.getByText(/pending composes/i);
+    test('renders ImagesTable child component', () => {
+        // check action loads
+        screen.getByTestId('create-image-action');
```
jgiardino, line 16: > Why was the test that checks the button removed?

I moved this check to this line. I really like the approach to use `.getByTestId` instead of other things, like `ByRole` or `ByText`, so I switched to that method here. One thing I like is that this uses `data-testid` on the element it's looking for, so it's obvious when working on UI updates which items are checked with testing, and it's easier to find where the item is defined in the tests.

src/App.js
```diff
@@ -30,7 +28,7 @@ class App extends Component {
 
     render () {
         return (
-            <Main>
+            <Main style={ { marginLeft: 0, padding: 0 } }>
```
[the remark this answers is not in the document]
jgiardino: In the main conversation thread you mentioned this:
> I'm happy to move the main tag to the landingpage in a followup.

@jkozol was helping me with this earlier this week. We realized that when `<Main>` is not included here and when it's included in LandingPage.js, the tests fail. So he raised a question about the `<Main>` component in a google chat (cc'ing us), about what it's purpose is. 

My workaround to keep the tests working were to keep `<Main>` here in App.js, but apply the same styles that insights apps use (as [shown here](https://github.com/RedHatInsights/insights-rbac-ui/blob/d829ae500e9ee477421899df674cea234fddaf52/src/App.js#L53)) to remove the extra padding. But the html rendered by `<Main>`—i.e. `<section className="pf-l-page__main-section pf-c-page__main-section">`—is still needed on LandingPage.js, so I include just that html instead of `<Main>` in that file.

I didn't fully understand the answer that @jkozol got back when he raised the question, so I would defer to you and him on what we ultimately want to do. But if the goal is to be able to use `<Main>` in LandingPage.js instead of here, then would we still want to move the css out to a separate file? I think it makes sense to move it if this is our long-term solution, but probably easier to rip this out if it's all in one chunk like it is now.

src/SmartComponents/ImagesTable/ImagesTable.js
jgiardino: Also, should this be something that's flagged with an eslint rule? I would have expected this to be caught, but maybe we're using defaults from the starter app?

jgiardino:
Thanks for the review, @Gundersanne! I answered some of your questions with more questions. ;-) Let me know what you suggest, and I'll push some updates.

src/SmartComponents/ImagesTable/ImagesTable.js
Gundersanne: yea I'll add this, it's `react/jsx-tag-spacing`

src/test/SmartComponents/LandingPage/LandingPage.test.js
Gundersanne: Oh right, yea that's fine then :)

src/App.js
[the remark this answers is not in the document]
Gundersanne: Alright, let's keep this as is then, and figure it out later.

Gundersanne (approved):
Thanks! Let's merge this. 

I'll enable the linting rule and push with `--fix`.

And for the main thing let's figure that out later, if other projects are doing inline style for just that element, then I think it's fine if we do that too for now.

PR #38, opened 2020-11-03
Initial test for CreateImageWizard

atodorov:
very basic, make sure we can render all of the expected text


Gundersanne (approved):
Nice! Thank you.

PR #43, opened 2020-11-05
Add more tests

atodorov:


src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1,26 +1,173 @@
...
+    test('clicking Next loads Target environment', () => {
+        const [ next, , ] = verifyButtons();
+        next.click();
+
+        screen.getByText('Destination');
+        screen.getByText('Secret access key');
+    });
+
+    // todo: Should we have the Back button here ?
+    // Maybe not b/c it doesn't lead anywhere
```
atodorov, line 68: @Gundersanne please comment on this

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1,26 +1,173 @@
...
+
+        screen.getByText('Register the system');
+    });
+
+    test('clicking Back loads Release', () => {
+        const [ , back, ] = verifyButtons();
+        back.click();
+
+        // note: there doesn't seem to be anything better on this screen
+        screen.getByText('Red Hat Enterprise Linux (RHEL) 8.2');
```
atodorov, line 100: Any idea if I can do better here ?

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1,26 +1,173 @@
...
+        anchor.click();
+    });
+
+    test('has 3 buttons', () => {
+        screen.getByRole('button', { name: /Create/ });
         screen.getByRole('button', { name: /Back/ });
         screen.getByRole('button', { name: /Cancel/ });
     });
+
+    // todo: add test for the Create button
```
atodorov, line 160: What is the Create button supposed to do? Maybe I have to mock the API to intercept the result from this action ?

atodorov (commented):
@Gundersanne this PR is mostly done, I need couple of answers before completing the work on clicking through the buttons on the pages.

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
Gundersanne: The back button should work? When I try it out on ci.clouddot it goes to the previous step.

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
Gundersanne: Hm is it possible to somehow get it via the `value` attribute? Alternatively I'd also be ok if it were just a check to make sure the `Select` is there.

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
Gundersanne: Yea, the create button sends the request and updates the store.

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
atodorov: This is the very first step - "Release" and the Back button here is disabled and IMO it should not even be present.

Gundersanne: Ah yea we mostly follow the designs here put forth by the UX team

image

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
atodorov: I've decided to leave this for my next PR b/c ATM I have no idea how to do it and b/c it is different from everything else that is going on in this test for now.

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
atodorov: I've added data-testid attribute and used that b/c the react testing library doesn't allow selecting elements by their ID. FTR they have philosophical opinion as to why that is which just makes our job a bit more miserable.

Gundersanne (approved):

PR #48, opened 2020-11-10
test: Make sure the status of a compose is called status

Gundersanne:
This was a bit confusing because in the inline documentation I wrote
used state as a key. However this changed to status pretty early on and
I forgot to update it.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -49,7 +49,7 @@ rules:
     - error
     - 4
     - SwitchCase: 1
-      MemberExpression: 0
+      MemberExpression: 1
       ImportDeclaration: 1
       ObjectExpression: 1
   react/jsx-curly-spacing:
--- a/src/test/SmartComponents/redux/reducers.test.js
+++ b/src/test/SmartComponents/redux/reducers.test.js
@@ -2,8 +2,8 @@ import { composeReducer } from '../../../SmartComponents/redux/reducers';
 import types from '../../../SmartComponents/redux/types';
 
 const compose = {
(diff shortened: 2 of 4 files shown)


src/test/SmartComponents/redux/reducers.test.js
```diff
@@ -3,7 +3,7 @@ import types from '../../../SmartComponents/redux/types';
 
 const compose = {
     'xxxx-xxxx-xxxx-xxxx': {
-        state: 'uploading',
+        status: 'uploading',
```
atodorov, line 6: See also `redux/actions.test.js`

atodorov (changes requested):
Overall looks good, 1 more file needs patching.

src/test/SmartComponents/redux/reducers.test.js
Gundersanne: Nice,  I also exchanged the key with a dummy uuid.

.eslintrc.yml
```diff
@@ -49,7 +49,7 @@ rules:
     - error
     - 4
     - SwitchCase: 1
-      MemberExpression: 0
+      MemberExpression: 1
```
Gundersanne, line 52: Multi-line statements would get indented on the same level with MemberExpression 0, I think have an indent makes it more readable.

atodorov: I was bashing my head against the wall yesterday with exactly the same problem. Thanks for figuring it out.

atodorov (approved):

PR #49, opened 2020-11-10
Update travis config file to rebase on master before test running

henrywang:


atodorov:
Why do we need this change ? 

From what I can see this is equivalent of rebasing before pushing an open PR so the only thing it saves is typing `git rebase`. In the case there are conflicts rebase will fail which will still require the author of a PR to do a manual rebase and fix the conflicts.

henrywang:
@atodorov, two reasons:
1. As you said, travis test will fail if there's conflict.
2. It saves typing `git rebase` not only for conflict checking but for things like code coverage, PR test fail due to source code change from latest PR.

Gundersanne (approved):
Good idea, thanks!

PR #50, opened 2020-11-11
Revert "Update travis config file to rebase on master before test running"

Gundersanne:
This reverts commit bd99a90faa8188eb8c9ada6b68c403fff6a800c2.

While this is common in cockpit's projects, osbuild projects refrain
from rebasing branches.

Rebasing on master as it happens to be when the test is running, leaves
you in a state where you neither know that the branch itself is working,
nor that it will work when you actually do merge it later, because
master might have changed in between.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,6 @@ node_js:
 cache:
   directories:
   - node_modules
-before_install:
-  - |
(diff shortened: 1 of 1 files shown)


larskarlitski (approved):
Thanks a lot!

atodorov (approved):

PR #55, opened 2020-11-19
More tests for CreateImageWizard

atodorov:


src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -338,5 +337,56 @@ describe('Click through all steps', () => {
...
+        // API request sent to backend
+        expect(composeImage).toHaveBeenCalled();
+
+        // returns back to the landing page
+        // await waitFor(
+        //    () => [
+        //        screen.getByText('Create a new image'),
+        //        screen.getByRole('button', { name: /Create image/ }),
+        //    ]
+        // );
```
atodorov, line 357: @Gundersanne in the source code I see `426: this.props.history.push('/landing');` but if we uncomment the wait above it fails, see https://travis-ci.com/github/osbuild/image-builder-frontend/builds/202712706. 

Also clicking the Create button in ci.cloud doesn't send me back to the landing page so I think it may be broken.

Gundersanne: Well the request needs to succeed before it sends you to the landing page, so if there's no api that's returning `201` with `{"id": uuid}` in  the response this is expected.

atodorov: Yup, that makes sense, but isn't the mock above returning what we need already ? From the coverage change report I can see execution traverses the lines after the mocked API returns successfully and the redirect on 426 seem to have been executed.

src/test/SmartComponents/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -338,5 +337,57 @@ describe('Click through all steps', () => {
...
+        create.click();
+
+        // API request sent to backend
+        expect(composeImage).toHaveBeenCalled();
+
+        // todo: assertion fails, can't find the button and we don't know why !
+        // returns back to the landing page
+        // await waitFor(
+        //    () => [
+        //        screen.getByText('Create a new image'),
```
Gundersanne, line 355: Hm I think this should be replaced here (and in verifyCancelButton) with something that is exclusive to the landing page, as the text "Create a new image" occurs on both the landing page and the wizard.

How about `screen.getByTestId('create-image-action')`, `images-table` as testId. Feel free to add a new one to on the landing page as well.

Gundersanne (approved):
Thanks!

PR #58, opened 2020-11-25
package.json: Update major versions of react and patternfly

Gundersanne:


atodorov (approved):
LGTM, tests don't crash.

@Gundersanne btw you may want to checkout Dependabot which can do these kinds of  updates automatically and open pull requests by itself.

PR #59, opened 2020-11-30
WIP: Experiment with cypress & cypress-react-select

atodorov:


atodorov:
We've discussed this and agreed not to pursue these types of tests for now.

PR #60, opened 2020-12-01
CreateImageWizard: Only ask for aws account id in upload step

Gundersanne:
image
---

The drop in coverage here is due to the removal of code in the wizard.

### DIFF
--- a/src/PresentationalComponents/CreateImageWizard/WizardStepUploadAWS.js
+++ b/src/PresentationalComponents/CreateImageWizard/WizardStepUploadAWS.js
@@ -1,55 +1,26 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { Form, FormGroup, FormSelect, TextInput, FormSelectOption, Title } from '@patternfly/react-core';
+import { Form, FormGroup, TextInput, Title } from '@patternfly/react-core';
 
 const WizardStepUploadAWS = (props) => {
-    const serviceOptions = [
-        { value: 'ec2', label: 'Amazon Elastic Compute Cloud (ec2)' },
-        { value: 's3', label: 'Amazon Simple Storage Service (s3)' },
-    ];
-
     return (
         <Form>
             <Title headingLevel="h2" size="xl">Upload to AWS</Title>
-            <FormGroup isRequired label="Access key ID" fieldId="amazon-access-id"
-                helperTextInvalid={ (props.errors['amazon-access-id'] && props.errors['amazon-access-id'].value) || '' }
+            <p>
+            Your image will be uploaded to a temporary account on Amazon Web Services.
+            The image will be shared with the account ID you provide below. <br />
+            Within the next 14 days you will need to copy the shared image to your own account.
+            After 14 days it will be unavailable and will have to be regenerated.
+            </p>
+            <FormGroup isRequired label="AWS account ID" fieldId="aws-account-id"
+                helperTextInvalid={ (props.errors['aws-account-id'] && props.errors['aws-account-id'].value) || '' }
                 validated={ (props.errors['amazon-access-id'] && 'error') || 'default' }>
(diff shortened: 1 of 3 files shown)


Gundersanne:
@jgiardino What do you think of the above info-alert text. I add the part about sharing because this has now changed.

jgiardino:
Good question! For reference, this is [an example](https://docs.google.com/presentation/d/1YUPsj01oafTfnTqlihf6-stGbuINfQSIN9K9a24-byY/edit#slide=id.g8fc6c996ad_1_43) for a similar workflow that I created a while back.

For this, I'd like @katierik and @andreasn to weigh in on whether the info alert is the right way to show this info, or some other method.

Also when chatting with @katierik and @andreasn, we were wondering if it makes sense to move the account ID to the first step, as a short term solution, since there isn't that much going on in this step right now. Not saying it has to be part of this PR, but just wanted to mention that.

src/PresentationalComponents/CreateImageWizard/WizardStepUploadAWS.js
```diff
@@ -1,55 +1,22 @@
...
             </FormGroup>
-            { props.upload.options.service === 's3' &&
-              <FormGroup isRequired label="Bucket" fieldId="amazon-bucket"
-                  helperTextInvalid={ (props.errors['amazon-bucket'] && props.errors['amazon-bucket'].value) || '' }
-                  validated={ (props.errors['amazon-bucket'] && 'error') || 'default' }>
-                  <TextInput value={ props.upload.options.bucket || '' } type="text" aria-label="amazon bucket" id="amazon-bucket"
-                      data-testid="aws-bucket" isRequired
-                      onChange={ value => props.setUploadOptions(Object.assign(props.upload.options, { bucket: value })) } />
-              </FormGroup> }
+            <Alert variant="info" isInline title="The image wil be uploaded to Amazon Web Services.
```
andreasn, line 18: typo: "wil" -> "will"

andreasn:
The [Patternfly guidelines](https://www.patternfly.org/v4/components/alert/design-guidelines) says _An alert is used to notify the user about a change in status or other event without blocking other actions in an interface. These guidelines talk about how to properly use different types of alerts in your designs._

This, together with it mostly being used in Cloud.rh for messages like these, so people get trained from other parts of the UI that something unusual is going on if they see these alerts:
image

So I think maybe a notification is not the correct element here.
Probably a simple text label without any other styling could work. And I think it's better if it comes before the input. So you know that what you put in the input will affect where it ends up.

andreasn:
This is a pretty good example of a explanatory text from another part of cloud.rh
image

Gundersanne:
hm @andreasn i used the designs form the 'IaaS, initial release'  slide. There's it's an "informational message". Anyway I put it on the agenda for tomorrow.

katierik:
After discussing:

* Change from an alert to instructional text. Text should be medium, 100 color; and placed above the field (so the order is Header, Text, Field) 
* Text update: “Your image will be uploaded to a temporary account on Amazon Web Services. The image will be shared with the account ID you provide below. | [Line break] | Within the next 14 days you will need to copy the shared image to your own account. After 14 days it will be unavailable and will have to be regenerated.” (Replace "14 days" with whatever the actual time is)

andreasn (approved):
Looks great now!

jgiardino:
This looks good to me! 

Just sharing some thoughts about this for future updates... It would be good to consider how the user can be informed about the expiration outside the creation flow. There is a [design for this](https://docs.google.com/presentation/d/1YUPsj01oafTfnTqlihf6-stGbuINfQSIN9K9a24-byY/edit#slide=id.g9551495f59_6_0) for the table view, but currently it's coupled with the link that takes you to AWS to access the image. If that link isn't doable in the foreseeable future, then it would be good to provide some other way of communicating how to find the image in AWS and along with when it expires.

PR #62, opened 2020-12-07
Wizard refactor

jgiardino:
Refactors wizard steps to more closely align to the [design document](https://docs.google.com/presentation/d/1YUPsj01oafTfnTqlihf6-stGbuINfQSIN9K9a24-byY/edit?usp=sharing), while also adopting some of the clouddot guidelines (i.e. adds a header to each step).

Also updates how js files are organized to align with recent refactoring made in [cockpit-composer](https://github.com/osbuild/cockpit-composer/tree/master/components/Wizard), while also adopting the SmartComponents and PresentationalComponents organization in the clouddot sample app. 

@Gundersanne I see that you made updates to the AWS form elements in #60. It looks like that PR replaces all of the AWS form inputs with the one for account ID. If that's the case, I think that probably what makes sense is to have the first step include release, target environment, and the AWS account ID. But @katierik and @andreasn might have suggestions for that. It would probably be good to chat sometime this week and figure out what approach makes sense.

image

### DIFF
--- a/src/PresentationalComponents/CreateImageWizard/WizardStepImageOutput.js
+++ b/src/PresentationalComponents/CreateImageWizard/WizardStepImageOutput.js
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Form, FormGroup, FormSelect, FormSelectOption, Title } from '@patternfly/react-core';
+
+const WizardStepImageOutput = (props) => {
+    const releaseOptions = [
+        { value: 'rhel-8', label: 'Red Hat Enterprise Linux (RHEL) 8.2' },
+    ];
+    const uploadOptions = [
+        { value: 'aws', label: 'Amazon Web Services' },
+    ];
+    return (
+        <Form>
+            <Title headingLevel="h2" size="xl">Image output</Title>
+            <FormGroup isRequired label="Release" fieldId="release-select">
+                <FormSelect value={ props.value } onChange={ value => props.setRelease(value) } isRequired
+                    aria-label="Select release input" id="release-select" data-testid="release-select">
+                    { releaseOptions.map(option => <FormSelectOption key={ option.value } value={ option.value } label={ option.label } />) }
+                </FormSelect>
+            </FormGroup>
+            <FormGroup isRequired label="Target environment" fieldId="upload-destination">
+                <FormSelect value={ props.upload.type || '' } id="upload-destination"
+                    data-testid="upload-destination" isRequired
+                    onChange={ value => props.setUpload({ type: value, options: props.upload.options }) } aria-label="Select upload destination">
+                    { uploadOptions.map(option => <FormSelectOption key={ option.value } value={ option.value } label={ option.label } />) }
+                </FormSelect>
+            </FormGroup>
+        </Form>
+    );
+};
+
+WizardStepImageOutput.propTypes = {
+    setRelease: PropTypes.func,
+    value: PropTypes.string,
+    upload: PropTypes.object,
+    setUpload: PropTypes.func,
+};
+
+export default WizardStepImageOutput;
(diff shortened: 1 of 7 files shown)


jgiardino:
As a follow up to these updates, the wizard doesn't fill the content area as it show, and I think this is related to how page sections are defined. I'm not sure we can follow the PF examples exactly as they are given how the clouddot creates page sections in the main region. This might be solvable by adding a flex layout to wrap the wizard header and body, and using the html classes provided for the in-page wizard for the body.

jgiardino:
@Gundersanne I see there are some tests I need to update, but I'm not sure how to run the tests locally. What's the command to get the test results locally?

Gundersanne:
@jgiardino `npm run tests:single` is the way to do it locally with readable-ish result output.

Gundersanne:
Also regarding #60 yea all that composer needs in future is the aws account id to share the ec2 snapshot/ami with. I'll put it on the agenda for the next meeting.

jgiardino:
Thanks, @Gundersanne! I'll work on updating the tests next. I also met with @katierik and @andreasn and these are some additional updates we discussed that I'll work on for this PR:

- make the first register radio button selected by default - this will force the user to either enter an activation key or choose to register later
- replace the flex layout on the Review step with a data list (`<dl>`) so that it has more semantics to associate the label with the value

Other updates we discussed for other PRs include:
- Moving the Wizard to a modal. The inline wizard is great if we have UI that requires a modal (e.g. if we want to enable the user to create a user account, like we do with cockpit-composer). But for now we don't. And for now there's a weird layout issue where the height of the wizard is 100% plus the height of the header, which forces the user to always scroll to the wizard footer buttons. So for now, we discussed that the best fix is to move the wizard to a modal.
- Check that the clouddot toolbar component is being used in the table view. I'm seeing an issue with the button being resized to fill with width. This component might address that.

jgiardino:
> Also regarding #60 yea all that composer needs in future is the aws account id to share the ec2 snapshot/ami with. I'll put it on the agenda for the next meeting.

I forgot to mention that we discussed this too. I think the consensus was that for now, we can combine this with the release and target environment in step 1, and not have a separate upload step for just the AWS account ID. Then later we can add a separate step for upload details if needed for future enhancements.

jgiardino:
@Gundersanne I was able to fix most of the failing tests, but I couldn't figure out why the last one was failing. Do you know why these last two tests are failing?

image

jgiardino:
@katierik and @andreasn I made these updates that we discussed:

- make the first register radio button selected by default - this will force the user to either enter an activation key or choose to register later
- replace the flex layout on the Review step with a data list (`<dl>`) so that it has more semantics to associate the label with the value

Here is an example of the updated Review screen. I had to add additional styles for the [data list](https://www.patternfly.org/v4/components/text#data-list) (`<dl>`) to make sure the labels have a consistent width in each section. I also increased the spacing around the section headers by one size, and decreased the space between data list items by one size. Otherwise things were kinda running together.

image

Gundersanne:
> @Gundersanne I was able to fix most of the failing tests, but I couldn't figure out why the last one was failing. Do you know why these last two tests are failing?
> 
> image

Hmm it seems you made this PR based on master, but we renamed the master branch to main. As a result there's a few commits that are missing. I'll rebase it and push it to yours, fixing the tests too.

If you're fine with it I'll also move all the individual wizard components under one folder. I think 1 folder in SmartComponents should have 1 folder in PresentationalComponents.

jgiardino:
> Hmm it seems you made this PR based on master, but we renamed the master branch to main. As a result there's a few commits that are missing. I'll rebase it and push it to yours, fixing the tests too.

Ah, ok. Thanks for catching that and fixing it!

> If you're fine with it I'll also move all the individual wizard components under one folder. I think 1 folder in SmartComponents should have 1 folder in PresentationalComponents.

Good call! That makes sense. I was trying to take what @jkozol did in cockpit-composer and incorporate that here, but what I ended up with definitely seemed off. I like your suggestion.

PR #63, opened 2020-12-16
Table view updates

jgiardino:
Updates include:
- Release column:
  - Adding a blue label
  - Adding a user-friendly short label (e.g. RHEL 8 vs rhel-8)
- Target column
  - Adding a user-friendly long label (e.g. Amazon Web Services vs aws)
- Adding more PF toolbar components to update the toolbar layout. This still leaves the clouddot `<TableToolbar>` component, which I'm not 💯% positive is the right component to use over PF's `<Toolbar>`, especially given the large margin before the Create button.

image


Gundersanne (approved):
Aw nice this looks amazing!

PR #64, opened 2020-12-16
Adds a Stack layout to have the wizard footer visible

jgiardino:
TLDR; before this update the user always has to scroll to get to
the footer actions in the wizard.

There's a weird nested page section thing happening which results
in the wizard body+footer being 100% of the height of the main content
area, but when you add the wizard header as a separate section above
the 100% height wizard body+footer, the header pushes the wizard
contents down so that the bottom portion of the wizard (equal to the
height of the header, and roughly the height of the footer) is always out
of view. This means that the user always has to scroll to get to the
footer.


Gundersanne (approved):
looks good!

PR #66, opened 2020-12-16
src: Replace Main with React.Fragment in App.js

Gundersanne:
We shouldn't apply any style at this point, and let any style be
determined by the page we're embedded in.

---

This doesn't affect the current layout:

image


jgiardino (commented):
LGTM! Thanks for updating this.

PR #69, opened 2020-12-17
Same folder name between presentational and smart components

Gundersanne:
This is just for consistency and clarity. Every folder under 'smartcomponents' should have a folder with the same name under 'presentationalcomponents' imo.


jkozol (approved):
This makes sense to me!

PR #74, opened 2021-02-02
src: add package step to wizard

jkozol:
The additional package selection step is added to the create image wizard. This step currently uses mock data set in packages.json. There are still a few things needed. Empty and loading state text needs to be added, actual data and routing needs to be supported, pagination needs to actual page through the data list. Tests need to be added.

selectPackagesV0

### DIFF
--- a/src/PresentationalComponents/CreateImageWizard/WizardStepPackages.js
+++ b/src/PresentationalComponents/CreateImageWizard/WizardStepPackages.js
@@ -1,0 +1,193 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {
+    Button, ButtonVariant,
+    DataList, DataListAction, DataListCell, DataListItem, DataListItemRow, DataListItemCells,
+    InputGroup,
+    Pagination,
+    Text, TextContent, TextInput,
+    Title,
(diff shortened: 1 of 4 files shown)


andreasn:
Are you planning on changing this to a dual list selector, as in Katie's latest wireframes (v4)?
https://www.patternfly.org/v4/components/dual-list-selector

jkozol:
@katierik this is my draft PR for package selection.

katierik:
Based on the discussion earlier & design changes I'll wait to review :)

jkozol:
So the design changes may be painful to implement. There is no current way to do it using Patternfly's react components. There may be a way using the html but it will be time intensive. I created a request for enhancement [here](https://github.com/patternfly/patternfly-react/issues/5436). I am working on other features for now.

jkozol:
@Gundersanne I just realized that we need the package `summary` not the `description`. I did not realize until now but the summary is much shorter than the description. 

**This is with the description:**
description


**This is with the summary:**
summary

Gundersanne:
ah I'll change that asap :)

jkozol:
Closed in favor of #86

PR #75, opened 2021-02-03
LandingPage: Show under construction page if image-builder returns 404

Gundersanne:
image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -5,11 +5,16 @@ import { connect } from 'react-redux';
 import { Routes } from './Routes';
 import './App.scss';
 
+import api from './api.js';
+import PermissionDenied from './PresentationalComponents/LandingPage/PermissionDenied';
+
 class App extends Component {
     constructor() {
         super();
 
-        this.state = {};
+        this.state = {
+            permission: true,
+        };
     }
 
     componentDidMount () {
@@ -19,6 +24,13 @@ class App extends Component {
         insights.chrome.auth.getUser().then(data => {
             this.setState({ identity: data.identity });
         });
+
+        api.getVersion().then(() => {
+            this.setState({ permission: true });
+        }).catch(() => {
+            this.setState({ permission: false });
+        });
+
     }
 
     componentWillUnmount () {
@@ -28,7 +40,7 @@ class App extends Component {
     render () {
         return (
             <React.Fragment>
-                <Routes childProps={ this.props } />
+                { this.state.permission ? <Routes childProps={ this.props } /> : <PermissionDenied /> }
             </React.Fragment>
         );
     }
--- a/src/PresentationalComponents/LandingPage/PermissionDenied.js
+++ b/src/PresentationalComponents/LandingPage/PermissionDenied.js
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import { InfoCircleIcon } from '@patternfly/react-icons';
+import { EmptyState, EmptyStateVariant, EmptyStateIcon, EmptyStateBody, Title } from '@patternfly/react-core';
+
+const PermissionDenied = () => {
+    return (
+        <EmptyState variant={ EmptyStateVariant.large } data-testid="empty-state-denied">
+            <EmptyStateIcon icon={ InfoCircleIcon } />
+            <Title headingLevel="h4" size="lg">
+            Image Builder is not quite ready
+            </Title>
+            <EmptyStateBody>
+            Image Builder is in early development and not ready for use yet.
+            If you&apos;re interested in trying it out once it reaches beta,
+            there will be a contact form up soon to notify anyone who signs up.
+            </EmptyStateBody>
+        </EmptyState>
+    );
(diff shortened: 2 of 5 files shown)


Gundersanne:
Ideas for where to direct the people interested in joining?

src/PresentationalComponents/LandingPage/PermissionDenied.js
```diff
@@ -0,0 +1,21 @@
...
+
+import { InfoCircleIcon } from '@patternfly/react-icons';
+import { EmptyState, EmptyStateVariant, EmptyStateIcon, EmptyStateBody, Title } from '@patternfly/react-core';
+
+const PermissionDenied = () => {
+    return (
+        <EmptyState variant={ EmptyStateVariant.large } data-testid="empty-state-denied">
+            <EmptyStateIcon icon={ InfoCircleIcon } />
+            <Title headingLevel="h4" size="lg">
+            Image Builder is not quite ready
```
larskarlitski, line 11: Love this!

src/PresentationalComponents/LandingPage/PermissionDenied.js
```diff
@@ -0,0 +1,21 @@
...
+            </Title>
+            <EmptyStateBody>
+            Image Builder is still very early in development and not quite ready for general use yet.
+            Would you be interested in testing it once it reaches beta, let us know and we&apos;ll be in touch once it&apos;s ready!
```
larskarlitski, line 15: I think I'd say "wider use" instead of "general" and drop the "very": "Image Builder is still early in development and not quite ready for wider use yet."

And maybe "if you're interested" instead of "would you be"? Not sure though.

My first idea was to ask people to use the pendo feedback button in the bottom right corner, but it doesn't seem to work. @andreasn do you have an alternative idea?

larskarlitski (commented):
Thank you!

src/PresentationalComponents/LandingPage/PermissionDenied.js
andreasn: I think a blue button in the middle of the page would work best, labelled `Sign up`
The very easiest and straight forward would be to link it to a Google form, but I'll ask in the Pendo slack channel if it's possible to use the feedback form.

Gundersanne:
Also @andreasn  I used the EmtyState patternfly element which isn't entirely according to your design, but I figured our usualy landing page uses the same element if there's no images being built.

I'm happy to change it to something with a white background though.

andreasn:
>  I used the EmtyState patternfly element which isn't entirely according to your design, but I figured our usualy landing page uses the same element if there's no images being built.

Looks good to me. It was a really quick and dirty design (the majority of the time was spent finding the best "under construction" image from the GeoCities era, finally settling on a Pokemon) and I think empty state looks good!

src/PresentationalComponents/LandingPage/PermissionDenied.js
andreasn: Would you make it start out like a question, but ends up being a statement, so I would suggest something like this:
`If you're interested in trying it out once it reaches beta, fill out your contact information in the sign up form, and we'll be in touch once it's ready.`

andreasn (changes requested):

src/PresentationalComponents/LandingPage/PermissionDenied.js
andreasn: Or if we use the same language as the form Margot is going to set up:
`If you're interested in trying it out once it reaches beta, fill out your contact details and we’ll reach out to you once the service is ready!`

larskarlitski:
@andreasn, we're merging this with a preliminary message and will address other comments in follow-ups.

larskarlitski:
Ok we decided to do this properly next week, after all. Sorry for the noise.

src/App.js
```diff
@@ -19,6 +24,13 @@ class App extends Component {
         insights.chrome.auth.getUser().then(data => {
             this.setState({ identity: data.identity });
         });
+
+        api.getVersion().then(() => {
+            this.setState({ permission: true });
+        }).catch(error => {
+            this.setState({ permission: error.response.status !== 404 });
```
larskarlitski, line 31: I would default to being more restrictive and setting permission to `false` for all errors. (If I understand axios' documentation correctly, this handler is called for any non-2xx response.)

For example, the backend not being reachable or throwing internal server errors should also show the PermissionDenied message.

larskarlitski (changes requested):
Thanks for updating this! Just a minor nit below.

larskarlitski:
Also, we might want to change codecov to give a green check if we only reduce coverage by a bit. This happens quite often when refactoring and shouldn't block things. In osbuild-composer, we're [using 5%](https://github.com/osbuild/osbuild-composer/blob/main/codecov.yml#L6).

larskarlitski (approved):
Thank youuuu :heart:

PR #76, opened 2021-02-04
Display wizard in modal

jkozol:
The wizard now displays inside a modal. Also the field `isInPage` is removed.

### DIFF
--- a/src/SmartComponents/CreateImageWizard/CreateImageWizard.js
+++ b/src/SmartComponents/CreateImageWizard/CreateImageWizard.js
@@ -218,7 +218,6 @@ class CreateImageWizard extends Component {
                     <Wizard
                         onNext={ this.onStep }
                         onGoToStep={ this.onStep }
-                        isInPage
                         steps={ steps }
                         onClose={ this.onClose }
                         onSave={ this.onSave } />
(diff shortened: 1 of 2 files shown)


KKoukiou (approved):
code changes look good, how to you test this locally?

Gundersanne (approved):

Gundersanne:
@KKoukiou  `npm run test` to run locally, or `npm run travis:verify` for linting as well. `npm run test:single` to make the output slightly more readable since it's not great :p

Gundersanne (commented):
Wait why do we want it as a modal?

jkozol:
Marking this as blocked until further discussion about design. @andreasn am I correct in my understanding that the Wizard should be a modal?

andreasn:
> am I correct in my understanding that the Wizard should be a modal?

Yup, that's what Katie recommended, as the rest of the Insights apps use modals for their wizards.

jkozol:
@andreasn thanks for the clarification!

@Gundersanne is it ok if I merge this?

andreasn:
> Wait why do we want it as a modal?

* This is what all other wizards on CRO does.
* Avoids (at least to some extent) the problem where the dropdowns become very, very wide on external monitors.
* Avoids the double left sidebar.
* Makes the application have less modes. Clearer distinction between the image list and the image creation.

Gundersanne (approved):
Alright, then this is good! Sorry for the confusion.

PR #77, opened 2021-02-05
CreateImageWizard: Change image type to ami

Gundersanne:
Once we have multiple image types we need to make this dynamic based on
upload type. But qcow2 doesn't work under any circumstance so let's use
ami.


larskarlitski (approved):
:+1:

PR #78, opened 2021-02-05
travis: Let main push to prod-beta as well

Gundersanne:
Once we fork prod in imagebuilder to be a separate version from stage, let's do the same with the frontend. Until that point let's just have main push to both stage and prod beta.

### DIFF
--- a/.travis/custom_release.sh
+++ b/.travis/custom_release.sh
@@ -4,7 +4,7 @@ set -x
 
 if [ "${TRAVIS_BRANCH}" = "main" ]
 then
-    for env in ci qa stage
+    for env in ci qa stage prod
     do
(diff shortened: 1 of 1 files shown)


jkozol:
@Gundersanne So eventually we will have a separate github branch for our production pushes? Or is this just about travis?

larskarlitski (approved):
Agreed. Seems simpler this way while we're ramping up.

Gundersanne:
@jkozol travis actually pushes the build to the redhatinsights/image-builder-frontend-build repo, and prod-beta (also here) is a separate branch. But as image-builder isn't yet forked from stage, i see no reason why the frontend should be (even though I did it idk why).

PR #82, opened 2021-02-17
add support for CentOS Stream 8

ondrejbudai:
image-builder has recently got support for building CentOS Stream 8 images.
This commit adds the option to build these images in the frontend.

I'm not entirely sure what I'm doing, so please review carefully. :) Also, I'm not sure if the backend part is deployed yet...


teg:
Thanks for this Ondrej, this is great! I would approve, but I fear that would be like the blind leading the blind, so I'll leave this in @jkozol and @Gundersanne's capable hands.

Gundersanne (approved):
Thanks!

PR #86, opened 2021-02-24
Select packages

jkozol:
The additional package selection step is added to the create image wizard. Users can now query for packages and select them from a list. These packages are not yet passed into the blueprint customizations and are just for display. Tests have been updated and added.

Currently, we are missing the empty state strings and the search bar error for too few characters. I will create issues to capture these but we will probably need to rewrite this component using html and patternfly's css classes so I don't think it is feasible before the demo is due.

@Gundersanne @katierik do you think we should alphabetically sort the list of selected packages or leave it in the user selected order?

packageSelection

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@babel/runtime": "^7.11.0",
     "@patternfly/patternfly": "^4.87.2",
-    "@patternfly/react-core": "^4.97.1",
+    "@patternfly/react-core": "^4.97.3",
     "@patternfly/react-table": "^4.23.1",
     "@redhat-cloud-services/frontend-components": "^2.4.23",
     "@redhat-cloud-services/frontend-components-utilities": "^2.2.7",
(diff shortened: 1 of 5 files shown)


teg (approved):
🥳

katierik:
User selected order! Its easier for users to find what they just did and confirm/undo/etc.
In other places we will show this list we will alphabetize, but when the user is actively working with it it is better to keep it in their order.

teg:
+1 for merging as is and dealing with the pagination in a follow-up. I figure scrolling past 100 entries is a pretty niche case, no?

Gundersanne:
> +1 for merging as is and dealing with the pagination in a follow-up. I figure scrolling past 100 entries is a pretty niche case, no?

What if you want to install a  `python3-*` package but can't remember the exact name?

katierik:
> > +1 for merging as is and dealing with the pagination in a follow-up. I figure scrolling past 100 entries is a pretty niche case, no?
> 
> What if you want to install a `python3-*` package but can't remember the exact name? 🤔
> 
> edit: please do merge as is :)

Hmmm maybe they should have thought of that before they *dared* install python! /s

Ideally we should load more on scroll.

Gundersanne:
I made #95 to support the pagination/scrolling.

PR #89, opened 2021-02-24
CreateImageWizard: add google upload step

jkozol:
The user can now specify their authentication settings for Google Cloud Platform.

Tests and validation are still needed for this PR.

emptyGoogle
googleDomain
googlePopover

### DIFF
--- a/src/PresentationalComponents/CreateImageWizard/WizardStepUploadGoogle.js
+++ b/src/PresentationalComponents/CreateImageWizard/WizardStepUploadGoogle.js
@@ -1,0 +1,148 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Form, FormGroup, TextList, TextListItem, Popover, Radio, TextContent, Text, TextInput, Title } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+
+const WizardStepUploadGoogle = (props) => {
+    const accountTypePopover = (
+        <Popover
+            hasAutoWidth
+            maxWidth='35rem'
+            headerContent={ 'Valid account types' }
+            bodyContent={ <TextContent>
+                <Text>The following account types can have an image shared with them:</Text>
+                <TextList>
+                    <TextListItem>
+                        <strong>Google account:</strong> A Google account represents a developer, an administrator,
+                        or any other person who interacts with Google Cloud. e.g., <em>`alice@gmail.com`</em>.
+                    </TextListItem>
+                    <TextListItem>
+                        <strong>Service account:</strong> A service account is an account for an application instead
+                        of an individual end user. e.g., <em>`myapp@appspot.gserviceaccount.com`</em>.
+                    </TextListItem>
+                    <TextListItem>
+                        <strong>Google group:</strong> A Google group is a named collection of Google accounts and
+                        and service accounts. e.g., <em>`admins@example.com`</em>.
+                    </TextListItem>
+                    <TextListItem>
+                        <strong>Google workspace domain/Cloud identity domain:</strong> A Google workspace or cloud identity
+                        domain represents a virtual group of all the Google accounts in an organization. These domains
+                        represent your organization&apos;s internet domain name. e.g., <em>`mycompany.com`</em>.
+                    </TextListItem>
+                </TextList>
+            </TextContent> }>
+            <button
+                type="button"
+                aria-label="Account info"
+                aria-describedby="google-account-type"
+                className="pf-c-form__group-label-help">
+                <HelpIcon />
+            </button>
+        </Popover>
+    );
+
+    return (
+        <Form>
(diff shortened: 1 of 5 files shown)


src/PresentationalComponents/CreateImageWizard/WizardStepUploadGoogle.js
```diff
@@ -0,0 +1,123 @@
...
+
+import { Form, FormGroup, TextList, TextListItem, Popover, Radio, TextContent, Text, TextInput, Title } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+
+const WizardStepUploadGoogle = (props) => {
+    const accountTypePopover = (
+        <Popover
+            hasAutoWidth
+            maxWidth='35rem'
+            headerContent={ 'Valid types of account emails' }
```
thozza, line 12: Although 3 out of 4 types are really emails, maybe we could reword this to "Valid account types"?

src/PresentationalComponents/CreateImageWizard/WizardStepUploadGoogle.js
```diff
@@ -0,0 +1,123 @@
...
+            bodyContent={ <TextContent>
+                <Text>The following types of emails can have an image shared with them:</Text>
```
thozza, line 14: Same feedback here as above. Personally I would use "The following account types ...".

thozza (approved):
This looks very nice. Great work! I added just two wording suggestions. I can not assess the code correctness, since I have almost no experience with JS...

src/PresentationalComponents/CreateImageWizard/WizardStepUploadGoogle.js
jkozol: Good point! I will change this.

src/PresentationalComponents/CreateImageWizard/WizardStepUploadGoogle.js
jkozol: Will update this too.

katierik:
LGTM! 🎉

src/PresentationalComponents/CreateImageWizard/WizardStepUploadGoogle.js
katierik: Updated in mocks as well

jkozol:
@Gundersanne This is ready for review. There are a couple minor issues with the PR. The upload state has gotten a little messy and should be refactored in the future. Also there is no validation. I have opened an issue for this https://github.com/osbuild/image-builder-frontend/issues/98

src/SmartComponents/CreateImageWizard/CreateImageWizard.js
```diff
@@ -240,34 +252,40 @@ class CreateImageWizard extends Component {
...
+        if (this.state.uploadDestinations.google) {
+            const upload_google = this.state.uploadGoogle;
+            delete upload_google.accountType;
+            let request = {
+                distribution: this.state.release,
+                image_requests: [
+                    {
+                        architecture: this.state.arch,
+                        image_type: 'gcp',
+                        upload_requests: [ upload_google ],
```
Gundersanne, line 281: Right this is getting a little confusing. I think it's best if we don't consider the upload (options) in the state to be directly usable in the request. And build the upload_request here in the onSave:

```
...
upload_requests: [{
    type: 'gcp',
    options: {
        share_with_accounts: upload_google.options.share_with_accounts
    }
}]
...
```

Can do that in a followup though because AWS would need the same treatment. That way we could also just flatten the upload state to just have `upload_google.share_with_accounts`, and not mess with the options all the time.

src/SmartComponents/CreateImageWizard/CreateImageWizard.js
```diff
@@ -54,8 +56,9 @@ class CreateImageWizard extends Component {
             },
             uploadGoogle: {
                 type: 'google',
```
Gundersanne, line 58: This should be `type: 'gcp',`

Gundersanne (changes requested):
Looks good so far, the type should be `gcp` though.

src/SmartComponents/CreateImageWizard/CreateImageWizard.js
jkozol: Fixed

Gundersanne (approved):
This looks good, I also created #100 to refactor the upload states.

I'd wait with merging this until we redeploy a composer version which includes gcp though, like we shouldn't enable things that don't work yet. Until then I'm adding the blocked label, and I'll get on deploying a new composer version now.

src/SmartComponents/CreateImageWizard/CreateImageWizard.js
teg: Tracked in #100

PR #91, opened 2021-02-25
CreateImageWizard: update review page

jkozol:
The review page now uses more patternfly components and has been split up a little so it is easier to add review fields for additional target environments.

### DIFF
--- a/src/PresentationalComponents/CreateImageWizard/WizardStepReview.scss
+++ b/src/PresentationalComponents/CreateImageWizard/WizardStepReview.scss
@@ -4,7 +4,6 @@
 // Increasing margins for h3 for better spacing and readability
 .pf-c-content h3 {
     margin-top: var(--pf-global--spacer--xl);
-    margin-bottom: var(--pf-global--spacer--md);
 }
 // Decreasing gap between dl items for better spacing and readability
 // Also setting a first column width to 25% instead of auto, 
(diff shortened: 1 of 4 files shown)


jkozol:
Switching to draft as I make changes

jkozol:
review

src/PresentationalComponents/CreateImageWizard/WizardStepReview.scss
```diff
@@ -14,4 +13,9 @@
         grid-template: 1fr / 25% 1fr;
         grid-gap: var(--pf-global--spacer--sm);
     }
+}
+
+#destination-header {
+    font-size: 18px;
+    margin-bottom: var(--pf-global--spacer--sm);
 }
```
Gundersanne, line 21: nitpick, but trailing newline please

Gundersanne (commented):
1 nitpick, and 1 question:

Is there an advantage over using the patternfly `Text*` components vs the regular `p/dt/dl/etc...` elements? Like do the pf components have extra styling applied? I'm fine with it  either way I just wonder what the difference is.

src/PresentationalComponents/CreateImageWizard/WizardStepReview.scss
jkozol: hehe i'm so bad at this one. Will fix.

jkozol:
> Is there an advantage over using the patternfly Text* components vs the regular p/dt/dl/etc... elements? Like do the pf components have extra styling applied? I'm fine with it either way I just wonder what the difference is.

@Gundersanne I am not completely sure. I have just seen plenty of their examples use it. If i had to guess it is because patternfly uses a few css variables over many components so using their Text* components will make sure all of our styles stay fairly consistent. @kkoukiou do you have a better answer for this?

Gundersanne (approved):
lgtm! thanks

PR #96, opened 2021-03-02
App: Wait for user active session before checking permission

Gundersanne:
This should fix the issue where if you go to the page directly, and are not logged in, you'd get the signup form after logging in even if you were authorized to use the service.

Essentially this waits until there's an active session.

https://user-images.githubusercontent.com/11140201/109670265-3aa0bc00-7b73-11eb-8a8e-c47045536dd0.mp4

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -13,7 +13,7 @@ class App extends Component {
         super();
 
         this.state = {
-            permission: true,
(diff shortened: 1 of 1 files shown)


andreasn (approved):
Much better. Thank you!

PR #101, opened 2021-03-05
Review and create feedback

Gundersanne:
on error:

https://user-images.githubusercontent.com/11140201/110132872-ee03ed80-7dcb-11eb-9ddf-d2afb5840491.mp4

notification:

https://user-images.githubusercontent.com/11140201/110132891-f1977480-7dcb-11eb-977b-ae21db44002d.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "@patternfly/react-core": "^4.97.3",
     "@patternfly/react-table": "^4.23.1",
     "@redhat-cloud-services/frontend-components": "^2.4.23",
+    "@redhat-cloud-services/frontend-components-notifications": "^3.0.4",
     "@redhat-cloud-services/frontend-components-utilities": "^2.2.7",
     "classnames": "^2.2.6",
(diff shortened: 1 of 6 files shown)


teg (approved):
nice :)

Gundersanne:
Don't merge :p

katierik:
Error looks 👍 
Notification toast looks 👍 
The time between clicking "create" and getting feedback is pretty long and with only the disable its still not quite communicative. Can we change the button label as seen [here](https://marvelapp.com/prototype/6g7dja9/screen/76873729)?
image

jkozol (approved):
LGTM!

jkozol:
@Gundersanne Create button text change looks good to me. The inline conditional is a little confusing but I can't think of a better way...

jkozol:
@Gundersanne I like that change. With validation being added the next button will have a few different states so it will be good to use an if else block.

jkozol (approved):
Looks good! Thanks for the updates.

PR #107, opened 2021-03-08
ImagesTable: Instead of the image type show the upload target

### LINKED ISSUE #105: target in the image listing page are different from the target names selected in the wizard
When creating an image from the wizard, you get asked to select between Amazon Web Services, Microsoft Azure and Google Cloud Platform.
But once the images are built, the targets gets listed as "vhd", "ami" etc. and without knowing the inner workings of Azure and Google, I find it quite hard to tell what one is what.

Gundersanne:
Fixes #105

---
image

---

TODO update the tests to check this


andreasn (approved):
I like this change!

teg (approved):

PR #108, opened 2021-03-08
ImagesTable: update empty-table message

teg:
Update the message now that we support CentOS Stream, Azure+GCP support as well as package selection.

### DIFF
--- a/src/SmartComponents/ImagesTable/ImagesTable.js
+++ b/src/SmartComponents/ImagesTable/ImagesTable.js
@@ -88,7 +88,8 @@ class ImagesTable extends Component {
                             Create an image
                         </Title>
                         <EmptyStateBody>
-                            Create RHEL OS images for deployment in Amazon Web Services. Images can include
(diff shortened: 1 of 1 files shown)


src/SmartComponents/ImagesTable/ImagesTable.js
```diff
@@ -88,7 +88,8 @@ class ImagesTable extends Component {
                             Create an image
                         </Title>
                         <EmptyStateBody>
-                            Create RHEL OS images for deployment in Amazon Web Services. Images can include
+                            Create OS images for deployment in Amazon Web Services, Azure and
```
Gundersanne, line 91: Maybe use "Microsoft Azure" here, whenever we use the full name we tend to use that.

teg: Ah, I meant to change that. Thanks for catching it!

teg: Done.

teg:
Replaced by #109

PR #109, opened 2021-03-10
ImagesTable: update empty-table message

teg:
Update the message now that we support CentOS Stream, Azure+GCP support as well as package selection.


jkozol (approved):
One super minor nitpick. I think the statement `Images can include an activation key to automate the registration process and a custom package set.` is slightly clearer. Approving anyways.

PR #114, opened 2021-03-11
CreateImageWizard: validate upload destination

### LINKED ISSUE #110: CreateImageWizard: require upload destination
We should require at least one cloud provider to be selected, if none is, disable the Next button

jkozol:
The user should not be able to continue without selecting an upload destination. The only issue here is that, while the Next button is disabled, the user can still select other steps in the sidebar. This fixes #110 

disableNextButtonDest

### DIFF
--- a/src/PresentationalComponents/CreateImageWizard/ImageWizardFooter.js
+++ b/src/PresentationalComponents/CreateImageWizard/ImageWizardFooter.js
@@ -14,24 +14,32 @@ const ImageWizardFooter = (props) => {
                     {({ activeStep, onNext, onBack, onClose }) => {
                         let nextButtonText = 'Next';
                         if (activeStep.name === 'Review') {
-                            nextButtonText = props.disable ? 'Creating...' : 'Create';
+                            nextButtonText = props.isSaveInProgress ? 'Creating...' : 'Create';
+                        }
+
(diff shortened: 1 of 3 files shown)


src/PresentationalComponents/CreateImageWizard/ImageWizardFooter.js
```diff
@@ -14,24 +14,33 @@ const ImageWizardFooter = (props) => {
                     {({ activeStep, onNext, onBack, onClose }) => {
                         let nextButtonText = 'Next';
                         if (activeStep.name === 'Review') {
-                            nextButtonText = props.disable ? 'Creating...' : 'Create';
+                            nextButtonText = props.isSaveInProgress ? 'Creating...' : 'Create';
+                        }
+
+                        let nextButtonIsDisabled = false;
```
Gundersanne, line 20: nitpicky, but you can avoid the conditional on the next line by just `let nextButtonIsDisabled = props.isSaveInProgress;`

jkozol: Fixed

teg (approved):
Nice!

Gundersanne (approved):
lgtm! thank you

Issue #112, opened 2021-03-11
CreateImageWizard: 


jkozol:
Default to register the system later. If registering is enabled, require an activation key.
(do we have a way to verify at least the format?)

PR #115, opened 2021-03-11
Validate subscription activation key

jkozol:
If register now is selected, an activation key is required. If no activation key is entered the Next and Create buttons are disabled.
Also, if the activation key is empty an error message will appear to prompt the user to enter an activation key.

The subscription errors variable is removed in favor of declaring the error message directly in the registration step component. The validateSubscription function is removed in favor of validating the activation key field on change.

This fixes #114 and is dependent on #112 

invalidActivationKey

### DIFF
--- a/src/PresentationalComponents/CreateImageWizard/WizardStepReview.js
+++ b/src/PresentationalComponents/CreateImageWizard/WizardStepReview.js
@@ -81,9 +81,9 @@ const WizardStepReview = (props) => {
         subscriptionReview = (<>
             <TextListItem component={ TextListItemVariants.dd }>Register the system on first boot</TextListItem>
             <TextListItem component={ TextListItemVariants.dt }>Activation key</TextListItem>
-            { props.subscriptionErrors['subscription-activation'] ? (
+            { !props.isValidSubscription || !props.subscription['activation-key'] ? (
                 <TextListItem component={ TextListItemVariants.dd }>
-                    <ExclamationCircleIcon className="error" /> { props.subscriptionErrors['subscription-activation'].value }
+                    <ExclamationCircleIcon className="error" /> { 'A value is required' }
                 </TextListItem>
             ) : (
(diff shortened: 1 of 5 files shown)


src/PresentationalComponents/CreateImageWizard/WizardStepReview.js
```diff
@@ -130,7 +130,7 @@ WizardStepReview.propTypes = {
     subscription: PropTypes.object,
     subscribeNow: PropTypes.bool,
     uploadAWSErrors: PropTypes.object,
-    subscriptionErrors: PropTypes.object,
+    isSubscriptionValid: PropTypes.object,
```
Gundersanne, line 133: Hm should this be `isValidSubscription`?

src/PresentationalComponents/CreateImageWizard/WizardStepReview.js
```diff
@@ -96,7 +96,7 @@ const WizardStepReview = (props) => {
     return (
         <>
             { (Object.keys(props.uploadAWSErrors).length > 0 ||
-               Object.keys(props.subscriptionErrors).length > 0) &&
+               !props.isSubscriptionValid) &&
```
Gundersanne, line 99: The property passed from the wizard is `isValidSubscription` no?

Gundersanne (commented):
Sorry for the late review, have a q.

src/PresentationalComponents/CreateImageWizard/WizardStepReview.js
jkozol: You are totally correct. Thanks for catching this. Fixed now.

src/PresentationalComponents/CreateImageWizard/WizardStepReview.js
jkozol: Fixed.

Gundersanne (approved):

PR #129, opened 2021-04-08
CreateImageWizard: Move steps into separate components

Gundersanne:


src/SmartComponents/redux/reducers.js
```diff
@@ -18,3 +18,33 @@ export function composeReducer(state = { }, action) {
...
+    let newState = Object.assign({}, state);
+    switch (action.type) {
+        case types.SET_RELEASE:
+            newState.release = { arch: action.arch, release: action.release };
+            return newState;
+        case types.SET_UPLOAD_DESTINATION:
+            newState.uploadDestinations = {
+                aws: action.aws,
+                azure: action.azure,
+                google: action.google,
```
jkozol, line 44: I personally am a fan of using spread operators. Any thoughts on using them through out the state?
 
```
switch (action.type) {
  case types.SET_RELEASE:
    return {
      ...state,
      release: {
        arch: action.arch,
        release: action.release 
      }
    };
```

src/SmartComponents/redux/reducers.js
```diff
@@ -18,3 +18,33 @@ export function composeReducer(state = { }, action) {
...
+        release: 'rhel-8',
+    },
+    uploadDestinations: {
+        aws: false,
+        azure: false,
+        google: false,
+    },
+};
+
+export function pendingCompose(state = initialPendingComposeState, action) {
```
jkozol, line 34: I think it would be good to split the reducers into different files. So `/reducers/composes` and `/reducers/pendingCompose`

src/SmartComponents/redux/reducers.js
```diff
@@ -18,3 +18,33 @@ export function composeReducer(state = { }, action) {
             return state;
     }
 }
+
+const initialPendingComposeState = {
+    release: {
+        arch: 'x86_64',
+        release: 'rhel-8',
```
jkozol, line 25: Maybe we could rename one of the releases so it isn't `release.release`? `release.distro`?

jkozol (commented):
Thanks for starting this work! A few comments on the new reducer. Overall great idea to split out the state and the different wizard components :)

src/store/actions/actions.js
```diff
@@ -7,6 +7,78 @@ function updateCompose(compose) {
     };
 }
 
+function setRelease({ arch, distro }) {
+    return {
+        type: types.SET_RELEASE,
+        arch,
+        distro,
```
jkozol, line 14: A general convention with redux actions is for the format to be:
```
{
  type: `type name`
  payload: object
}
```
so here it would be:
```
{
  type: types.SET_RELEASE,
  payload: {
    arch,
    distro,
  }
}
```

src/store/index.js
```diff
@@ -17,7 +18,8 @@ export function init (store = {}, ...middleware) {
     ]);
 
     registry.register({
-        composes: composeReducer,
+        composes,
+        pendingCompose,
```
jkozol, line 22: I am not sure how well it will play with the clouddot ReducerRegistry but I believe it would be good to combine the reducers into one reducer using [combineReducers](https://redux.js.org/api/combinereducers). This can be done in an index.js file in the reducers directory like how we combine and export our actions in their index.js

jkozol (commented):
Thanks for this work. A couple of recommendations. If you would prefer to implement these changes in future PRs I am fine approving this now.

src/test/testUtils.js
```diff
@@ -3,19 +3,15 @@ import { Router } from 'react-router-dom';
 import { Provider } from 'react-redux';
 import { render } from '@testing-library/react';
 import { createMemoryHistory } from 'history';
-import configureStore from 'redux-mock-store';
+//import configureStore from 'redux-mock-store';
```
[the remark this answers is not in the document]
Gundersanne: nope :) i'll remove it

src/store/index.js
Gundersanne: Right so the [ReducerRegistry](https://github.com/RedHatInsights/insights-frontend-components/blob/master/src/Utilities/ReducerRegistry.js) already combines any reducers we register.

I think that's also what allows us to just pass a `store` object to init() which follows the general store structure.

src/SmartComponents/CreateImageWizard/WizardStepReview.js
```diff
@@ -0,0 +1,155 @@
...
+            subscriptionReview = (<>
+                <TextListItem component={ TextListItemVariants.dd }>Register the system on first boot</TextListItem>
+                <TextListItem component={ TextListItemVariants.dt }>Activation key</TextListItem>
+                { !this.props.isValidSubscription || !this.props.subscription['activation-key'] ? (
+                    <TextListItem component={ TextListItemVariants.dd }>
+                        <ExclamationCircleIcon className="error" /> { 'A value is required' }
+                    </TextListItem>
+                ) : (
+                    <TextListItem component={ TextListItemVariants.dd } type="password">
+                        {'*'.repeat(this.props.subscription['activation-key'].length)}
```
jkozol, line 96: `'activation-key'` needs to be `activationKey`

jkozol (changes requested):

jkozol (approved):
Great work! LGTM!

PR #147, opened 2021-04-16
Compose state refactor

jkozol:
When a compose is started the api call is no longer handled by the CreateImageWizard onSave function. Instead, the CreateImageWizard calls the composeStart thunk. This function calls the api and handles the response. If successful, the compose is added to the store. Otherwise, an error is added to the store.

The store's compose object now has a list of the compose ids and an object containing key/value pairs mapping a compose id to the compose for all composes. This "normalized" state will allow more efficiency when selecting individual composes or iterating through all composes.

The compose objects in the store now match the composeRequest object instead of having a shape unique to the UI. This can be changed in the future if image-builder's api returns compose objects of a different format.

Tests are updated for new compose format and action/reducer types.


Gundersanne (approved):
This is lovely! Thank you.

PR #152, opened 2021-04-20
Add documentation links

jkozol:
Links to documentation are added. Some of the styling does not exactly match the mockups but is the default styling for the patternfly components. @katierik any thoughts?
wizardDocumentation
documentationLanding


Gundersanne (approved):
Good with me :)

katierik:
That's the "light" color, right? That passes color contrast standards. All good.

PR #153, opened 2021-04-20
src: Remove SmartComponents/PresentationalComponents split

Gundersanne:
Just have a directory per component.


jkozol:
I think we should still have a `components` directory where the CreateImageWizard, ImagesTable, and LandingPage directories live. Like we do with the `store` directory where all of the redux directories/files are.

jkozol (approved):
LGTM!

PR #155, opened 2021-04-21
Fetch all composes and update status in redux

jkozol:
A user's prior composes will now be displayed on load. The `/compose` api route is now used to fetch them and the get status api call is now made in redux instead of in the ImagesTable component. Also, the store's compose object has been slightly modified to match the format returned by `/compose`.


src/api.js
```diff
@@ -11,6 +11,16 @@ async function composeImage(body) {
     return request.data;
 }
 
+async function getComposes(limit, offset) {
+    const params = new URLSearchParams({
```
Gundersanne, line 15: I had on idea this existed, neat!

src/store/actions/actions.js
```diff
@@ -35,6 +35,25 @@ export const composeStart = (composeRequest) => async dispatch => {
...
+    payload: { id, status }
+});
+
+export const composeGetStatus = (id) => async dispatch => {
+    const request = await api.getComposeStatus(id);
+    dispatch(composeUpdatedStatus(id, request.image_status));
+};
+
+export const composeGetAll = () => async dispatch => {
+    // only search for the first 10 composes for now
```
Gundersanne, line 49: technically it's the most recent 10

Gundersanne (approved):
Looking good, thank you!

PR #157, opened 2021-04-22
CreateImageWizard: remove nested p tag

jkozol:
The patternfly Wizard component's description is wrapped in a p tag. The patternfly Text component also uses a p tag. A p tag should not be nested inside another p tag.


Gundersanne (approved):
ayy good catch

PR #163, opened 2021-04-27
Add pagination to the composes list and ImagesTable toolbar

jkozol:
Improve the compose actions to enable pagination and display a paginated list of composes.
pagination1


Gundersanne (commented):
Looks good! Works locally, any chance of expanding the imagestable test though?

jkozol:
@Gundersanne Added tests for the new count state and a test to check that pagination renders. Testing pagination more thoroughly is a little tricky since it is very much a patternfly component. At some point I would like to split the toolbar into a separate component and can test rendering it with a default state.

PR #164, opened 2021-04-27
CreateImageWizard: Show previously selected packages

Gundersanne:
This should fix the bug where chosen packages would disappear from the
ui when switching steps and returning to the packages step.


jkozol (approved):
lgtm

PR #165, opened 2021-04-29
chrome 2.0

Gundersanne:


config/dev.webpack.config.js
```diff
@@ -1,13 +1,23 @@
...
+    useFileHash: false,
 });
 
+plugins.push(
+    require('@redhat-cloud-services/frontend-components-config/federated-modules')({
+        root: resolve(__dirname, '../'),
+        useFileHash: false,
+        moduleName: 'image_builder',
+        exposes: {
+            './RootApp': resolve(__dirname, '../src/bootstrap-dev'),
```
karelhala, line 16: This one should point to `AppEntry.js` otherwise you'll end with 2 apps in virtual DOM.

config/dev.webpack.config.js
```diff
@@ -1,13 +1,23 @@
-
 const { resolve } = require('path');
 const config = require('@redhat-cloud-services/frontend-components-config');
 const { config: webpackConfig, plugins } = config({
     rootFolder: resolve(__dirname, '../'),
     debug: true,
-    https: false,
     port: 8002,
+    useFileHash: false,
```
karelhala, line 7: Add `isChrome2: false` for time being in here to use fallback render. Once the code hits production and we update the `main.yml` config to use the module you can remove it from here.

karelhala (commented):
Looking good! Just a few minor improvements so the app renders and works properly.

config/dev.webpack.config.js
Gundersanne: Hmm it seems `isChrome2` is part of insights-chrome, and not part of `frontend-components-config`. `skipChrome: true` works though.

karelhala (approved):

karelhala:
Any update on this? @Gundersanne it should be OK to merge.

jkozol (approved):
I was able to build and run this as expected.

PR #174, opened 2021-05-09
Add full-stack development environment

teg:
This adds a Dockerfile and docker-compose.yml to make development of the full image builder stack easier.

This requires the other image builder projects to be checked out, but otherwise should be self contained and not depend on any other local state.

Note: cloud secrets have still not been hooked up, so pushing to the cloud environments will not work yet.


teg:
This combines work from @Gundersanne @achilleas-k and @gicmo.

profiles/local-frontend-and-api-with-identity.js
```diff
@@ -106,15 +106,13 @@ const authPlugin = (req, res, target) => {
 
 
 
-routes[`/beta/${SECTION}/${APP_ID}`] = { host: `http://localhost:${FRONTEND_PORT}` };
-routes[`/${SECTION}/${APP_ID}`]      = { host: `http://localhost:${FRONTEND_PORT}` };
-routes[`/beta/apps/${APP_ID}`]       = { host: `http://localhost:${FRONTEND_PORT}` };
-routes[`/apps/${APP_ID}`]            = { host: `http://localhost:${FRONTEND_PORT}` };
-routes[`/api/${APP_ID}`]             = { host: `http://localhost:${API_PORT}` };
+routes[`/beta/${SECTION}/${APP_ID}`] = { host: `http://frontend:${FRONTEND_PORT}` };
```
Gundersanne, line 109: Is it possible to set default values for these? I think we'd want this to work also without compose.

Gundersanne (commented):
I love this thank you :)

teg:
I decided to move this out to: https://github.com/teg/devtools (which, if we agree we want it, we should move to the osbuild org). What's left is just the needed Dockerfile, and a tweak to package.json, but I have no idea if that's the right thing to do, so still leaving it as WIP.

Gundersanne (approved):
LGTM, thank you! As discussed I would be fine with storing the compose file inside of the FE repo as well, as this is the only project that makes use of the full stack in the compose file.

teg:
Thanks! I'll use my day of learning to respin this :)

teg:
@Gundersanne I now put it all back here!

Gundersanne (approved):
lovely :smiley_cat:

Gundersanne:
@teg what do you mean by the 'cloud secrets have not been hooked up yet'?

teg:
>  @teg what do you mean by the 'cloud secrets have not been hooked up yet'? 

The worker can't push the images, I need to mount in the aws/azure/gcp secrets and set the right env var.

PR #178, opened 2021-05-11
src: Have separate AppEntry points for dev and prod

Gundersanne:


jkozol (commented):
What is the purpose of this change? As far as I can tell we don't pass in undefined middleware. I am unsure how to verify this works as expected.

Gundersanne:
@jkozol I took a different approach that makes more sense I think.

jkozol (approved):
Looks good to me! I assume we will have a separate commit in the prod branch importing ImageBuilder from AppEntry?

PR #179, opened 2021-05-12
Add link to google cloud

jkozol:
The images list now contains a link to the uploaded image. Only google cloud images platform images currently display the link following a successful upload.

This does not match the complete designs for the page but I would prefer to wait to implement the full design. There is no support for multiple uploads and the status information is still fairly limited. 

imageLink

### DIFF
--- a/src/Components/ImagesTable/ImageLink.js
+++ b/src/Components/ImagesTable/ImageLink.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Button } from '@patternfly/react-core';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+
+const ImageLink = (props) => {
+    const uploadStatus = props.imageStatus ? props.imageStatus.upload_status : undefined;
+    if (uploadStatus) {
+        let url = '';
+        if (uploadStatus.type === 'gcp') {
+            url = 'https://console.cloud.google.com/compute/imagesDetail/projects/' +
+                uploadStatus.options.project_id +
+                '/global/images/' +
+                uploadStatus.options.image_name;
+        }
+
+        return (
+            <Button
+                component="a"
+                target="_blank"
+                variant="link"
+                icon={ <ExternalLinkAltIcon /> }
+                iconPosition="right"
+                isInline
+                href={ url }>
+                    View uploaded image
+            </Button>
+        );
+    }
+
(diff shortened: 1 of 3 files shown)


teg:
Neat! I wasn't sure a direct link would work! Does this look good to you @thozza ?

thozza:
First of all, thanks @jkozol for working on this change!

Unfortunately, the link is quite useless from my point of view and I do believe that we should rather display the project ID and image name as strings, than as a URL. At least for now. 

The main reason is that the URL does not clearly display the image's project ID, which is needed to use the image using `gcloud` or API. And `gclud` and API are currently the only ways to use the image due to https://issuetracker.google.com/issues/185800732. Sure, the project ID is in the URL, but that is not ideal and not obvious from my point of view. When the user would go to the image page, there is literally nothing that they can accomplish with the image using the Web UI.

teg:
would it still make sense to include the link so people can look at the image info google probides, or is there nothing useful there?

thozza:
> would it still make sense to include the link so people can look at the image info google probides, or is there nothing useful there?

It looks like this:
Screenshot from 2021-05-13 14-46-31

The only useful information is IMO the image name. There are additional information, but I don't consider them very useful. The most problematic thing from my point of view that when users click on "Create Instance" on such image page, they'll get error. :confused:

jkozol:
I am not sure what a user would or wouldn't find useful on this page. I agree, however, that we should be careful about linking the user to the image as long as `Create Instance` is broken since this will confuse users about whether or not the image creation and upload was successful. 

@teg do you think any of the info beyond the project ID and image name are useful?

I believe that eventually this link would be useful so for now I am going to mark this as a draft and as blocked. I can open a separate PR to display project ID and image name.

teg:
@jkozol I think that's a good call.

jkozol:
@katierik do you have any thoughts on displaying this google cloud information?

katierik:
In the full design, I've got google cloud accessible by copying a CLI command: https://marvelapp.com/prototype/6g7dja9/screen/76945621 

This information that the link that GCP has does not work when clicking "execute image" is awful. I agree that it is bad usability to link the user to a place that they will immediately encounter errors. 

Will respond here in a bit with a design.

jkozol:
This is very out of date and probably not useful. If needed will just fully rewrite since the rebasing won't be worth it anyways.

PR #201, opened 2021-05-19
CreateImageWizard: Pass organization as a number

Gundersanne:
This should be changed in the api. But will take some more work. For now
just pass it as a number, that is what the api expects at this point in time.


teg (approved):
Thanks for the quick fix!

PR #216, opened 2021-05-27
CreateImageWizard: use your image instead of the image

### LINKED ISSUE #190: Always refer to the image as "your image"
In the review step for instance it's referred to as "the image"

jkozol:
This fixes #190. I left the term "the image" for the google and aws steps since we already specify it as "your image" and duplicate wording sounds weird.


Gundersanne (approved):
Sorry, missed this. Thank you!

PR #218, opened 2021-05-31
ImagesTable: add gcp info

jkozol:
When a gcp image is successfully uploaded the user needs to know the project id and image name. These are now displayed in the status section of the UI.

This is a minimal solution just to display the info until dropdowns for the image info can be implemented.

@katierik Do you have any thoughts on how to simply display this information for now?

gcp

### DIFF
--- a/src/Components/ImagesTable/ImageLink.js
+++ b/src/Components/ImagesTable/ImageLink.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { Button } from '@patternfly/react-core';
+import { Button, TextContent, Text, TextVariants, Popover } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 const ImageLink = (props) => {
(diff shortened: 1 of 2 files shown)


thozza:
LGTM! From functionality point of view, this looks good. I'm not qualified to assess the UX point of view, but personally I would probably like more if the headers were in the first column and values in the second (compared to being in the rows). But that's just my unqualified suggestion.

thozza (approved):

Gundersanne:
AAh tihs needs a rebase now @jkozol  :( But otherwise this looks good

katierik:
My understanding of the GCP process is that the image is shared with a specific email, and that email gets a notification that the image is shared. -- Does the email link have the same problems as the direct link we can provide?

I propose telling the user *where* to find the image by telling them where it has been shared to, and then providing the additional details. 
This layout also has the advantage of being 1 line long, which corresponds to the way the others will be displayed.

image

jkozol:
@katierik I like the idea of using a popover. Do you think it would be fine to only state "Shared with "exampleemail@gmail.com" inside of the popover so the style remains similar to the links?

katierik:
Sounds good to me, updated:
image

(Pulled the instructions from here: https://cloud.google.com/deployment-manager/docs/configuration/using-images-from-other-projects-for-vm-instances )

thozza:
> My understanding of the GCP process is that the image is shared with a specific email, and that email gets a notification that the image is shared. -- Does the email link have the same problems as the direct link we can provide?

Hmm. Did you actually see such email? My experience so far is that no notifications emails are sent when an image is shared.

> I propose telling the user _where_ to find the image by telling them where it has been shared to, and then providing the additional details.

The tricky part is that the user won't see the shared image in the Web UI at all. It won't show up in any of the wizards where images can be chosen by the user. So they won't find the image anywhere. For this reason, our documentation states to use "gcloud" in GCP Shell with the **image name** and **project ID**.

thozza:
> gcpDetails

This looks amazing!

PR #221, opened 2021-06-01
ImagesTable: add azure image link

jkozol:
When an azure image is successfully uploaded the user is provided a link to view the image in the azure web portal.

azure

### DIFF
--- a/src/test/Components/ImagesTable/ImagesTable.test.js
+++ b/src/test/Components/ImagesTable/ImagesTable.test.js
@@ -231,7 +231,11 @@ const store = {
                             image_type: 'vhd',
                             upload_request: {
                                 type: 'azure',
-                                options: {}
+                                options: {
+                                    tenant_id: 'test-tenant-id',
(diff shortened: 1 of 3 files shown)


src/test/Components/ImagesTable/ImagesTable.test.js
```diff
@@ -51,6 +51,37 @@ const store = {
                     }
                 },
             },
+            '77fa8b03-7efb-4120-9a20-da66d68c4494': {
```
Gundersanne, line 54: Bit confused with the diff in this file, did you just move it up ?

Gundersanne (commented):
just have 1 q

src/test/Components/ImagesTable/ImagesTable.test.js
jkozol: ah ya I moved it when playing around with the tests. I can move it back to improve the diff since the location doesn't matter.

Gundersanne (approved):
yeee thx :)

PR #224, opened 2021-06-02
store: Insert new composes on top of the table

### LINKED ISSUE #161: insert new composes on top of the overview table
New composes are appended to the table, should be inserted in front, as most recent composes should be on top.

Gundersanne:
Fixes #161


jkozol (approved):
This works as expected. Thanks!

PR #226, opened 2021-06-03
src: Rename RHEL 8.3 to just RHEL 8

Gundersanne:
We should pull these labels/descriptions from the service, but let's do
that separately.


teg (approved):
This makes sense. If ever we want to specify minor versions it should be explicitly for EUS supported ones (when the customer has the right subs)

PR #234, opened 2021-06-10
devel: update image builder link to point to UI

jkozol:


devel/README.md
```diff
@@ -40,6 +40,6 @@ docker compose up
 ```
 
 Access the service through the GUI:
-[http://prod.foo.redhat.com:1337/beta/](http://prod.foo.redhat.com:1337/beta/), or
+[https://prod.foo.redhat.com:1337/apps/image-builder](https://prod.foo.redhat.com:1337/apps/image-builder), or
```
teg, line 43: Should it not be `beta/`?

jkozol: I see that there are actually a few links that bring me to image builder. I updated this to point to `https://prod.foo.redhat.com:1337/beta/insights/image-builder` which I think makes the most sense.

teg: Did you intend to push a new version? :)

jkozol: hehe you are correct :)

teg (approved):

PR #265, opened 2021-07-13
CreateImageWizard: Only show Registration section in Review if applic…

### LINKED ISSUE #252: DDF Registration step shows up under review even if there's no registration
image

ochosi:
…able

Closes #252

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ReviewStep.js
+++ b/src/Components/CreateImageWizard/formComponents/ReviewStep.js
@@ -39,20 +39,22 @@ const ReviewStep = () => {
                     </TextListItem>
                 </TextList>
             </>}
-            <Text component={ TextVariants.h3 }>Registration</Text>
-            <TextList component={ TextListVariants.dl } data-testid='review-image-registration'>
-                <TextListItem component={ TextListItemVariants.dt }>Subscription</TextListItem>
-                <TextListItem component={ TextListItemVariants.dd }>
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -39,20 +39,22 @@ const ReviewStep = () => {
...
-                        'Register the system on first boot' :
-                    registerValues?.[getState()?.values?.['register-system']?.title]
-                    }
-                </TextListItem>
-                <TextListItem component={ TextListItemVariants.dt }>Activation key</TextListItem>
-                <TextListItem component={ TextListItemVariants.dd } type="password">
-                    {'*'.repeat(getState()?.values?.['subscription-activation']?.length)}
-                </TextListItem>
-            </TextList>
+            {getState()?.values?.['register-system'] === 'subscribe-now-radio' && <>
```
Gundersanne, line 42: Also worth checking the `release` value to see if it contains `rhel`.

Gundersanne (approved):
Thank you! This only partially fixes the problem. You can still get it into a state where it shows the subscription without needing it to by first filling it out to the end for rhel, and then going back and changing the image output-release field to centos.
image

I'm ok with merging this already if you prefer however, as it is a marked improvement.

Gundersanne:
Let's address this in a followup. Might do a release later today and this should be in it.

PR #276, opened 2021-07-21
Fix review page

ochosi:
Previously one could trick the state machine into showing the subscription section by first filling in the subscription information, then going back to the image type and changing it to CentOS.

Now we check if "rhel" is part of the image type (to hopefully also cover future RHEL offerings).
This may have to be extended in case RHEL for Edge or some other offering doesn't contain the "rhel" string.


Gundersanne (approved):
Thanks!

PR #278, opened 2021-07-21
Align all values on Review step

### LINKED ISSUE #194: CreateImageWizard: Align all values on the review step
image

ochosi:
Closes #194 

Please note that this is just a minimal adjustment to the review page layout. I would personally vote for making the subsections in the "Target environment" section stand out more (are we actually bound to using h3 there or can we bump everything by one level and use h3 for the environment titles themselves?).

I can do so in a follow-up PR or even in this one if you like.
image


Gundersanne (approved):
Very nice! Thank you :)

PR #281, opened 2021-08-04
Fix deployments

Gundersanne:
TODO:
- [x] make sure the devel container still works.


Gundersanne:
I ran it locally with both podman-compose and docker-compose, so should be good I think :)

Gundersanne:
@ochosi if you wanna check this out if it's easier to run it now. If you feel like trying out the compose file in the devel/ dir that would be great too :)

ochosi:
Nice work - the deployment works neatly now! (tested with proxy 3.2.2)

teg:
Looks nice!

Regarding the dockerfile: I think this will make the caching worse, if I understand correctly: the npm install will now be re run on every change to the working directory

Gundersanne:
> Looks nice!
> 
> Regarding the dockerfile: I think this will make the caching worse, if I understand correctly: the npm install will now be re run on every change to the working directory

Ah! I didn't think of that, i'll revert that change

teg (approved):

PR #285, opened 2021-08-09
devel: Privileged and cap_add are mutually exclusive

Gundersanne:


teg (approved):
I guess we should be able to drop `privileged` if we set the right `cap_add` things? But this is just for development so 🤷

Gundersanne:
Ah I just wanted to close this >< There's also some issue with selinux labels/relabeling of volumes, I think the ovlume mounts need `:z` depending on policy.

I'm ashamed to admit i dont actually have selinux running, but i'll revisit this in a fedora vm or something later, as people have still been running into issues.

PR #302, opened 2021-09-03
CreateImageWizard: Use rhel minor versions everywhere

Gundersanne:
Fallout from switching to minor versions everywhere.


teg (approved):

teg (approved):

teg (approved):

ochosi (approved):
Looks good to me!
Also checked and the conditional for the review step that I added (registration) is still fine with this :)

Issue #350, opened 2021-10-18
Change azure authentication redirect uri


jkozol:
Currently the azure redirect URI looks like:

 https://login.microsoftonline.com/<target-tenant-id>/oauth2/v2.0/authorize?client_id=b94bb246-b02c-4985-9c22-d44e66f657f4&scope=openid&response_type=code&response_mode=form_post&redirect_uri=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fnativeclient

The redirect uri should instead point to https://portal.azure.com

PR #351, opened 2021-10-18
CreateImageWizard: update azure redirect

jkozol:
On authorization of image builder on azure, the user will be redirected to portal.azure.com. The response mode is changed to query instead of post.

Fixes: #350


teg:
Probably should say Fixes #350 :)

teg (approved):
Thanks!

Gundersanne (approved):

PR #396, opened 2021-11-02
CreateImageWizard: sort packages by relevance

### LINKED ISSUE #256: packages: ordering search results
A big improvement can be made here. When searching for `vi` for instance, the packages get listed in alphabetical order, and anything that contains `vi` gets listed.

1. List exact matches first
2. Then list packages that start with the entered letters

And I think case insensitive would be best here.

kingsleyzissou:
Sort the package results in the CreateImageWizard first by exact matches and then by package results
that start with the same letters as the search term. Fixes #256

Before (search term: "libvirt"):

before_packages

After:
after_packages

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Packages.js
+++ b/src/Components/CreateImageWizard/formComponents/Packages.js
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect, } from 'react';
+import React, { useState, useRef, useEffect, useCallback } from 'react';
 import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
 import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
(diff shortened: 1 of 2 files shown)


jkozol (changes requested):
This works great when sorting on package load. But, if I move packages to the chosen list and then back to the available list they are added to the bottom of the list. Could you also add this sort to the moveSelected and moveAll functions when packages are added back to the available packages list?

kingsleyzissou:
Thanks for the review. Yeah of course I can add the sort to those functions

kingsleyzissou:
@jkozol I added in some tests to check that now too and improved the sort a little bit, because it wasn't really sorting the relevant results alphabetically.

jkozol (approved):
Thanks for this. Looks good to me!! Sorting works great :)

PR #422, opened 2021-11-11
constants: Introduce RHEL_8 constant

Gundersanne:
Bump rhel-84 to rhel-85.

---

Ideally this would be part of #6 ; where we'd pick the latest rhel version that hte service supports dynamically, but this takes priority.


teg (commented):
Thanks for this!

Gundersanne:
@teg actually not correct, pushing a fix

teg:
Notice my foresight in realising I was not competent enough to approve.

jkozol (approved):
I cannot currently verify a successful image build but this looks correct to me and the POSTed compose object is as expected

PR #436, opened 2021-11-18
Update React Router to v6

jkozol:
This required first updating to v5. This was based on the updates made to the insights starter app. Beyond that the React Router's v6 upgrade guide was followed: https://reactrouter.com/docs/en/v6/upgrading/v5


Gundersanne (approved):
works nicely, thank you!

PR #450, opened 2021-11-20
CreateImageWizard: fix Enter key for packages search

### LINKED ISSUE #448: packages search enter button doesn't work

jkozol:
The addEventListener call needs to have the useCapture field set to true. The data-driven-forms patternfly 4 wizard will continue to the next step when the Enter key is pressed. If we add our keydownHandler as an event listener with useCapture set to True, the keydownHandler will trigger before the wizards handler. This will effectively override the wizard's handling of the Enter key so that it can be used to search when the user has either package search bar in focus.

Fixes #448 

See: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener


teg (approved):
Sounds good. Thanks!

PR #454, opened 2021-11-21
CreateImageWizard: fix add all packages

### LINKED ISSUE #447: packages "add all" button doesn't work
the double arrow `»` which adds all packages from the left hand column doesn't work, 0 packages chosen on the review screen.

jkozol:
When adding all packages we no longer depend on the state update. Instead, we initialize an object with the chosen packages and pass that to both the step's state and the form's state. Using the step state to set the form's state was causing the form's state to be set from the prior step state and not the state with the new package list.

Fixes #447


Gundersanne (approved):
works!

PR #456, opened 2021-11-22
CreateImageWizard: don't obscure activation key

chloenayon:
Activation key was previously obscured by stars, instead just show it
(both in the input and in the review step) as it isn't secret.


jkozol:
Closed in favor of #458

PR #467, opened 2021-11-25
CreateImageWizard: Show "no package results are found" when no packages are found

### LINKED ISSUE #463: Show "no package results are found" when no packages are found
Searching for something which does not exists as, doesn't show that there are no results.

image

KKoukiou:
Fixes #463


jkozol:
@KKoukiou This looks great! But, could you add the same for the `packagesChosen` list? If you'd prefer we can also do this in a followup since this does work fine as is.

Gundersanne (approved):
Ah thank you so much :3 I'll copy you for the right-hand column <3

PR #477, opened 2021-11-29
CreateImageWizard: Fix move selected packages

Gundersanne:
If fromAvailable is false, packagesavailable can be undefined if it's
the first package that's being moved over.


Gundersanne:
Broken after rebasing on top of #467

Gundersanne:
Closed in favour of #476

PR #475, opened 2021-11-29
CreateImageWizard: only display upload review for selected destinations

jkozol:
Fixes #460

PR #479, opened 2021-11-29
CreateImageWizard: display message for invalid chosen packages filter

jkozol:
Testing this turned out to be a little more complicated than I would have liked so I will append the tests later.

### DIFF
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
@@ -49,10 +49,28 @@ const mockPkgResult = {
     ]
 };
 
-const searchForPackages = async (searchbox, searchTerm) => {
+const mockPkgResultEmpty = {
+    meta: { count: 0 },
+    links: { first: '', last: '' },
+    data: null
+};
+
+const searchForAvailablePackages = async (searchbox, searchTerm) => {
     userEvent.type(searchbox, searchTerm);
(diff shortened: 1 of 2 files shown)


jkozol:
Rebased on top of #475

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -558,6 +576,41 @@ describe('Step Packages', () => {
...
+        await searchForAvailablePackages(searchboxAvailable, 'test');
+        expect(getPackages).toHaveBeenCalledTimes(1);
+
+        screen.getByRole('button', { name: /Add all/ }).click();
+
+        searchboxChosen.click();
+        await searchForChosenPackages(searchboxChosen, 'asdf');
+        screen.getByText('No packages found');
+        // We need to clear this input in order to not have sideeffects on other tests
+        await searchForChosenPackages(searchboxChosen, '');
```
Gundersanne, line 612: Can we move this to the beforeEach?

jkozol: Specifically 612? or the whole block? beforeEach wouldn't work because it is also effecting other test groups/`describe`. I could put it in a afterEach but it is only used once and this side effect should hopefully disappear when I rewrite how we render the components as part of the eslint update

Gundersanne (approved):

PR #507, opened 2021-12-16
Refactor tests

jkozol:
The goal of this PR is to improve the style of our tests per the rules from eslint's plugin for the testing library. This PR includes a version bump for eslint-plugin-testing-library from 3.10.2 to 5.0.1. There were many error types so I devoted one commit to each error class. If you would like I can squash some of them together but it was easiest for me to do this one error type at a time.

### DIFF
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
@@ -355,17 +355,13 @@ describe('Step Registration', () => {
         const registrationRadio = screen.getByLabelText('Embed an activation key and register systems on first boot');
         userEvent.click(registrationRadio);
 
-        await waitFor(() => {
-            const organizationId = screen.getByLabelText('Organization ID');
-            expect(organizationId).toHaveValue('5');
-            expect(organizationId).toBeDisabled();
-            // can't getByLabelText b/c the label contains an extra <span>
-            // with a `*` to denote required
-            const activationKey = screen.getByTestId('subscription-activation');
-            expect(activationKey).toHaveValue('');
-            expect(activationKey).toBeEnabled();
-            // expect(activationKey).toBeRequired(); DDF does not support required fields
-        });
+        const organizationId = await screen.findByLabelText('Organization ID');
+        expect(organizationId).toHaveValue('5');
+        expect(organizationId).toBeDisabled();
+
+        const activationKey = await screen.findByTestId('subscription-activation');
+        expect(activationKey).toHaveValue('');
+        expect(activationKey).toBeEnabled();
 
         userEvent.type(screen.getByTestId('subscription-activation'), '012345678901');
         screen.getByRole('button', { name: /Next/ }).click();
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
@@ -336,9 +336,7 @@ describe('Step Registration', () => {
     beforeEach(async() => {
         window.HTMLElement.prototype.scrollTo = function() {};
 
-        await act(async () => {
-            history = renderWithReduxRouter(<CreateImageWizard />).history;
-        });
+        history = renderWithReduxRouter(<CreateImageWizard />).history;
 
         // select aws as upload destination
         const awsTile = screen.getByTestId('upload-aws');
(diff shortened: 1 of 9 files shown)


src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -692,7 +672,7 @@ describe('Step Review', () => {
             name: 'Amazon Web Services'
         });
         userEvent.click(buttonRegistration);
-        screen.getByText('Register the system on first boot');
+        screen.getByText('Register the system later');
```
Gundersanne, line 745: This does disable testing the subscription tab on the review page when adding one, no?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -598,21 +603,30 @@ describe('Step Packages', () => {
...
         expect(getPackages).toHaveBeenCalledTimes(2);
-        expect(availablePackages.children.length).toEqual(2);
-        expect(chosenPackages.children.length).toEqual(1);
+
+        availablePackagesItems = within(availablePackagesList).getAllByRole('option');
+        chosenPackagesItems = within(chosenPackagesList).getAllByRole('option');
+        expect(availablePackagesItems).toHaveLength(2);
+        // Knowing if it is in document isn't enough. We want a specific length of 1 so ignore rule.
+        // eslint-disable-next-line jest-dom/prefer-in-document
+        expect(chosenPackagesItems).toHaveLength(1);
```
Gundersanne, line 685: I guess better still would be to look for the package name in the chosen/available package list. This preserves current coverage so it's ok though.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -355,17 +355,21 @@ describe('Step Registration', () => {
...
-            const organizationId = screen.getByLabelText('Organization ID');
-            expect(organizationId).toHaveValue('5');
-            expect(organizationId).toBeDisabled();
             // can't getByLabelText b/c the label contains an extra <span>
             // with a `*` to denote required
-            const activationKey = screen.getByTestId('subscription-activation');
-            expect(activationKey).toHaveValue('');
-            expect(activationKey).toBeEnabled();
-            // expect(activationKey).toBeRequired(); DDF does not support required fields
+            activationKey = screen.getByTestId('subscription-activation');
```
Gundersanne, line 397: Could replace this with await screen.findByTestId?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -355,17 +355,21 @@ describe('Step Registration', () => {
         const registrationRadio = screen.getByLabelText('Embed an activation key and register systems on first boot');
         userEvent.click(registrationRadio);
 
+        let organizationId;
+        await waitFor(() => {
+            organizationId = screen.getByLabelText('Organization ID');
```
Gundersanne, line 388: Could replace with await screen.findByLabelText?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -107,11 +107,8 @@ afterAll(() => {
 });
 
 describe('Create Image Wizard', () => {
-    beforeEach(async () => {
```
Gundersanne, line 107: https://jestjs.io/docs/api#beforeeachfn-timeout beforeEach should be fine to use as a setup no? `beforeAll` isn't used.

Gundersanne (commented):
Looks good! A few questions.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
jkozol: So, according to [the docs](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-render-in-setup.md), it is best to not render in the beforeEach. Figuring out why this is a rule is a little tricky and I would also be fine setting a rule to ignore this case. From what I've read we shouldn't separate the render from our tests. The render call should be made in each test as it is when calling `setUp` now. This supposedly helps with cleanup and, overall, this rule is meant to also encourage writing fewer but longer tests.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
jkozol: I'm not completely sure what your question is. I tried to switch the default selection for registration to no registration since it is simpler to test and for the tests that aren't directly checking registration I think it is better for it to be simpler. The subscription tab displays regardless of what is selected in the step and we test the registration key in the `should allow choosing activation keys` test

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
jkozol: Good point. I now also check for the package name within the chosen packages list.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
jkozol: Yep! Updated.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
jkozol: Yep! Updated.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
Gundersanne: Oh I missed line 800 too, it thought the first boot option wasn't tested now, sorry for the confusion

Gundersanne (approved):
lgtm!

PR #532, opened 2022-01-12
CreateImageWizard: conditionally render Registration tab

chloenayon:
At the final review step, only render the Registration tab if the user has selected RHEL 8 as the release.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ReviewStep.js
+++ b/src/Components/CreateImageWizard/formComponents/ReviewStep.js
@@ -125,42 +125,43 @@ const ReviewStep = () => {
                         }
                     </List>
                 </Tab>
-                <Tab eventKey={ 1 } title={ <TabTitleText>Registration</TabTitleText> } data-testid='tab-registration'>
-                    {getState()?.values?.['register-system'] === 'register-later-radio-button' &&
-                        <TextContent>
-                            <TextList component={ TextListVariants.dl }>
-                                <TextListItem component={ TextListItemVariants.dt }>
-                                    Subscription
-                                </TextListItem>
-                                <TextListItem component={ TextListItemVariants.dd }>
-                                    Register the system later
-                                </TextListItem>
-                            </TextList>
-                        </TextContent>
-                    }
-                    {getState()?.values?.['register-system'] === 'subscribe-now-radio' &&
-                    getState()?.values?.release.includes('rhel') &&
-                        <TextContent>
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -98,6 +98,7 @@ const ReviewStep = () => {
                         }
                     </List>
                 </Tab>
+                {releaseValues?.[getState()?.values?.release] === 'Red Hat Enterprise Linux (RHEL) 8' &&
```
jkozol, line 101: Nice start! I would recommend something more similar to line 118. 

`getState()?.values?.release.includes('rhel')`

This allows us to more generally check if we are building a rhel system or not.

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -98,6 +98,7 @@ const ReviewStep = () => {
                         }
                     </List>
                 </Tab>
+                {getState()?.values?.release.includes('rhel') === 'Red Hat Enterprise Linux (RHEL) 8' &&
```
jkozol, line 101: This will always evaluate to False. [.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) returns a boolean. So testing if it equals 'Red Hat Enterprise Linux (RHEL) 8' will always return false.

jkozol:
Also, please remove `getState()?.values?.release.includes('rhel') &&` from line 116 since it is now duplicated.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -741,6 +768,22 @@ describe('Step Review', () => {
             name: 'Packages'
         });
     });
+
+    test('has two tabs for centos', async () => {
+        setUpCentOS();
+
+        const buttonTarget = screen.getByTestId('tab-target');
+        const buttonSystem = screen.getByTestId('tab-system');
+
```
jkozol, line 777: Could you add a non-existence check here? A good way to do that is you could query by testid using the react testing library and then check that it is not in the document using jest-dom.

jkozol (changes requested):
The test is looking better. It looks like you dropped the original commit while rebasing. Could you cherry-pick it back in?

jkozol (approved):
Thanks for the updates! Looks great :)

PR #535, opened 2022-01-12
Add new image types

jkozol:
The image types `vsphere`, `guest-image`, and `image-installer` are now supported by the CreateImageWizard. They require no additional customizations and support the current registration and packages customizations.

Screenshot from 2022-01-12 19-33-26

newImageTypes


Gundersanne (changes requested):
Thanks!

Just one request, for every image type we should in the wizard we should add a test in `CreateImageWizard.test.js`  under `Click through all steps` ; and then make sure the generated request matches our expectation.

jkozol:
@Gundersanne testing is added :)

Gundersanne (approved):
Thanks!

PR #543, opened 2022-01-19
Registration step improvement

jkozol:
Update the registration options to include registering without insights and improve the design. The organization id is moved to the review step.

Screenshot from 2022-01-19 21-47-14
Screenshot from 2022-01-19 21-46-54
Screenshot from 2022-01-19 21-46-39


jkozol:
I would like to add a test to also check for insights being disabled but I am going to include that in a separate work on testing more possible image creations. #545 545

Gundersanne (approved):
Ah very nice!

PR #571, opened 2022-02-11
Fetch activation keys

jkozol:
Opening this as a a draft PR because I am not sure of the best testing strategy here. I load the activation keys by an async api call. In our tests we need to mock this call as I have done in the beforeAll function. Also, we need to await on the registration step to make sure the promise has resolved before continuing. So, any test that navigates to the Registration page must be async. @Gundersanne what are your thoughts on this? Do you see a better way to test this?

This commit: https://github.com/RedHatInsights/image-builder-frontend/pull/571/commits/a50fb2126816bf1e4be8d1ffda124e95a4cc4f1c shows best what I am talking about.


Gundersanne:
I think it's ok if the test skips registration in all cases except the 'Step Registration' and 'with valid values' tests.

And there I'd wait until one of the options is available. In case of the 'with valid values' test I'd select an option and verify it as it's done currently by comparing the request.

There's no problem with tests being async, as jest will just await it.

jkozol:
@Gundersanne So I don't believe we can skip the test registration step unless we switch to testing a CentOS image. We test the Wizard as a whole so have to click through the tests. Any navigation to the Registration step triggers the api call. Any reason to not by default test against CentOS?

jkozol:
I can also just switch all the tests that require navigating past the registration page over to async but it feels a little weird to do this just to mock an api call the tests don't test.

Gundersanne:
> @Gundersanne So I don't believe we can skip the test registration step unless we switch to testing a CentOS image. We test the Wizard as a whole so have to click through the tests. Any navigation to the Registration step triggers the api call. Any reason to not by default test against CentOS?

I'd just not wait on the api call if it's not the registration step and click the register later radio button. Or will those buttons only render once the api call has been complete?

jkozol:
So we don't have to wait on the api call but we get a bunch of error messages about `Warning: An update to ActivationKeys inside a test was not wrapped in act(...).` I believe this is because the component updates after our test completes since the promise will return at some point. So any time we navigate to the registration page we will need to `await`. 

https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

jkozol:
Screenshot from 2022-02-09 19-02-29
Screenshot from 2022-02-09 19-02-17

jkozol:
Screenshot from 2022-02-11 13-39-38
Screenshot from 2022-02-11 13-39-18

src/Components/CreateImageWizard/formComponents/ActivationKeys.js
```diff
@@ -0,0 +1,70 @@
...
+                typeAheadAriaLabel="Select activation key">
+                {isLoading &&
+                    <SelectOption isNoResultsOption={ true } data-testid='activation-keys-loading'>
+                        <Spinner isSVG size="lg" />
+                    </SelectOption>
+                }
+                {activationKeys.map((key, index) => (
+                    <SelectOption
+                        key={ index }
+                        value={ key.id }
```
Gundersanne, line 53: I'd actually make the key.name the value and the description the id. Not sure the id is even exposed on access.

jkozol: I was unsure about that. So if a user was to type you think they would type in the descriptive value?

teg: Agreed. I don't think the id is ever used except internally by RH, right? I.e., the thing we embed in our images is the name, not the ID, so the customers would not see them? (not sure I understand how that makes sense, but that's another matter)

jkozol: Okay cool. Changed.

jkozol: I switched them but should we remove the id completely?

teg: Hm... Maybe? If it is not shown anywhere else, I would. Are the names forced to be unique?

jkozol:
Screenshot from 2022-02-11 18-55-14

src/Components/CreateImageWizard/formComponents/ActivationKeys.js
jkozol: Only displaying the names now.

jkozol: Per this mock: https://marvelapp.com/prototype/6g7dja9/screen/76602546

src/constants.js
```diff
@@ -1,2 +1,3 @@
 export const IMAGE_BUILDER_API = '/api/image-builder/v1';
+export const INISGHTS_API = '/api/rhsm/v2';
```
Gundersanne, line 2: Can we rename this to RHSM_API? 'insights' is a little vague.

src/Components/CreateImageWizard/steps/registration.js
```diff
@@ -79,7 +77,7 @@ export default {
                 {
                     type: validatorTypes.REQUIRED,
                 },
-            ],
```
Gundersanne, line 82: nitpick, trailing comma diff. I prefer with trailing commas whenever possibl.

Gundersanne (changes requested):
Looking good, one last comment.

src/Components/CreateImageWizard/steps/registration.js
jkozol: fixed

Gundersanne (approved):

PR #577, opened 2022-02-14
Filesystem customization

Gundersanne:


Gundersanne:
Test still needs to be added for draggable rows if we want to keep those.

Gundersanne:
I removed the popover on the toggle page, as automatic partitioning might not result in image-builder using lvm. For any manual configuration LVM is used so the lvm popover there can stay.

teg:
LVM support hasn't actually landed yet, so maybe hold that line from the UI? Ideally we would have that "real soon now", but don't think it is worth blocking the UI on it.

jkozol (commented):
This is really impressive. Thanks for putting all the work into getting this table working with the drag! I noticed a couple bugs when playing around with it. I am happy to have them as a followup since this is already quite full and they are minor. Once the lvm text is dropped I'm happy to merge this.

jkozol:
Here the bugs I noticed:

If I create several duplicate entries, drag an entry, and delete any entry, the duplicate alert goes away.

Screenshot from 2022-02-25 21-33-54

The review screen "view partitions" popover should be larger to fit minimum size.

Screenshot from 2022-02-25 21-31-08

Also, what happens when a user enters a strange partitioning scheme? Such as having no root directory. Would image-builder create the image with necessary partitions not to the specs of the user? If so we should have a message stating that?

For instance here a user would create a partition table with no partitions:

Screenshot from 2022-02-25 21-59-37

Gundersanne:
Removed the LVM references. Fixed some rendering issues (added keys to mountpoint & sizeunit components), added a validation for '/', and next is disabled if no partitions are defined.

Gundersanne:
> LGTM! The one question I have is should we include a message telling the user that the mount order might not be the same as they have placed?

Definitely something to discuss, put it on the agenda.

PR #589, opened 2022-02-22
CreateImageWizard: Get all matching packages from package search

lucasgarfield:
When searching for a package, all matching packages are now returned. First an
attempt is made using the api's default limit and if there are more
matching packages than the default limit a second request is made with
an increased limit. To facilitate this, api.getPackages() now accepts an
optional limit parameter. Retrieving all matching packages is necessary
because of the sorting logic.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Packages.js
+++ b/src/Components/CreateImageWizard/formComponents/Packages.js
@@ -43,31 +43,34 @@ const Packages = ({ defaultArch, ...props }) => {
 
     const searchResultsComparator = useCallback((searchTerm) => {
         return (a, b) => {
+            a = a.name.toLowerCase();
+            b = b.name.toLowerCase();
+
             // check exact match first
-            if (a.name === searchTerm) {
+            if (a === searchTerm) {
                 return -1;
             }
 
-            if (b.name === searchTerm) {
+            if (b === searchTerm) {
                 return 1;
             }
 
             // check for packages that start with the search term
-            if (a.name.startsWith(searchTerm) && !b.name.startsWith(searchTerm)) {
+            if (a.startsWith(searchTerm) && !b.startsWith(searchTerm)) {
                 return -1;
             }
 
-            if (b.name.startsWith(searchTerm) && !a.name.startsWith(searchTerm)) {
+            if (b.startsWith(searchTerm) && !a.startsWith(searchTerm)) {
                 return 1;
             }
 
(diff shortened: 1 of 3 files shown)


src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -768,6 +790,30 @@ describe('Step Packages', () => {
...
+            .mockImplementation(() => Promise.resolve(mockPkgResultAlpha));
+
+        await searchForAvailablePackages(searchbox, 'test');
+        expect(getPackages).toHaveBeenCalledTimes(1);
+
+        const availablePackagesList = screen.getByTestId('available-pkgs-list');
+        const availablePackagesItems = within(availablePackagesList).getAllByRole('option');
+        expect(availablePackagesItems).toHaveLength(3);
+        const [ firstItem, secondItem, thirdItem ] = availablePackagesItems;
+        //expect(firstItem).toHaveTextContent('lib-testlib-test package summary');
```
Gundersanne, line 835: rogue comment

Gundersanne (commented):
This is looking very nice! Would be happy to merge as is tbh. It's definitely big enough so wouldn't push any more commits on top, de0a7972f71b5392cf2e18c6434c39340638127b can even be split out to a separate PR.

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -88,18 +91,30 @@ const Packages = ({ defaultArch, ...props }) => {
...
+        const args = [
             getState()?.values?.release,
             getState()?.values?.architecture || defaultArch,
             packagesSearchName.current
-        );
-        if (data) {
-            const packagesAvailableFiltered = filterPackagesAvailable(data);
+        ];
+        let { data, meta } = await api.getPackages(...args);
+        if (data && meta.count === data.length) {
```
jkozol, line 101: Thanks to [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) you don't need to check if data exists. You can just do `data?.length`

jkozol (approved):
Aside from @Gundersanne's comment this looks great. Thanks for adding the tests and good commit messages. My change is very minor and isn't needed for merge. Just a nice way to dig into objects that might be defined.

jkozol (approved):
Wonderful work here!

PR #626, opened 2022-02-28
Add details step with Image Name field

jkozol:
A user can now enter an optional custom image name. If the user enters a name it is displayed in the table, otherwise, the uuid is displayed.

Screenshot from 2022-02-28 22-20-56
Screenshot from 2022-02-28 22-20-36

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -19,6 +19,7 @@ import {
     packages,
     imageOutput,
     fileSystemConfiguration,
+    details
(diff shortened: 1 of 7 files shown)


teg:
The description says "user name", which caused me pause ;)

That aside, this looks really great! I assume the image names will be shown in the overview of built images too?

jkozol:
@teg oops haha. This is why I shouldn't PR when tired.

Here is a screen of the Images Table:

Screenshot from 2022-03-01 13-12-37

Gundersanne (commented):
Should we have a default name? Like $distro_$date ?

Issue #592, opened 2022-02-22
Header question mark


jkozol:
Space between question mark and title really tight

![image15](https://user-images.githubusercontent.com/11712857/155201067-a227a4ea-ef9b-401d-aff2-e9e1ebfe82e0.png)

PR #631, opened 2022-03-01
LandingPage: tight space between header and icon

chloenayon:


jkozol:
Okay. So here is my process to figuring this out.

1. Look at what we do in other popovers.
2. I notice that the other popovers set the className of `pf-c-form__group-label-help`
3. Inspect this element for css properties. I do this using the browser html/css console inspector. But, this can also be found in the patternfly documentation.
4. While inspecting I notice that the padding for this class is set by: `--pf-c-form__group-label-help--PaddingLeft`
5. According to the documentation for the form group label found [here](https://www.patternfly.org/v4/components/form/#css-variables). This corresponds to patternfly's `--pf-global--spacer--xs` css variable.
6. I read patternfly's spacing utility's documentation [here](https://www.patternfly.org/v4/utilities/spacing).
7. This shows me that I can adding a className of `pf-u-pl-xs` to get a left padding that would be equal to what we set for other popovers.
8. I visually review this and compare it to the designs. The popover is shown on several design slides such as [this one](https://marvelapp.com/prototype/6g7dja9/screen/83986025).
9. I notice that the spacing is still too small for the designs so I change the spacer from `xs` to `sm`.
10. This may still not be the correct spacing so I will @ katie with a screenshot on the PR.

src/Components/LandingPage/LandingPage.js
```diff
@@ -51,12 +51,12 @@ https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/up
                                     Contribute on GitHub
                             </Button>
                         </TextContent> }>
-                        <button
-                            type="button"
+                        <Button
+                            variant="plain"
                             aria-label="About image builder"
-                            className="pf-c-form__group-label-help">
+                            className="pf-c-form__group-label-help" >
```
jkozol, line 57:
```suggestion
                            className="pf-u-pl-sm" >
```

jkozol (changes requested):

jkozol (approved):
lgtm!

jkozol:
Fixes: #592

PR #636, opened 2022-03-03
Spur/image output instructions

### LINKED ISSUE #596: Add image output instructions
https://marvelapp.com/prototype/6g7dja9/screen/80928504

image21

lucasgarfield:
Resolves #596

Got a little ahead of myself with what I thought would be a simple fix... I realized we use the documentation button a lot. Four times currently and it is possible to imagine more in the future. It's a lot of code for a simple button so I refactored it into its own component; this helps dry the code and improves the form schema readability. Should also make maintaining it easier in the future if the URL changes, and if we need separate documentation URLs at some point in the future it should be easier to add that functionality as well.

Couple of questions:
1. Do we have any 'standard' for our imports? They don't seem to be in any order... If chaos rules, that's fine, just want to check and make sure.
2. If we like the idea of having this as a component, where should it live? Right now it is in CreateImageWizard/formComponents but I would also like to use it in the ImagesTable and LandingPage as well. Is it okay for ImagesTable and LandingPage to import form components from CreateImageWizard or should the component live in a higher level directory? Maybe we need to make a directory for shared components?

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -2,8 +2,7 @@ import React from 'react';
 import ImageCreator from './ImageCreator';
 import { useNavigate } from 'react-router-dom';
 import componentTypes from '@data-driven-forms/react-form-renderer/component-types';
-import { Button } from '@patternfly/react-core';
-import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import DocumentationButton from '../sharedComponents/DocumentationButton';
 import './CreateImageWizard.scss';
 import { useDispatch } from 'react-redux';
 import api from '../../api';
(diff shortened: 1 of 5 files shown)


jkozol (dismissed):
Great work on this. 

1. Chaos rules in an semi organized fashion. I usually put React imports first (i.e react router), then library imports (i.e. patternfly), then imports from within our own project (i.e. .scss file)

2. I like the idea of having the DocumentationButton in a top level Components folder since it is shared by the Wizard, Table, and Landing Page. Feel free to name it as you think makes sense.

Feel free to do 2. in this PR or in another one.

lucasgarfield:
I named the top level components folder sharedFormComponents, I'm not sure if it necessarily has to include *form* components but that naming keeps it consistent with /CreateImageWizard/formComponents so it made sense to me.

I think we also need to have a discussion about formalizing Image Builder's description, I also noticed that in the popover it references making RHEL images... I didn't fix that with this PR because it is still in line with the UI prototypes. I'll make a separate issue for that discussion, let's just merge this one in after review.

jkozol (changes requested):
Can you name the folder to just `sharedComponents`. Only the Wizard is a form so these components aren't necessarily form components.

lucasgarfield:
Good point! Done.

Issue #617, opened 2022-02-25
Tab title should be Image Builder not image_builder


jkozol:
![Screenshot from 2022-02-25 22-01-10](https://user-images.githubusercontent.com/11712857/155802427-ec4790d8-1c46-4aa0-9e64-5794d57003e9.png)

PR #640, opened 2022-03-07
App: fix tab title

kingsleyzissou:
The tab title for image builder in crc was being diplayed as `image_builder`.
This commit fixes issue #617

tab_title


jkozol (approved):
Perf! Thanks!

PR #641, opened 2022-03-07
Recreate image builds

jkozol:
This still needs some updates so opening as a draft.

Particularly the package description and filesystem size need work. Also, I am a little stuck on how to get the nav bar to behave when starting at the review step. Finally, the action kebab has too much padding.

Any thoughts?

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import ImageCreator from './ImageCreator';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useLocation } from 'react-router-dom';
 import componentTypes from '@data-driven-forms/react-form-renderer/component-types';
 import DocumentationButton from '../sharedComponents/DocumentationButton';
 import './CreateImageWizard.scss';
@@ -198,6 +198,7 @@ const onSave = (values) => {
 const CreateImageWizard = () => {
     const dispatch = useDispatch();
     const navigate = useNavigate();
+    const location = useLocation();
     return <ImageCreator
         onClose={ () => navigate('/') }
         onSubmit={ ({ values, setIsSaving }) => {
@@ -261,7 +262,8 @@ const CreateImageWizard = () => {
                     ]
                 }
             ]
-        } } />;
+        } }
+        initialValues={ {} } />;
 };
 
 export default CreateImageWizard;
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -79,7 +79,7 @@ const onSave = (values) => {
         requests.push(request);
     }
 
-    if (values['target-environment']?.google) {
+    if (values['target-environment']?.gcp) {
         let share = '';
         switch (values['google-account-type']) {
             case 'googleAccount':
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -6,6 +6,7 @@ import DocumentationButton from '../sharedComponents/DocumentationButton';
 import './CreateImageWizard.scss';
 import { useDispatch } from 'react-redux';
 import api from '../../api';
+import { UNIT_KIB, UNIT_MIB, UNIT_GIB } from '../../constants';
 import { composeAdded } from '../../store/actions/actions';
 import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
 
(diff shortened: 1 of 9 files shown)


src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -195,9 +195,118 @@ const onSave = (values) => {
...
+                case 'domain':
+                    formState['google-account-type'] = 'domain';
+                    formState['google-domain'] = account;
+                    break;
+            }
+        }
+
+        // customizations
+        // packages
+        // TODO: add package description
```
Gundersanne, line 261: Could either omit the description in this case or we need to make a /packages call for each package and see what the description is.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -195,9 +195,118 @@ const onSave = (values) => {
...
+        formState['selected-packages'] = composeRequest?.customizations?.packages?.map(packageName => {
+            return ({
+                name: packageName
+            });
+        });
+
+        // filesystem
+        // TODO: intelligent unit selection
```
Gundersanne, line 269: https://github.com/RedHatInsights/image-builder-frontend/commit/6aed45346eb495e85223029cb48eab1099a4ad43#diff-dbd247166ab10488d571ec72c6213aab6fe73f84f43538a34c4e972ce418b738R24-R36 could do something similar to this?

Gundersanne:
Regarding the nav, is it possible to change the wizardcontext https://data-driven-forms.org/provided-mappers/wizard?mapper=mui#wizardcontext to get it to mark all steps as enabled?

src/Components/CreateImageWizard/CreateImageWizard.js
jkozol: Ya I was thinking of just adding a `fetchPackageDescription` call that will return the description for the specified package name.

jkozol:
@Gundersanne 

> Regarding the nav, is it possible to change the wizardcontext https://data-driven-forms.org/provided-mappers/wizard?mapper=mui#wizardcontext to get it to mark all steps as enabled?

Ya I think this is the best way as well. Sadly the documentation isn't the best and I haven't been able to figure out which of these we need and which index our review step would be :( I will keep playing around and mark this ready once I have a solution.

src/Components/CreateImageWizard/CreateImageWizard.js
jkozol: This is fixed.

src/Components/CreateImageWizard/CreateImageWizard.js
jkozol: The description is now fetched.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -195,9 +196,163 @@ const onSave = (values) => {
...
+
+        // customizations
+        // packages
+        formState['selected-packages'] = composeRequest?.customizations?.packages?.map(packageName => {
+            const packageDescription = getPackageDescription(composeRequest?.distribution, imageRequest?.architecture, packageName);
+            return ({
+                name: packageName,
+                description: packageDescription
+            });
+        });
```
jkozol, line 312: I have done something wrong here but am unsure what exactly to do. This returns an object of:

```
{
  "name": "openssh",
  "description": Promise {
    <state>: "fullfilled",
    <value>: "An open source implementation of SSH protocol version 2"
  }
}
```

I am unsure exactly why. I would expect the async/await function to resolve to the value since the state is "fullfilled". But, it doesn't. I thought about awaiting the `getPackageDescription` call but then this would end up requiring await/async all up the function chain. 

Any thoughts on what I've done wrong here?

Gundersanne: An async function always returns a promise. Even if it's resolved on return.

jkozol: Ok so turns out `.map` doesn't play with async as I thought it would. In order to resolve the "Promises" `forEach` worked better.

jkozol:
https://user-images.githubusercontent.com/11712857/159027814-c313dcd7-92ef-42f1-9616-5ca107e1d391.mp4

jkozol:
:man_facepalming:

Gundersanne (approved):
Very nice :)

katierik:
The interaction is 👍 , one note is that for the dropdown menu the text should be black, not blue.

jkozol:
@katierik Things changed since the screenshots. Here's a video:


https://user-images.githubusercontent.com/11712857/159773128-9c882c43-f161-4e24-939d-bca8acda95b3.mp4

PR #644, opened 2022-03-16
Images table hooks

jkozol:
The ImagesTable component is now a functional instead of class component. This is useful because it allows us to use react hooks in the component.

This is rebased on top of #643


lucasgarfield (approved):
Looks great, Jacob! I tested it locally and everything seems to work as it should. I am a total novice at Redux but as far as I can tell there is a one to one correspondence between the code that was removed and the code that was added so we should be good to go!

PR #652, opened 2022-03-23
Update style lint

jkozol:


Gundersanne:
Why is the style import no longer required? I think I just copied this over from https://github.com/RedHatInsights/frontend-starter-app ; which now imports style from -utilities.

jkozol:
I based that removal off of the documentation on the npm page: https://www.npmjs.com/package/@redhat-cloud-services/frontend-components

jkozol:
I have asked on the consoledot slack

Gundersanne:
Makes sense, I've never read the documentation :D

Gundersanne (approved):

jkozol:
@Gundersanne we don't need to import the styles anymore. Importing from utilities provides some sass utils methods but since we haven't needed those yet I don't see a reason to import them.

PR #668, opened 2022-03-24
Spur/keyboard accessibility

lucasgarfield:


src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
```diff
@@ -67,6 +75,7 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
                                 className='provider-icon'
                                 src={ '/apps/frontend-assets/partners-icons/microsoft-azure-short.svg' } /> }
                             onClick={ () => handleSetEnvironment('azure') }
+                            onKeyDown = { (e) => handleKeyDown(e, 'azure') }
```
jkozol, line 78: Can you also add the same the same on key handling for the checkbox. The tiles select with Space or Enter but the checkbox only selects with Space.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1399,3 +1399,127 @@ describe('Click through all steps', () => {
...
+    test('autofocus on each step first input element', async () => {
+        setUp();
+
+        // Details
+        const detailsInput = screen.getByRole('textbox', { name: /image name/i });
+        expect(detailsInput).toHaveFocus();
+        clickNext();
+
+        // Image output
+        // const releaseSelect = screen.getByRole('button', { name: /open menu/i });
```
jkozol, line 1442: Can you remove the commented tests?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1399,3 +1399,127 @@ describe('Click through all steps', () => {
...
+
+        // Registration
+        const registerRadio = screen.getByRole('radio', {
+            name: /register and connect image instances with red hat/i
+        });
+        expect(registerRadio).toHaveFocus();
+        clickNext();
+
+        // File system configuration
+        // const automaticPartioningButton = screen.getByRole('button', {
```
jkozol, line 1473: Here too.

jkozol (changes requested):
Overall looks good and works much nicer with the keyboard. Just a couple changes and then good to merge.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
lucasgarfield: Yes, definitely! I couldn't decide whether it was best to remove them or leave them as 'futures', to be un-commented when those steps do have autofocus. But they're so simple, it's probably better to just rewrite them when that time comes.

src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
lucasgarfield: It might actually be better to remove the ability to select the tiles with Enter and use Space for everything. Thoughts?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
jkozol: I'd actually create an issue in github to track it. The code/tests might change by the time you would implement these so better to track the need for them at a higher level (issues) since you may need to implement them differently anyways.

src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
lucasgarfield: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

It actually seems it is atypical for a checkbox to allow Enter, so I think going Space only might be best.

jkozol: Hahah was just reading the same. Yes lets go Space for everything. The tiles aren't visibly checkboxes but they function as checkbox.

jkozol: Those mozilla docs on accessibility are great :)

jkozol: https://webaim.org/techniques/keyboard/ is also helpful. University led accessibility project.

jkozol (dismissed):

jkozol (approved):

PR #669, opened 2022-03-24
CreateImageWizard: target env validation (fixes #575)

### LINKED ISSUE #575: Fix validation on image output page
Selecting and unselecting an image type will allow the user to progress with no image type selected.

kingsleyzissou:
Validation is not enforced on the image output step when selecting and then deselecting an image type. This
is because the only validation which is set is `Required`. However, after deselecting an image type the input becomes
a key-value object with all the values set to false. This commit adds a custom validator to check that at least one
value in the object is set to true.

Peek 2022-03-24 15-57


jkozol (changes requested):
Awesome! Works just as expected. Can you add a simple test just to make sure that unselecting the targets disables the next button.

jkozol (approved):
Looks good :)

PR #680, opened 2022-04-01
CreateImageWizard: Change valid prefixes to match composer

Gundersanne:
Adds all valid prefixes defined in composer, removes invalid ones, and
disables the TextInput for the root partition. This makes customized
filesystems less error prone.

https://bugzilla.redhat.com/show_bug.cgi?id=2070761

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/MountPoint.js
+++ b/src/Components/CreateImageWizard/formComponents/MountPoint.js
@@ -13,7 +13,7 @@ import path from 'path';
 
 const MountPoint = ({ ...props }) => {
     // check '/' last!
-    const validPrefixes = [ '/home', '/opt', '/srv', '/usr/local', '/var', '/boot', '/' ];
+    const validPrefixes = [ '/home', '/opt', '/srv', '/usr', '/usr/local', '/var', '/app', '/data', '/' ];
     const [ isOpen, setIsOpen ] = useState(false);
     const [ prefix, setPrefix ] = useState('/');
     const [ suffix, setSuffix ] = useState('');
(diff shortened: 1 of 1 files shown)


Gundersanne:
See https://github.com/osbuild/osbuild-composer/blob/main/internal/distro/rhel85/distro.go#L44 for the current list. I guess we should just make this track rhel GA.

src/Components/CreateImageWizard/formComponents/MountPoint.js
```diff
@@ -13,7 +13,7 @@ import path from 'path';
 
 const MountPoint = ({ ...props }) => {
     // check '/' last!
-    const validPrefixes = [ '/home', '/opt', '/srv', '/usr/local', '/var', '/boot', '/' ];
+    const validPrefixes = [ '/home', '/opt', '/srv', '/usr', '/usr/local', '/var', '/app', '/data', '/' ];
```
lucasgarfield, line 16: Is /usr/local still a valid prefix since it is a sub directory of a valid prefix? It isn't in the current list:
https://github.com/osbuild/osbuild-composer/blob/main/internal/distro/rhel85/distro.go#L44

teg: Any sub directory of a valid prefix should be valid (that was the idea anyway ;) ).

lucasgarfield (approved):
Looks good!

src/Components/CreateImageWizard/formComponents/MountPoint.js
Gundersanne: So any mounts starting with the prefixes in the `mountpointAllowList`, save for `/`, are still good (see https://github.com/osbuild/osbuild-composer/blob/main/internal/distro/rhel85/distro.go#L418). 

Just to check:
image

^ this indeed succeeded: 
image

PR #681, opened 2022-04-01
Draggable row fixes

Gundersanne:


jkozol (approved):
Nice catch on this bug! This fix works and the code cleanup makes sense.

PR #682, opened 2022-04-04
CreateImageWizard: Add clear button to package search inputs

### LINKED ISSUE #604: Available package search input textbox is not wide enough
The available package search input textbox is not wide enough and this becomes obvious when it is in focus as shown below:

image

This behavior doesn't always happen, I am not sure why yet. I have observed it in both Firefox and Chromium.

lucasgarfield:
Resolves #604. The available and chosen package search inputs now have a
clear button which resets the search and list elements to their default
states. This fixes a cosmetic issue related to the focus-visible width
being smaller than the search input width.

This is accomplished by passing an onClear prop to the search input.
This makes state management is more complicated and to facilitate this feature the
chosenFilter was converted from a useState hook to a useRef hook.

The functions used to add all and selected packages were split from two
functions into four functions. The functions appear similiar but this is only
superficial; the logic is actually quite different and there is very
little code reuse. Four separate functions will be easier to understand and
maintain.

This commit includes two other modifications:
1. If a filter is in place, it is now enforced when adding additional
   packages.
2. Package sorting behavior is now consistent. Previously, packages would be
   sorted if 'add all' was used but not if 'add selected' was used. Now
   the default behavior is to always sort the package list.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Packages.js
+++ b/src/Components/CreateImageWizard/formComponents/Packages.js
@@ -4,7 +4,6 @@ import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
 import api from '../../../api';
 import PropTypes from 'prop-types';
 import {
-    Button,
     DualListSelector,
     DualListSelectorPane,
     DualListSelectorList,
@@ -25,12 +24,13 @@ const removePackagesDisplayFields = (packages) => packages.map((pack) => ({
 const Packages = ({ defaultArch, ...props }) => {
     const { change, getState } = useFormApi();
     const { input } = useFieldApi(props);
-    const packagesSearchName = useRef();
+    const [ packagesSearchName, setPackagesSearchName ] = useState(undefined);
+    const [ filterAvailable, setFilterAvailable ] = useState(undefined);
+    const [ filterChosen, setFilterChosen ] = useState(undefined);
     const [ packagesAvailable, setPackagesAvailable ] = useState([]);
     const [ packagesAvailableFound, setPackagesAvailableFound ] = useState(true);
     const [ packagesChosen, setPackagesChosen ] = useState([]);
     const [ packagesChosenFound, setPackagesChosenFound ] = useState(true);
-    const [ filterChosen, setFilterChosen ] = useState('');
     const [ focus, setFocus ] = useState('');
 
     // this effect only triggers on mount
@@ -78,24 +78,29 @@ const Packages = ({ defaultArch, ...props }) => {
         };
     });
 
-    const sortPackages = useCallback((packageList) => {
-        const sortResults = packageList.sort(searchResultsComparator(packagesSearchName.current));
+    const setPackagesAvailableSorted = (packageList, filter = filterAvailable) => {
+        const sortResults = packageList.sort(searchResultsComparator(filter));
         setPackagesAvailable(sortResults);
-    });
+    };
+
+    const setPackagesChosenSorted = (packageList) => {
+        const sortResults = packageList.sort(searchResultsComparator(filterChosen));
+        setPackagesChosen(sortResults);
+    };
 
     // filter the packages by name
-    const filterPackagesAvailable = useCallback((packageList) => {
+    const filterPackagesAvailable = (packageList) => {
         return packageList.filter((availablePackage) => {
             // returns true if no packages in the available or chosen list have the same name
             return !packagesChosen.some((chosenPackage) => availablePackage.name === chosenPackage.name);
         });
-    });
+    };
 
     const getAllPackages = async () => {
         const args = [
             getState()?.values?.release,
             getState()?.values?.architecture || defaultArch,
-            packagesSearchName.current
+            packagesSearchName
         ];
         let { data, meta } = await api.getPackages(...args);
         if (data?.length === meta.count) {
@@ -108,11 +113,13 @@ const Packages = ({ defaultArch, ...props }) => {
 
     // call api to list available packages
     const handlePackagesAvailableSearch = async () => {
+        setFilterAvailable(packagesSearchName);
+
         const packageList = await getAllPackages();
         if (packageList) {
             const packagesAvailableFiltered = filterPackagesAvailable(packageList);
-            sortPackages(packagesAvailableFiltered);
-            setPackagesAvailableFound(true);
+            setPackagesAvailableSorted(packagesAvailableFiltered, packagesSearchName);
+            setPackagesAvailableFound(packagesAvailableFiltered.length ? true : false);
         } else {
             setPackagesAvailable([]);
             setPackagesAvailableFound(false);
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -156,60 +163,99 @@ const Packages = ({ defaultArch, ...props }) => {
         };
     });
 
-    // move selected packages
-    const moveSelected = (fromAvailable) => {
-        const sourcePackages = fromAvailable ? packagesAvailable : packagesChosen;
-        const destinationPackages = fromAvailable ? packagesChosen : packagesAvailable;
+    const areFound = (filter, packageList) => {
+        return filter === undefined ? true :
```
jkozol, line 167: No need to change this. Just in general I think that once inline conditionals start getting chained together they get hard to read. I would probably use an if else here or add a comment to improve the readability. But, this is correct and no need to change it.

jkozol, line 167: No need to change this. Just in general I think that once inline conditionals start getting chained together they get hard to read. I would probably use an if else here or add a comment to improve the readability. But, this is correct and no need to change it.

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -156,60 +163,99 @@ const Packages = ({ defaultArch, ...props }) => {
...
+            packageList.some(pack => pack.name.includes(filter)) ? true : false;
+    };
+
+    const isHidden = (filter, pack) => {
+        const filterActive = filter === undefined ? false : true;
```
jkozol, line 172: I don't believe this is necessary. In logic I believe undefined evaluates to false. So you can just check for `filterActive && !pack...`

jkozol, line 172: I don't believe this is necessary. In logic I believe undefined evaluates to false. So you can just check for `filterActive && !pack...`

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -156,60 +163,99 @@ const Packages = ({ defaultArch, ...props }) => {
...
+
+        updateState(updatedPackagesAvailable, updatedPackagesChosen);
+    };
+
+    const moveAllToChosen = () => {
+        const newPackagesChosen = packagesAvailable.reduce((result, pack) => {
+            pack.selected = false;
+            pack.isHidden = isHidden(filterChosen.current, pack);
+            return [ ...result, pack ];
+        }, []);
```
jkozol, line 228: Using `.map` would be better here than reduce since each step is independent of the previous ones.

jkozol, line 228: Using `.map` would be better here than reduce since each step is independent of the previous ones.

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -156,60 +163,99 @@ const Packages = ({ defaultArch, ...props }) => {
...
+    };
+
+    const moveAllToAvailable = () => {
+        const updatedPackagesChosen = packagesChosen.filter(pack => pack.isHidden);
+
+        const newPackagesAvailable = filterAvailable.current === undefined ? [] :
+            packagesChosen.reduce((result, pack) => {
+                pack.selected = false;
+                return (!pack.isHidden && pack.name.includes(filterAvailable.current) ? [ ...result, pack ] : [ ...result ]);
+            }, []);
```
jkozol, line 243: I would also write this using `.map`. It should make the return statement a bit clearer.

jkozol, line 243: I would also write this using `.map`. It should make the return statement a bit clearer.

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -25,12 +24,13 @@ const removePackagesDisplayFields = (packages) => packages.map((pack) => ({
 const Packages = ({ defaultArch, ...props }) => {
     const { change, getState } = useFormApi();
     const { input } = useFieldApi(props);
-    const packagesSearchName = useRef();
+    const packagesSearchName = useRef(undefined);
+    const filterAvailable = useRef(undefined);
+    const filterChosen = useRef(undefined);
```
jkozol, line 29: Why `useRef` instead of `useState`

jkozol, line 29: Why `useRef` instead of `useState`

jkozol (commented):
Generally looks good. Just a few questions and nitpicks.

src/Components/CreateImageWizard/formComponents/Packages.js
lucasgarfield: Agreed 100% on readability, I went ahead and changed this!

src/Components/CreateImageWizard/formComponents/Packages.js
lucasgarfield: Nice catch, I'm still getting comfortable with JS truthiness. Done.

lucasgarfield: Nice catch, I'm still getting comfortable with JS truthiness. Done.

src/Components/CreateImageWizard/formComponents/Packages.js
lucasgarfield: Agreed, done!

lucasgarfield: Agreed, done!

src/Components/CreateImageWizard/formComponents/Packages.js
lucasgarfield: Duplicate comment, resolving.

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -78,24 +78,29 @@ const Packages = ({ defaultArch, ...props }) => {
         };
     });
 
-    const sortPackages = useCallback((packageList) => {
-        const sortResults = packageList.sort(searchResultsComparator(packagesSearchName.current));
+    const setPackagesAvailableSorted = (packageList, filter = filterAvailable) => {
```
lucasgarfield, line 81: I had to add an optional filter parameter here since `filterAvailable` is now a state.

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -108,22 +113,24 @@ const Packages = ({ defaultArch, ...props }) => {
...
     // call api to list available packages
     const handlePackagesAvailableSearch = async () => {
+        setFilterAvailable(packagesSearchName);
+
         const packageList = await getAllPackages();
         if (packageList) {
             const packagesAvailableFiltered = filterPackagesAvailable(packageList);
-            sortPackages(packagesAvailableFiltered);
-            setPackagesAvailableFound(true);
+            setPackagesAvailableSorted(packagesAvailableFiltered, packagesSearchName);
```
lucasgarfield, line 121: This is why `setPackagesAvailableSorted()` needs an optional `filter` parameter, we need to use `packagesSearchName` here as the default `filterAvailable` would be the value from the previous render and therefore stale.

jkozol (approved):
Works great!

PR #686, opened 2022-04-07
CreateImageWizard: fix label spacing

### LINKED ISSUE #597: Label spacing is wrong
Spacing between labels – In mocks it is 4px between label & what is being labeled, and then 8px between the content.

image22

chloenayon:
Decrease the spacing below labels from 8px to 4px in accordance with the mocks.
Fixes #597

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/TargetEnvironment.scss
+++ b/src/Components/CreateImageWizard/formComponents/TargetEnvironment.scss
@@ -1,3 +1,7 @@
+.pf-c-form__group-label {
+    --pf-c-form__group-label--PaddingBottom: var(--pf-global--spacer--xs);
+}
+
 .tiles {
     display: flex;
 }


src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
```diff
@@ -41,7 +41,7 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
     return (
         <>
             <FormGroup isRequired={ isRequired } label={ label } data-testid="target-select">
-                <FormGroup label={ <Text component={ TextVariants.small }>Public cloud</Text> } data-testid="target-public">
+                <FormGroup label={ <small>Public cloud</small> } data-testid="target-public">
```
jkozol, line 44: This change is unnecessary for the fix

jkozol, line 44: This change is unnecessary for the fix

jkozol (commented):
Nice find on the fix. Looks good to me. I wonder if we should change all the label spacing. @katierik any thoughts? It seems other field labels have a margin of 8px. Should we change all or is this 4px spacing specific to the image output labels?


~~Also, @chloenayon can you increase the spacing between the tiles or create a seperate issue for tracking that. "8px between the content"~~

jkozol (dismissed):

chloenayon:
with the change: 

Screenshot from 2022-04-08 13-33-05

jkozol (approved):

PR #691, opened 2022-04-11
CreateImageWizard/FileSystem: Default to 0 if size input is NaN

Gundersanne:
This gets rid of prop warnings during tests.


jkozol (approved):
Thanks for the fix. Test logs look cleaner.

jkozol (approved):

PR #696, opened 2022-04-12
Webpack conf cleanup

Gundersanne:


config/dev.webpack.config.js
```diff
@@ -4,15 +4,14 @@ const { GitRevisionPlugin } = require('git-revision-webpack-plugin');
...
-    deployment: process.env.BETA ? 'beta/apps' : 'apps',
     useProxy: true,
     proxyVerbose: true,
-    debug: true,
     env: `${process.env.STAGE ? 'stage' : 'prod'}-${process.env.BETA ? 'beta' : 'stable'}`,
-    appUrl: process.env.BETA ? '/beta/insights/image-builder' : '/insights/image-builder',
 };
 
 const { config: webpackConfig, plugins } = config({
+    deployment: process.env.BETA ? 'beta/apps' : 'apps',
```
jkozol, line 13: Should the prod config also default to apps? Currently it is `beta/apps` or not default. Is default apps?

config/dev.webpack.config.js
```diff
@@ -4,15 +4,14 @@ const { GitRevisionPlugin } = require('git-revision-webpack-plugin');
...
+    appUrl: process.env.BETA ? '/beta/insights/image-builder' : '/insights/image-builder',
```
jkozol, line 14: Also, why do we not need to declare `appUrl` in the prod config?

jkozol (approved):
Overall makes sense. Just a couple questions

config/dev.webpack.config.js
Gundersanne: Yea when we deploy to prod-stable it should also default to `apps`. I'll add this, one less thing to think about when we get there.

config/dev.webpack.config.js
Gundersanne: Ugh this is a mistake sorry, appURL is proxy sepcific; deployment isn't. Changing this.

jkozol (approved):

PR #705, opened 2022-04-14
CreateImageWizard: move name image step to end

jkozol:
The Details step is renamed to Name Image and moved to the last step before Review. Also, the validation is removed since all utf-8 characters are supported for image names.

Screenshot from 2022-04-14 18-13-35


lucasgarfield:
I'm sure this has already been discussed but is there a need for any validation on the length of the name? What about validation for invalid names? With all utf-8 characters supported there aren't many invalid names I can think of but for instance an image named ' ' (just a space) will fail and the user will receive a 400 status code back. Do we still want to try to catch invalid names such as ' ' in the form before submission?

Gundersanne:
> I'm sure this has already been discussed but is there a need for any validation on the length of the name? What about validation for invalid names? With all utf-8 characters supported there aren't many invalid names I can think of but for instance an image name with spaces in it will fail and the user will receive a 400 status code back. Maybe we should keep validation and check for spaces?

Spaces are allowed though, no?

A max length is a good idea however, should be enforced on api level first and foremost.

lucasgarfield:
> > I'm sure this has already been discussed but is there a need for any validation on the length of the name? What about validation for invalid names? With all utf-8 characters supported there aren't many invalid names I can think of but for instance an image name with spaces in it will fail and the user will receive a 400 status code back. Maybe we should keep validation and check for spaces?
> 
> Spaces are allowed though, no?
> 
> A max length is a good idea however, should be enforced on api level first and foremost.

Oh you're right, spaces are allowed, I was receiving a 400 error for a different unknown reason but it was right after a successful build so I just assumed it was due to the space.

lucasgarfield (approved):
Everything looks good to me, nice job!

Gundersanne (approved):

PR #709, opened 2022-04-21
CreateImageWizard: fix validation for activation keys

### LINKED ISSUE #624: Activation key needs validation
The user should not be able to continue with no activation key selected.

jkozol:
The validators are passed values by the useFieldApi hook. Adding this to the activation key component allows the `required` validation to work.

Fixes #624


lucasgarfield (approved):
Nice fix, I'm not 100% up to speed on the useFieldApi and useFormApi hooks but as far as I can tell everything is as it should be!

PR #710, opened 2022-04-21
.github: Sync branches workflow

Gundersanne:
This workflow syncs the main branch with the stage-stable branch.
    
    It also includes a production target which needs to be dispatched
    manually. The production target pushes the input ref to both prod-beta
    and prod-stable branches.
    
    This results in the main branch always being deployed on stage-beta and
    stage-stable, and the prod-beta and prod-stable branches being synced on
    release.

---

~~With 4 different branches now (main, stage-stable, prod-beta, prod-stable), we should have a more formal way of releasing to a specific target.~~

~~Should we have prod-beta == stage-stable always?~~

---

On push to main main -> stage-beta sync.

Release sync $sourc ref -> prod-beta == prod-stable

### DIFF
--- a/.github/workflows/sync-branches.yml
+++ b/.github/workflows/sync-branches.yml
@@ -1,0 +1,79 @@
+name: Sync branches
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      source:
+        description: Source ref (branch or sha)
+        required: true
+        type: string
+      target:
+        description: Target branch
+        required: true
+        type: choice
+        options:
+          - production
+          - stage-stable
+
+
+jobs:
+  check:
+    name: Check source and target validity
+    timeout-minutes: 1
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


Gundersanne:
See https://github.com/Gundersanne/image-builder-frontend/actions/workflows/sync-branches.yml for some example runs

Gundersanne:
Suggestion from @teg, main == stable-stage == stable-beta ; and we keep the 1 separate prod release target where prod-beta == prod-stage.

ochosi:
Love this! :heart: 
Makes it easy to switch to scheduled releases as a next step :grin:

jkozol:
How do you think we should differentiate between prod-beta and prod-stable then? Push the ref to beta whenever main updates to keep stage == prod-beta?

Gundersanne:
Well we could have stable and beta always be the same version, and only differentiate between prod and stage. beta vs non-beta will just have runtime differences like in #712 (we probably should add a good way to  test this though, like maybe run the test suite once as beta and once as non-beta, currently we only test beta). 

And then main == stage and prod is separate, but prod-beta == prod-stable. But in that case the sync-branches workflow really needs to become just a release workflow which always pushes main to stage-stable; and on manual trigger pushes the same ref to prod-beta and prod-stable.

jkozol (approved):

PR #711, opened 2022-04-22
README: update start scripts

kingsleyzissou:
Update the start scripts to match the updated scripts in the `package.json`. Additionally, add instructions
for forwarding the domains to localhost for the stage and prod since this is only mentioned in the `devel/`
readme.


Gundersanne (approved):
Woops, my bad. Thanks for fixing :)

PR #713, opened 2022-04-22
Fix docker stack

kingsleyzissou:
- minor fixes to the devel stack configuration to match the changes in other repos
- switch to a client secret instead of an offline token
- update the frontend entrypoint command
- add profiles to the metrics containers so they are only launched when desired

### DIFF
--- a/devel/docker-compose.yml
+++ b/devel/docker-compose.yml
@@ -125,6 +125,7 @@ services:
      - "backend"
      - "frontend"
   prometheus:
+    profiles: ["metrics"]
     image: prom/prometheus:latest
     ports:
       - "9000:9090"
@@ -140,6 +141,7 @@ services:
     command:
       - "--config.file=/config/prometheus.yml"
   grafana:
+    profiles: ["metrics"]
     image: grafana/grafana:latest
     ports:
       - "3000:3000"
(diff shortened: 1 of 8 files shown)


devel/docker-compose.yml
```diff
@@ -26,14 +26,10 @@ services:
...
       - ${WORKER_CONFIG_DIR}/osbuild-worker.toml:/etc/osbuild-worker/osbuild-worker.toml:z
-      - ${WORKER_CONFIG_DIR}/token:/etc/osbuild-worker/token:z
+      - ${WORKER_CONFIG_DIR}/secret:/etc/osbuild-worker/secret:z
     environment:
       - CACHE_DIRECTORY=/var/cache/osbuild-composer
     privileged: true
-    cap_add:
-      - MKNOD
-      - SYS_ADMIN
-      - NET_ADMIN
```
Gundersanne, line 36: Are these not needed to build an image?

Gundersanne (dismissed):
lgtm, just a small q

devel/docker-compose.yml
kingsleyzissou: Ah yes, you're right. I was getting a weird error for a bit where `privileged` and `cap_add` were clashing. I'll fix that now

Gundersanne (approved):

PR #718, opened 2022-04-27
ImagesTable: add error status

kingsleyzissou:
Add error information for failed image builds. The error details are only displayed
for failed builds and the error highest up the error chain is displayed in the expandable
section of the image.

### DIFF
--- a/src/Components/ImagesTable/ImageBuildErrorDetails.js
+++ b/src/Components/ImagesTable/ImageBuildErrorDetails.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from '@patternfly/react-core';
+
+const useGetErrorReason = (err) => {
+    if (!err?.reason) {
+        return 'An unknown error occured';
+    }
+
+    if (err.details?.reason) {
+        return err.details.reason;
+    }
+
+    return err.reason;
+};
+
+const ErrorDetails = ({ status }) => {
+    if (!status || status.status !== 'failure') {
+        return <></>;
+    }
+
+    const reason = useGetErrorReason(status.error);
+
+    return (
+        <div className='pf-u-mt-sm'>
+            <strong>Status</strong>
+            <Alert variant="danger" title="Image build failed" isInline isPlain />
+            <p className='pf-u-danger-color-200 pf-u-w-33-on-md'>{reason}</p>
+        </div>
+    );
+};
+
+ErrorDetails.propTypes = {
+    status: PropTypes.object,
+};
+
+export default ErrorDetails;
--- a/src/Components/ImagesTable/ImagesTable.js
+++ b/src/Components/ImagesTable/ImagesTable.js
@@ -15,6 +15,7 @@ import ImageBuildStatus from './ImageBuildStatus';
 import Release from './Release';
 import Target from './Target';
 import ImageLink from './ImageLink';
+import ErrorDetails from './ImageBuildErrorDetails';
 
 const ImagesTable = () => {
     const [ page, setPage ] = useState(1);
@@ -189,6 +190,7 @@ const ImagesTable = () => {
                                             <ExpandableRowContent>
                                                 <strong>UUID</strong>
                                                 <div>{ id }</div>
+                                                <ErrorDetails status={ compose.image_status } />
                                             </ExpandableRowContent>
                                         </Td>
                                     </Tr>
--- a/src/test/Components/ImagesTable/ImagesTable.test.js
+++ b/src/test/Components/ImagesTable/ImagesTable.test.js
@@ -180,6 +180,12 @@ const store = {
                 },
                 image_status: {
                     status: 'failure',
+                    error: {
+                        reason: 'A dependency error occured',
+                        details: {
+                            reason: 'Error in depsolve job'
+                        }
+                    }
                 },
             },
             'ca03f120-9840-4959-871e-94a5cb49d1f2': {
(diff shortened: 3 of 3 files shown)


kingsleyzissou:
image

src/Components/ImagesTable/ImageBuildErrorDetails.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    if (err.details && err.details.reason) {
+        return err.details.reason;
+    }
+
+    return err.reason;
+};
+
+const FailureIcon = () => (
```
lucasgarfield, line 18: We could also use a PatternFly `<Alert>` here, specifically the inline plain variation:
https://www.patternfly.org/v4/components/alert/

src/Components/ImagesTable/ImageBuildErrorDetails.js
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from 'prop-types';
+import { Flex } from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
+
+const useGetErrorReason = (err) => {
+    if (!err || !err.reason) {
+        return 'An unknown error occured';
+    }
+
+    if (err.details && err.details.reason) {
```
lucasgarfield, line 11: We could use optional chaining here instead, i.e. `err.details?.reason`, couldn't we? Obviously the short circuit `&&` totally works and makes sense but for consistency in the code base I think optional chaining is generally preferred when possible, at least that's the impression I have from Jacob from my previous code reviews ;).

src/Components/ImagesTable/ImageBuildErrorDetails.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Flex } from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
+
+const useGetErrorReason = (err) => {
+    if (!err || !err.reason) {
```
lucasgarfield, line 7: Should we use optional chaining here as well? i.e. `!err2?.reason`?

src/Components/ImagesTable/ImageBuildErrorDetails.js
```diff
@@ -0,0 +1,49 @@
...
+        return <React.Fragment />;
+    }
+
+    const reason = useGetErrorReason(status.error);
+
+    return (
+        <div className='pf-u-mt-sm'>
+            <strong>Status</strong>
+            <FailureIcon />
+            <p className='pf-u-danger-color-100 pf-u-w-33-on-md'>
```
lucasgarfield, line 38: Using PatternFly's `<Alert>` would also take care of this, although by default the `<Alert>` text isn't colored red (only the header) so we would still need some CSS.

src/Components/ImagesTable/ImageBuildErrorDetails.js
```diff
@@ -0,0 +1,49 @@
...
+        <div className="pf-u-mr-sm">
+            <ExclamationCircleIcon className="error" />
+        </div>
+        <strong className='pf-u-danger-color-100'>Image build failed</strong>
+    </Flex>
+);
+
+const ErrorDetails = ({ status }) => {
+    if (!status || status.status !== 'failure') {
+        return <React.Fragment />;
```
lucasgarfield, line 29: Would the short syntax for fragments `</>` work here? We use both throughout the code base but </>, I'm not sure whether we prefer one over the other... The short syntax is a newer feature...

lucasgarfield (commented):
Looks great! Just a few comments, hope it isn't too nit-picky :smile:

src/Components/ImagesTable/ImageBuildErrorDetails.js
jkozol: Yep this works as expected but `err.details?.reason` accomplishes the same with less code.

src/Components/ImagesTable/ImageBuildErrorDetails.js
jkozol: So both work. We should ideally only use one. I'm more in favor of the short `<> </>`

src/Components/ImagesTable/ImageBuildErrorDetails.js
```diff
@@ -0,0 +1,49 @@
...
+            <strong>Status</strong>
+            <FailureIcon />
+            <p className='pf-u-danger-color-100 pf-u-w-33-on-md'>
+                { reason }
+            </p>
+        </div>
+    );
+};
+
+ErrorDetails.propTypes = {
```
lucasgarfield, line 45: This is not a critical comment and I am not questioning it... but just for my own knowledge, why do we need this?

src/Components/ImagesTable/ImageBuildErrorDetails.js
kingsleyzissou: I prefer `<></>`, I just saw that one of the other components was using `<React.Fragment>` so went with that for consistency. I'll update.

src/Components/ImagesTable/ImageBuildErrorDetails.js
kingsleyzissou: Think it's required by the linter to validate the type of props being passed to the component. Think it's just for type safety.

src/Components/ImagesTable/ImageBuildErrorDetails.js
kingsleyzissou: Ahhhh I didn't think of that. I'll update.

src/Components/ImagesTable/ImageBuildErrorDetails.js
kingsleyzissou: I guess it makes things slightly simpler. Will have a look

src/Components/ImagesTable/ImageBuildErrorDetails.js
kingsleyzissou: Haha my javascript is a bit rusty. Will fix

src/Components/ImagesTable/ImageBuildErrorDetails.js
lucasgarfield: Oh, good point... I agree that using the `<p>` tag is the way to go here, it is true to the mockup and consistent with the UUID alignment.

lucasgarfield (approved):
Looks good to me!

jkozol (approved):

PR #719, opened 2022-04-27
.github: Add schutzbot token to gh workflow

Gundersanne:


jkozol (approved):

Gundersanne:
This did not work because you can't just overwrite the token by setting the env, forgot that. The solution is to use the separate token in the remote url. I'll followup tomorrow.

PR #722, opened 2022-05-05
ImagesTable: support additional image type labels

jkozol:
The rhel-edge-commit and rhel-edge-installer image types had no target label. This is now added. Also, if other image types are created that do not have a descriptive label then default to using the image type.

### DIFF
--- a/src/Components/ImagesTable/Target.js
+++ b/src/Components/ImagesTable/Target.js
@@ -28,7 +28,7 @@ const Target = ({ composeId }) => {
     target = targetOptions[compose.uploadType];
   }
 
-  return <>{target}</>;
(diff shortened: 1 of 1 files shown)


Gundersanne (approved):
Neat!

Gundersanne:
> Should we display images at all if we don't have explicit support for them? We can't build edge images through our UI, so maybe displaying them isn't right either?
> 
> @bronaghs you might want to be aware that the edge images are currently dispelled in our UI.

We should at least disable the recreate image button for them yea, and the download link.

Gundersanne (dismissed):

croissanne (dismissed):
Let's just do this ._. thank you!

croissanne (approved):

PR #724, opened 2022-05-12
.github: Rename sync-branches workflow ref check job

Gundersanne:
Make the name of the check job more like the name of the sync job.

---

very important patch, it just annoyed me suddenly .-.


ochosi (approved):
Makes sense!

PR #729, opened 2022-05-31
CI: add trigger for IQE pipelines

jrusz:
This modifies the Jenkinsfile used for release to also send a POST
requested to a webhook url which will then trigger test pipelines in IQE
Jenkins.

That Jenkinsfile is quite stable and will probably not see any more changes as the frontend deployement is moving away from it. For now if we want to trigger IQE test pipelines after a deploy of new frontend this is the way to do it.


jrusz:
Ugh, I think I want this to go to `stage-stable` branch, I'll open a different PR.

PR #730, opened 2022-05-31
CI: add trigger for IQE pipelines

jrusz:
This modifies the Jenkinsfile used for release to also send a POST
requested to a webhook url which will then trigger test pipelines in IQE
Jenkins.

That Jenkinsfile is quite stable and will probably not see any more changes as the frontend deployement is moving away from it. For now if we want to trigger IQE test pipelines after a deploy of new frontend this is the way to do it.

### DIFF
--- a/.travis/58231b16fdee45a03a4ee3cf94a9f2c3
+++ b/.travis/58231b16fdee45a03a4ee3cf94a9f2c3
@@ -1,0 +1,78 @@
+@Library("github.com/RedHatInsights/insights-pipeline-lib@v3")
+import groovy.json.JsonSlurper
+
+node {
+  stage ("deploy") {
+
+    checkout scm
+    withCredentials(bindings: [sshUserPrivateKey(credentialsId: "cloud-netstorage",
+                      keyFileVariable: "privateKeyFile",
+                      passphraseVariable: "",
+                      usernameVariable: "")]) {
+
+      String APP_NAME = "image-builder"
+      String BRANCH = env.BRANCH_NAME.replaceAll("origin/", "")
+
+      if (BRANCH == "prod-stable") {
+        PREFIX = ""
+      } else if (BRANCH == "prod-beta") {
+        PREFIX = "beta/"
+      } else if (BRANCH == "qa-stable" || BRANCH == "stage-stable") {
+        PREFIX = "stage/"
+      } else if (BRANCH == "qa-beta" || BRANCH == "stage-beta") {
+        PREFIX = "stage/beta/"
+      } else {
+        error "Bug: invalid branch name, we only support (prod/qa/stage)-(beta/stable) and we got ${BRANCH}"
+      }
+
+      // Write build info into app.info.json
+      // We have the src info there already
+      def app_info = readJSON file: "./app.info.json"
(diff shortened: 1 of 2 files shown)


teg:
Sorry if this is a naive question, but could you explain how the IQE test pipeline fits in? What tests does it run, and is this going to be used for things like auto promotion or something like that? I'm sure I have read this at some point, but might also be good to document / explain for outsiders to understand.

> That Jenkinsfile is quite stable and will probably not see any more changes as the frontend deployement is moving away from it. For now if we want to trigger IQE test pipelines after a deploy of new frontend this is the way to do it.

I didn't understand this, could you explain a bit more? Are you saying that the Jenkinsfile is being deprecated in favour of the web hook to the IQE test pipeline?

jrusz:
> Sorry if this is a naive question, but could you explain how the IQE test pipeline fits in? What tests does it run, and is this going to be used for things like auto promotion or something like that? I'm sure I have read this at some point, but might also be good to document / explain for outsiders to understand.
> 

The IQE pipeline runs just basic sanity checks and one e2e scenario against `console.redhat.com` The test suite lives in here:
https://gitlab.cee.redhat.com/insights-qe/iqe-image-builder-backend-plugin & https://gitlab.cee.redhat.com/insights-qe/iqe-image-builder-frontend-plugin and it's run in IQE Jenkins currently twice a day. I want to run the tests as soon as deployment of new app on `console.redhat.com` happens. It's not going to be used for any promotions AFAIK. 

> > That Jenkinsfile is quite stable and will probably not see any more changes as the frontend deployement is moving away from it. For now if we want to trigger IQE test pipelines after a deploy of new frontend this is the way to do it.
> 
> I didn't understand this, could you explain a bit more? Are you saying that the Jenkinsfile is being deprecated in favour of the web hook to the IQE test pipeline?

Yeah my understanding is that the whole Jenkins is currently legacy and it's going to be replaced by new frontend-deploy container run in app-sre environment. And no it's not being deprecated now by this, I just don't have any other way to put in the `curl` command, this was suggested to me if I want to do it. All the frontend apps share the same Jenkinsfile and it hasn't been changed for months so it should be fine to have our own. Alternatively I can just modify it with `sed` after it's downloaded from the common repo, but I figured it's going to be easier like this.

I also plan to add some conditions for branches, right now I want to test it for stage only.

teg:
Thanks @jrusz !

What happens to the IQE results? Where are they reported? Even if no auto promotion is planned, is it a possibility?

jrusz:
> Thanks @jrusz !
> 
> What happens to the IQE results? Where are they reported? Even if no auto promotion is planned, is it a possibility?

They are reported to `#osbuild-image-builder-service` channel on Slack. The auto promotion is certainly a possibility in app-sre environments, but first Insights QE teams would have to finish their migrations I think, there are already attempts to run the tests there in containers but there are still some issues AFAIK. I didn't look into this much deeper because we decided we don't want to move the current IQE pipelines from their environment where we know they're working and doing what we need :)

jrusz:
@Gundersanne could you please review?

Gundersanne (approved):

PR #736, opened 2022-06-02
CI: Enable IQE trigger for prod branch

jrusz:
Inserting different webhook url based on the deployed branch in the
modified Jenkinsfile. Also renaming it to make more sense for us.


Gundersanne (approved):
Thanks!

PR #738, opened 2022-06-06
CreateImageWizard: Use stage candlepin urls on stage consoledot 

Gundersanne:


teg (commented):
I'm not 100% sure if this is the right thing to do. As discussed on a call, this means we can only test artefacts from staging if on the VPN (and using the proxy?). Do we really want/need to use staging candlepin?

Gundersanne:
> I'm not 100% sure if this is the right thing to do. As discussed on a call, this means we can only test artefacts from staging if on the VPN (and using the proxy?). Do we really want/need to use staging candlepin?

So the frontend activation keys that are loaded are the stage ones which can only be used against stage candlepin.

And you can only use the staging service from the vpn, I don't think it's a huge issue either way tbh.

jkozol (approved):

PR #745, opened 2022-06-27
[JSFIX] Major version upgrade of react-redux from version 7.2.6 to 8.0.2

jsfix-updater:


bastilian:
This can be closed without hesitation.

PR #753, opened 2022-07-14
Devel housekeeping

kingsleyzissou:
A few small changes to the devel docker stack. The docker stack needed a few
small changes in order to run and test some of the functionality locally.

This PR depends on:
https://github.com/osbuild/osbuild-composer/pull/2832

### DIFF
--- a/devel/docker-compose.yml
+++ b/devel/docker-compose.yml
@@ -5,6 +5,13 @@ services:
     build:
       context: ../../osbuild-composer
       dockerfile: ./distribution/Dockerfile-ubi
+    entrypoint:
+      [
+        "python3",
+        "/opt/entrypoint.py",
+        "--remote-worker-api",
+        "--composer-api"
+      ]
     volumes:
       - ${COMPOSER_CONFIG_DIR}/osbuild-composer.toml:/etc/osbuild-composer/osbuild-composer.toml:z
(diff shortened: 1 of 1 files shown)


devel/config/worker/osbuild-worker.toml
```diff
@@ -1,4 +1,4 @@
 [authentication]
-oauth_url = "http://fauxauth:8888/token?refresh_token=42"
+oauth_url = "http://fauxauth:8888/token?refresh_token=15885990"
```
Gundersanne, line 2: This isn't your personal org_id is it?

Gundersanne (commented):
thank you for maintaining this! Happy to merge, just 1 q before we do though.

devel/config/worker/osbuild-worker.toml
kingsleyzissou: I have no idea I can check after lunch. I think it's the default one that's set in Chloe's grafana PR though

kingsleyzissou: Ok checked. It's not my `org-id`

https://github.com/osbuild/osbuild-composer/pull/2756/files#diff-46ebf2e61d769f87c2459d6a997727fc5e16efa81f5a9b498406c8c3bf2ac4c6R2703-R2705

Gundersanne (approved):
thank you!

PR #761, opened 2022-08-04
devel/gen-dashboard: use python3 explicitly in the shebang

ondrejbudai:
So people don't have to install python-unversioned-command on Fedora. It saves you 11 kB of disk space, worth it. :P


kingsleyzissou (approved):
11kb is too good to be true :)

PR #762, opened 2022-08-04
ImagesTable: add option to download a raw compose request

ondrejbudai:
In order to help people transition from using GUI to use the API directly,
it's helpful to give them an easy way to inspect the whole raw compose
request.

This commit adds a new button next to each compose that simply downloads
the original compose request in a json format. This request can then be
directly piped into the IB API to build a new image.

## Notes

Manipulation DOM directly when using React? Crazy, right? It works though but I'm open to suggestions.

## Demo
https://www.youtube.com/watch?v=pr-6evaue_4
(I failed to embed gif, sry.)

## Static picture
image

Original idea by Troy Dawson

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.js
+++ b/src/Components/ImagesTable/ImagesTable.js
@@ -128,6 +128,18 @@ const ImagesTable = () => {
           state: { composeRequest: compose.request, initialStep: 'review' },
         }),
     },
+    {
+      title: (
+        <a
+          href={`data:text/plain;charset=utf-8,${encodeURIComponent(
+            JSON.stringify(compose.request)
+          )}`}
+          download="request.json"
+        >
+          Download compose request (.json)
+        </a>
+      ),
+    },
   ];
 
   // the state.page is not an index so must be reduced by 1 get the starting index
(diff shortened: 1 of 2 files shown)


src/Components/ImagesTable/ImagesTable.js
```diff
@@ -128,6 +128,25 @@ const ImagesTable = () => {
           state: { composeRequest: compose.request, initialStep: 'review' },
         }),
     },
+    {
+      title: 'Download compose request (.json)',
```
lucasgarfield, line 132: It's actually possible to use anchor tags in the action title, so we could do something like this if we want to avoid manipulating the DOM:
image

ondrejbudai: Isn't the formatting different though? 
image

lucasgarfield: We could use css to adjust the formatting, however, it might be okay as is.

Generally we try to keep things as close to the PatternFly examples as possible. Consider this example:
https://www.patternfly.org/v4/components/table#composable-actions

The link is blue with underlining on over since it is an anchor tag. And it makes sense to use an anchor tag for this purpose, downloading a resource, in my opinion.

In any case, the design team will have the final say and we can just ask them what they want us to do!

ondrejbudai:
Rewritten in the way @lucasgarfield suggested (and credited him) and added a simple test.

This is how the button looks now:
image

Gundersanne (commented):
Needs a `npm run lint:js:fix` but other than that lgtm :)

ondrejbudai:
Fixed, thanks!

Gundersanne (approved):
lgtm! thank you

PR #786, opened 2022-08-25
Wizard: Match the case of System configuration with other step titles

### LINKED ISSUE #768: Wizard: System configuration step sentence case
`System configuration` should match the case of the other step titles.

image

regexowl:
Fixes #768. This matches System configuration case with the case of the other step titles as per recent SPUR.


lucasgarfield (approved):
Looks good to me!

PR #790, opened 2022-08-29
Wizard: Fix UI issues for Microsoft Azure Target environment

### LINKED ISSUE #772: Wizard: Azure Target environment step UI issues
Current:
image

To do:
1. Increase the size of the Oauth 2.0 link
2. Remove the `Authorizing an Azure account` expandable
3. Remove the `Destination` heading (neither GCP nor AWS have this heading, so this makes Azure consistent with them)

Updated mocks: https://marvelapp.com/prototype/6g7dja9/screen/82465786

regexowl:
Fixes #772. This updates description of the step according to recent SPUR.
The `Authorizing an Azure account` expandable was removed (both from the step and `ImageCreator.js`) as well as `Destination` heading. Size of the Oauth 2.0 link was changed from small to default.


regexowl:
I've just noticed there was styling for the expandable that doesn't seem to be used anywhere else, so I've deleted that as well.

lucasgarfield (approved):
This looks great! Nice work!

PR #791, opened 2022-08-30
Add Schutzbot and Sonarqube

jrusz:
This adds the ability to use our Schutzbot Gitlab CI and run Sonarqube
scan there. We have pretty much the exact same thing in weldr-client
repo and use it only for Sonarqube. This could also be used in the
future if there is any need to use our own CI.

The added scan is just informative and is by no means supposed to be
used to gate PRs, there will be just one more link to
check the results in case anyone is interested.


.github/workflows/trigger-gitlab.yml
```diff
@@ -0,0 +1,59 @@
+# inspired by rhinstaller/anaconda
+
+name: Trigger GitLab CI
+
+on:
+  pull_request:
+    branches:
+      - "*"
```
Gundersanne, line 8: Why on pull requests? Afaik we barely check the results on main for the other projects.

jrusz: Idk, I guess for consistency sake? We have it like that in all other repos. I know nobody checks the results but I just want to have them in case we ever decide we want them... It takes a few minutes and spins one aws machine if I don't count the github runner...

jrusz: I've changed it to run just on `main`

jrusz:
I've tried merging this PR on my fork and it seems to run only on `main` https://github.com/jrusz/image-builder-frontend/actions

It just fails on not having the correct key as I don't have access to it and mine didn't work :sweat_smile:

jrusz:
Ok I made a total mess out of my fork, but the action is working. I removed the branch checkout and kept 0 fetch depth cause shallow push is not allowed to Gitlab. I've also managed to test it with a temp key (I didn't realize that if I have password set on my ssh key it will copy it with it) and the push was successful https://github.com/jrusz/image-builder-frontend/runs/8130120449?check_suite_focus=true 

It's now running in Gitlab CI, https://gitlab.com/redhat/services/products/image-builder/ci/image-builder-frontend/-/pipelines/628436366

jrusz:
The scanner is not working out of the box, need to adjust a bit.

jrusz:
Looks good now, was just missing `nodejs` package https://sonarqube.corp.redhat.com/dashboard?id=osbuild%3Aimage-builder-frontend

jrusz:
Saw just one error during the analysis, but idk what it means...
```
ERROR: Browserslist: caniuse-lite is outdated. Please run:
ERROR: npx browserslist@latest --update-db
ERROR: 
ERROR: Why you should do it regularly:
ERROR: https://github.com/browserslist/browserslist#browsers-data-updating
```

jrusz:
I tried adding the update but it's still unhappy, the analysis is working regardless so I guess it's good enough as is https://gitlab.com/redhat/services/products/image-builder/ci/image-builder-frontend/-/jobs/2964320196

Gundersanne (approved):
LGTM, thank you!

Gundersanne (approved):

PR #792, opened 2022-08-31
Redux: Add Redux Toolkit

lucasgarfield:
This commit adds Redux Toolkit as a dependency. Adding Redux Toolkit is
beneficial because it will allow us to use the current best-practice Redux
development patterns and tools as well as reduce the number of
dependencies. Redux Toolkit is backwards compatible with existing Redux code,
and therefore works with all of the current actions and reducers.

The store is now created using Redux Toolkit's configureStore(). Previously, a
custom store creator developed by Red Hat Insights was used. However,
this is actually not required for Insights apps, and creating the store using
configureStore() is necessary to take advantage of Redux Toolkit. For
instance, the store can now be inspected using the Redux devtools in the
browser.

This commit removes the redux-logger middleware. It is no longer
necessary, as the Redux development tools can now be used to easily inspect and
reason about the redux store and its state.

The Thunk middleware dependency has also been removed, as Thunks are already
included in Redux Toolkit.

The redux-promise-middleware dependency has been left in place for now,
but its functionality is also available in Redux Toolkit and it may be
considered for removal in the future.

Using Redux Toolkit will also allow us to move to the `Slice` pattern when
defining actions/reducers in the future if we wish. This will make
writing, reasoning about, and debugging the code related to the Redux
store much easier.

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -1,38 +1,14 @@
-import { ReducerRegistry } from '@redhat-cloud-services/frontend-components-utilities/ReducerRegistry';
+import { configureStore } from '@reduxjs/toolkit';
+import composes from './reducers/composes';
 import promiseMiddleware from 'redux-promise-middleware';
-import thunk from 'redux-thunk';
 import { notificationsReducer } from '@redhat-cloud-services/frontend-components-notifications/redux';
 
-import composes from './reducers/composes';
+export const reducer = {
+  composes: composes,
+  notifications: notificationsReducer,
+};
 
-let registry;
+export const middleware = (getDefaultMiddleware) =>
+  getDefaultMiddleware().concat(promiseMiddleware);
 
-export function init(store = {}, ...middleware) {
-  if (!registry) {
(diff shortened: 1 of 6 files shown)


src/store/index.js
```diff
@@ -1,38 +1,14 @@
-import { ReducerRegistry } from '@redhat-cloud-services/frontend-components-utilities/ReducerRegistry';
+import { configureStore } from '@reduxjs/toolkit';
+import composes from './reducers/composes';
```
Gundersanne, line 2: Should we come up with a policy on import ordering or something? Like separate libraries from in-tree imports etc..

Gundersanne (approved):
Tried it, yes please! Thank you :)

src/store/index.js
lucasgarfield: Yes, that's a very good idea... and we should try to put the policy into ESLint.

Gundersanne (approved):
LGTM, ty!

PR #798, opened 2022-09-09
ESLint: Set rules for import order

### LINKED ISSUE #795: Import sorting
We should create (or perhaps to save time, find) a good policy for import order of packages. Ideally we would be able to use ESLint to enforce it and autofix the current codebase.

regexowl:
Fixes #795. This commit adds rules to `.eslintrc.yml` which enforce sorting of the import declarations.

Added plugin `eslint-plugin-import` allows to sort the import declarations based on their types (builtin, external, internal, etc.) and rule `sort-imports` then sorts the declarations inside these groups alphabetically.


package.json
```diff
@@ -64,6 +64,7 @@
     "babel-plugin-transform-imports": "2.0.0",
     "css-loader": "6.7.1",
     "eslint": "8.13.0",
+    "eslint-plugin-import": "^2.26.0",
```
Gundersanne, line 67: We to to prefer exact versions and let dependabot update versions one by one. So `"2.26.0"` without caret.

Gundersanne (changes requested):
Thank you!

2 small comments.

Also, when implementing a rule like this, it needs to be applied to the source. Currently this breaks CI as the linting job fails with the new rules. So in a 2nd commit do a linting run, fixing everything. You can use `npm run lint:js:fix` to get 99% of the way there, it'll also give us an idea of what the changes look like.

package.json
regexowl: Got it, fixed.

.eslintrc.yml
```diff
@@ -4,3 +4,19 @@ globals:
...
+    ignoreDeclarationSort: true
+  import/order:
+  - error
+  - groups:
+      - builtin
+      - external
+      - internal
+      - sibling
+      - parent
+      - index
```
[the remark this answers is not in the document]
regexowl: Newline added, thanks Sanne :)

regexowl:
I've added a commit with the fixed files from  #799.

PR #802, opened 2022-09-13
ImagesTable: Add pagination to the bottom of the table

### LINKED ISSUE #594: Need images pagination on buttom as well as top
When a user scrolls down the images list they should see pagination buttons at the bottom.

regexowl:
Fixes #594. This adds pagination to the bottom of the images list so the user doesn't have to scroll up to use it.

Tests were also updated to reflect the change.

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.js
+++ b/src/Components/ImagesTable/ImagesTable.js
@@ -19,6 +19,7 @@ import {
   EmptyStateBody,
   EmptyStateSecondaryActions,
   Pagination,
+  PaginationVariant,
   Toolbar,
   ToolbarContent,
   ToolbarItem,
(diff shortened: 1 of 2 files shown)


src/Components/ImagesTable/ImagesTable.js
```diff
@@ -286,6 +286,25 @@ const ImagesTable = () => {
                 );
               })}
           </TableComposable>
+          <Toolbar>
+            <ToolbarContent>
+              <ToolbarItem
+                variant="pagination"
+                align={{ default: 'alignRight' }}
+              >
+                <Pagination
```
jkozol, line 295: Can you add `variant={PaginationVariant.bottom}` so the dropdowns are dropups.

jkozol (changes requested):
Great work. Other than the small change this works as expected :)

src/Components/ImagesTable/ImagesTable.js
regexowl: @jkozol Added! :)

jkozol (approved):

PR #807, opened 2022-09-14
CreateImageWizard: Drop isBeta() check for centos

Gundersanne:
The accompanying text allows us to enable this for non-beta.


lucasgarfield (approved):
Nice!

PR #817, opened 2022-09-21
test/ImagesTable: use redux store and mocked api

Gundersanne:
Instead of supplying the composes directly in the store, let the component call the api.


lucasgarfield (approved):
Very nice, I like this approach much more than what we were doing previously!

lucasgarfield (approved):

PR #819, opened 2022-09-29
Redux: Modernize Redux store

lucasgarfield:
This commit modernizes the Redux store to bring it in line with the current recommended best practices/patterns. It is possible because of a recent previous commit that added Redux Toolkit as a dependency.

For detailed information on what modern Redux entails, see the Redux docs: https://redux.js.org/introduction/getting-started#learn-modern-redux-livestream

Practically speaking, this means a huge reduction in boilerplate code. Maintaining and extending the code will be much easier. All Redux logic is now encapsulated by 'slices'. Reducers are defined in the slices, and action creators and action types are generated automatically.

Redux Toolkit includes Immer, which greatly simplifies writing reducer logic much simpler - state updates in Redux must be immutable, but working with Javascript objects in an immutable fashion is clumsy, requiring gratuitious use of the spread ... operator. With Immer, the state can be updated as if mutable updates were allowed. Reducer logic has been changed to take advantage of this in this commit.

This commit also removes a large amount of unused code. Fingers crossed that nothing breaks. The tests do pass, afterall...

One other minor change... composesGet was renamed to fetchComposes and composeGetStatus was renamed to fetchComposeStatus. This is in line with the Redux documentation examples.


Gundersanne (approved):
Thank you! Removing so much dead code feels good too :)

PR #826, opened 2022-10-06
App: explicitly pass the store to the NotificationsPortal

Gundersanne:
Because of mismatched redux versions between NotificationPortal and our app, the former couldn't automatically discover the store.

Also drop reregister of notificationsReducer, we already register it in store.


Gundersanne:
image
it'll resolve this error

jrusz (approved):

PR #827, opened 2022-10-10
ImagesTable: Indicate expiration of aws.s3 targets

regexowl:
This adds indication of expiring presigned urls for aws.s3 targets. Previously the targets remained marked as `Ready` until the builds disappeared after 14 days.

When an image is ready, the countdown in hours is shown in `Status` column. After that the image is marked as `Expired` and the download link changes to `Recreate image`.

### DIFF
--- a/src/Components/ImagesTable/ImageLinkDirect.js
+++ b/src/Components/ImagesTable/ImageLinkDirect.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useNavigate } from 'react-router-dom';
 
 import {
   Button,
@@ -9,8 +10,10 @@ import {
   TextVariants,
 } from '@patternfly/react-core';
 import { DownloadIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
+import { resolveRelPath } from '../../Utilities/path';
 
 const ImageLinkDirect = (props) => {
+  const navigate = useNavigate();
   const fileExtensions = {
     vsphere: '.vmdk',
     'guest-image': '.qcow2',
@@ -101,19 +104,40 @@ const ImageLinkDirect = (props) => {
         </Popover>
       );
     } else if (uploadStatus.type === 'aws.s3') {
-      return (
-        <Button
-          component="a"
-          target="_blank"
-          variant="link"
-          icon={<DownloadIcon />}
-          iconPosition="right"
-          isInline
-          href={uploadStatus.options.url}
-        >
-          Download {fileExtensions[props.imageType]}
-        </Button>
-      );
+      if (!props.isExpired) {
+        return (
+          <Button
+            component="a"
+            target="_blank"
+            variant="link"
+            icon={<DownloadIcon />}
+            iconPosition="right"
+            isInline
+            href={uploadStatus.options.url}
+          >
+            Download {fileExtensions[props.imageType]}
+          </Button>
+        );
+      } else {
+        return (
+          <Button
+            component="a"
+            target="_blank"
+            variant="link"
+            onClick={() =>
+              navigate(resolveRelPath('/imagewizard'), {
+                state: {
+                  composeRequest: props.recreateImage,
+                  initialStep: 'review',
+                },
+              })
+            }
+            isInline
+          >
+            Recreate image
+          </Button>
+        );
+      }
     }
   }
 
@@ -123,6 +147,8 @@ const ImageLinkDirect = (props) => {
 ImageLinkDirect.propTypes = {
   imageStatus: PropTypes.object,
   imageType: PropTypes.string,
+  isExpired: PropTypes.bool,
+  recreateImage: PropTypes.object,
   uploadOptions: PropTypes.object,
 };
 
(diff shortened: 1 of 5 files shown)


Gundersanne (changes requested):
This works nicely! I do get an error when creating an image however.

`compose.created_at` could be undefined when we first queue a compose.

image

When we create an image, we add the compose immediately in the store. But the data in the store is incomplete:
https://github.com/RedHatInsights/image-builder-frontend/blob/4c1a6c223dff9279cbf87b6f06a44798ad7615e7/src/Components/CreateImageWizard/CreateImageWizard.js#L429-L437

The 2 options I see are making the expiration deal with an undefined imageCreatedAt string. Or we stop inserting the new compose alltogether on creation, and just rely on it being there when we fetch all composes.

The latter would be good probably, but isn't trivial as we currently rely on the api for ordering of the composes in the overview table as well. So we'd need to make changes to the way we add composes to the store.

src/test/Components/ImagesTable/ImagesTable.test.js
```diff
@@ -309,7 +313,7 @@ const mockStatus = {
   },
   '4873fd0f-1851-4b9f-b4fe-4639fce90794': {
     image_status: {
-      status: 'success',
+      status: 'expired',
```
croissanne, line 316: https://github.com/osbuild/image-builder/blob/891b573d509e9ad77e93b9e3ee58f63141322ec9/internal/v1/api.yaml#L371

`expired` is not a state that image builder can return, see the above list.

src/test/Components/ImagesTable/ImagesTable.test.js
```diff
@@ -401,12 +405,19 @@ describe('Images Table', () => {
 
       // render the expected <ImageLink /> and compare the text content for a link
       render(
-        <ImageLink
-          imageStatus={mockStatus[compose.id].image_status}
-          uploadOptions={
-            compose.request.image_requests[0].upload_request.options
-          }
-        />,
+        <BrowserRouter>
```
croissanne, line 408: Why include BrowserRouter

src/test/Components/ImagesTable/ImagesTable.test.js
```diff
@@ -401,12 +405,19 @@ describe('Images Table', () => {
...
+          <ImageLink
+            imageStatus={mockStatus[compose.id].image_status}
+            uploadOptions={
+              compose.request.image_requests[0].upload_request.options
+            }
+            isExpired={
+              mockStatus[compose.id].image_status.status === 'expired'
+                ? true
+                : false
+            }
```
croissanne, line 418: This is now checking the expired state based on the mocked image status; not on the delta to created_at.  I'd check based on the delta since that's what we expect here.

croissanne (changes requested):
Thank you, just some testing questions left.

src/test/Components/ImagesTable/ImagesTable.test.js
regexowl: `ImageLink` component uses `useNavigate` to process recreation of an image. When I've tried to run the tests without a router they kept failing with: `Error: useNavigate() may be used only in the context of a <Router> component.`

croissanne: oh wow, that must've been there for a while then :) Ok that's fine.

src/test/Components/ImagesTable/ImagesTable.test.js
regexowl: Thank you! Understood, the `expiring` and `expired` statuses exist only in the front end and don't map with api 1:1. I've updated the tests to accommodate that.

src/test/Components/ImagesTable/ImagesTable.test.js
regexowl: Updated. In the case of expired images the tests check for a match with a text content.

The tests are now a mix of rendering and text matching. When I've tried to drop the rendering entirely, I've encountered a problem with creating the test output to match against. For a state from the mockupStatus to be mapped correctly, it would need to go through `setComposeStatus()` function in `ImagesTable.js` first and after that through `ImageBuildStatus.js`

src/Components/ImagesTable/ImageLinkDirect.js
```diff
@@ -101,19 +103,40 @@ const ImageLinkDirect = (props) => {
...
+          </Button>
+        );
+      } else {
+        return (
+          <Button
+            component="a"
+            target="_blank"
+            variant="link"
+            onClick={() =>
+              navigate('/imagewizard', {
```
croissanne, line 127: This needs to be `navigate(resolveRelPath('/imagewizard'),...` due to https://github.com/RedHatInsights/image-builder-frontend/pull/833

croissanne (changes requested):
Last comment, small change required due to the way our pathing changed. Otherwise LGTM

src/Components/ImagesTable/ImageLinkDirect.js
regexowl: Got it, fixed :)

croissanne (approved):
Thank you!

PR #828, opened 2022-10-11
Frontend container migration

jrusz:
Adding necessary scripts for migrating our frontend component to container.

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+# name of app-sre "application" folder this component lives in; needs to match for quay
+export COMPONENT="image-builder"
+# IMAGE should match the quay repo set by app.yaml in app-interface
+export IMAGE="quay.io/cloudservices/image-builder-frontend"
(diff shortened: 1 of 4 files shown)


pr_check.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+# name of app-sre "application" folder this component lives in; needs to match for quay
+export COMPONENT="image-builder"
+# IMAGE should match the quay repo set by app.yaml in app-interface
+export IMAGE="quay.io/cloudservices/image-builder-frontend"
```
ochosi, line 9: Out of curiosity, does this container have to be built manually with the build_deploy.sh script? Or will this end up in a GitHub Action?

jrusz: It should be handled by the @app-sre-bot you can see the results of it already in the checks here as `ci.ext.devshift.net PR build` :)

ezr-ondrej: yeah, it's automatically built by jenkins :) `build_deploy.sh` on merge to main and `pr_check.sh` on every PR push.

jrusz:
I think this is good now, worked fine when I tried deploying it. @croissanne could you take a look at this please?

croissanne (approved):
lgtm, thank you!

croissanne (approved):

PR #830, opened 2022-10-13
ImagesTable: Remove the icon from the download link

### LINKED ISSUE #803: Remove the Download Icon
Screen Shot 2022-09-13 at 10 32 41 PM

The download icon was kept on the right-hand side to be consistent with the launch instance icon, but it's an incorrect pattern and will have to be repeated every time we introduce a new primary action.

Most icons associated with actions should be on the left, and the download should too!

regexowl:
Fixes #803. This removes the download icon and adds brackets around the file extension.

### DIFF
--- a/src/Components/ImagesTable/ImageLinkDirect.js
+++ b/src/Components/ImagesTable/ImageLinkDirect.js
@@ -9,7 +9,7 @@ import {
   TextContent,
   TextVariants,
 } from '@patternfly/react-core';
-import { DownloadIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { resolveRelPath } from '../../Utilities/path';
 
 const ImageLinkDirect = ({ uploadStatus, ...props }) => {
@@ -106,12 +106,10 @@ const ImageLinkDirect = ({ uploadStatus, ...props }) => {
           component="a"
           target="_blank"
           variant="link"
-          icon={<DownloadIcon />}
-          iconPosition="right"
           isInline
           href={uploadStatus.options.url}
         >
-          Download {fileExtensions[props.imageType]}
+          Download ({fileExtensions[props.imageType]})
         </Button>
       );
     } else {


src/test/Components/ImagesTable/ImagesTable.test.js
```diff
@@ -421,6 +421,7 @@ describe('Images Table', () => {
           <BrowserRouter>
             <ImageLink
               imageStatus={mockStatus[compose.id].image_status}
+              imageType={compose.request.image_requests[0].image_type}
```
lucasgarfield, line 424: I was able to get the test to pass without this line, what's it for?

regexowl: The tests kept failing without the line, but you're right, they pass now. Seems like changes from PR #831 fixed the previous problem. 

But at the same time I'm newly getting a warning both with and without the line, as `imageName` is now marked as required. I'm not entirely sure why

lucasgarfield: I'm planning on a huge refactor of the <ImagesTable /> component soon which will include hugely simplifying all of the child components (current plan is to only pass in `composeId` as the sole prop and then have the child component itself get everything it needs using Redux selectors). So I'm not super worried about this but it might be worth removing because it's confusing.

Let's do get the warning fixed though.

regexowl: The two ways to fix the warning I can see right now would be providing the tests with an empty strings for the required fields or removing the `isRequired` for the two props in `ImageLink.js` on lines 45 and 46. 

I'm leaning towards the empty strings in tests. Although it adds a bit of a clutter to the testing file the needed functionality of the code wouldn't be changed.

lucasgarfield: Empty strings in the tests will be fine... this will all get cleaned up in the refactor I'm planning anyway (and this is a good example of why the refactor I want to do is needed).

jkozol:
@regexowl Can you rebase this?

regexowl:
> @regexowl Can you rebase this?

@jkozol Rebased

jkozol (approved):

PR #831, opened 2022-10-14
Add the provisioning Button

ezr-ondrej:
This is only draft, it needs error handling and cleanup and documentation on how to test it, but I'm too exited not to share ^_^

[theButton.webm](https://user-images.githubusercontent.com/2884324/195860627-34fb8a18-1f2d-40d9-8194-799760348a6a.webm)


src/Components/ImagesTable/ImageLink.js
```diff
@@ -1,126 +1,45 @@
...
-          component="a"
-          target="_blank"
-          variant="link"
-          icon={<ExternalLinkAltIcon />}
-          iconPosition="right"
-          isInline
-          href={url}
-        >
-          View uploaded image
+  if ((!error && !insights.chrome.isProd()) || insights.chrome.isBeta()) {
```
croissanne, line 20: I think this should also check if the image type/upload type is 'aws' no?

ezr-ondrej: right, good catch :+1:

src/Components/ImagesTable/ImageLink.js
```diff
@@ -1,126 +1,49 @@
...
-      const url =
-        'https://portal.azure.com/#@' +
-        props.uploadOptions.tenant_id +
-        '/resource/subscriptions/' +
-        props.uploadOptions.subscription_id +
-        '/resourceGroups/' +
-        props.uploadOptions.resource_group +
-        '/providers/Microsoft.Compute/images/' +
-        uploadStatus.options.image_name;
+    if (!error) {
```
croissanne, line 23: Wait now all the conditions are gone? Will we get an error if it's the wrong image type or stage-stable?

ezr-ondrej: it's double `if` the upper condition gates by existence of provisioning app in the environment, if the environment lacks provisioning, we'd error out on `useLoadModule`, so that needs to be gated by the provisioning existence.

src/Components/ImagesTable/ImageLink.js
```diff
@@ -1,126 +1,49 @@
...
 
 const ImageLink = (props) => {
-  const fileExtensions = {
-    vsphere: '.vmdk',
-    'guest-image': '.qcow2',
-    'image-installer': '.iso',
-  };
+  const scalprum = useScalprum();
+  const hasProvisionig = scalprum.initialized && scalprum.config?.provisioning;
+  if (hasProvisionig && props.imageType === 'ami') {
```
ezr-ondrej, line 10: first gating makes sure we won't error out on `useLoadModule` which is only safe to use, if the loaded app is in the environment.

src/Components/ImagesTable/ImageLink.js
croissanne: oh that's scalprum? (unfamiliar with it).

croissanne (dismissed):
LGTM, thank you :100:

src/Components/ImagesTable/ImageLink.js
croissanne: lgtm then :)

croissanne (approved):

PR #839, opened 2022-10-21
Aws multiple regions

lucasgarfield:
This PR exposes the API's clone endpoints in the frontend. 

At this time, only AWS images support cloning (or 'sharing', as it is referred to in the frontend). An AWS composes' clones are displayed in a sub-table located in an expandable row.

Composes can be cloned (shared) to other regions via a new modal.

This PR also continues the process of modernizing our Redux store through the use of selectors for data normalization.


croissanne (approved):
lgtm, thank you!

PR #840, opened 2022-10-22
Hide launch button until image is built

ezr-ondrej:
Move the conditional render on the main ImageLink component.
Uses destruct for props to streamline the usage of props.

### DIFF
--- a/src/Components/ImagesTable/ImageLinkDirect.js
+++ b/src/Components/ImagesTable/ImageLinkDirect.js
@@ -12,7 +12,7 @@ import {
 import { DownloadIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { resolveRelPath } from '../../Utilities/path';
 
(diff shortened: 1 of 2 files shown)


src/Components/ImagesTable/ImageLinkDirect.js
```diff
@@ -10,145 +9,115 @@ import {
...
   const fileExtensions = {
     vsphere: '.vmdk',
     'guest-image': '.qcow2',
     'image-installer': '.iso',
   };
 
-  const uploadStatus = props.imageStatus
-    ? props.imageStatus.upload_status
-    : undefined;
-  if (uploadStatus) {
```
ezr-ondrej, line 26: GitHub makes it very difficult to see, but I've just removed the if from here and aligned the code.
Only other change is removal of the `props.`

croissanne (approved):
lgtm, ty!

Issue #805, opened 2022-09-14
Download compose request (.json) should not be a link action


arburka:
While you download a "link" with this action, the action is to download.

The link action is only for navigation. If this action took me to the URL in the .json it would be correct.
<img width="297" alt="Screen Shot 2022-09-13 at 11 32 49 PM" src="https://user-images.githubusercontent.com/88330245/190053882-4de39c45-ab6c-4572-a29d-49f0cc45280b.png">

PR #843, opened 2022-10-25
imagestable: request.json link now non-styled

supakeen:
By skipping the rendering of the <a /> we can still download a data-URI formatted link.

This resolves: #805.


lucasgarfield (approved):
Great job!

PR #845, opened 2022-10-25
CreateImageWizard: fix tile hover issue

### LINKED ISSUE #777: Wizard: Target environment buttons bug
The AWS, GCP, and Azure buttons lose their outlines when selected. That is not the correct behavior, they should retain their outlines. Interestingly, this only occurs when they are selected using the mouse - select them with the keyboard (using the spacebar) and the outline remains.
image
image

kingsleyzissou:
Fix #777: there was an issue with hovering over tiles in the create image wizard due
to a deprecated css variable. This commit resolves the issue


lucasgarfield (approved):
Brilliant.

PR #849, opened 2022-11-01
[Provisioning-wizard] Fix react warning and use AsyncComponent

amirfefer:
This PR fixes the following react warning
```
Warning: React has detected a change in the order of Hooks called by ImageLink. 
This will lead to bugs and errors if not 
fixed. For more information, read the Rules of Hooks
```
caused by a hook that is under a condition

This also adds `AsyncComponent` wrapper for loading federated modules. instead of using `@scalprum/react-core` directly

### DIFF
--- a/src/Components/ImagesTable/ImageLink.js
+++ b/src/Components/ImagesTable/ImageLink.js
@@ -1,7 +1,7 @@
-import React, { Suspense } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import { Button } from '@patternfly/react-core';
-import { useLoadModule, useScalprum } from '@scalprum/react-core';
+import { Button, Spinner } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


src/Components/ImagesTable/ImageLink.js
```diff
@@ -11,41 +11,27 @@ const ImageLink = ({
   imageStatus,
   ...props
 }) => {
-  const scalprum = useScalprum();
-  const hasProvisionig = scalprum.initialized && scalprum.config?.provisioning;
```
ezr-ondrej, line 15: @amirfefer this was kinda important, because in some environment (test and prod ATM) we do not have the provisioning app and thus provisioning manifest present. This check made sure we render the fallback component in those environments.

ezr-ondrej (commented):
I believe we fail on missing provisioning manifest

lucasgarfield:
Closing this as the React warning was resolved by moving the hooks outside of the conditional in 5c37e3b4.

PR #851, opened 2022-11-02
api: use content-sources api for package search

jkozol:
The Content Sources service is now used for handling package info and search. The api call to this service requires a list of repo urls. The `repos.json` file tracks the supported repositories for each distro. This is then mapped into a url array for whichever distro we are searching for packages. The response body is similar to what existed in image-builder but the key `package_name` needs to be mapped to just `name`. Also, there is no pagination in Content Sources for `/rpms/names` so we can currently only fetch a limit of 20 packages.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Packages.js
+++ b/src/Components/CreateImageWizard/formComponents/Packages.js
@@ -108,17 +108,24 @@ const Packages = ({ defaultArch, ...props }) => {
   };
 
   const getAllPackages = async () => {
-    const args = [
-      getState()?.values?.release,
-      getState()?.values?.architecture || defaultArch,
-      packagesSearchName,
-    ];
-    let { data, meta } = await api.getPackages(...args);
-    if (data?.length === meta.count) {
-      return data;
-    } else if (data) {
-      ({ data } = await api.getPackages(...args, meta.count));
-      return data;
+    // if the env is stage beta then use content-sources api
+    // else use image-builder api
+    if (!insights.chrome.isProd() && insights.chrome.isBeta()) {
+      const args = [getState()?.values?.release, packagesSearchName];
+      return await api.getPackagesContentSources(...args);
+    } else {
+      const args = [
+        getState()?.values?.release,
+        getState()?.values?.architecture || defaultArch,
+        packagesSearchName,
+      ];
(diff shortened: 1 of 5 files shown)


src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -108,18 +108,8 @@ const Packages = ({ defaultArch, ...props }) => {
...
-    ];
-    let { data, meta } = await api.getPackages(...args);
-    if (data?.length === meta.count) {
-      return data;
-    } else if (data) {
-      ({ data } = await api.getPackages(...args, meta.count));
-      return data;
-    }
+    const args = [getState()?.values?.release, packagesSearchName];
+    return await api.getPackages(...args);
```
croissanne, line 112: I think we should only use the content-service on stage-beta for now; which can be checked against the insights.chrome object (isBeta && !isProd).

src/repos.json
```diff
@@ -0,0 +1,46 @@
+{
+  "rhel-86": [
```
croissanne, line 2: Let's use the constant [RHEL_8] here, we'll only ever support rhel-8, so might as well.

src/repos.json
```diff
@@ -0,0 +1,46 @@
...
+      "name": "rhel-86-google-compute-engine",
+      "url": "https://packages.cloud.google.com/yum/repos/google-compute-engine-el8-x86_64-stable"
+    },
+    {
+      "distribution_arch": "x86_64",
+      "name": "rhel-86-google-cloud-sdk",
+      "url": "https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64"
+    }
+  ],
+  "rhel-90": [
```
croissanne, line 24: [RHEL-9]

croissanne (changes requested):
lgtm, I do think we should restrict it to beta only though

src/Components/CreateImageWizard/formComponents/Packages.js
jkozol: Not the cleanest since the data responses, and therefore the handling, is different. But, it should differentiate where to call now.

croissanne (approved):
LGTM, thank you!

PR #856, opened 2022-11-07
ImagesTable: Fix Recreate image button

regexowl:
This removes an extra `/` from the path for the `navigate` function to be able to resolve relative path correctly.


regexowl:
@croissanne I've just noticed that the `Recreate image` button for the expired images doesn't work on consoledot, this should fix that.

croissanne (approved):
thank you!

croissanne (approved):

PR #860, opened 2022-11-10
ci: add tags to AWS instances

lavocatt:
The new terraform runners support an extended lists of tags to get a greater precision on the stats we can extract from AWS cost center.


ondrejbudai (approved):

ondrejbudai (approved):

ondrejbudai:
(rebased)

PR #865, opened 2022-11-11
Wizard: Add additional information for an activation key in Review step

### LINKED ISSUE #822: Add additional information for a selected activation key
When user selects an activation key for registration, wizard should show additional information connected to the activation key.

The displayed fields will help user determine if they selected the correct key.

This will be needed for the `Review` and `Registration` (https://github.com/RedHatInsights/image-builder-frontend/issues/603) steps.

regexowl:
Fixes #822. This adds component `ActivationKeyInformation` with additional information for an activation key on a Registration tab of Review step. The added fields are:
- Name
- Role
- SLA
- Usage
- Additional repositories

Organization ID is added to the Activation key popover.

Function `getActivationKey` was added to `api` to get an additional information about activation key from RHSM api based on the name of the key.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
+++ b/src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
@@ -1,0 +1,139 @@
+import React, { useEffect, useState } from 'react';
+import {
+  Text,
+  TextContent,
+  TextList,
+  TextListItem,
+  TextListItemVariants,
+  TextListVariants,
+  TextVariants,
+} from '@patternfly/react-core';
+import { useFormApi } from '@data-driven-forms/react-form-renderer';
+import { Button, Popover } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+import {
+  TableComposable,
+  Tbody,
+  Td,
+  Th,
+  Thead,
+  Tr,
+} from '@patternfly/react-table';
+import api from '../../../api';
+
+const ActivationKeyInformation = () => {
+  const { getState } = useFormApi();
+
+  const activationKey = getState()?.values?.['subscription-activation-key'];
+  const [role, setRole] = useState(undefined);
+  const [serviceLevel, setServiceLevel] = useState(undefined);
+  const [usage, setUsage] = useState(undefined);
+  const [additionalRepositories, setRepositories] = useState(undefined);
+
+  useEffect(() => {
+    const fetchKeyInformation = async () => {
+      const data = await api.getActivationKey(activationKey);
+      setRole(data?.role);
+      setServiceLevel(data?.serviceLevel);
+      setUsage(data?.usage);
+      setRepositories(data?.additionalRepositories);
+    };
+    fetchKeyInformation();
+  }, []);
+
+  return (
+    <>
+      <TextContent>
+        <TextList component={TextListVariants.dl}>
+          <TextListItem component={TextListItemVariants.dt}>Name:</TextListItem>
+          <TextListItem component={TextListItemVariants.dd}>
+            {activationKey}
+          </TextListItem>
+          <TextListItem component={TextListItemVariants.dt}>Role:</TextListItem>
+          <TextListItem component={TextListItemVariants.dd}>
+            {role || 'Not defined'}
+          </TextListItem>
+          <TextListItem component={TextListItemVariants.dt}>SLA:</TextListItem>
+          <TextListItem component={TextListItemVariants.dd}>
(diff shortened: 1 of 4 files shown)


src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
```diff
@@ -0,0 +1,138 @@
...
+                      aria-label="Additional repositories table"
+                      variant="compact"
+                    >
+                      <Thead>
+                        <Tr>
+                          <Th>Name</Th>
+                        </Tr>
+                      </Thead>
+                      <Tbody data-testid="additional-repositories-table">
+                        {additionalRepositories?.map((r, ri) => (
```
lucasgarfield, line 112: This is personal preference but maybe call these (repo, index)? For me, there is some mental overhead required to convert `r` to `repo`. But we don't have a style guide that specifies either way and this is also fine if you prefer it.

lucasgarfield (changes requested):
This is looking great!!!

I think we should probably add test coverage for this before we merge it... The new API call can be mocked, take a look at how we use `jest.spyOn().mockImplementation()` in `ImagesTable.test.js`.

src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
```diff
@@ -0,0 +1,138 @@
...
+            {usage || 'Not defined'}
+          </TextListItem>
+          <TextListItem component={TextListItemVariants.dt}>
+            Additional <br /> repositories:
+            <Popover
+              bodyContent={
+                <TextContent>
+                  <Text>
+                    The core repositories for your operating system version are
+                    always enabled and do not need to be explicitally added to
```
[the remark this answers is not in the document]
regexowl: Fixed!

src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
regexowl: Fair, updated for better readability :)

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -85,6 +88,7 @@ const ReviewStep = () => {
       (async () => {
         const userData = await insights?.chrome?.auth?.getUser();
         const id = userData?.identity?.internal?.org_id;
+        setOrgId(id);
```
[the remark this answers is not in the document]
regexowl: Updated

regexowl:
@lucasgarfield Thanks for the comments! I've updated the code and added test coverage for the shown additional information.

src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
```diff
@@ -0,0 +1,139 @@
...
+              >
+                <HelpIcon />
+              </Button>
+            </Popover>
+          </TextListItem>
+          <TextListItem
+            component={TextListItemVariants.dd}
+            className="pf-u-display-flex pf-u-align-items-flex-end"
+          >
+            {additionalRepositories?.length === 0 ? (
```
lucasgarfield, line 93: If `additionalRepositories` is `undefined` for whatever reason, then we will render the `<Popover>` component... which I don't think we want. What if this was something like: `additionalRepositories?.length > 0 ? (<Popover ... />) : ('None')`? That way, we only render the popover with the additional repositories table if we are 100% sure there are additional repositories.

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -315,22 +317,34 @@ const ReviewStep = () => {
...
+                            {getState()?.values?.[
+                              'subscription-organization-id'
+                            ] !== undefined ? (
+                              `Your organization's ID is ${
+                                getState()?.values?.[
+                                  'subscription-organization-id'
+                                ]
+                              }`
+                            ) : (
+                              <Spinner />
```
lucasgarfield, line 330: If you disable the line of code that sets `'subscription-organization-id'` so you can see what happens if it is `undefined`, the popover looks like this:

spinner

What if we tried something like `Your organization's ID is <Spinner size="sm" />.`?

src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
regexowl: Fixed :+1:

lucasgarfield (approved):

PR #867, opened 2022-11-15
src/constants: switch to rhel-87 and rhel-91

croissanne:
- [ ] ib production has support for 87 and 91


jkozol (approved):
Builds as expected :)

But, I think we should create issues to discuss how to treat old versions. As you can see in this screenshot, the label for RHEL 9.0 turns from `RHEL 9` to `rhel-90` and same for 86.

Screenshot from 2022-11-15 20-57-58



Also, the old images can still be recreated but there is no notice of the distro change.

Screenshot from 2022-11-15 20-57-14

PR #869, opened 2022-11-16
DocumentationButton: fix documentation url

croissanne:
The old URL pointed to the on-prem documentation, not the service docs.


ochosi (approved):
LGTM!

PR #873, opened 2022-11-22
Add provisioning-fronternd integration script

amirfefer:
The provisioning-wizard is a separate frontend [app](https://github.com/RHEnVision/provisioning-frontend) which is consumed for launching images in image-builder. In some cases, developers would like to develop and integrate both apps at once.

run `npm run dev:hms`
this script will
    -  git clone `provisioning-frontend` and `cloud-services-config` to parent directory (if needed)
    -  copy files and install npm modules
    -  add local routes to webpack-proxy 
    -  run `provisioning-frontend` as federated module
    -  run http server to serve the cloud-config locally
    -  run image builder in stage-beta

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,19 @@ The UI should be running on
 https://prod.foo.redhat.com:1337/beta/insights/image-builder/landing.
 Note that this requires you to have access to either production or stage (plus VPN and proxy config) of insights.
 
+### Develop with provisioning-frontend
+
+The provisioning-wizard is a separate frontend [app](https://github.com/RHEnVision/provisioning-frontend) which is consumed for launching images in image-builder. In some cases, developers would like to develop and integrate both apps at once.
+
+run `npm run dev:hms`
+this script will
+    -  git clone `provisioning-frontend` and `cloud-services-config` to parent directory (if needed)
+    -  copy files and install npm modules
+    -  add local routes to webpack-proxy 
+    -  run `provisioning-frontend` as federated module
+    -  run http server to serve the cloud-config locally
+    -  run image builder in stage-beta
+ 
 ## Backend Development
 
 To develop both the frontend and the backend you can again use the proxy to run both the
--- a/package.json
+++ b/package.json
@@ -95,7 +95,8 @@
     "test": "TZ=UTC jest --verbose --no-cache",
     "test:single": "jest --verbose -w 1",
     "build": "webpack --config config/prod.webpack.config.js",
-    "verify": "npm-run-all build lint test"
+    "verify": "npm-run-all build lint test",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -95,7 +95,8 @@
     "test": "TZ=UTC jest --verbose --no-cache",
     "test:single": "jest --verbose -w 1",
     "build": "webpack --config config/prod.webpack.config.js",
-    "verify": "npm-run-all build lint test"
+    "verify": "npm-run-all build lint test",
+    "dev:provisioning": "DEV_ROUTES_PATH=config/routes.dev.config.js ./devel/provisioning.sh"
```
ezr-ondrej, line 99: Could we make the routes configuration optional? maybe on a flag, or two npm scripts?
I see value in both consuming the local backend and backend from stage, but if I were to choose, I'd go with using the stage backend and only documenting usage of `DEV_ROUTES_PATH`. If documented I think it's reasonable to be running:
`DEV_ROUTES_PATH=config/routes.dev.config.js npm run dev:provisioning`
to connect to local backends.

ezr-ondrej (commented):
I'd love to have this, tho I'd call it `dev:hms` instead and make the routes to local backend optional.

package.json
amirfefer: the stagde backend is by default, we can add in a follow-up PR a flag for routing the local API as well

README.md
```diff
@@ -71,6 +71,19 @@ The UI should be running on
 https://prod.foo.redhat.com:1337/beta/insights/image-builder/landing.
 Note that this requires you to have access to either production or stage (plus VPN and proxy config) of insights.
 
+### Develop with provisioning-frontend
+
+The provisioning-wizard is a separate frontend [app](https://github.com/RHEnVision/provisioning-frontend) which is consumed for launching images in image-builder. In some cases, developers would like to develop and integrate both apps at once.
+
+run `npm run dev:provisioning`
```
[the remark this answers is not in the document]
amirfefer: @ezr-ondrej, changed to `dev:hms` :)

ezr-ondrej:
Nice, works for me, it's quite opinionated, but might be helpful.
@croissanne WDYT? I'm happy to have this in provisioning-frontend :)

croissanne (changes requested):
2 small comments

ideally we can integrate this setup with the docker compuse setup in devel as well, though we'd need sources running there too. That way it would be possible to run the whole image-builder stack + provisioning.

ezr-ondrej:
@amirfefer can we close this now, given we are going the container compose way? :)

lucasgarfield:
Stale PR, closing.

PR #878, opened 2022-12-01
CreateImageWizard: custom buttons for each step

croissanne:
This adds custom buttons for each step, where each button has a id taking the format `$step-(next|previous|cancel)-button`.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/CustomButtons.js
+++ b/src/Components/CreateImageWizard/formComponents/CustomButtons.js
@@ -4,14 +4,36 @@ import { FormSpy } from '@data-driven-forms/react-form-renderer';
 import WizardContext from '@data-driven-forms/react-form-renderer/wizard-context';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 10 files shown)


lucasgarfield:
Next button size changes when going from registration to file system configuration, haven't started investigating why yet but it jumped out at me: 

next button size

croissanne:
> Next button size changes when going from registration to file system configuration, haven't started investigating why yet but it jumped out at me:

it's the same component but somehow the filesystem one is lacking pf-m-progress O.o

lucasgarfield:
Okay this is really nit-picky but it kind of bugs me that the `Next` button ends up being wider than the `Back` button for no real reason - after all, the loading spinner will appear when the button says `Create image`, not `Next`... In the Patternfly example they are the same size, after all.

Here's an alternate solution I think I like more - pass in `null` if we aren't on the review step. This makes the `Back` and `Next` the same size. Obviously this also requires removing the `isLoading` prop from the FSC component.

image

lucasgarfield (approved):
Nice! I bet this is going to end up being super useful ;) !!!

PR #880, opened 2022-12-01
fix(HMSPROV-377) - disable image's launch button when image is not ready

amirfefer:
before: 
Screen Shot 2022-12-01 at 18 44 46

after:
Screen Shot 2022-12-01 at 18 42 51

### DIFF
--- a/src/Components/ImagesTable/ImageLink.js
+++ b/src/Components/ImagesTable/ImageLink.js
@@ -26,10 +26,15 @@ const ProvisioningLink = ({ imageId, isExpired, isInClonesTable }) => {
   );
 
   if (!error) {
-    image = image.isClone ? parent : image;
+    const provisionedImage = image.isClone ? parent : image;
     return (
       <Suspense fallback="loading">
-        <Button variant="link" isInline onClick={() => openWizard(true)}>
+        <Button
+          isDisabled={image.status !== 'success'}
+          variant="link"
+          isInline
+          onClick={() => openWizard(true)}
+        >
           Launch
         </Button>
         {wizardOpen && (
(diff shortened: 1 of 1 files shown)


src/Components/ImagesTable/ImageLink.js
```diff
@@ -26,20 +26,25 @@ const ProvisioningLink = ({ imageId, isExpired, isInClonesTable }) => {
   );
 
   if (!error) {
-    image = image.isClone ? parent : image;
+    const provisionedImage = image.isClone ? parent : image;
     return (
       <Suspense fallback="loading">
-        <Button variant="link" isInline onClick={() => openWizard(true)}>
+        <Button
+          isDisabled={image.status !== 'success'}
```
croissanne, line 33: Can we move this to/solve this in ImageLink directly? It currently returns null if there's no uploadstatus, but might as well solve it for all cases, and require that the uploadstatus is defined and has succeeded.

croissanne (changes requested):
One comment, otherwise lgtm.

Currently we don't disable the button, we just omit it in case it's not ready. Let's just omit it in case of failure as well.

src/Components/ImagesTable/ImageLink.js
```diff
@@ -26,20 +26,25 @@ const ProvisioningLink = ({ imageId, isExpired, isInClonesTable }) => {
   );
 
   if (!error) {
-    image = image.isClone ? parent : image;
+    const provisionedImage = image.isClone ? parent : image;
```
lucasgarfield, line 29: This is a super nit-pick but now that we don't re-assign `image` can we make it a `const` (on line 11)?

ezr-ondrej:
Implementing in #954 it was a one liner :)

ezr-ondrej:
#954 is in, so I guess we can close this one @lucasgarfield ? :)

PR #883, opened 2022-12-05
CreateImageWizard: use 'aws' image type instead of 'ami'

croissanne:
'ami' was deprecated a while ago, and is kept around as an alias of 'aws'.


jkozol (dismissed):
Seems to work well. Does the api send an image type of `aws` and `azure` back? If so we should change all of the `ami` and `vhd` type checks in the the store and components.

croissanne:
> Seems to work well. Does the api send an image type of `aws` and `azure` back? If so we should change all of the `ami` and `vhd` type checks in the the store and components.

It sends  back whatever was requested,  and woops, you're right; marking as draft :facepalm:

teg (dismissed):

jkozol (approved):

PR #888, opened 2022-12-20
Content sources

lucasgarfield:
Adds support for Content Sources.


jkozol:
Screenshot from 2022-12-22 16-11-58

I don't think we should list the rhel subscriptions as third party.

jkozol:
Screenshot from 2022-12-22 16-17-09

If packages are selected and then the third party sources are removed should we remove the packages from third party?

croissanne:
> Screenshot from 2022-12-22 16-11-58
> 
> I don't think we should list the rhel subscriptions as third party.

Did you define these yourself in settings/content?

croissanne (approved):

croissanne:
Amazing, let's hold off on merging until next year though :) Thank you <3

lucasgarfield:
> Screenshot from 2022-12-22 16-17-09
> 
> If packages are selected and then the third party sources are removed should we remove the packages from third party?

This is tricky. Neither image-builder nor osbuild-composer actually maps packages to a specific repository. The request schema (and therefore the JSON blob we work with) has a field for package names, and a field for repositories (under payload.customizations). So while it would be a nice feature to have, it is not straightforward to implement.

lucasgarfield:
> Screenshot from 2022-12-22 16-11-58
> 
> I don't think we should list the rhel subscriptions as third party.

This will be fixed by #894

PR #895, opened 2023-01-03
Wizard: Remove clear handler from Release select on Image Output step

### LINKED ISSUE #893: Clearing the release select doesn't seem to clear the selection and is confusing
In the "Create Image modal" in the first step where the "Release" can be selected there is one option already pre-selected. There is a "clear button" next to it which when clicked doesn't seem to actually clear the selection as it's still shown but prevents clicking on "Next" and also the "Registration" step is hidden.
image

And after clicking the clear button it looks like this:
image

So the Release still seems to be selected but I am unable to continue which seems confusing. Maybe it would make sense to hide the "clear button"? Or at least somehow show that nothing is selected there once cleared.

lucasgarfield:
This commit fixes #893 by removing the ability to clear the release select on the image output step (this is the select where the user chooses from RHEL 8, RHEL 9, CentOS 8 or 9, etc...).

Originally the ability to clear the select was added so that the default state (where CentOS options are hidden behind a 'speedbump') could be returned to but this seems unnecessary in retrospect and introduces opportunities for bugs (as seen with the validation in #893) so it seems wise to simply remove the ability to clear altogether.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ImageOutputReleaseSelect.js
+++ b/src/Components/CreateImageWizard/formComponents/ImageOutputReleaseSelect.js
@@ -24,11 +24,6 @@ const ImageOutputReleaseSelect = ({ label, isRequired, ...props }) => {
     setIsOpen(false);
   };
 
-  const handleClear = () => {
-    change(input.name, null);
-    setShowDevelopmentOptions(false);
-  };
-
   const handleExpand = () => {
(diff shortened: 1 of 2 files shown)


regexowl (approved):
Nice catch, removing the clear handler makes perfect sense :+1:

jrusz:
Amazing, thanks for the quick fix!

PR #900, opened 2023-01-05
Revert "build(deps): bump react-router-dom from 6.4.2 to 6.6.1"

lucasgarfield:
Reverts RedHatInsights/image-builder-frontend#892

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "react": "17.0.2",
         "react-dom": "17.0.2",
         "react-redux": "8.0.5",
-        "react-router-dom": "6.6.1",
+        "react-router-dom": "6.4.2",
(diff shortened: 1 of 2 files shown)


teg (approved):
Would be good to add a test (in a follow-up!) to avoid this (whatever it was) happening again. Also, the PR description should mention that this was a regression, or other details you might know - for posterity.

lucasgarfield:
> Would be good to add a test (in a follow-up!) to avoid this (whatever it was) happening again. Also, the PR description should mention that this was a regression, or other details you might know - for posterity.

Ah, good point. I don't think I can easily change the commit message on this PR because it's from dependabot's branch... next time. I'll at least make an issue with the console error messages so we can figure out why this broke stage beta.

PR #901, opened 2023-01-05
ImagesTable: Fix status bug

### LINKED ISSUE #899: Sharing an image to a new region also resets the status of the original image
image
The image in us-east-1 should have had a ready status.

lucasgarfield:
Fixes #899. The status of images (parent images and clones) in the clones table was displayed incorrectly - the 'highest priority' (e.g. failure > success) was displayed for all images.

This was due to a bug in a conditional in the ImageBuildStatus component. In the main images table, rows for AWS images should display the highest priority status of *all* images. A single failed clone should cause the status of the row in the main images table to be failure, even if the parent compose status is successful.

This logic was incorrectly being applied to *all* statuses. This commit fixes this - from now on, this logic is only used for rows in the main images table.

### DIFF
--- a/src/Components/ImagesTable/ImageBuildStatus.js
+++ b/src/Components/ImagesTable/ImageBuildStatus.js
@@ -20,7 +20,7 @@ import {
 } from '../../store/composesSlice';
 import { hoursToExpiration } from '../../Utilities/time';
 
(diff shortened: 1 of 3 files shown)


regexowl (approved):
LGTM! :+1:

PR #897, opened 2023-01-03
Enable provisioning wizard in Ephemeral environment

ezr-ondrej:

PR #926, opened 2023-01-12
Enable provisioning wizard in Ephemeral environment

lucasgarfield:
Supersedes #897 

Fixed a minor typo, `initiated` to `initialized` that was blocking merging. Not having provisioning in the ephemeral environment is a blocker for @akhil-jha.

### DIFF
--- a/src/Components/ImagesTable/ImageLink.js
+++ b/src/Components/ImagesTable/ImageLink.js
@@ -1,6 +1,7 @@
 import React, { Suspense, useState } from 'react';
 
 import { Button } from '@patternfly/react-core';
+import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 import { useLoadModule, useScalprum } from '@scalprum/react-core';
 import PropTypes from 'prop-types';
 import { useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


akhil-jha:
Just giving more info, I can create images in ephemeral, it's just I cannot launch provisioning wizard. `Launch` opens up the aws page right now. :)

ezr-ondrej:
Sorry, I tried to do the parent PR from vacation, not a good idea :/
So to add more details, the `useChrome()`  and `useScalprum()` are new react hooks and some documentation can be found here: http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/chrome/chrome-api#Chrome
I needed to be told about these by the frontend team and then read source code anyway to understand how to use it, but I think it does what we want now.

I'm hoping it's temporary as we might not want to limit on isBeta() in the future and thus we will be able to simplify the condition.

Thanks @lucasgarfield for taking this over! :orange_heart:

lucasgarfield:
No worries @ezr-ondrej! I figured you were probably just busy getting caught up after the break. Thanks for the insights into `useChrome()` and `useScalprum()`... One thing I did notice was that with a recent upgrade to `@redhat-cloud-services` we are now getting console warnings about `insights.chrome` use being deprecated (and to replace it with `useChrome()`).

croissanne (approved):

PR #930, opened 2023-01-13
Wizard: Update order of releases on the Image output step

### LINKED ISSUE #909: Release order in Image output should be reversed
Currently, the order of the releases on the Image output screens goes from oldest to newest. Should these not go from newest to oldest?
Screen Shot 2023-01-09 at 2 19 14 PM

regexowl:
Fixes https://github.com/RedHatInsights/image-builder-frontend/issues/909. Fixes https://github.com/RedHatInsights/image-builder-frontend/issues/918.

This updates the order of the releases on the Image output step to sort from newest to oldest. The RELEASES Object was converted to Map which remembers insertion order.

The color of the `Show options for further development of RHEL` option in the Release dropdown select was also changed to blue.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -8,12 +8,12 @@ export const UNIT_KIB = 1024 ** 1;
 export const UNIT_MIB = 1024 ** 2;
 export const UNIT_GIB = 1024 ** 3;
 
-export const RELEASES = {
-  [RHEL_8]: 'Red Hat Enterprise Linux (RHEL) 8',
-  [RHEL_9]: 'Red Hat Enterprise Linux (RHEL) 9',
-  'centos-8': 'CentOS Stream 8',
-  'centos-9': 'CentOS Stream 9',
-};
+export const RELEASES = new Map([
+  [RHEL_9, 'Red Hat Enterprise Linux (RHEL) 9'],
+  [RHEL_8, 'Red Hat Enterprise Linux (RHEL) 8'],
+  ['centos-9', 'CentOS Stream 9'],
+  ['centos-8', 'CentOS Stream 8'],
+]);
 
 export const DEFAULT_AWS_REGION = 'us-east-1';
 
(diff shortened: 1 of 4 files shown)


src/constants.js
```diff
@@ -9,10 +9,10 @@ export const UNIT_MIB = 1024 ** 2;
 export const UNIT_GIB = 1024 ** 3;
 
 export const RELEASES = {
-  [RHEL_8]: 'Red Hat Enterprise Linux (RHEL) 8',
   [RHEL_9]: 'Red Hat Enterprise Linux (RHEL) 9',
-  'centos-8': 'CentOS Stream 8',
+  [RHEL_8]: 'Red Hat Enterprise Linux (RHEL) 8',
   'centos-9': 'CentOS Stream 9',
+  'centos-8': 'CentOS Stream 8',
```
croissanne, line 15: I'm not sure we should rely on the insertion order of keys in an object. I think theoretically for strings it should work, but most js implementations will reorder keys which are integers.

croissanne: https://github.com/tc39/proposal-for-in-order so as long as we use for...in there is a guaranteed ordering from 2020 it seems.

lucasgarfield: `Map` objects preserve insertion order. I'm not sure if it would work as a drop in replacement for the current object but it might be a bit cleaner overall than trying to implement some funky sorting logic in the actual components.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

regexowl: Hmm, will give the issue another look. I like the idea of changing the object to the map one, so I'll look into that. I'll switch this PR to draft for now.

regexowl: Alright, the RELEASES Object was converted to Map. When we want to add another release option in the future, the order in which we add it should be the very same as that which will be shown in the Release dropdown menu.

croissanne (approved):

PR #945, opened 2023-01-25
Aws sources

lucasgarfield:
This PR adds the ability to configure an AWS target using the Sources service available on Insights instead of manually entering an AWS account ID. 

Note that the Sources API is not queried directly, rather the Provisioning API provides endpoints to obtain sources and their associated information (such as the account id). The Provisioning endpoints are available via RTK Query hooks, and mocked in the tests using Mock Service Worker.

### DIFF
--- a/src/store/apiSlice.js
+++ b/src/store/apiSlice.js
@@ -6,10 +6,42 @@ export const apiSlice = createApi({
   reducerPath: 'api',
   baseQuery: fetchBaseQuery({ baseUrl: '' }),
   endpoints: (builder) => ({
-    getSources: builder.query({
-      query: () => PROVISIONING_SOURCES_ENDPOINT,
+    getAWSSources: builder.query({
+      async queryFn(_arg, _queryApi, _extraOptions, fetchWithBQ) {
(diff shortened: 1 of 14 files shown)


src/store/apiSlice.js
```diff
@@ -6,10 +6,42 @@ export const apiSlice = createApi({
...
+        // requests so that the React components can call simply call a single hook,
+        // useGetAWSSourcesQuery().
+
+        const awsSources = await fetchWithBQ(
+          `${PROVISIONING_SOURCES_ENDPOINT}?provider=aws`
+        );
+
+        if (awsSources.error) return { error: awsSources.error };
+
+        const awsAccountIds = await Promise.allSettled(
```
lucasgarfield, line 23: I struggled how to decide between using Promise.all() or Promise.allSettled() here... any thoughts?

croissanne (commented):
As discussed off-thread, I think gating this behind beta would be nice until we get the overview + clones sorted out.

Other than that, this is fantastic :D Thank you!

croissanne (approved):
lgtm!

PR #952, opened 2023-02-07
Dev Deps: Bump Jest to 29.4.1

lucasgarfield:
This commit bumps Jest to 29.4.1. This requires adding two additional development dependencies.

As of Jest 28 `jest-environment-jsdom` is no longer shipped with Jest by default, and has therefore been added as a dependency.

`uuid` was also added as a dev dependency. Jest 29 introduced some compatibility issues with `uuid` that were fixed in `uuid@9`: https://github.com/uuidjs/uuid/issues/451#issuecomment-1206284480

We import and use `uuid` in our tests, but as it was a dependencies dependency we did not control which version was installed. Version 8 (what was installed) causes Jest tests to fail like so:
```
    SyntaxError: Unexpected token 'export'

    > 29 | import { v4 as uuidv4 } from 'uuid';
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,8 +52,9 @@
         "git-revision-webpack-plugin": "5.0.0",
         "history": "5.3.0",
         "identity-obj-proxy": "3.0.0",
-        "jest": "27.5.1",
+        "jest": "29.4.1",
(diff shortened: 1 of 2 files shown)


croissanne (approved):
thanks!

PR #954, opened 2023-02-07
Do not show Launch button on copied images

ezr-ondrej:
This is touching the same code as #880.

I might include that PR here as well if @amirfefer won't object.


lucasgarfield:
Adding #880 here seems like a great idea :+1:

ezr-ondrej:
> Adding #880 here seems like a great idea +1

Added as second commit, souded as simple one liner, but not sure if I did not miss anything :)

ezr-ondrej:
> will **increase** coverage by `0.09%`

Oh yeah :tada: I love removing code xD

ezr-ondrej:
Did I brek it? :)

src/Components/ImagesTable/ImageLink.js
```diff
@@ -9,13 +9,9 @@ import { useSelector } from 'react-redux';
 import ImageLinkDirect from './ImageLinkDirect';
 
 import { selectImageById } from '../../store/composesSlice';
-import { selectComposeById } from '../../store/composesSlice';
 
 const ProvisioningLink = ({ imageId, isExpired, isInClonesTable }) => {
   let image = useSelector((state) => selectImageById(state, imageId));
```
lucasgarfield, line 14: Let's declare `image` as a `const` now that we no longer reassign it.

lucasgarfield (changes requested):
Looks and works great, thanks Ondrej! Just one minor nit-picky request.

src/Components/ImagesTable/ImageLink.js
ezr-ondrej: sure! nice find :)
I wonder why linter is not bugging me for that, I'd expect this to be lintable.

lucasgarfield: Because we don't have that rule in our eslintrc :wink: good point, I'll add it.

lucasgarfield (approved):

PR #958, opened 2023-02-09
ESLint: Add `prefer-const` rule

lucasgarfield:
Require const declarations for variables that are never reassigned after being declared. If a variable is never reassigned, using the const declaration is better. const declaration tells readers, “this variable is never reassigned,” reducing cognitive load and improving maintainability.


regexowl (approved):
Nice :+1:

PR #959, opened 2023-02-09
config: update module name in webpack config

croissanne:
Due to merging the openshift and chrome plugin systems there's additional requirements on the format of the module name.

PR #960, opened 2023-02-09
config: remove module name in webpack config (#959)

Hyperkid123:
Due to merging the openshift and chrome plugin systems there's additional requirements on the format of the module name.

Camelcase is the default, so the values can just be removed from the config.


lucasgarfield (dismissed):

lucasgarfield:
Disabling auto merge because I want to double check something before we merge.

croissanne:
Don't make merge requests straight to the prod-beta branch, use the sync-branches workflow.

PR #949, opened 2023-02-02
Show 'Open Service' badge and popover

ochosi:
As part of our 'Open Source Services' initiative, we'll add a badge that indicates that our service is open. The popover will give additional context as well as a link to our architectural docs.

image

PR #961, opened 2023-02-09
WIP: Add open source alert

ochosi:
This is based on top of #949. It adds a dismissable alert to the landing page of the service that highlights the new badge.

1. Not sure if this is the way to properly close the alert.
2. I'm waiting for feedback from UX if dismissing the alert should dismiss it 'forever' for a particular user.

Related to: COMPOSER-1899

### DIFF
--- a/src/Components/LandingPage/LandingPage.js
+++ b/src/Components/LandingPage/LandingPage.js
@@ -2,7 +2,7 @@
 
 import React, { Component } from 'react';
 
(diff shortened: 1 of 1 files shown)


ochosi:
So the requirements for the banner are as follows:
1. Should be dismissable (have a close button)
2. Should not show up again once dismissed
3. Should not show up again after the session ends (so only visible for one session per user)

ochosi:
Andy suggested he would be able to do this with Pendo, so let's close this one.

PR #963, opened 2023-02-10
Wizard: Get distribution repositories from an api endpoint

regexowl:
Previously the base repository links for each distribution were read from `repos.js` file. This gets the needed information from a call to api endpoint `architectures/{distribution}`. The data is fetched via RTK Query and is then filtered by an architecture. For now the x86_64 architecture is hardcoded, this will change when multiple architectures get available.

### DIFF
--- a/src/store/apiSlice.js
+++ b/src/store/apiSlice.js
@@ -1,6 +1,6 @@
 import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react';
 
-import { PROVISIONING_SOURCES_ENDPOINT } from '../constants';
+import { IMAGE_BUILDER_API, PROVISIONING_SOURCES_ENDPOINT } from '../constants';
 
 export const apiSlice = createApi({
   reducerPath: 'api',
@@ -41,7 +41,15 @@ export const apiSlice = createApi({
         return awsSources;
       },
     }),
+    getArchitecturesByDistribution: builder.query({
+      query: (distribution) =>
+        `${IMAGE_BUILDER_API}/architectures/${distribution}`,
+    }),
   }),
 });
 
-export const { useGetAWSSourcesQuery, usePrefetch } = apiSlice;
+export const {
+  useGetAWSSourcesQuery,
+  useGetArchitecturesByDistributionQuery,
+  usePrefetch,
+} = apiSlice;
--- a/src/test/mocks/handlers.js
+++ b/src/test/mocks/handlers.js
@@ -44,4 +44,154 @@ export const handlers = [
       }
     }
   ),
+  rest.post(
+    baseURL.concat('/api/content-sources/v1/rpms/names'),
+    async (req, res, ctx) => {
+      const { search } = await req.json();
+      if (search === 'test') {
+        return res(
+          ctx.status(200),
+          ctx.json([
+            {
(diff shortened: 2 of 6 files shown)


src/store/apiSlice.js
```diff
@@ -7,7 +7,12 @@ export const apiSlice = createApi({
     getSources: builder.query({
       query: () => '/provisioning/v1/sources',
     }),
+    getArchitecturesByDistribution: builder.query({
+      query: (distribution) =>
+        `/image-builder/v1/architectures/${distribution}`,
```
lucasgarfield, line 12: This is absolutely correct, but just want to give a warning that in #953 I changed the baseUrl from '/api' to '' so this will need to become '/api/image-builder/...' if #953 gets merged or if you rebase onto #953.

regexowl:
Alright, this one is finally ready for review!

Note about tests:
To mock the api response the tests are now using `msw` instead of previously used jest `mockImplementation`. That means that at the same time, the new way of api mocking and the old way are now coexisting within the test files.

This was causing problems in tests:
- Step Packages -> chosen packages can be reset after filtering
- Click through all steps -> with valid values

both in beta and non-beta tests as the `mockImplementation` was overriding the results of `msw`. For this reason I've updated values the tests are looking for in these cases to fit the mocked results of 
```javascript
const getPackages = jest

.spyOn(api, 'getPackagesContentSources')

.mockImplementation(() =>

Promise.resolve(mockPkgResultAlphaContentSources)

);
```

As the tests are checking for the presence of a package and are not working with it further this shouldn't have any effect on their reliability.

Unification of the mocks might be an issue on its own, what do you think?

src/store/apiSlice.js
```diff
@@ -41,7 +41,15 @@ export const apiSlice = createApi({
         return awsSources;
       },
     }),
+    getArchitecturesByDistribution: builder.query({
```
croissanne, line 44: Mostly because i don't really understand, what's the difference between adding it under `endpoints:` and adding it as a field directly? 

Reading https://redux-toolkit.js.org/rtk-query/api/createApi it seems they add all queries under `endpoints`.

croissanne: oh wait nvm, I can't read. It's added under endpoints :facepalm:

src/test/mocks/handlers.js
```diff
@@ -44,4 +44,154 @@ export const handlers = [
...
+        return res(
+          ctx.status(200),
+          ctx.json([
+            {
+              arch: 'x86_64',
+              repositories: [
+                {
+                  baseurl:
+                    'http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/',
+                  rhsm: true,
```
croissanne, line 178: nitpick, centos doesn't require rhsm (not really important here since it's just mocks).

croissanne (approved):
1 nitpick, lgtm otherwise! Thank you! We should maybe think about switching over to content sources altogether soon, so we can drop those repos finally.

src/test/mocks/handlers.js
regexowl: Thanks! Updated and I'll keep that in mind :+1:

croissanne (approved):
lgtm!

PR #965, opened 2023-02-13
feat(HMS-776): Add provider to provisioning wizard

amirfefer:
needed for https://github.com/RHEnVision/provisioning-frontend/pull/195

### DIFF
--- a/src/Components/ImagesTable/ImageLink.js
+++ b/src/Components/ImagesTable/ImageLink.js
@@ -10,6 +10,15 @@ import ImageLinkDirect from './ImageLinkDirect';
 
 import { selectImageById } from '../../store/composesSlice';
 
+const getImageProvider = ({ imageType }) => {
+  switch (imageType) {
+    case 'aws' || 'ami':
+      return 'aws';
+    default:
+      'aws';
+  }
+};
+
 const ProvisioningLink = ({ imageId, isExpired, isInClonesTable }) => {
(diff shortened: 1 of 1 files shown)


src/Components/ImagesTable/ImageLink.js
```diff
@@ -74,7 +75,9 @@ const ImageLink = ({ imageId, isExpired, isInClonesTable }) => {
 
   if (
     hasProvisioning &&
-    (image.imageType === 'aws' || image.imageType === 'ami')
+    (image.imageType === 'gcp' ||
```
ezr-ondrej, line 78: @amirfefer I don't think this we want to expose it to everyone yet, this is not done feature, it needs quite some work to be completed. could we do it only if a feature flag is enabled?

ezr-ondrej: I'd vote for removing the gcp for now, just adding the providerType so we are sure Provisioning Wizard does not crash.

amirfefer: agree, i removed the gcp for now, adding only `provider`
in a follow-up PR , i'll add a feature flag for gcp provider.

ezr-ondrej (dismissed):
LGTM :+1:

@croissanne @lucasgarfield could you take a look pls? :)

ezr-ondrej (approved):

croissanne (approved):

PR #966, opened 2023-02-16
src/repos: Fix urls

croissanne:
These URLs are used for package searching with the content-service.

- Remove the google cloud repos, these are added only for gcp
- Fix centos 9 URLs

### DIFF
--- a/src/repos.js
+++ b/src/repos.js
@@ -15,16 +15,6 @@ export const repos = {
       name: 'appstream',
       url: 'https://cdn.redhat.com/content/dist/rhel8/8.7/x86_64/appstream/os',
     },
-    {
-      distribution_arch: 'x86_64',
-      name: 'google-compute-engine',
-      url: 'https://packages.cloud.google.com/yum/repos/google-compute-engine-el8-x86_64-stable',
-    },
-    {
-      distribution_arch: 'x86_64',
-      name: 'rhel-86-google-cloud-sdk',
-      url: 'https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64',
-    },
   ],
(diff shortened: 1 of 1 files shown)


regexowl (dismissed):

ondrejbudai (approved):
Thanks!

src/repos.js
```diff
@@ -64,42 +44,17 @@ export const repos = {
...
-    {
-      name: 'google-compute-engine',
-      distribution_arch: 'x86_64',
-      url: 'https://packages.cloud.google.com/yum/repos/google-compute-engine-el9-x86_64-stable',
-    },
-    {
-      name: 'google-cloud-sdk',
-      distribution_arch: 'x86_64',
-      url: 'https://packages.cloud.google.com/yum/repos/cloud-sdk-el9-x86_64',
+      url: 'https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/',
```
regexowl, line 57: I'm getting "No packages found" with `https`, but with `http` it seems to work

croissanne: ugh, pushed local changes when fixing the linter error :facepalm:

teg: Very odd! Accessing the URL directly works for me.

croissanne: It's more that the content service saved it as http, and doesn't recognize the https url => no packages if the https url is used.

regexowl (approved):

src/repos.js
teg: Ah!

teg: We probably should avoid non https by default, no? At least if https is available. Is Justin aware of this quirk?

PR #970, opened 2023-02-22
Jest: Add mock service worker server commands to jest.setup.js

lucasgarfield:
This commit makes the `msw` server globally available to all test files. Given that the majority of our test files will ultimately make calls to APIs using RTK Query (calls are made in both the table and the wizard), making these globally available in jest.setup.js is reasonable.


regexowl (approved):
LGTM :+1:

PR #981, opened 2023-03-01
Wizard: Update activation keys link text and URI

lucasgarfield:
This commit updates the link on the registration step so users are directed to the exact place where they can manage keys.

Because managing keys through the customer portal will soon become deprecated (keys will be managed through the keys service on insights) the text was also changed to be more general.


regexowl (commented):
Nice one. Two small nitpicks - I'd probably leave out the word "here" as we don't seem to use it in any other links. 

And we could probably also remove the surrounding empty tags as there's only a button in the label now.

What do you think?

regexowl:
Absolutely, makes perfect sense then :)

regexowl (approved):
LGTM

PR #982, opened 2023-03-01
ShareImageModal: Add support for sources

lucasgarfield:
This commit allows sharing (cloning) composes that use the sources service to resolve the AWS account ID.


croissanne (approved):
thank you!

PR #983, opened 2023-03-01
Landingpage: Switch to CodeBranchIcon

ochosi:
Use the CodeBranchIcon instead of just CodeIcon for the Open Source Services popover button.

This is a follow-up of https://github.com/RedHatInsights/image-builder-frontend/pull/949

Here's how the new icon looks:
image


lucasgarfield (approved):
Wow, shiny!

PR #1001, opened 2023-03-14
Wizard: Update the Review step to have expandable sections

### LINKED ISSUE #919: Update Review step to have collapsable sections
Instead of the tabs, show all review sections on a single page with collapsable sections.

Current:
Screen Shot 2023-01-10 at 4 22 56 PM

Proposed:
https://www.sketch.com/s/d7aa6d29-fca0-4283-a846-09cc5fd10612/a/09P2xEV

regexowl:
Fixes #919.

This updates the Review step as per proposed mocks.

The tabs were replaced by expandable sections containing all the information relevant to a specific step.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.scss
+++ b/src/Components/CreateImageWizard/CreateImageWizard.scss
@@ -47,7 +47,7 @@
 }
 
 .pf-u-min-width {
-    --pf-u-min-width--MinWidth: 11ch;
+    --pf-u-min-width--MinWidth: 18ch;
 }
 
 .pf-u-max-width {
--- a/src/Components/CreateImageWizard/formComponents/ReviewStep.js
+++ b/src/Components/CreateImageWizard/formComponents/ReviewStep.js
@@ -2,88 +2,35 @@ import React, { useEffect, useState } from 'react';
 
 import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
 import {
-  Button,
-  DescriptionList,
-  DescriptionListDescription,
-  DescriptionListGroup,
-  DescriptionListTerm,
-  List,
-  ListItem,
-  Popover,
-  Spinner,
-  Tab,
-  Tabs,
-  TabTitleText,
+  ExpandableSection,
   Text,
   TextContent,
-  TextList,
-  TextListItem,
-  TextListItemVariants,
-  TextListVariants,
   TextVariants,
 } from '@patternfly/react-core';
-import { HelpIcon } from '@patternfly/react-icons';
+
 import {
-  TableComposable,
-  Tbody,
-  Td,
-  Th,
-  Thead,
-  Tr,
-} from '@patternfly/react-table';
-import PropTypes from 'prop-types';
+  ContentList,
+  FSCList,
+  ImageDetailsList,
+  ImageOutputList,
+  RegisterLaterList,
+  RegisterNowList,
+  TargetEnvAWSList,
+  TargetEnvAzureList,
+  TargetEnvGCPList,
+  TargetEnvOtherList,
+} from './ReviewStepTextLists';
 
-import ActivationKeyInformation from './ActivationKeyInformation';
-
-import { RELEASES, UNIT_GIB, UNIT_MIB } from '../../../constants';
-import { useGetAWSSourcesQuery } from '../../../store/apiSlice';
 import isRhel from '../../../Utilities/isRhel';
-import { googleAccType } from '../steps/googleCloud';
-
-const FSReviewTable = ({ ...props }) => {
-  return (
-    <TableComposable
-      aria-label="File system configuration table"
-      variant="compact"
-    >
-      <Thead>
-        <Tr>
-          <Th>Mount point</Th>
-          <Th>File system type</Th>
-          <Th>Minimum size</Th>
-        </Tr>
-      </Thead>
-      <Tbody data-testid="file-system-configuration-tbody-review">
-        {props.fsc.map((r, ri) => (
-          <Tr key={ri}>
-            <Td className="pf-m-width-30">{r.mountpoint}</Td>
-            <Td className="pf-m-width-30">xfs</Td>
-            <Td className="pf-m-width-30">
-              {r.size}{' '}
-              {r.unit === UNIT_GIB
-                ? 'GiB'
-                : r.unit === UNIT_MIB
-                ? 'MiB'
-                : 'KiB'}
-            </Td>
-          </Tr>
-        ))}
-      </Tbody>
-    </TableComposable>
-  );
-};
-
-FSReviewTable.propTypes = {
-  fsc: PropTypes.arrayOf(PropTypes.object).isRequired,
-};
 
 const ReviewStep = () => {
-  const [activeTabKey, setActiveTabKey] = useState(0);
-  const [minSize, setMinSize] = useState();
+  const [isExpandedImageOutput, setIsExpandedImageOutput] = useState(false);
+  const [isExpandedTargetEnvs, setIsExpandedTargetEnvs] = useState(false);
+  const [isExpandedFSC, setIsExpandedFSC] = useState(false);
+  const [isExpandedContent, setIsExpandedContent] = useState(false);
+  const [isExpandedRegistration, setIsExpandedRegistration] = useState(false);
+  const [isExpandedImageDetail, setIsExpandedImageDetail] = useState(false);
   const { change, getState } = useFormApi();
-
-  const { data: awsSources, isSuccess: isSuccessAWSSources } =
-    useGetAWSSourcesQuery();
 
   useEffect(() => {
     const registerSystem = getState()?.values?.['register-system'];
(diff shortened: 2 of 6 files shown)


lucasgarfield:
>in the mocks the packages popover in Content includes also release and version of the packages, do we have that information somewhere? Would we like to add it?

Both the content sources and image builder endpoints we get back only the name and summary. I think what you've done for now is fine and we can look into adding it later.

>for now the Red Hat packages and 3rd party packages are merged as we seem to work with those just as selected-packages without differentiating where the package comes from. Would we want to split them as shown in the mocks?

The challenge here would be 'Recreate Image'. The json blob we use for 'Recreate Image' only contains a list of packages and a list of repositories, but there is no mapping between them. That's also one of the reasons that the list of packages is actually shared between the Red Hat packages step and the Custom Repositories packages step.

Splitting them is of course technically possible but it would be a lot of work to implement - for now let's just keep them together.

lucasgarfield (commented):
Wow, this looks really nice! I'm still reviewing, just have a couple of initial things to bring up:
1. We changed the language from '3rd party repository' to 'custom repository' because some of the custom repositories are actually 1st party (from RH). These mocks were probably created before we made that change, so let's also make it here.
2. This PR matches the mocks perfectly but there are some capitalization inconsistencies I noticed that might be easier to just fix now instead of later. We have discussed this previously for the nav bar which I believe has the correct pattern - first word capitalized, following words lowercase (unless they are proper nouns). E.g. File system configuration, not File System Configuration. Can we make these changes for the headings on the Review step as well? E.g. Image details, not Image Details? There is also inconsistency on this step by itself as it has Image Details but also has File system configuration.
3. The popovers with dynamic lists should have a maximum height - I don't think that made it into this PR yet, did it?

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -47,7 +47,7 @@
 }
 
 .pf-u-min-width {
-    --pf-u-min-width--MinWidth: 11ch;
```
lucasgarfield, line 50: What was this needed for? I'm wondering why it was 18 to begin with - what components relied on it being 18?

Would it be better to inline the css for things like this so it is explicit?

I put in the .pf-u-max-width class at 26rem recently for the sources step and this is actually now making me think I should probably revisit that, as well.

regexowl: I've been using it to unify width of the description terms to align the lists into equal columns.

lucasgarfield: Ah, that's right. I need to revisit how I did the layout on the sources step as I think I might have been relying on this as well... even if it turns out to be fine, I think that this is a good sign that we should avoid these global width changes when they are needed for specific components... Let's see if we can find a good inline solution so we can eliminate these global overrides.

lucasgarfield: We can do this in a follow up PR, I'll create an issue so we don't forget.

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -113,392 +63,106 @@ const ReviewStep = () => {
...
-      </DescriptionList>
-      <Tabs
-        isFilled
-        activeKey={activeTabKey}
-        onSelect={handleTabClick}
-        className="pf-u-w-75"
+      <ExpandableSection
+        toggleContent={'Image output'}
+        onToggle={onToggle}
+        isExpanded={isExpanded['image-output']}
```
lucasgarfield, line 73: I'm not 100% sure about this. `isExpanded` is a boolean, so `isExpanded['image-output']` should return undefined... right? 

The expandable sections work fine when you actually run this in the browser and in the tests, though! Which is kind of surprising.

Still, I worry that this could lead to bugs in the future. Even though it works, we probably shouldn't be treating a boolean like an object... 

Maybe the solution is to give each expandable section its own useState hook?

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -2,83 +2,33 @@ import React, { useEffect, useState } from 'react';
...
-
-FSReviewTable.propTypes = {
-  fsc: PropTypes.arrayOf(PropTypes.object).isRequired,
-};
-
 const ReviewStep = () => {
-  const [activeTabKey, setActiveTabKey] = useState(0);
+  const [isExpanded, setIsExpanded] = useState(false);
+
   const [minSize, setMinSize] = useState();
```
lucasgarfield, line 32: It seems like minSize is only used by the <FSCList minSize={minSize} /> component. Would it make sense to push this state 'down' into the <FSCList /> component instead of managing it here in the parent?

lucasgarfield: Lines 48-62 in the useEffect hook are also related to this and could be handled in the `<FSCList />` component, I think.

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -113,392 +63,106 @@ const ReviewStep = () => {
...
+        )}
+      </ExpandableSection>
+      <ExpandableSection
+        toggleContent={'File system configuration'}
+        onToggle={onToggle}
+        isExpanded={isExpanded['file-system-configuration']}
+        isIndented
+        data-testid="file-system-configuration-expandable"
+      >
+        <FSCList minSize={minSize} />
```
lucasgarfield, line 128: Mentioned in another comment above but we could just manage minSize inside of the `<FSCList />` component itself couldn't we?

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -113,392 +63,106 @@ const ReviewStep = () => {
...
+            isSuccessAWSSources={isSuccessAWSSources}
+          />
         )}
-        <Tab
-          eventKey={2}
-          title={<TabTitleText>System configuration</TabTitleText>}
-          data-testid="tab-system"
-        >
+        {getState()?.values?.['target-environment']?.gcp && (
+          <TargetEnvGCPList googleAccType={googleAccType} />
```
lucasgarfield, line 93: Could we import googleAccType inside the `<TargetEnvGCPList />` component?

src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
```diff
@@ -0,0 +1,514 @@
...
+            <>
+              {' '}
+              <Popover
+                position="bottom"
+                headerContent="Partitions"
+                hasAutoWidth
+                minWidth="30rem"
+                bodyContent={
+                  <FSReviewTable
+                    fsc={getState().values['file-system-configuration']}
```
lucasgarfield, line 243: Could we call getState().values['file-system-configuration'] inside of the `<FSReviewTable />` instead of passing it in as a prop?

src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
```diff
@@ -0,0 +1,514 @@
...
+        <TextListItem component={TextListItemVariants.dd}>
+          Built image will be available for download
+        </TextListItem>
+      </TextList>
+      <br />
+    </>
+  );
+};
+
+export const FSCList = ({ ...props }) => {
```
lucasgarfield, line 215: I think here we can eliminate {...props} entirely if we just move minSize state into this component, but in general, would you be okay with _not_ using prop destructuring? 

We don't have a style guide (maybe we should work on that, or at least get an eslint plugin) but I have started to really prefer explicitly declaring props whenever possible, in my opinion it reduces the cognitive load and makes it easier to understand components. With {...props} I am not sure what is actually getting passed in, plus it also makes it possible to pass in un-used props that cause additional re-renders for no good reason (inefficient). I'd like to use {...props} only when absolutely necessary. 

That's my personal preference but maybe there is also a good reason for using the destructuring I'm not aware of... what do you think?

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -0,0 +1,112 @@
...
+        >
+          <Thead>
+            <Tr>
+              <Th>Mount point</Th>
+              <Th>File system type</Th>
+              <Th>Minimum size</Th>
+            </Tr>
+          </Thead>
+          <Tbody data-testid="file-system-configuration-tbody-review">
+            {props.fsc.map((r, ri) => (
```
lucasgarfield, line 32: Would it be easier for someone unfamiliar with this code to understand what we're mapping over if we used a more descriptive variable name? It isn't immediately obvious to me what `r` stands for...

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -0,0 +1,112 @@
...
+    <Panel isScrollable>
+      <PanelMain maxHeight="30ch">
+        <TableComposable aria-label="Packages table" variant="compact">
+          <Thead>
+            <Tr>
+              <Th>Name</Th>
+            </Tr>
+          </Thead>
+          <Tbody data-testid="packages-tbody-review">
+            {props.packages.map((r, ri) => (
```
lucasgarfield, line 64: More descriptive variable name could help people unfamiliar with the code here, too. If we're mapping over packages, maybe something like `package` or even `pkg` would make the code a bit easier to understand?

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -0,0 +1,112 @@
...
+              </Tr>
+            ))}
+          </Tbody>
+        </TableComposable>
+      </PanelMain>
+    </Panel>
+  );
+};
+
+export const PackagesTable = ({ ...props }) => {
```
lucasgarfield, line 53: So in the parent component that renders this component, it is called like this: `<PackagesTable packages={getState()?.values['selected-packages']} />`

We could just call `const packages={getState()?.values['selected-packages']}` inside of this component itself.

I think that reduces the cognitive load needed to understand the component... When the packages come in as a prop, someone reading the code might ask themselves "well where did those packages come from?". Then they would need to refer to the parent component (a context switch) to find out that the come from the form state, then come back to this component (another context switch).

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -0,0 +1,112 @@
...
+          aria-label="Custom repositories table"
+          variant="compact"
+        >
+          <Thead>
+            <Tr>
+              <Th>Name</Th>
+            </Tr>
+          </Thead>
+          <Tbody data-testid="repositories-tbody-review">
+            {props.repositories.map((r, ri) => (
```
lucasgarfield, line 90: Here I can figure out that `r` is probably `repository` but could `repo` be a bit easier to understand on someone's first time reading the code?

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -0,0 +1,112 @@
...
+              </Tr>
+            ))}
+          </Tbody>
+        </TableComposable>
+      </PanelMain>
+    </Panel>
+  );
+};
+
+export const RepositoriesTable = ({ ...props }) => {
```
lucasgarfield, line 76: Could we get rid of the props by calling `const repositories={getState()?.values?.['custom-repositories']}` inside this component?

lucasgarfield:
image

It seems like under Content, when there are 0 'Additional Red Hat and 3rd party packages' it is out of alignment with the number of 'Custom repositories' because the value displayed for 'Custom repositories' is a button with padding.

Also, for some reason, the color of the 'Custom repositories' button is grey... shouldn't it be blue to indicate that it is clickable?

src/Components/CreateImageWizard/formComponents/ReviewStep.js
regexowl: It makes perfect sense to move those, done!

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -113,392 +63,106 @@ const ReviewStep = () => {
...
+        toggleContent={'Target environments'}
+        onToggle={onToggle}
+        isExpanded={isExpanded['target-environments']}
+        isIndented
+        data-testid="target-environments-expandable"
+      >
+        {getState()?.values?.['target-environment']?.aws && (
+          <TargetEnvAWSList
+            awsSources={awsSources}
+            isSuccessAWSSources={isSuccessAWSSources}
```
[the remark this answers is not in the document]
regexowl: Done!

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
regexowl: Updated :+1:

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -0,0 +1,120 @@
...
+  Tr,
+} from '@patternfly/react-table';
+import PropTypes from 'prop-types';
+
+import { UNIT_GIB, UNIT_MIB } from '../../../constants';
+
+export const FSReviewTable = () => {
+  const { getState } = useFormApi();
+  const fsc = getState().values['file-system-configuration'];
+  console.log(fsc);
```
lucasgarfield, line 20: Missed this one ;)

regexowl: Oops, removed!

PR #1002, opened 2023-03-14
Allow launch wizard for Azure images

ezr-ondrej:
Adds feature flagged Azure support.

Depends on https://github.com/RHEnVision/provisioning-frontend/pull/224


src/Components/ImagesTable/ImageLink.js
```diff
@@ -35,9 +38,9 @@ const ProvisioningLink = ({ imageId, isExpired, isInClonesTable }) => {
   );
 
   const provider = getImageProvider(image);
-  if (!error && image.share_with_sources) {
+  if (!error) {
```
croissanne, line 41: I would prefer to keep this actually. I think we should only link to provisioning if the image was shared with a source.

ezr-ondrej: this was extracted to a "allowed for provider" function.

src/Components/ImagesTable/ImageLink.js
```diff
@@ -84,13 +88,22 @@ const ImageLink = ({ imageId, isExpired, isInClonesTable }) => {
...
   if (!uploadStatus || image.status !== 'success') return null;
 
-  if (
-    hasProvisioning &&
-    (image.imageType === 'aws' || image.imageType === 'ami')
-  ) {
+  const provisioningLinkEnabled = (image) => {
+    switch (getImageProvider(image)) {
+      case 'aws':
+        return !!image.share_with_sources;
```
croissanne, line 94: The double not is pretty confusing, can we just change this to length > 0?

src/Components/ImagesTable/ImageLink.js
croissanne: missed that, sorry!

croissanne (approved):
Thanks!

PR #1008, opened 2023-03-15
Wizard: Fix sources select bug

### LINKED ISSUE #1007: Source name select is shown behind wizard buttons
I think this image says it all:

image

I saw this in stage beta.

lucasgarfield:
Fixes https://github.com/RedHatInsights/image-builder-frontend/issues/1007. A bug in Patternfly causes dropdown and select menus to
appear underneath the Wizard footer when they are appended to their
parent (which is done to allow them to 'pop out' of the wizard and avoid
scrolling inside of the wizard's inner window).

The bug is discussed here:
https://github.com/patternfly/patternfly/issues/4794#issuecomment-1451187628

For now, the best solution seems to be adjusting the wizard's z-index
values as the other solutions discussed have a negative impact on
accessibility and keyboard navigation.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.scss
+++ b/src/Components/CreateImageWizard/CreateImageWizard.scss
@@ -59,3 +59,22 @@ ul.pf-m-plain {
     padding-left: 0;
     margin-left: 0;
 }
+
+/*
+These z-index values prevent a bug where dropdowns that are appended to their
+parent (in order to allow them to "pop out" of the modal and prevent scrolling
+inside the modal's inner window) appear 'underneath' the wizard footer. This is
+a known bug in Patternfly 4. When it is fixed (likely in PF5), these z-index
+values can be removed.
+*/
+
+.pf-c-wizard__header,
+.pf-c-wizard__main {
+    z-index: auto !important;
+}
+
+.pf-c-wizard__toggle,
+.pf-c-wizard__nav,
+.pf-c-wizard__footer {
+    z-index: 1 !important;
+}


lucasgarfield:
Draft PR until I can verify that it works - console.stage.redhat.com is down so I am unable to create sources to verify that it works at the moment.

jrusz (commented):
CI seems unhappy, both builds failed... And I can't really test it since this feature is not available in ephemeral...

lucasgarfield:
I found a good discussion about this on the Patternfly Github repo, using their solution seems to work well:
image

jrusz:
Ugh, something changed in one of the `pr_check.sh` scripts and it's now failing... There is some `--label image-type=single` in the docker command as the most suspicious change, I'm going to have to investigate what's up.

lucasgarfield:
Going to hold off on this PR for a bit, discussed this in the weekly UX/dev sync and we may go with the scrollable modal instead of pop-out to be more consistent across RH Insights services.

jrusz:
> Going to hold off on this PR for a bit, discussed this in the weekly UX/dev sync and we may go with the scrollable modal instead of pop-out to be more consistent across RH Insights services.

Does that mean that this whole widget might change? If so do you have an ETA on when it could happen?

lucasgarfield:
I will do a quick fix this afternoon that will cause it to behave exactly like the select on the next step used to select an activation key. I don't think it would affect your tests - it will still be the same component, it will just appear in a different place in the DOM.

jrusz:
@lucasgarfield any updates on this? It would be nice to have fixed as if there are 4 or more sources defined our UI tests fail on selecting and I have to go there and do some manual cleanup.
It's not a blocker, this usually happens when the test teardown functions are not executed but it's annoying regardless...

lucasgarfield:
Superseded by https://github.com/RedHatInsights/image-builder-frontend/pull/1118

PR #1009, opened 2023-03-15
Wizard: Expose beta features in qa environment

lucasgarfield:
The QE team has requested that all beta features be exposed in the ephemeral environment (insights.chrome.getEnvironment() === 'qa').

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -31,6 +31,7 @@ import { UNIT_GIB, UNIT_KIB, UNIT_MIB } from '../../constants';
 import { useGetArchitecturesByDistributionQuery } from '../../store/apiSlice';
 import { composeAdded } from '../../store/composesSlice';
 import { fetchRepositories } from '../../store/repositoriesSlice';
+import isBeta from '../../Utilities/isBeta';
 import isRhel from '../../Utilities/isRhel';
 import { resolveRelPath } from '../../Utilities/path';
 import DocumentationButton from '../sharedComponents/DocumentationButton';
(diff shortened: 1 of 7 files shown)


jrusz (commented):
Ok looks great, I can see the fatures enabled in ephemeral, thank you!

Testing in ephemeral is going to be failing for a bit now but we're working with Alex on adjusting the test suite already and this should help us get it faster. 

Do you agree @atodorov that we should get this merged now and we'll make it green again by finishing our changes?

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -262,7 +262,7 @@ const getPackageDescription = async (release, arch, repoUrls, packageName) => {
   let pack;
   // if the env is stage beta then use content-sources api
   // else use image-builder api
-  if (insights.chrome.isBeta()) {
+  if (insights.chrome.isBeta() || insights.chrome.getEnvironment() === 'qa') {
```
atodorov, line 265: I would prefer to see this as a helper function instead of adjusting it everywhere. 

Otherwise looks good to me.

atodorov:
> Do you agree @atodorov that we should get this merged now and we'll make it green again by finishing our changes?

I don't have a strong preference for the order in which we merge this change. Merging now will probably make it a bit easier for us when working on the new tests and we could see them passing internally before merging into the test suite.

src/Components/CreateImageWizard/CreateImageWizard.js
lucasgarfield: Good point... the more beta features we add the bigger the chance of a footgun going off... I'll refactor on Friday when I'm back.

jrusz:
I've found out that I can't search for additional red hat packages with this change, nothing shows up. The custom packages search work and I can add packages from epel for example.

jrusz:
> I've found out that I can't search for additional red hat packages with this change, nothing shows up. The custom packages search work and I can add packages from epel for example.

This turned out to be an issue with `sources` app. They don't introspect all the repos in ephemeral but they've enabled 8.7 baseos repo for us to test with so I'll go try that now.

jrusz (approved):
Looks good to me. I have the corresponding MR in the test suite ready so we can merged this whenever :)

PR #1031, opened 2023-03-24
config/dev: run proxy without VPN for production envs

croissanne:
This allows users to develop against production without having access to the internal network.

https://github.com/RedHatInsights/frontend-components/tree/62f37029ca0f5f7e9b0fd61815fe62c96503bcbf/packages/config#running-prod-proxy-without-vpn


ochosi (approved):
\o/ Works like a charm! :heart_eyes_cat:

PR #1033, opened 2023-03-27
ESLint: Add `no-console` rule

regexowl:
This adds `no-console` rule to throw an error when using any of the `console` methods as those are used only for debugging purposes.


lucasgarfield (approved):
I think this is a great idea, should save us some headaches. Thanks!

PR #1035, opened 2023-03-28
Add sources selection for Azure

ezr-ondrej:
Fixes https://issues.redhat.com/browse/HMS-1511

It now depends on https://github.com/osbuild/image-builder/pull/672

[Screencast from 2023-03-29 16-03-23.webm](https://user-images.githubusercontent.com/2884324/228563824-beb5a5d2-5371-4b31-99d3-1068036fa4fa.webm)


src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -501,20 +503,21 @@ const formStepHistory = (composeRequest) => {
...
 
 const CreateImageWizard = () => {
-  const awsTarget = insights.chrome.isBeta() ? awsTargetBeta : awsTargetStable;
-
+  const chrome = useChrome();
   const dispatch = useDispatch();
   const navigate = useNavigate();
   const location = useLocation();
 
+  const awsTarget = chrome.isBeta() ? awsTargetBeta : awsTargetStable;
```
lucasgarfield, line 511: We just merged https://github.com/RedHatInsights/image-builder-frontend/pull/1009 which adds an isBeta() helper function... the helper function is useful because it also exposes these features in the qa environment. Let's use it here! :)

ezr-ondrej: Totally, great helper! :)
It's bit of a shame it can't use the `useChrome` hook as frontend team tries to push on removing the dom `chrome`, but I guess that won't happen anytime soon :+1:

ezr-ondrej: Yeah, I got the same feeling, shame, but maybe there will be a solution out there for us soon (might be good to bring it up with the frontend team, they were very helpful whenever I approached them)

src/Components/CreateImageWizard/steps/msAzure.beta.js
```diff
@@ -0,0 +1,255 @@
...
+      condition: {
+        when: 'azure-type',
+        is: 'azure-type-manual',
+      },
+    },
+    {
+      component: componentTypes.TEXT_FIELD,
+      name: 'azure-tenant-id',
+      className: 'pf-u-w-50',
+      'data-testid': 'azure-tenant-id',
```
lucasgarfield, line 156: Would it make the tests a bit more robust/explicit if we gave these two fields different test ids? Like `azure-tenant-id-source` and `-manual`?

src/Components/CreateImageWizard/steps/msAzure.beta.js
```diff
@@ -0,0 +1,255 @@
...
+      condition: {
+        when: 'azure-type',
+        is: 'azure-type-manual',
+      },
+    },
+    {
+      component: componentTypes.TEXT_FIELD,
+      name: 'azure-subscription-id',
+      className: 'pf-u-w-50',
+      'data-testid': 'azure-subscription-id',
```
lucasgarfield, line 201: Same comment as above re: unique test ids.

src/Components/CreateImageWizard/steps/msAzure.beta.js
```diff
@@ -0,0 +1,255 @@
...
+      ],
+      condition: {
+        when: 'azure-type',
+        is: 'azure-type-manual',
+      },
+    },
+    {
+      component: 'azure-resource-groups',
+      name: 'azure-resource-group',
+      className: 'pf-u-w-50',
```
lucasgarfield, line 239: It doesn't seem like the `pf-u-w-50` class gets used for selects. I think it is picking up its width from .pf-c-select__toggle which is 100% by default.

lucasgarfield:
This is really great, amazing how quickly you put it together! I think the last thing to do is to add support for the "Recreate image" feature. We just need to update the Azure section of the requestToState() function so that the Wizard's initial state is initialized properly.

https://github.com/RedHatInsights/image-builder-frontend/blob/4405d2ddd2eb3839a46abedaeb8ba5dacc3351dc/src/Components/CreateImageWizard/CreateImageWizard.js#L292

src/Components/CreateImageWizard/steps/msAzure.beta.js
```diff
@@ -0,0 +1,255 @@
...
+        {
+          type: validatorTypes.REQUIRED,
+        },
+      ],
+      condition: {
+        when: 'azure-type',
+        is: 'azure-type-source',
+      },
+    },
+    // TODO check oauth2 thing too here?
```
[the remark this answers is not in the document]
ezr-ondrej: It's copy pasted, no idea what it means, so I've kept it ^_^

src/Components/CreateImageWizard/steps/msAzure.beta.js
ezr-ondrej: I've made it to work, if the className prop is passed to the select directly and not to the FormGroup, it works as expected (it's actually exactly what other components do as well anyways :)

I've then changed the class to `pf-u-max-width`, that you use in AWS, to be consistent.

src/Components/CreateImageWizard/steps/msAzure.beta.js
ezr-ondrej: yup, make sense, I've change the ids :+1:

ezr-ondrej:
I've addressed your comments and also added alert for the case when sources are fetched successfuly, but the selected source fails to fetch details (the message expects the error to be in the contacting the provider, which is probably most likely to be hit by users).

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -151,6 +152,13 @@ const onSave = (values) => {
   }
 
   if (values['target-environment']?.azure) {
+    const upload_options =
+      values['azure-type'] === 'azure-type-source'
+        ? { source_id: [values['azure-sources-select']] }
```
lucasgarfield, line 157: Azure only supports sharing to a single source, so let's just use `source_id: values['azure-sources-select']` instead of putting it in a list.

ezr-ondrej: ah, yeah, makes sense, no way to target multiple accounts for azure, right? :)

ezr-ondrej: Fixed :)

lucasgarfield:
We have merged https://github.com/osbuild/image-builder/pull/672 and so you can test this front end PR locally against the stage proxy (`npm run stage-beta`).

It's looking great, just a couple of issues I've run into. Both are related to the 'Recreate image' action.

First, on the review step (after clicking 'Recreate image'), the tenant id and subscription id don't appear. We just need to call the useQuery hook in the review step component. Take a look at the `TargetEnvAWSList` component for inspiration. We will add the hook to the `TargetEnvAzureList` component.
https://github.com/RedHatInsights/image-builder-frontend/blob/451b5784d56967514a6f5bffdbba4a3a28ef1fbe/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js#L62
https://github.com/RedHatInsights/image-builder-frontend/blob/451b5784d56967514a6f5bffdbba4a3a28ef1fbe/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js#L159
image

Second, after clicking 'Recreate image', if I click back to the Azure step, the Wizard crashes... the error message indicates it has something to do with the value of `currentStep` but I'm still not sure what the problem is. I am investigating.

ezr-ondrej:
I'm happy to investigate the failure as well, but let me know if you figure out what's causing it ;)

lucasgarfield:
The bug causing the crash was actually a latent bug not related to this PR. I've created #1068 to fix it.

src/Components/CreateImageWizard/steps/msAzure.beta.js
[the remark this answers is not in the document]
lucasgarfield: Ha, fair enough. We can keep it for now for consistency.

ezr-ondrej:
I've just noticed, you introduced some changes, I've pulled the latest I hope, but just in case, could you check?
I've addressed the Review step now, so we see source name for source selection

lucasgarfield (approved):
Excellent work, Ondrej!

ezr-ondrej:
Thanks @lucasgarfield for all the support through the process :orange_heart:

PR #1039, opened 2023-03-30
Wizard: Migrate activation keys calls to RTK Query

regexowl:
This migrates calls to RHSM endpoints `activation_keys` and `/activation_keys/{name}` to RTK Query.

Tests were also updated to use Mock Service Worker instead of previous Jest mock function.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
+++ b/src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
@@ -1,7 +1,10 @@
-import React, { useEffect, useState } from 'react';
+import React, { useContext } from 'react';
 
 import { useFormApi } from '@data-driven-forms/react-form-renderer';
+import WizardContext from '@data-driven-forms/react-form-renderer/wizard-context';
 import {
+  Alert,
+  Spinner,
   Text,
   TextContent,
   TextList,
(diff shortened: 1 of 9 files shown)


src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
```diff
@@ -21,27 +22,15 @@ import {
...
-      const data = await api.getActivationKey(activationKey);
-      setRole(data?.role);
-      setServiceLevel(data?.serviceLevel);
-      setUsage(data?.usage);
-      setRepositories(data?.additionalRepositories);
-    };
-    fetchKeyInformation();
-  }, []);
+  const { data: activationKeyInfo, isSuccess: isSuccessActivationKeyInfo } =
+    useGetActivationKeyInformationQuery(activationKey);
```
lucasgarfield, line 33: activationKey could theoretically be `undefined`, and in that case our query would return an error. `isSuccess` would remain false, and we would just show spinners indefinitely.

We might want to also extract `isError` and handle the error state separately (not show the spinner, instead display an error message). We could choose whether or not to render the spinners based on `isFetching`. 

We could also use the query's skip option to skip the query entirely if activationKey === undefined:
https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching

But I still think it would be good to handle the error state properly. What if, for instance, the key has been deleted? It would be good to notify the user in that case.

lucasgarfield:
Really, really nice work! 

Just one thing, do you think you could bring this component in alignment with the AWS step? I have been planning on using that one as the model for things like this, where we populate a dropdown via an API call.

Specifically, I mean that we should refetch the list of keys every time the dropdown is opened and show a spinner at the bottom of the list indicating it is fetching. Also, in addition to error state on the review page as I mentioned in my comment, it would be good to have an error state on the registration step itself.

Here are screenshots of what I mean from the AWS step (when using the sources option):
sources refetch

Error state:
image

src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
regexowl: Great point, added error handling!

image

regexowl:
Thanks a lot for the comments!

The dropdown menu is updated and refetches the keys every time it is open. Also added the error state to the Registration step as well.

PR #1040, opened 2023-03-30
Wizard: Remove prop validation for `ReviewStepTables` and `ReviewStepTextLists`

regexowl:
This removes prop validation that was left behind after the props were removed during a Review step updates PR review.


lucasgarfield (approved):
Nice catch! :)

PR #1045, opened 2023-03-31
CreateImageWizard: show error details on save

croissanne:
Show details when the wizard sends a request that results in an error. Now that AWS and Azure have a maximum size, this needs to be shown.

---


regexowl:
The code looks good. What would be a good scenario to test this on?

croissanne:
> The code looks good. What would be a good scenario to test this on?

Azure or AWS image, and in the FS customizations enter 12 GB

regexowl (approved):
Nice, works well! Thank you :)

PR #1052, opened 2023-04-04
Wizard: Disable remove button for the root partition

### LINKED ISSUE #910: Manual File system configuration issues
1. Should we allow them to disable the /root partition? It's required to continue. Why do we let them delete it? 
2. If I delete all partitions, then the next button remains enabled, but once I click is disables and gives no feedback.
3. Why do you allow duplicate partitions to be added if I only have one? Why not remove them from the dropdown?

https://user-images.githubusercontent.com/88330245/211394200-f98b8242-b203-4c50-97cf-20dbd875980e.mov

https://user-images.githubusercontent.com/88330245/211394216-94fdf630-70e8-4ff3-afb4-f5c9dec5f63d.mov

https://user-images.githubusercontent.com/88330245/211394223-103663d0-5dff-4ff8-bc7d-09aac5e7bbdf.mov

regexowl:
Fixes #910. This disables the remove button and select for the root partition as that one is required to continue with creating an image.

The third point mentioned in the issue is removing mount points that have already been chosen from the dropdown to prevent adding duplicate mountpoints.

This point was not implemented as we need to allow the user to create multiple sub directories under the same mount point if needed.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
+++ b/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
@@ -389,6 +389,7 @@ const FileSystemConfiguration = ({ ...props }) => {
                       icon={<MinusCircleIcon />}
                       onClick={() => removeRow(row.id)}
                       data-testid="remove-mount-point"
+                      isDisabled={row.mountpoint === '/' ? true : false}
                     />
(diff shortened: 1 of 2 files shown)


croissanne (approved):
LGTM!

PR #1059, opened 2023-04-06
Wizard: Add "refine search" warning to Packages step

### LINKED ISSUE #913: Show additional results than search returned if needed
An issue we ran into when searching additional Red Hat packages is that the results are limited to 100 results, but there could be more based on your search criteria. We should inform them that not all results are displayed and to refine their search (vs. allowing them to think the package might not exist.)

An exception to this is if a result has an exact match as well as over 100 search results. We show the same pkg AND prompt them to refine their search to get more results.

Mocks:
https://www.sketch.com/s/d7aa6d29-fca0-4283-a846-09cc5fd10612/a/DPm10z2

regexowl:
Fixes #913.

This adds status bars to the `DualListSelectorPane` on the Packages step. The status bar indicates how many packages were found and how many of those have been selected.

Warning for too many returned results was also added. When an exact match is found during a search with over 100 results, it is shown together with the warning.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Packages.js
+++ b/src/Components/CreateImageWizard/formComponents/Packages.js
@@ -10,6 +10,8 @@ import React, {
 import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
 import WizardContext from '@data-driven-forms/react-form-renderer/wizard-context';
 import {
+  Alert,
+  Divider,
   DualListSelector,
   DualListSelectorControl,
   DualListSelectorControlsWrapper,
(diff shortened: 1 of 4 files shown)


regexowl:
The exact match can be tested for example on "perl" and "python3".

It seems the placement of the "too many results" warning icon as per mocks is possible for [`TextInput`](https://www.patternfly.org/v4/components/text-input), but not for [`SearchInput`](https://www.patternfly.org/v4/components/search-input) which we currently use. Replacing the search inputs would require more changes, so just wanted to check, if that's what we want. 
 
 The warning icon is placed next to the `HelperText` warning for now.

lucasgarfield:
Whoa, this is awesome Klara! It works really well.

Just one little thing I noticed, the warning triangle icon appears in a different place here than the designs (maybe the designs changed since you did this lol)...

This PR:
image

The designs:
image

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -264,25 +268,68 @@ const Packages = ({ getAllPackages, isSuccess }) => {
     setFilterChosen('');
   };
 
+  const ExactMatch = ({ pkgList, search }) => {
```
lucasgarfield, line 271: This actually seems to be inside of the <Packages> component. It works, but we don't have nested components anywhere else in the codebase, so maybe for the sake of consistency we should move it outside of <Packages>? (It can stay in Packages.js, of course...)

regexowl: You're right, I've moved the `ExactMatch` outside of the `Packages` component :+1:

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -31,6 +33,32 @@ import api from '../../../api';
...
+  search,
+  chosenPackages,
+  selectedAvailablePackages,
+  handleSelectAvailableFunc,
+}) => {
+  const match = pkgList.find((pkg) => pkg.name === search);
+  return (
+    <DualListSelectorListItem
+      data-testid={`exact-match-${match.name}`}
+      key={match.name}
```
lucasgarfield, line 47: I think that we technically don't really need a key here, since this list item is not part of an array:
https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

`Keys tell React which array item each component corresponds to, so that it can match them up later. This becomes important if your array items can move (e.g. due to sorting), get inserted, or get deleted. A well-chosen key helps React infer what exactly has happened, and make the correct updates to the DOM tree.`

regexowl: Yeah, I think you're right. I'll remove it.

lucasgarfield (approved):
Excellent!!!

PR #1062, opened 2023-04-06
Provisioning only launch

ezr-ondrej:
~Depends on https://github.com/RHEnVision/provisioning-frontend/pull/238~


croissanne:
We need to be sure that also for the link, we can select a region. Or show a link for each region. But this would be amazing <3

ezr-ondrej:
You're right, I've missed that, but we can easily add the region select to the empty state :)

ezr-ondrej:
The empty state just got merged, I believe this is ready to go in? :)

lucasgarfield:
Absolutely love this! I can't wait until the Launch Wizard is in production and we can just get rid of our old launch popover completely.

lucasgarfield (approved):
Thank you!

PR #1069, opened 2023-04-12
Wizard: Use useChrome hook to set beta feature flag

lucasgarfield:
Calling insights.chrome is deprecated. The commits in this PR:

(1) replace isBeta() function with custom useGetEnvironment() hook
(2) adapt App.js to useChrome() hook (stop calling insights.chrome directly)
(3) replace all calls in code to insights.chrome with calls to either useChrome() or useGetEnvironment() hooks

This PR is important, because calling the deprecated insights.chrome functions results in a massive amount of warnings flooding the console, which makes it very difficult to find real/important error messages.

Chrome API docs can be found here:
http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/chrome/chrome-api

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -302,7 +302,7 @@ const getPackageDescription = async (
 };
 
 // map the compose request object to the expected form state
-const requestToState = (composeRequest, distroInfo, isBeta) => {
+const requestToState = (composeRequest, distroInfo, isBeta, isProd) => {
   if (composeRequest) {
     const imageRequest = composeRequest.image_requests[0];
     const uploadRequest = imageRequest.upload_request;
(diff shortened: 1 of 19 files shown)


src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -544,11 +544,16 @@ const CreateImageWizard = () => {
...
 
   const awsTarget = isBeta() ? awsTargetBeta : awsTargetStable;
   const msAzureTarget = isBeta() ? msAzureTargetBeta : msAzureTargetStable;
-  let initialState = requestToState(composeRequest, distroInfo, isBeta());
+  let initialState = requestToState(
+    composeRequest,
+    distroInfo,
+    isBeta(),
+    isProd()
+  );
```
lucasgarfield, line 556: Not very happy with how many arguments this function takes. Might need to do some refactoring in the future. Open to suggestions if anything seems obvious to you, dear reviewer.

jkozol: Can we put the is beta check into our global store?

jkozol (approved):
Works for me and seems to reduce the error messages.

PR #1072, opened 2023-04-13
test: Increase azure test timeout

ezr-ondrej:
The Jenkins job takes too long to execute the Azure tests.

This is a hotfix, we should rather speed up the tests.


lucasgarfield:
Would you mind adding a note to the commit message and description that we specifically need these because the Jenkin's job is failing? Might be useful in the future after we forget why we put these in ;)

lucasgarfield:
Also, I'd like to investigate using Jest's `--runInBand` option, apparently it can make a big difference in test speeds on resource constrained systems:
https://jestjs.io/docs/troubleshooting

lucasgarfield:
Yeah I totally agree, I'd love to do a deep dive and figure out why this specific test is causing problems but there's too much else to do right now. Let's just get this in so that failed CI isn't blocking us from making progress. :smile:

lucasgarfield (approved):
Thanks, hopefully we won't have to up these again... :wink:

PR #1076, opened 2023-04-14
CreateImageWizard: support verifying repository metadata

croissanne:
blocked on:
- [x] composer release
- [x] image-builder release


ochosi:
See MRs 65151 and 65154 against `app-interface`.

lucasgarfield (approved):
Nice work, I manually tested against a source with metadata verification set to true I made on stage and can confirm everything is working (from the front end's perspective)!

PR #1079, opened 2023-04-17
Add beta prompts

shaded-enmity:
This PR adds two prompts depending if user is in beta or not:
- Beta
  - Shows buttons to start quickstarts for AWS and Azure launch and Content sources
- Non-beta
  - Invites user to switch to beta to try out launch and content sources

Includes to-beta prompt from @croissanne
Includes quickstarts from @arburka (https://github.com/RedHatInsights/quickstarts/pull/84)

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useMemo } from 'react';
 
 import componentTypes from '@data-driven-forms/react-form-renderer/component-types';
 import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
@@ -544,6 +544,11 @@ const CreateImageWizard = () => {
 
   const handleClose = () => navigate(resolveRelPath(''));
 
+  const appendTo = useMemo(
+    () => document.querySelector('.pf-c-page.chr-c-page'),
+    []
+  );
+
   useEffect(() => {
     if (isBeta()) {
       dispatch(fetchRepositories());
@@ -610,6 +615,9 @@ const CreateImageWizard = () => {
           {
             component: componentTypes.WIZARD,
             name: 'image-builder-wizard',
+            ModalProps: {
+              appendTo,
+            },
             className: 'imageBuilder',
             isDynamic: true,
             inModal: true,
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -28,7 +28,7 @@ import {
 
 import './CreateImageWizard.scss';
 import api from '../../api';
-import { UNIT_GIB, UNIT_KIB, UNIT_MIB } from '../../constants';
+import { UNIT_GIB, UNIT_KIB, UNIT_MIB, MODAL_ANCHOR } from '../../constants';
 import { useGetArchitecturesByDistributionQuery } from '../../store/apiSlice';
 import { composeAdded } from '../../store/composesSlice';
 import { fetchRepositories } from '../../store/repositoriesSlice';
@@ -544,10 +544,7 @@ const CreateImageWizard = () => {
 
   const handleClose = () => navigate(resolveRelPath(''));
 
-  const appendTo = useMemo(
-    () => document.querySelector('.pf-c-page.chr-c-page'),
-    []
-  );
+  const appendTo = useMemo(() => document.querySelector(MODAL_ANCHOR), []);
 
   useEffect(() => {
     if (isBeta()) {
--- a/src/Components/LandingPage/LandingPage.js
+++ b/src/Components/LandingPage/LandingPage.js
@@ -1,14 +1,16 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 
 import {
   Alert,
   AlertActionCloseButton,
   Button,
+  ExpandableSection,
   Popover,
   Text,
   TextContent,
 } from '@patternfly/react-core';
 import {
+  ArrowRightIcon,
   HelpIcon,
   CodeBranchIcon,
   ExternalLinkAltIcon,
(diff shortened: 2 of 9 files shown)


lucasgarfield:
Awesome! One thing I noticed is that when the Wizard is open, it is no longer possible to see or scroll the quickstart, which makes following the steps very difficult...
quickstart and modal question

I'm not sure what the 'best' way to fix this is... one possible solution I found is setting the z-index of the quick start. Adding a high z-index (e.g. 9999) to the element with id of pf-drawer-panel-2 allows you to access both the modal and the quick start:
with z index

Then we would just need to dynamically adjust the size of the modal based on whether the quickstart is open or not and its size... thoughts?

shaded-enmity:
@lucasgarfield Thanks for taking a look and pointing out the issue with z index - @Hyperkid123 can you please take a look?

Hyperkid123:
We will have to append the modal to a different element than a body. It will require a minor change in DDF. I'll open a PR there in a little bit.

This is the outcome with the fix:
screenshot-prod foo redhat com_1337-2023 04 17-12_21_44

Hyperkid123:
OK This PR enables the `appendTo` props in `ModalProps`: https://github.com/data-driven-forms/react-forms/pull/1376

Once merged a new version is released, you can use this change to apply the new Modal root:
```diff
diff --git a/src/Components/CreateImageWizard/CreateImageWizard.js b/src/Components/CreateImageWizard/CreateImageWizard.js
index 0f8d140..72774c3 100644
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useMemo } from 'react';
 
 import componentTypes from '@data-driven-forms/react-form-renderer/component-types';
 import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
@@ -527,6 +527,11 @@ const CreateImageWizard = () => {
 
   const handleClose = () => navigate(resolveRelPath(''));
 
+  const appendTo = useMemo(
+    () => document.querySelector('.pf-c-page.chr-c-page'),
+    []
+  );
+
   useEffect(() => {
     if (isBeta()) {
       dispatch(fetchRepositories());
@@ -593,6 +598,9 @@ const CreateImageWizard = () => {
           {
             component: componentTypes.WIZARD,
             name: 'image-builder-wizard',
+            ModalProps: {
+              appendTo,
+            },
             className: 'imageBuilder',
             isDynamic: true,
             inModal: true,

```

arburka:
@Hyperkid123, @lucasgarfield , @shaded-enmity, this is a really nice fix!

Can you make sure the **Sources**, **Add Repo**, and **Launch** all have the same capability? 

You may have already done this, but the code above only had `the name: 'image-builder-wizard'`, in it.

shaded-enmity:
@Hyperkid123 Thanks for the quick fix! Updated the code and it's behaving correctly.

Can you please take a look at the point raised by @arburka - it seems like we need to address how modals interoperate with quickstarts in general.

shaded-enmity:
Tests have passed 🍿 @lucasgarfield can you please take another look?

lucasgarfield:
I just noticed that the quickstarts seem to break the `Share to Regions` modal and `Recreate image` features. Gif shows behavior with and without quickstart:

sharetoregionbroken
recreatebroken

src/Components/LandingPage/LandingPage.js
```diff
@@ -18,15 +20,33 @@ import {
...
 import ImagesTable from '../ImagesTable/ImagesTable';
-import './LandingPage.scss';
 import DocumentationButton from '../sharedComponents/DocumentationButton';
 
 export const LandingPage = () => {
   const [showBetaAlert, setShowBetaAlert] = useState(true);
+  const [showHint, setShowHint] = useState(true);
+
+  const { quickStarts } = useChrome();
+  const activateQuickstart = (qs) => quickStarts.toggle(qs.metadata.name);
```
ezr-ondrej, line 40: Is there a reason why we're not calling on the quickstarts currently defined on the platform? Duplicating the content feels not great.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -606,6 +611,9 @@ const CreateImageWizard = () => {
           {
             component: componentTypes.WIZARD,
             name: 'image-builder-wizard',
+            ModalProps: {
+              appendTo,
```
ezr-ondrej, line 615: This feels not great, but I guess there's no way around it currently? Why exactly does it break otherwise?

ezr-ondrej (commented):
Some questions. I think some of the mechanisms are not great, but we can merge and figure out how to make it better after if needed.

---

@croissanne speaking

ezr-ondrej:
@shaded-enmity also @arburka already got the Content of the quickstarts in the quickstarts repo in https://github.com/RedHatInsights/quickstarts/pull/84 shall we use that instead of copying the full content here?

src/Components/LandingPage/LandingPage.js
shaded-enmity: AFAIK those are available only on stage. But let me check.

src/Components/CreateImageWizard/CreateImageWizard.js
shaded-enmity: sure

shaded-enmity: updated

src/Components/LandingPage/LandingPage.js
shaded-enmity: no still not there

ezr-ondrej: It is fine if it's just in stage, Martin said he can push it within minutes. That should not be a reason to copy content

shaded-enmity: I spoke with him on Tuesday and he mentioned it's blocked by QE at the moment. 

@Hyperkid123 Hey Martin, is there an ETA when we can switch to using `quickStarts.activateQuickstart()`?

shaded-enmity:
@croissanne @ezr-ondrej ping

With @lucasgarfield work now merged 🎉 we should be ready to go. This needs to be merged & redeployed together with https://github.com/RHEnVision/provisioning-frontend/pull/249

As for the duplication of the QS content - once I hear from @Hyperkid123 that we can now use the platform QSs and the new API, I'll open another pull request to incorporate that.

croissanne (approved):
lgtm, thank you!

lucasgarfield (approved):
Awesome work, everything seems to be working perfectly on our end.

ezr-ondrej:
I'm ready to push the button on provisioning end as well :+1:

PR #1083, opened 2023-04-18
src: Remove `repos.js`

regexowl:
This removes `repos.js` file with the repository definitions as it's no longer needed after #963 was merged.


lucasgarfield (approved):
Love cleaning up! Thanks Klara :)

PR #1085, opened 2023-04-19
ImagesTable: Fix bug where Launch link incorrectly displayed

lucasgarfield:
This commit fixes a bug where the Launch link (which opens the Provisioning wizard) was incorrectly displayed for all image types.

The bug is currently in production beta, so this commit is needed for the hotfix. Changes made are minimal, only what is necessary to fix the bug - we still need to discuss the getImageProvider() function (the original source of the bug) with the Provisioning team.

### DIFF
--- a/src/Components/ImagesTable/ImageLink.js
+++ b/src/Components/ImagesTable/ImageLink.js
@@ -18,6 +18,7 @@ const getImageProvider = ({ imageType }) => {
     case 'azure':
       return 'azure';
     default:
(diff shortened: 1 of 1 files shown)


lucasgarfield:
The bug:
image

This PR:
image

regexowl (approved):
Seems to be working well, download and recreate links are back!

ochosi (approved):

PR #1086, opened 2023-04-19
ImagesTable: Add popovers with error details for failed builds

regexowl:
This removes error details from the image detail and moves them to popovers activated by clicking on "Image build failed" status.

Popovers were also added for clones which didn't include any error details previously.

### DIFF
--- a/src/Components/ImagesTable/ImageBuildStatus.js
+++ b/src/Components/ImagesTable/ImageBuildStatus.js
@@ -1,6 +1,13 @@
 import React from 'react';
 
-import { Flex } from '@patternfly/react-core';
+import {
+  Alert,
+  Button,
+  Flex,
+  Panel,
+  PanelMain,
+  Popover,
+} from '@patternfly/react-core';
 import {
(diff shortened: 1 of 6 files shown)


src/Components/ImagesTable/ImageBuildStatus.js
```diff
@@ -136,7 +160,37 @@ export const ImageBuildStatus = ({ imageId, isImagesTableRow }) => {
         messages[status].map((message, key) => (
           <Flex key={key} className="pf-u-align-items-baseline pf-m-nowrap">
             <div className="pf-u-mr-sm">{message.icon}</div>
-            {message.text}
+            {status === 'failure' ? (
+              <Popover
```
lucasgarfield, line 164: This is really nice! I like the idea of the scrollable pane. Do you think we could put this in its own React component (it can still live in this file) just so that the conditional rendering logic is a bit easier to understand at first glance?

lucasgarfield (approved):
Thank you!!!

PR #1089, opened 2023-04-21
LaunchWizard: Add large variant prop to set width of wizard correctly

lucasgarfield:
Without this prop, the wizard's width expands to fill the entire window.


regexowl (approved):
Nice one, works great!

PR #1091, opened 2023-04-21
Quickstarts: Add check for the last step of AWS and Azure

regexowl:
This adds the "Check your work" review for the 3rd step of AWS and Azure.


lucasgarfield (approved):
Thank you!

PR #1097, opened 2023-04-25
ImagesTable: Update ClonesTable

regexowl:
This updates ClonesTable according to the new mocks.

### DIFF
--- a/src/Components/ImagesTable/ClonesTable.js
+++ b/src/Components/ImagesTable/ClonesTable.js
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { ClipboardCopy } from '@patternfly/react-core';
 import {
   TableComposable,
   Tbody,
@@ -13,7 +14,6 @@ import { useSelector } from 'react-redux';
 
 import { ImageBuildStatus } from './ImageBuildStatus';
 
-import { useGetAWSSourcesQuery } from '../../store/apiSlice';
 import {
   selectClonesById,
   selectComposeById,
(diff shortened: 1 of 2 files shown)


regexowl:
Question regarding the AMIs - are those the same as UUID of the AWS clones?

lucasgarfield:
> Question regarding the AMIs - are those the same as UUID of the AWS clones?

No, the AMI is different than the UUID for the clones (and their parent composes, too). If you use the `selectCloneById` selector from the `clonesSlice` it has an `ami` property you can use to get the ami. :smile:

lucasgarfield:
This looks really great! I wish we could ensure that the column alignment was consistent. For instance, if you have two AWS images expanded the resulting output is a bit jarring:

image

I brought this up in a design meeting once and there was some discussion about the drawbacks of fixing column widths. I think we can merge this for now (nice work!) and implement a solution for consistent column locations in a future PR.

lucasgarfield:
Had a chat with @arburka about column widths. The solution actually ended up being a lot easier than I thought it would be to implement, so I say we put it in this PR.

image
image

regexowl:
Amazing! Thanks a lot @lucasgarfield 
I've updated the PR to include the column widths.

lucasgarfield:
Love it! Thanks!

lucasgarfield (approved):

regexowl:
Oh no :scream_cat: created a revert PR

jrusz:
> Oh no scream_cat created a revert PR

Thank you :) ... We really need to coordinate better on breaking changes

PR #1104, opened 2023-04-28
Embed payload repositories as custom repositories

croissanne:
custom repositories are added to /etc/yum.repos.d

---
image

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Repositories.js
+++ b/src/Components/CreateImageWizard/formComponents/Repositories.js
@@ -99,14 +99,30 @@ const BulkSelect = ({
   );
 };
 
-// Utility function to convert from Content Sources to Image Builder API schema
-const convertSchemaToImageBuilder = (repo) => {
+// Utility function to convert from Content Sources to Image Builder payload repo API schema
+const convertSchemaToIBPayloadRepo = (repo) => {
   const imageBuilderRepo = {
     baseurl: repo.url,
     rhsm: false,
   };
   if (repo.gpg_key) {
     imageBuilderRepo.gpgkey = repo.gpg_key;
+    imageBuilderRepo.check_gpg = true;
+    imageBuilderRepo.check_repo_gpg = repo.metadata_verification;
+  }
+
+  return imageBuilderRepo;
+};
+
+// Utility function to convert from Content Sources to Image Builder custom repo API schema
+const convertSchemaToIBCustomRepo = (repo) => {
+  const imageBuilderRepo = {
+    id: repo.uuid,
+    name: repo.name,
+    baseurl: [repo.url],
+  };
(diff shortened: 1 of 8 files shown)


lucasgarfield:
I just want to make sure I have this correct. 

We use the language 'Custom Repositories' in the Wizard. And for those instances, we are using the camelCase `customRepositories` variable. Variables stored in the form state use kebab-case, which was previously `custom-repositories` and is now `payload-repositories`. This brings it in alignment with the cloud API variable name, `payload_repositories`. 

Given we that the name of the step in the Wizard is "Custom Repositories", it makes sense to keep `customRepositories` and not convert it to `payloadRepositories`.

Just want to double check that this is intentional and not an oversight!

src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -99,7 +99,7 @@ const BulkSelect = ({
   );
 };
 
-// Utility function to convert from Content Sources to Image Builder API schema
+// Utility function to convert from Content Sources to Image Builder payload repo API schema
 const convertSchemaToImageBuilder = (repo) => {
```
lucasgarfield, line 103: Can we change this to convertSchemaToIBPayloadRepo?

lucasgarfield:
Looks great, everything makes sense! I would just like to request that we rename one function so things are crystal clear for our future selves. :smile:

croissanne:
> I just want to make sure I have this correct.
> 
> We use the language 'Custom Repositories' in the Wizard. And for those instances, we are using the camelCase `customRepositories` variable. Variables stored in the form state use kebab-case, which was previously `custom-repositories` and is now `payload-repositories`. This brings it in alignment with the cloud API variable name, `payload_repositories`.
> 
> Given we that the name of the step in the Wizard is "Custom Repositories", it makes sense to keep `customRepositories` and not convert it to `payloadRepositories`.
> 
> Just want to double check that this is intentional and not an oversight! My first instinct would have been to use sed to do the find replace and it obviously would miss this. smile

exactly, I just wanted to map the variable in the form state to the one we'll send in the api call. But happy to be told otherwise if the naming is less confusing. Plus we can always change it.

src/Components/CreateImageWizard/formComponents/Repositories.js
croissanne: absolutely

lucasgarfield:
> exactly, I just wanted to map the variable in the form state to the one we'll send in the api call. But happy to be told otherwise if the naming is less confusing. Plus we can always change it.

I think it is totally fine. Form state variable names match API, UI variables match UI.

lucasgarfield (approved):
Looks great! Nice work! :)

PR #1116, opened 2023-05-03
Wizard: Add filtering to repositories

lucasgarfield:
This commit adds filtering based on architecture and version to the repositories table in the wizard.

Apart from filtering, this commit now shows *all* repositories, regardless of their status. Previously, invalid repositories were not displayed. A future PR will add a status column and edge case handling (how to handle an invalid repo, etc...). For now, invalid repos are displayed and can be selected.

RTK Query is now used to manage the state of the repositories. Previously, the repositories were fetched in a useEffect hook upon opening the wizard.

MSW is now used instead of jest mocking for the tests involving repositories. The repositories test fixture now contains a function that mimics the content sources API.


src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -281,131 +315,177 @@ const Repositories = (props) => {
...
+      <EmptyStateBody>
+        Custom repositories managed via the Red Hat Insights Repositories app
+        will be available here to select and use to search for additional
+        packages.
+      </EmptyStateBody>
+      <Button
+        variant="primary"
+        component="a"
+        target="_blank"
+        href={isBeta() ? '/beta/settings/content' : '/settings/content'}
```
regexowl, line 475: Might need to be changed to `/preview/settings/content`

regexowl:
One thing I'm thinking about - in case there's a longer list of repositories, do we want to scroll the entire content of the step or just the list? Might be nice to be able to filter and refresh the list without having to scroll all the way up, what do you think?

Also this is just a "cosmetic" change so we can take a look at it separately.

Screenshot from 2023-05-04 10-19-29

lucasgarfield:
> One thing I'm thinking about - in case there's a longer list of repositories, do we want to scroll the entire content of the step or just the list? Might be nice to be able to filter and refresh the list without having to scroll all the way up, what do you think?
> 
> Also this is just a "cosmetic" change so we can take a look at it separately.
> 
> Screenshot from 2023-05-04 10-19-29

Hm, yes that is a great idea to make the only table scrollable, I like it!

src/Components/CreateImageWizard/formComponents/Repositories.js
lucasgarfield: Nice catch!

src/Components/CreateImageWizard/steps/repositories.js
```diff
@@ -1,13 +1,32 @@
...
+  const { isBeta } = useGetEnvironment();
+  return (
+    <Button
+      component="a"
+      target="_blank"
+      variant="link"
+      iconPosition="right"
+      isInline
+      icon={<ExternalLinkAltIcon />}
+      href={isBeta() ? '/beta/settings/content' : '/settings/content'}
```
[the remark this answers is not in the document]
lucasgarfield: Good catch!

regexowl (approved):
Works great, thanks! :star:

lucasgarfield:
Currently failing IQE tests but @jrusz has issued an MR that should get them passing: https://gitlab.cee.redhat.com/insights-qe/iqe-image-builder-plugin/-/merge_requests/40

PR #1119, opened 2023-05-04
ImagesTable: Ensure `created_at` date for a new image

regexowl:
This adds a temporary value of current date to `created_at` for newly created images.

This is not the most elegant solution and will be reworked after the ImagesTable refactor.


lucasgarfield (approved):
I actually think this is a perfectly elegant solution. The actual response from the API only contains the id of the compose. This is essentially an optimistic update, which we also do for the request and the image_status.

Great work!

PR #1120, opened 2023-05-04
Wizard: Prefetch data for external API calls

lucasgarfield:
This makes the wizard feel much 'faster' for a better user experience. The implementation uses RTK Query for prefetching. For most API requests, prefetching is done when the 'next' button is hovered over on the previous step.


regexowl (approved):
Works like magic, thank you! :sparkle:

PR #1121, opened 2023-05-04
Add Zero State

johnsonm325:


src/Router.js
```diff
@@ -8,14 +12,52 @@ const LandingPage = lazy(() => import('./Components/LandingPage/LandingPage'));
...
-      </Route>
-    </Routes>
+    <Suspense
+      fallback={
+        <Bullseye>
+          <Spinner size="xl" />
+        </Bullseye>
+      }
+    >
+      {!hasSystems ? (
```
croissanne, line 42: I think we should show the zero state based on whether or not the user has images, not systems.

src/Router.js
```diff
@@ -8,14 +12,52 @@ const LandingPage = lazy(() => import('./Components/LandingPage/LandingPage'));
 const CreateImageWizard = lazy(() =>
   import('./Components/CreateImageWizard/CreateImageWizard')
 );
+const INVENTORY_TOTAL_FETCH_URL = '/api/inventory/v1/hosts';
```
croissanne, line 15: We generally keep those in /src/constants

croissanne (changes requested):
Thank you! I like the look of the new ZeroState page.

I have a few questions though

1. could you add some more information in the PR description / commit message? 
2. Can you make the format of the commit message to be:
```
Router: Add Zero State page

$body
``` ?

Happy to drag this over the finish line ourselves though.

PR #1122, opened 2023-05-05
ImagesTable/ShareImageModal: Update sharing options for AWS

regexowl:
This disables sharing to new regions in case of failed build of the parent image.

The regions that always end in failure were also disabled in the `ShareImageModal`.

Screenshot from 2023-05-05 09-27-53

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.js
+++ b/src/Components/ImagesTable/ImagesTable.js
@@ -145,6 +145,7 @@ const ImagesTable = () => {
     {
       title: 'Share to new region',
       onClick: () => navigate(resolveRelPath(`share/${compose.id}`)),
+      isDisabled: compose?.image_status?.status === 'success' ? false : true,
     },
     ...actions(compose),
   ];
--- a/src/Components/ShareImageModal/RegionsSelect.js
+++ b/src/Components/ShareImageModal/RegionsSelect.js
@@ -43,7 +43,8 @@ export const selectRegionsToDisable = createSelector(
 const prepareRegions = (regionsToDisable) => {
   const regions = AWS_REGIONS.map((region) => ({
     ...region,
-    disabled: regionsToDisable.has(region.value),
+    disabled:
+      regionsToDisable.has(region.value) || region?.disableRegion === true,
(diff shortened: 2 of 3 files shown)


src/Components/ImagesTable/ImagesTable.js
```diff
@@ -267,7 +267,8 @@ const ImagesTable = () => {
                       </Td>
                       <Td>
                         {compose.request.image_requests[0].upload_request
-                          .type === 'aws' ? (
+                          .type === 'aws' &&
+                        compose?.image_status?.status === 'success' ? (
```
lucasgarfield, line 271: This hides the 'Share to region' action. Could we just disable it instead?

I checked the PF source code and I think that if you add an `isDisabled: true` property to the `awsActions` (line 144) it should work.

src/Components/ShareImageModal/RegionsSelect.js
```diff
@@ -43,7 +43,8 @@ export const selectRegionsToDisable = createSelector(
 const prepareRegions = (regionsToDisable) => {
   const regions = AWS_REGIONS.map((region) => ({
     ...region,
-    disabled: regionsToDisable.has(region.value),
+    disabled:
+      regionsToDisable.has(region.value) || region?.disableRegion === true,
```
lucasgarfield, line 47: Could we add this logic to the `selectRegionsToDisable` selector (line 26) instead?

src/Components/ImagesTable/ImagesTable.js
regexowl: Fixed

src/Components/ShareImageModal/RegionsSelect.js
lucasgarfield: This seems like it's going to be kind of tricky, we will refactor when we migrate to RTKQ.

regexowl: This is a bit tricky, let's keep it after the refactor

lucasgarfield (approved):
Thank you!

PR #1124, opened 2023-05-05
ClonesTable: Remove `created_at` column from the clones table

regexowl:
This removes the `created_at` column from the clones table as per the latest designs.


lucasgarfield:
Running into same issue where different UUID lengths for different parent composes result in different column widths.

Here was how we decided to fix it for the details PR:
image

Maybe we could do spacing of 40:20:20:20 for UUID:Account:Region:Status?

lucasgarfield (approved):
Nice!!!

PR #1126, opened 2023-05-05
Wizard: Adjust AWS launch popover for images created in preview

lucasgarfield:
If an image does not contain a share_with_accounts property, launch links do not appear in the launch popover. This happens when a user creates an AWS image in beta/preview, and then clicks Launch in stable.

A message is now displayed that explains the image was created using Preview features, and the link to Preview has been clarified to be more explicit.

image


regexowl (approved):
Works great, thanks!

PR #1127, opened 2023-05-05
Wizard: Add Azure launch link for images created in preview

lucasgarfield:
If an image does not contain a subscription_id property, the link to view the image in Azure will be malformed. This happens when a user creates an Azure image in beta/preview using sources, and then clicks the 'View uploaded image' link in stable.

Now, if an Azure image was created using sources (as evidenced by presence of a source_id in request's upload options), the Launch button appears and its popover has a link to Preview.

image

### DIFF
--- a/src/Components/ImagesTable/ImageLinkDirect.js
+++ b/src/Components/ImagesTable/ImageLinkDirect.js
@@ -2,6 +2,7 @@ import React from 'react';
 
 import {
   Button,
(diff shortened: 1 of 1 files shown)


regexowl (approved):
The code looks good. I don't have any Azure images/source so didn't test in the frontend, but from the screenshot looks like exactly what we need.

PR #1136, opened 2023-05-15
Wizard: Image name validation

regexowl:
This adds validation of an image name based on a pattern, which was chosen as an intersection of the naming constrains of hyperscalers.

The pattern is based on the one for naming GCP images as those naming guidelines are the most restrictive. With added minimal length restriction of AWS.

Sources:
- [Azure](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules)
- [GCP](https://cloud.google.com/compute/docs/naming-resources)
- [AWS](https://docs.aws.amazon.com/cli/latest/reference/ec2/create-image.html) and
Screenshot from 2023-05-15 15-00-04

As the length of the name for AMI differs between the CLI reference and the screenshot of the error the stricter rule was chosen.

### DIFF
--- a/src/Components/CreateImageWizard/steps/imageName.js
+++ b/src/Components/CreateImageWizard/steps/imageName.js
@@ -32,8 +32,10 @@ export default {
       autoFocus: true,
       validate: [
         {
-          type: validatorTypes.MAX_LENGTH,
-          threshold: 100,
+          type: validatorTypes.PATTERN,
+          pattern: /^[a-z][-a-z0-9]{1,61}[a-z0-9]$/,
+          message:
+            'The image name can be 3-63 characters long. It can contain lowercase letters, digits and hyphens, has to start with a letter and cannot end with a hyphen.',
         },
       ],
     },
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.beta.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.beta.test.js
@@ -740,7 +740,7 @@ describe('Click through all steps', () => {
     const nameInput = screen.getByRole('textbox', {
       name: 'Image name',
     });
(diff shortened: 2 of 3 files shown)


croissanne (approved):
lgtm!

croissanne (approved):

regexowl:
@jrusz I think this could possibly break tests. It'll change the pattern for validation of the image name to lowercase letters, numbers and hyphens only (new pattern is `^[a-z][-a-z0-9]{1,61}[a-z0-9]$`). 

The images that are build for the testing include underscores in the name as well, right? So it probably wouldn't get past the validation on the Name image step. Would it be possible to change the name of the test images?

cc @croissanne

croissanne:
> @jrusz I think this could possibly break tests. It'll change the pattern for validation of the image name to lowercase letters, numbers and hyphens only (new pattern is `^[a-z][-a-z0-9]{1,61}[a-z0-9]$`).
> 
> The images that are build for the testing include underscores in the name as well, right? So it probably wouldn't get past the validation on the Name image step. Would it be possible to change the name of the test images?
> 
> cc @croissanne

I think the notification engine is currently broken in eph ;_; easiest to just check the iqe plugin directly to see what name they use

jrusz:
> @jrusz I think this could possibly break tests. It'll change the pattern for validation of the image name to lowercase letters, numbers and hyphens only (new pattern is `^[a-z][-a-z0-9]{1,61}[a-z0-9]$`).
> 
> The images that are build for the testing include underscores in the name as well, right? So it probably wouldn't get past the validation on the Name image step. Would it be possible to change the name of the test images?
> 
> cc @croissanne

Yeah it seems it does use something like `image-under-test-2023-07-10_11-59-39_536988` so there is an underscore. I'll go ahead and change it somehow.

PR #1143, opened 2023-05-23
Reduce test warnings

croissanne:


regexowl:
Trying to understand this. The goal is to log info about a specific error being suppressed but not the error itself, is that right?
 
Screenshot from 2023-05-24 11-55-50

croissanne:
> 

ah wait, it should have muted the red bit I think. MAybe I got it wrong somewhere, I only wanted to show the 'Suppressed error:...' line.

croissanne:
@regexowl ok now it should actually suppress :facepalm:

src/test/jest.setup.js
```diff
@@ -0,0 +1,63 @@
...
+    this.console = console;
+    this.filters = [filter1];
+  }
+
+  logSuppressedError(err) {
+    this.console.info('Suppressed error: ', err);
+  }
+
+  filter(...args) {
+    for (let fn of this.filters) {
```
regexowl, line 36: It does and it's glorious! You'll need to switch `let` for `const` here to make the linter happy, but other than that looks good to me. Thank you!

regexowl (approved):
Amazing, thank you!

PR #1144, opened 2023-05-24
test: Move button clicking functions to `testUtils`

regexowl:
This moves the `clickBack`, `clickCancel`, `clickNext`, `getNextButton` and `verifyCancelButton` functions from the tests to `testUtils`.


lucasgarfield:
Very nice start on the enhanced test utilities idea!

Should we be using the new userEvent.click() for these, as well? Obviously these work fine, but for the sake of consistency it might be a good idea...

regexowl:
@lucasgarfield Good point - updated! Added some async/await for good measure.

croissanne (approved):

PR #1146, opened 2023-05-29
MSW: Integrate MSW into browser

lucasgarfield:
This commit integrates Mock Service Worker into the browser. This will allow us to use MSW to mock network requests at the API level from a browser using the same MSW fixtures that we are currently using for our testing. This means we will be able to have the same environment for debugging, development, and testing if desired.

To develop using MSW, the developer will now use `npm run stage-beta:msw`. The dev server will start as normal, but now any request made that has a corresponding fixture will be intercepted and the fixture used to generate the response.

This adds an MSW=TRUE environment variable to the run script. When this environment variable is present, the dev webpack config will copy mockServiceWorker.js into the build directory (/dist) using CopyPlugin and will add some necessary headers to the dev server's responses using DefinePlugin.

### DIFF
--- a/src/AppEntry.js
+++ b/src/AppEntry.js
@@ -5,6 +5,26 @@ import { Provider } from 'react-redux';
 import App from './App';
 import { store } from './store';
 
+if (process.env.NODE_ENV === 'development' && process.env.MSW === true) {
+  // process.env.MSW is set in the webpack config using DefinePlugin
+  const { worker } = require('./test/mocks/browser');
+  worker.start({
+    serviceWorker: {
+      url: '/beta/apps/image-builder/mockServiceWorker.js',
+      options: {
+        /* 
+        Service workers can only intercept requests made from within their scope.
+        mockServiceWorker.js is served from /beta/apps/image-builder/, which becomes
+        the worker's default scope. Set scope to '/' so that all requests are in scope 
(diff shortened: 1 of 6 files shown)


src/mockServiceWorker.js
```diff
@@ -0,0 +1,303 @@
+/* eslint-disable */
```
lucasgarfield, line 1: In case it isn't obvious, this file is programmatically generated by MSW and doesn't need to be reviewed. ;)

src/AppEntry.js
```diff
@@ -5,6 +5,26 @@ import { Provider } from 'react-redux';
...
+  const { worker } = require('./test/mocks/browser');
+  worker.start({
+    serviceWorker: {
+      url: '/beta/apps/image-builder/mockServiceWorker.js',
+      options: {
+        /* 
+        Service workers can only intercept requests made from within their scope.
+        mockServiceWorker.js is served from /beta/apps/image-builder/, which becomes
+        the worker's default scope. Set scope to '/' so that all requests are in scope 
+        and can be intercepted. Note that the the Service-Worker-Allowed header must
```
regexowl, line 19: > Note that the the Service-Worker-Allowed

Just a nitpick - duplicate "the"

regexowl (approved):
This looks and works great! Amazing job Lucas!! :tada:
Now we just need to write ourselves some more handlers

PR #1147, opened 2023-05-29
Mock Service Worker: Relative handler URLs

lucasgarfield:
This commit removes the base URL from the MSW handlers. This is
necessary in order for MSW to integrate into the browser (future commit)
and does not adversely affect the current test integration with MSW.


regexowl (approved):
Works great, thank you!

PR #1149, opened 2023-05-30
add tabs to image builder table

mgold1234:
This commit adds tabs to image builder table - https://issues.redhat.com/browse/THEEDGE-3275.
This is necessary to integrate edge frontend images on the same page. one tab for image builder and one tab for edge images.
when adding these tabs to 'landingpage' we will cause the 'create image' button will appear under tabs instead of the upper section - related to this story - https://issues.redhat.com/browse/THEEDGE-3354.
and prevent the duplicate toolbar that will show in edge image tab - related to that story https://issues.redhat.com/browse/THEEDGE-3361

---

The webpack changes are for using local federated modules.

### DIFF
--- a/src/Components/LandingPage/LandingPage.js
+++ b/src/Components/LandingPage/LandingPage.js
@@ -6,6 +6,9 @@ import {
   Button,
   ExpandableSection,
   Popover,
+  Tabs,
+  Tab,
+  TabTitleText,
   Text,
   TextContent,
 } from '@patternfly/react-core';
@@ -17,11 +20,13 @@ import {
   PageHeaderTitle,
 } from '@redhat-cloud-services/frontend-components';
 import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
+import { useFlag } from '@unleash/proxy-client-react';
 import { Outlet } from 'react-router-dom';
 
 import './LandingPage.scss';
 
 import { useGetEnvironment } from '../../Utilities/useGetEnvironment';
+import EdgeImagesTable from '../edge/ImagesTable';
 import ImagesTable from '../ImagesTable/ImagesTable';
 import DocumentationButton from '../sharedComponents/DocumentationButton';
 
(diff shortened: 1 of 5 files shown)


mgold1234:
pr is blocked until this pr https://github.com/RedHatInsights/image-builder-frontend/pull/1152 will merge

src/Components/LandingPage/LandingPage.js
```diff
@@ -32,122 +37,149 @@ export const LandingPage = () => {
...
+              Build an Image with Custom Content
+            </Button>
+          </p>
+        </ExpandableSection>
+      )}
+      <ImagesTable />
+    </section>
+  );
+  return (
+    <Suspense fallback="loading...">
```
lucasgarfield, line 132: I think I would prefer to wrap only the imported module (specifically `<EdgeImagesTable>`) in `<Suspense >` for now... 

Just a bit of context on why: we aren't using `<Suspense>` in this project (apart from the other imported module, the Launch wizard). For cases where we need to load data, we are currently using RTK Query hooks and conditional rendering. RTK Query does not support `<Suspense>` at the moment because, according to the React docs, "Suspense-enabled data fetching without the use of an opinionated framework is not yet supported".

src/Components/LandingPage/LandingPage.js
```diff
@@ -32,122 +37,149 @@ export const LandingPage = () => {
...
-          </ExpandableSection>
+            <Tab
+              eventKey={0}
+              title={<TabTitleText>Traditional (RPM - DNF)</TabTitleText>}
+            >
+              {traditionalImageList}
+            </Tab>
+            <Tab
+              eventKey={1}
+              title={<TabTitleText>Immutable - (OSTree)</TabTitleText>}
```
lucasgarfield, line 172: I think that this should be `Immutable (OSTree)` to match the designs. Basically, remove the `-`. :sweat_smile:

mgold1234: thank you @lucasgarfield removed

src/Components/LandingPage/LandingPage.js
mgold1234: thank you for your explanation:)

lucasgarfield (approved):
I think this is a great start! Nice work (both you and colleagues) Michal!

PR #1155, opened 2023-06-01
Webpack: Run proxy without VPN for production environments

lucasgarfield:
This allows users to develop against production without having access to the internal network.

https://github.com/RedHatInsights/frontend-components/tree/62f37029ca0f5f7e9b0fd61815fe62c96503bcbf/packages/config#running-prod-proxy-without-vpn

This was previously introduced with #49b3f04 and then reverted because it seemed to cause problems with developing against stage. However, I am unable to replicate that - stage is working fine at the moment with these changes in place. Perhaps it was a coincidence - in any case, we should try again and if problems with stage arise, we will investigate them and fix them. It is important to be able to develop against the front-end outside of the VPN in order to support our open source service objectives.

### DIFF
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -4,6 +4,9 @@ const config = require('@redhat-cloud-services/frontend-components-config');
 
 const webpackProxy = {
   useProxy: true,
+  // Configure useAgent and bounceProd to allow access to prod from outside the VPN
+  useAgent: process.env.STAGE ? true : false,
+  bounceProd: process.env.STAGE ? false : true,
   proxyVerbose: true,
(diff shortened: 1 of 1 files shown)


regexowl:
Both prod-beta and prod-stable end up in this lock screen for me. It's the same being on and off VPN. Stage works just fine. If I remember it right, the behavior was the same when we were trying these changes previously.

Screenshot from 2023-06-01 12-59-58

lucasgarfield:
Very odd, it was working for me before I created the pull request and now I am experiencing the same thing. I will reach out to the consoledot folks for help. Changing to draft until this is resolved.

PR #1157, opened 2023-06-01
LandingPage: Update Open Source Badge

lucasgarfield:
The open source badge now lives in @redhat-insights/frontend-components.


ochosi (approved):
Wonderful work - thanks for this!
Just tried it locally against prod-beta and it works as expected.

PR #1158, opened 2023-06-01
Revert "ImagesTable: Update ClonesTable"

regexowl:
Reverts RedHatInsights/image-builder-frontend#1097


regexowl:
@lucasgarfield @croissanne First time reverting a PR "the proper way" as it breaks the hms-integration testing, could I get some eyes on it, just to be sure I didn't mess it up?

cc @jrusz

jrusz:
Thanks, navigation fails now as it changed in ephemeral, I need to fix that (at least it's finally unified again). I'll check this out manually but IMO a revert like this is ok...

PR #1159, opened 2023-06-05
ImagesTable: Add image details and update ClonesTable

regexowl:
This adds expandable rows details for each image and updates ClonesTable according to the new [mocks](https://www.sketch.com/s/d7aa6d29-fca0-4283-a846-09cc5fd10612/a/qbbyGg3).

### DIFF
--- a/src/Components/ImagesTable/ClonesTable.js
+++ b/src/Components/ImagesTable/ClonesTable.js
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { ClipboardCopy } from '@patternfly/react-core';
 import {
   TableComposable,
   Tbody,
@@ -13,7 +14,6 @@ import { useSelector } from 'react-redux';
 
 import { ImageBuildStatus } from './ImageBuildStatus';
 
-import { useGetAWSSourcesQuery } from '../../store/apiSlice';
 import {
   selectClonesById,
   selectComposeById,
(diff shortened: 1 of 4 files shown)


regexowl:
:wave: @jrusz 
Here's the PR with both changes to the clones table (included in the reverted https://github.com/RedHatInsights/image-builder-frontend/pull/1097) and the updated image details

jrusz:
> wave @jrusz Here's the PR with both changes to the clones table (included in the reverted #1097) and the updated image details

Thank you! I'll work on adjusting the test suite right away.

jrusz:
Hi @regexowl is the `Source` supposed to be the ID or the name of it? From what I can see it shows the ID but in the sketch I saw a name... 
image

lucasgarfield:
> Hi @regexowl is the `Source` supposed to be the ID or the name of it? From what I can see it shows the ID but in the sketch I saw a name... image

@jrusz, for now let's assume that the name should be displayed. The source id is only used internally. We will just need to update the code in this PR so that we resolve the name from the id before displaying it.

regexowl:
Hey @jrusz, the Source is indeed supposed to show the name of the source. I'll update the code. Thank you for letting me know about this!

regexowl:
Displaying of the source name is fixed!

@jrusz We will be ready to merge after the review, would that be alright from your side?

jrusz:
> Displaying of the source name is fixed!
> 
> @jrusz We will be ready to merge after the review, would that be alright from your side?

Yeah go ahead, there are unrelated issues in ephemeral so I won't make it work now. Thanks for letting me know!

jrusz:
I may have found another issue. There is no information shown when the image build is done via API, is that expected? The following two images are pretty much the same, the UI one is still building though.
image

regexowl:
Alright, this should do the trick. Thanks a lot! Those were great points :+1:

lucasgarfield (approved):
Looks great!

PR #1161, opened 2023-06-05
Wizard: Hotfix for recreate image rate limiting issue

lucasgarfield:
Related to HMS-1886: Clicking the recreate button on a build with a lot of packages results in the user being rate limited.

Package summaries are obtained by querying the Content Sources API. A large number of packages requires a large number of requests and leads to the rate limiting problem when using the 'Recreate image' function.

This commit fixes the problem by not requesting or displaying package summaries when 'Recreate image' is used. (Note that newly added packages will still request and display a summary.)

The 'Recreate image' function will soon be replaced by blueprints. This hotfix should carry us forward until then.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -263,49 +263,6 @@ const parseSizeUnit = (bytesize) => {
   return [size, unit];
 };
 
-const getDistributionRepoUrls = (distributionInformation) => {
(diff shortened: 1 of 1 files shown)


regexowl (approved):
Seems to be working well and it also has a nice "side effect" - the complete list of packages is displayed immediately. 

When I've tried to recreate an image with 115 packages, I didn't get 403'ed. But when clicking between the packages and custom packages steps the lists were slowly updating as the requests were resolving. This eliminates the problem.

PR #1171, opened 2023-06-09
Wizard: Enable Azure and GCP provisioning links in ephemeral

lucasgarfield:
Unleash is being used to determine whether or not to use the Launch wizard for Azure and GCP images. Unleash is not yet supported in our ephemeral environments, but we want our ephemeral environments to have parity with beta. This commit ensures that the Launch wizard opens for Azure and GCP images in ephemeral.


jrusz (approved):
Thank you!

PR #1172, opened 2023-06-14
ImagesTable: Remove `Image name` Spinner for failed Azure images

regexowl:
This removes Spinner next to the Image name when a build of an Azure image fails.


lucasgarfield (approved):
Getting the bugs out!

PR #1175, opened 2023-06-15
ImagesTable: Update Image commands popover for GCP images

regexowl:
This adds "Launch an instance" and "Save a copy" commands to the Image commands popover for GCP images.

Mocks: https://www.sketch.com/s/d7aa6d29-fca0-4283-a846-09cc5fd10612/a/qbbyGg3

### DIFF
--- a/src/Components/ImagesTable/ImageLinkDirect.js
+++ b/src/Components/ImagesTable/ImageLinkDirect.js
@@ -2,11 +2,12 @@ import React from 'react';
 
 import {
   Button,
+  ClipboardCopy,
+  ClipboardCopyVariant,
   Divider,
   Popover,
   Text,
   TextContent,
-  TextVariants,
 } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
@@ -18,6 +19,14 @@ import { RegionsPopover } from './RegionsPopover';
 import { selectImageById } from '../../store/composesSlice';
 import { resolveRelPath } from '../../Utilities/path';
 import BetaLabel from '../sharedComponents/BetaLabel';
+
+const launchInstanceCommand = (uploadStatus) => {
+  return `gcloud compute instances create ${uploadStatus.options.image_name}-instance --image-project ${uploadStatus.options.project_id} --image ${uploadStatus.options.image_name}`;
+};
+
+const saveCopyCommand = (uploadStatus) => {
+  return `gcloud compute images create ${uploadStatus.options.image_name}-copy --source-image-project ${uploadStatus.options.project_id} --source-image ${uploadStatus.options.image_name}`;
+};
 
 const ImageLinkDirect = ({ imageId, isExpired, isInClonesTable }) => {
   const navigate = useNavigate();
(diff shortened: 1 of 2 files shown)


lucasgarfield:
The copy action doesn't seem to be working for me, when I click `copy` and paste the results somewhere, `[object Object]` is what was on the clipboard.

regexowl:
Hmm, yes, that makes sense. 

We'll have to choose between the formatted text with a different copy method or a `ClipboardCopy` with plain text as any html tags within will break the copy function.

I'd say the second option is better as it's consistent with the usage of a `ClipboadCopy` in the image details.

croissanne (approved):
amazing

src/Components/ImagesTable/ImageLinkDirect.js
```diff
@@ -110,37 +110,44 @@ const ImageLinkDirect = ({ imageId, isExpired, isInClonesTable }) => {
...
             </Text>
+            <ClipboardCopy
+              hoverTip="Copy"
+              clickTip="Copied"
+              variant="inline-compact"
+              ouiaId="gcp-save-copy"
+              isBlock
+            >
+              gcloud compute images create MY_IMAGE_NAME --source-image-project
+              PROJECT_ID_FROM_RESPONSE --source-image IMAGE_NAME
```
croissanne, line 144: Wait I think the variables need to be filled in actually 
image

src/Components/ImagesTable/ImageLinkDirect.js
```diff
@@ -110,37 +110,44 @@ const ImageLinkDirect = ({ imageId, isExpired, isInClonesTable }) => {
...
             </Text>
+            <ClipboardCopy
+              hoverTip="Copy"
+              clickTip="Copied"
+              variant="inline-compact"
+              ouiaId="gcp-launch-instance"
+              isBlock
+            >
+              gcloud compute instances create INSTANCE_NAME --image-project
+              PROJECT_ID_FROM_RESPONSE --image IMAGE_NAME --zone GCP_ZONE
```
croissanne, line 130: Variables need to filled in with the details

croissanne (changes requested):

src/Components/ImagesTable/ImageLinkDirect.js
regexowl: @croissanne Ok, I wasn't sure about this. Some of the variables are straightforward. If I get it right, the `INSTANCE_NAME` and `MY_IMAGE_NAME` would be identifiers set by user independently, right?

If that's correct we could do a partial fill. But what about `GCP_ZONE`? Is there a default we could use?

regexowl: It seems to set `ZONE` based on an environment variable if the tag is not defined, maybe? Can we just leave that part out?

croissanne: Indeed, but we can set the project_id, the IMAGE_NAME, and then we can prefill their name as "IMAGE_NAME-copy" maybe?


For the instance maybe

INSTANCE_NAME=IMAGE_NAME-instance ; GCP_ZONE we can leave as blank I guess.

regexowl: Sounds good, I've pushed the changes!

croissanne:
image
can we increase the padding a bit? In the designs the grey areas are slightly larger.

Also either remove the `\`s or add newlines? I think it would be nice if it was a bit structured somehow as:

```
create instance \
        composer-api... \
        --image-poroject.... \
        --....
```

or just 

```
create instance composer-api... --image-project... --image .... 
```

Not sure what would look better.

regexowl:
@croissanne Added the padding.

I've removed \s from the commands. Structured command would be nicer, but adding formatting to the text seems to break the copy function. :(

croissanne:
image

Maybe the margin was a bad idea, sorry :/ I also can't get it to increase the grey zone without pushing the icon out.

The only alternative I can think of is completely different from the designs. 
image

IG let's just remove the margin ? idk @arburka

lucasgarfield:
Just want to point out that Sanne's proposed fix is a standard Patternfly pattern, see here:
https://www.patternfly.org/v4/components/clipboard-copy/#read-only-expanded-by-default

I say we go for it!

croissanne:
> The popover grows/shrinks with the expansion, correct?
 

yup, but I think maybe default expanded would be good? (isReadOnly, isExpanded and variant=ClipboardCopyVariant.expansion})

regexowl:
Thanks for the comments everyone! I've tried to change the ClipboardCopy to the read only expanded by default variant. 

The popover does adjust to expanding/shrinking the text fields. My vote would also go to expanded by default as I feel it's more transparent than having to click another button to see the complete command before copying it.

lucasgarfield:
As per @jrusz, CI is currently failing due to issues regarding the CPU limit to spawn the test pod.

I don't believe that the IQE tests would touch this part of the code, so I am merging.

ezr-ondrej:
@regexowl, @croissanne this has disabled the Launch button for GCP, should we rather move these commands to Launch and keep the button opening the Provisioning wizard?

CC @adiabramovitch

croissanne:
> @regexowl, @croissanne this has disabled the Launch button for GCP, should we rather move these commands to Launch and keep the button opening the Provisioning wizard?
> 
> CC @adiabramovitch

Ah good point. But we also want this in stable. I'd love for these commands to be integrated into the launch wizard and then showing this when there's no source (in preview).

ezr-ondrej:
> I'd love for these commands to be integrated into the launch wizard and then showing this when there's no source (in preview).

I'd love that as well! :) make sense, I'll talk to @adiabramovitch how to proceed, we will just need to make sure the Launch button shows up in preview, so the check should be on the feature flag IMO, but that's an easy fix ;)

PR #1179, opened 2023-06-19
GCP: rename Email label to Principal

lzap:
The GCP account identifier is not email, it can be, but not always. For example service accounts do have identifiers which look like emails, but they are not. This is also confirmed by the GCP docs. This patch renames that label to "Principal" which is also used in the GCP docs.

Now, to avoid confusion, I am keeping (e.g. e-mail address) just to not put too much stress on the customer trying to figure this out.

Warning: I don’t know React or any JS library pretty much so this might break all tests - review with care.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
+++ b/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
@@ -153,7 +153,7 @@ export const TargetEnvGCPList = () => {
           {googleAccType?.[getState()?.values?.['google-account-type']] ===
           'Domain'
             ? 'Domain'
-            : 'Email address'}
+            : 'Principal'}
         </TextListItem>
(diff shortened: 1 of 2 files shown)


lucasgarfield:
Thanks, Lukáš! I think it's a great idea to bring the UI in line with the GCP docs.

We have some validation on that field in the Wizard, as well... right now, it _must_ be an e-mail address. We'll probably want to fix the validation at the same time.

Can you provide a link to the docs? I'd like to determine what kind of validation we should actually perform (if any).

lzap:
Hey, docs: https://cloud.google.com/iam/docs/overview

The principal always looks like email and in many cases it is, but for example for service account principal it is actually not. It looks like `something@something.gserviceaccount.com` but if you send an email there it is not delivered. So I think you can keep the validators as-is.

PR #1180, opened 2023-06-19
Initial Typescript setup

lucasgarfield:
The first commit provides the basic plumbing needed for our project to work with Typescript.
The second commit adds global type definitions for the API endpoints we are currently querying with RTK Query.


tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "noImplicitAny": true,
```
kingsleyzissou, line 4: 👏👏👏

PR #1186, opened 2023-06-20
build(deps): Regress `data-driven-forms` libraries to version 3.20.5

regexowl:
This reverts an update of:
- `@data-driven-forms/pf4-component-mapper`
- `@data-driven-forms/react-form-renderer`

from version 3.20.8 to version 3.20.5.

The reason being a bug ending up in an "Node.removeChild: The node to be removed is not a child of this node"

Steps to reproduce the bug:
1. Open Image Wizard with a Create image button
2. Close the Wizard with Cancel button
3. Repeat two or three times

Also can be reproduced with:
1. Click on Kebab in the image row of the Images Table
2. Choose Recreate image
3. Close the Wizard with Cancel button

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,8 +8,8 @@
       "name": "image-builder",
       "version": "1.1.0",
       "dependencies": {
-        "@data-driven-forms/pf4-component-mapper": "3.20.8",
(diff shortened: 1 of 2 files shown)


regexowl:
Screenshot of the error:

error

PR #1192, opened 2023-06-21
Components/ReviewStepTextLists: remove useEffect

lavocatt:
According to the documentation (https://react.dev/learn/you-might-not-need-an-effect) useEffect should only be used to synchronize with external systems.

In the FSC component showing the partition table useEffect was used to compute a value based on the form state.

The documentation present two relatable usecases to what we were doing in the component:
- https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes
- https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

This commit uses the presented strategies to externalise in a new component the computation of the MinSize. Because it has for dependency the list of partitions and the manual setting, it will get redrwan every time these values are changed in the form state.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
+++ b/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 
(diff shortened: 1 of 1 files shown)


regexowl (approved):
This looks great! Thanks @lavocatt!

PR #1199, opened 2023-06-27
eslintrc: Add React hooks plugin

lucasgarfield:
This plugin enforces the React 'rules of hooks'. See here: https://legacy.reactjs.org/docs/hooks-rules.html

(These are the legacy docs but I believe the information is still accurate.)

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -6,6 +6,7 @@ globals:
   mount: 'readonly'
 plugins: 
   - import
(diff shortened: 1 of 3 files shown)


lucasgarfield:
Draft PR for now until we have fixed the errors (and ideally the warnings).

lucasgarfield:
Closing as this is covered by #1303.

PR #1206, opened 2023-06-28
pr_check: Enable testing with downstream MR images

jrusz:


jrusz:
Changes moved to the cicd-tools repo so that other teams can use this as well https://github.com/RedHatInsights/cicd-tools/pull/10

I'm just adding some more variables, especially the `IQE_IMAGE_TAG` is important now as we need a default set by us.

lucasgarfield (approved):
LGTM

PR #1208, opened 2023-06-29
ImagesTable: Fix the GCP launch link

regexowl:
This fixes the bug when Launch button got replaced with Image command button even in the Preview.


ezr-ondrej:
I think that would be correct! :)
And we'll work on adding the launch commands on our side, when there is no source to launch directly :)

PR #1213, opened 2023-07-03
Wizard: change the image name

### LINKED ISSUE #1196: Update Image Name step to Details as per mocks
Adjust the texts and content of the Image Name step as per [mocks](https://www.sketch.com/s/d7aa6d29-fca0-4283-a846-09cc5fd10612/a/Axek07M). 

Screenshot from 2023-06-27 09-33-29

mgold1234:
fixes #1196
this commit update the wording on image name step
when user create new image.


src/Components/CreateImageWizard/steps/imageName.js
```diff
@@ -20,15 +20,18 @@ export default {
       name: 'plain-text-component',
       label: (
         <p>
-          Optionally enter a name for your image. All images will have a UUID.
+          Optionally enter a name and description to identify your image later
```
regexowl, line 23: Let's remove "and description" for now as we are not adding the text input for Description in this commit.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1810,7 +1810,7 @@ describe('Keyboard accessibility', () => {
     clickNext();
 
     // Name
-    const nameInput = screen.getByRole('textbox', { name: /image name/i });
+    const nameInput = screen.getByRole('textbox', { name: /Image Name/i });
```
regexowl, line 1813: We can leave the previous `{ name: /image name/i }` here as the `/i` means the check is case insensitive.

regexowl:
Perfect! Just two small nitpicks and a then a rebase

regexowl:
Looks good! It only needs a small lint fix :)

regexowl:
@jrusz Changes in this PR will affect most of the texts on the "Name image" step. Could that cause any problems with the integration tests?

jrusz:
> @jrusz Changes in this PR will affect most of the texts on the "Name image" step. Could that cause any problems with the integration tests?

So the tests don't seem to care as they don't really check if they're on the "Image name" step :smile: ... But I am going to prepare an update regardless. Thanks for bringing this to my attention @regexowl :slightly_smiling_face:

jrusz:
Btw, what is that description good for? I don't see it anywhere after I create the image...

regexowl:
Thanks for checking this @jrusz!

The description should be an optional field for adding more information about the image. For now it should show on the Review step when recreating it as far as I can tell from the mocks.

regexowl:
Thank you @mgold1234, this looks great! Can you please rebase and update the image name `helperText` with "The image name can be 3-63 characters long. It can contain lowercase letters, digits and hyphens, has to start with a letter and cannot end with a hyphen."? We've recently updated the name validation :)

regexowl (approved):
Looks great, thank you! :sparkles:

PR #1214, opened 2023-07-03
CreateImageWizard: Add support for vsphere-ova imagetype

croissanne:
The OVA type is also the default when checking VMWare now.

image

I chose different tooltip text for the vmdk, I found the text supplied in the mocks not super informative, so I decided to stress the usage of it as well.

I also changed the tooltip text on the ova sligthly, mostly to indicate that it also includes the vmdk.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
+++ b/src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
@@ -5,10 +5,14 @@ import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
 import {
   Checkbox,
   FormGroup,
+  Popover,
+  Radio,
   Text,
+  TextContent,
   TextVariants,
   Tile,
 } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
 
 import { usePrefetch } from '../../../store/apiSlice';
@@ -20,6 +24,7 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
     aws: false,
     azure: false,
     gcp: false,
+    'vsphere-ova': false,
     vsphere: false,
     'guest-image': false,
     'image-installer': false,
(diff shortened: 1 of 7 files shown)


src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
```diff
@@ -32,129 +37,203 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
...
+              <Popover
+                maxWidth="30rem"
+                position="right"
+                bodyContent={
+                  <TextContent>
+                    <Text>
+                      An OVA file is a virtual appliance used by virtualization
+                      platforms such as VMWare vSphere. It is a package that
+                      contains files used to describe a virtual machine, which
+                      includes an VMDK image, OVF descriptor file and a manifest
```
regexowl, line 160: Should it be "a VMDK image"?

src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
```diff
@@ -32,129 +37,203 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
...
+            isSelected={environment.azure}
+            isStacked
+            isDisplayLarge
           />
-        </FormGroup>
+        </div>
       </FormGroup>
-    </>
+      <FormGroup
+        label={<Text component={TextVariants.small}>Private cloud</Text>}
```
regexowl, line 124: Can we put slightly more space between the public cloud buttons and the "Private cloud" text?

src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
```diff
@@ -32,129 +37,203 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
...
+            </>
+          }
+          onChange={(checked) => {
+            handleSetEnvironment('vsphere-ova', checked);
+            handleSetEnvironment('vsphere', !checked);
+          }}
+          isChecked={environment['vsphere-ova']}
+          isDisabled={!(environment.vsphere || environment['vsphere-ova'])}
+        />
+        <Radio
```
regexowl, line 177: Just a thought - could there be a scenario when the user needs to create both types of VMWare images and we would need checkboxes instead of a radio button?

regexowl:
The .OVA image will need to be added to `targetOptions` in `Target.js`

image

src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
croissanne: I thought radios were more apprioriate, because the ova is just a tar file which contains the vmdk inside of it. So you can just extract the vmdk with tar -xvf composer-api.ova.

regexowl: Aah, got it! Radios make perfect sense then. :+1:

src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
croissanne: I added a `pf-u-mt-sm`, lmk if you think `pf-u-mt-md` would be better

regexowl: `pf-u-mt-sm` feels right, thanks!

regexowl (approved):
Looks great! :sparkles:

PR #1188, opened 2023-06-20
Use RTKQ for compose and clone statuses

lucasgarfield:

PR #1219, opened 2023-07-06
ImagesTable: refactor target to TypeScript

mgold1234:
This commit replace Proptype to TypeScript for Target component as part of improving the code to use TypeScript and enjoy the benefits of it.


src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+  aws: 'Amazon Web Services',
+  'aws.s3': 'Amazon Web Services',
+  azure: 'Microsoft Azure',
+  gcp: 'Google Cloud Platform',
+  vsphere: 'VMWare',
+  'guest-image': 'Virtualization - Guest image',
+  'image-installer': 'Bare metal - Installer',
+  'edge-commit': 'edge-commit',
+  'edge-installer': 'edge-installer',
+  ami: 'ami',
```
croissanne, line 20: This should be the same as the `aws` text.

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+  useGetComposeStatusQuery,
+} from '../../store/apiSlice';
+
+type TargetProps = {
+  composeId: string;
+};
+
+const targetOptions = {
+  aws: 'Amazon Web Services',
+  'aws.s3': 'Amazon Web Services',
```
croissanne, line 12: For s3 targets we rely on the image type, so this should never be used.

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+  gcp: 'Google Cloud Platform',
+  vsphere: 'VMWare',
+  'guest-image': 'Virtualization - Guest image',
+  'image-installer': 'Bare metal - Installer',
+  'edge-commit': 'edge-commit',
+  'edge-installer': 'edge-installer',
+  ami: 'ami',
+  'rhel-edge-commit': 'rhel-edge-commit',
+  'rhel-edge-installer': 'rhel-edge-installer',
+  vhd: 'vhd',
```
croissanne, line 23: Same, we never use the upload type for vhd, always the image type.

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+type TargetProps = {
+  composeId: string;
+};
+
+const targetOptions = {
+  aws: 'Amazon Web Services',
+  'aws.s3': 'Amazon Web Services',
+  azure: 'Microsoft Azure',
+  gcp: 'Google Cloud Platform',
+  vsphere: 'VMWare',
```
croissanne, line 15: This needs a rebase as this changed, + `vsphere-ova` was added.

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+  'rhel-edge-installer': 'rhel-edge-installer',
+  vhd: 'vhd',
+};
+
+const Target = ({ composeId }: TargetProps) => {
+  const { data: clones, isSuccess: isSuccessClones } =
+    useGetClonesQuery(composeId);
+  const { data: composeStatus, isSuccess: isSuccessTarget } =
+    useGetComposeStatusQuery(composeId);
+  const target = composeStatus?.request?.image_requests[0];
```
croissanne, line 31: I think this variable name is confusing, maybe call it imageReq?

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  let tar;
+
+  if (isSuccessTarget) {
```
croissanne, line 35: This is a change in behaviour, we didn't check if the target succeeded before. Why is this necessary?

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+      tar = targetOptions[target?.image_type];
+    } else if (target?.upload_request?.type === 'aws') {
+      if (isSuccessClones) {
+        tar =
+          targetOptions[target?.upload_request?.type] +
+          ` (${clones.meta.count !== 0 ? clones.meta.count + 1 : 1})`;
+      }
+    } else {
+      tar = targetOptions[target?.upload_request?.type];
+    }
```
croissanne, line 46: I think we can actually simplify this by only specialcasing the `aws` and `ami` image_types. If it's `aws` or `ami`, check for clones. Otherwise just use the `image_type` in the image request.

I'd try to drop any use of `upload_request.type` if possible.

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }
+  if (tar) {
+    return tar;
+  }
+  return targetOptions[target?.image_type] || null;
```
croissanne, line 51: Note that we never returned null: ` return <>{target ? target : compose.imageType}</>;`. Default to the raw `image_type`. In case we don't have the type in `targetOptions`.

croissanne (changes requested):
Thank you!

The use of typescript isn't the biggest change in this commit imo, it's supporting more image types in Target, plus there seems to actually be some change in the behaviour like defaulting to `null`.

Also it fails to compile for me.

src/Components/ImagesTable/Target.tsx
regexowl: aws.s3 can be safely left out, but if we leave out vhd, TypeScript shows a warning `Property 'vhd' does not exist on type`. 

We are getting the types from the ImageTypes enum in `imageBuilderSchema.ts` which was auto-generated by openapi-typescript. The vhd type is a part of ImageTypes values in `image-builder` so we could either leave it here even thought we never use it or remove it from the api specification as well to silence the error. Or at least I think those would be the options

src/Components/ImagesTable/Target.tsx
regexowl: Hmm, this is not really checking for the success of the target, rather for the success of fetching the compose information as the `useSelector()` hook is replaced by fetching the data via RTKQuery hook. This ensures we are not trying to read an image type from a request that is not resolved yet, ending up with undefined type.

A change in the name of the variable might be in place from `isSuccessTarget` to `isSuccessComposeStatus`

src/Components/ImagesTable/Target.tsx
regexowl: +1

It would also allow to change the `tar` back to `target` which I'd say would be preferable.

src/Components/ImagesTable/Target.tsx
mgold1234: ami doesnt exist in the schema  : UploadTypes: "aws" | "gcp" | "azure" | "aws.s3";
do I need to add it?

src/Components/ImagesTable/Target.tsx
mgold1234: done @regexowl thank you

mgold1234:
It looks like typeScript  slows down the loading of Target column, comparing it to Proptype :(

src/Components/ImagesTable/Target.tsx
mgold1234: change the name to isSuccessComposeStatus

src/Components/ImagesTable/Target.tsx
regexowl: I've just pushed a PR (https://github.com/RedHatInsights/image-builder-frontend/pull/1247) with updates to our fixtures. With the changes this PR should pass tests. 

What we should probably still do though is try to avoid returning null as that could cause unforeseen problems later. Either with something like this:
```
if (target) {
	return target
} else {
	return composeStatus?.request?.image_requests[0]?.image_type
}
```
or with a ternary operator similar to the one we've used before:
```
return <>{target ? target : composeStatus?.request?.image_requests[0]?.image_type}</>;
```

The problem is both break the tests as of now even though everything renders without any problems. Still if we need to "bend" something, I feel it should rather be tests than the code. What do you think?

I'll continue looking into the failing tests.

lucasgarfield:
The target values load slowly because we are making duplicate API requests (plus the usual slowness caused by using the proxy).

I will work on getting #1188 merged and that should greatly increase the loading speed.

lucasgarfield:
Blocked until #1188 is in.

PR #1224, opened 2023-07-10
test/ImagesTable: Remove `scrollTo is not a function` console error

regexowl:
This eliminates the `Error: Uncaught [TypeError: stepBody.scrollTo is not a function]` console error by mocking `scrollTo` as a prototype.

The error was caused by `scrollTo` not being implemented in JSDOM.


lucasgarfield (approved):
Yay, another console warning taken care of! We'll get them all eventually!

PR #1229, opened 2023-07-11
Wizard: Remove `ModalProps` from schema

regexowl:
These changes were introduced in #1079 to ensure that quickstarts stay visible even after opening a Wizard modal as they are both top-most components.

`ModalProps` which were used for this purpose were introduced in version 3.20.6 of `@data-driven-forms/pf4-component-mapper`. Since then we had to regress the version to 3.20.5 as it was crashing the application on closing of a modal. The issue linked to this can be found here: https://github.com/data-driven-forms/react-forms/issues/1389

As we are now using mapper version that does not allow to use `ModalProps` the functionality of the workaround was lost and it was causing following console.error: `Warning: React does not recognize the ModalProps prop on a DOM element.`

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 
 import componentTypes from '@data-driven-forms/react-form-renderer/component-types';
 import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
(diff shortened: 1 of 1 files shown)


regexowl:
This is how the page looks with version 3.20.6 which causes crashes:

Screenshot from 2023-07-11 13-42-33

And this is how it looks now without `ModalProps` available:

Screenshot from 2023-07-11 13-43-42

I've opened an issue to address this: https://github.com/RedHatInsights/image-builder-frontend/issues/1230

croissanne (approved):

PR #1231, opened 2023-07-11
feat(RHIF-255): Add Zero State to image builder

adonispuente:
This PR adds zero state to image builder. 
To test use an account with zero systems, or change the logic for what the state is supposed to show and verify things look correct. 

This zero state should pop up when there are 0 images on the account.

Fixes https://issues.redhat.com/browse/HMS-2184


croissanne:
Thank you! A few thoughts:

1. Now that edge is also integrating, I think we should figure out how those two are supposed to play together.
2. I wonder if we should show it when linking directly to the wizard
3. The beta prompt/popup is no longer visible if there's no images

src/Router.js
```diff
@@ -11,28 +15,80 @@ const LandingPage = lazy(() => import('./Components/LandingPage/LandingPage'));
...
+    }
+  }, [hasSystems]);
   const edgeParityFlag = useFlag('edgeParity.image-list');
   return (
-    <Routes>
-      <Route path="*" element={<LandingPage />}>
-        <Route path="imagewizard/:composeId?" element={<CreateImageWizard />} />
-        <Route path="share/:composeId" element={<ShareImageModal />} />
-      </Route>
+    <Suspense
```
lucasgarfield, line 35: I don't think that I want to do this at the <Router>... maybe we can move this to <LandingPage> instead?

lucasgarfield:
Hi @adonispuente, would you mind squashing the second commit into the first one?

lavocatt:
It would be nice to wait up until https://github.com/RedHatInsights/image-builder-frontend/pull/1320 is merged so that we can use the new hook to know how many images are in the images table.

src/Router.js
```diff
@@ -1,17 +1,102 @@
...
+  }
+  const composes = data.data;
+  switch (from) {
+    case 'ZeroState':
+      // navigating to the LandinPage is only allowed from the ZeroState page
+      if (composes.length > ZERO_IMAGES) {
+        navigate(resolveRelPath(''));
+      }
+      break;
+    case 'LandinPage':
```
lucasgarfield: Small typo here and in a few of the comments, this should be LandingPage (with a g :wink:).

src/Router.js
```diff
@@ -1,17 +1,102 @@
...
+ *
+ * caveat:
+ * This component takes care of displaying itself the CreateImageWizard. Meaning
+ * that the wizard is not being displayed using the navigation system. This is
+ * valid for-now™ [2023-09] because the wizard is being displayed in a modal.
+ * This would most likely have to get updated when the wizard will get changed
+ * to an InPage mode.
+ */
+const AppZeroState = () => {
+  const [wizardOpenned, setWizardOpenned] = useState(false);
```
lucasgarfield: It isn't an official convention in our code base yet but I usually like to start boolean variable names with `is`, what do you think of `isWizardOpened`?

lavocatt: Very good idea.

src/Components/ShareImageModal/RegionsSelect.tsx
```diff
@@ -82,7 +82,7 @@ const RegionsSelect = ({
   });
 
   if (!isSuccess) {
-    return undefined;
```
lucasgarfield, line 85: Returning `undefined` should actually be allowed now, after e5896d8d9c56713a0a3668e80f44547830812175. You might need to update your dependencies with an `npm ci` to get your tsserver to not complain about it.

src/Components/ShareImageModal/ShareImageModal.tsx
```diff
@@ -32,7 +32,7 @@ const ShareToRegionsModal = () => {
 
   if (!composeId) {
     handleClose();
-    return undefined;
```
lucasgarfield, line 35: See comment above, `npm ci` should allow you to return undefined.

src/Router.tsx
```diff
@@ -59,43 +58,42 @@ const AppZeroState = () => {
   );
 };
 
-const ZERO_IMAGES = 0;
-
```
lucasgarfield: As in previous commit, please change to `LandingPage` with small 'g'. :wink:

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "./dist/",
     "noImplicitAny": true,
-    "module": "es6",
+    "module": "esnext",
```
lucasgarfield, line 5: Could this have some implications on browser compatibility? What doesn't work with `es6`?

src/Router.tsx
```diff
@@ -59,43 +58,42 @@ const AppZeroState = () => {
...
+type ZeroStateNavigationType = {
+  from: 'ZeroState' | 'LandinPage';
```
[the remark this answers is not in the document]
lavocatt: How man, I'm happy you get to see those typos :smile:  How about the `Landin'Page` (just kidding)

tsconfig.json
lavocatt: I have no clue :shrug:. If you change it back, eslint's gonna rant about our current lazy imports not being possible. So I guess we were not compatible before ? But without knowing it.

src/Router.tsx
```diff
@@ -59,43 +58,42 @@ const AppZeroState = () => {
...
+        break;
+      case 'LandinPage':
+        // navigating to the ZeroState page is only allowed from the LandinPage
+        if (composes.length === ZERO_IMAGES) {
+          navigate(resolveRelPath('zerostate'));
+        }
+        break;
+    }
   }
+  return <></>;
```
lucasgarfield, line 95: We should be able to return `undefined` now.

src/Components/ShareImageModal/ShareImageModal.tsx
lavocatt: It's working for eslint, not in my nvim, it's weird.

lavocatt: Right, I updated my coc modules, it's now fine.

src/Router.tsx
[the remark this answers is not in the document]
lucasgarfield: :rofl: Landin'Page sounds great!

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "./dist/",
     "noImplicitAny": true,
-    "module": "es6",
+    "module": "es2020",
```
lucasgarfield, line 5: I'd like to do just a bit more research into this, but my initial understanding is es2020 should be very close to es6 so this is probably fine. Let's add a comment here as to why (`es6 does not allow dynamic imports, which we use for React lazy() loading/code splitting`? These config files tend to become very opinionated in mysterious ways as time passes.

lavocatt:
@jrusz this PR might break the IQE tests

jrusz:
> @jrusz this PR might break the IQE tests

There are errors in the unit tests. The image didn't get built because of that so I can't test the change.

jrusz:
Also that commit doesn't look quite right. Idk about the content, but the message is definitely from a different commit.

lavocatt:
ah! I messed up a rebase it seems .. I went too fast!

lavocatt:
The unit tests are fixed @jrusz all good for you test it.

jrusz:
> The unit tests are fixed @jrusz all good for you test it.

Thanks! I'll get on it.

jrusz:
Lol it actually passed :smile: ... Since I know already where to update the ouiaID I'm going to do that and then change it a little bit to make it more stable.

lavocatt (changes requested):
Quoting @croissanne: 
> Theoretically there are 3 "empty" states:  an empty page state (returning user who's images expired), the zeroState, and the unentitled page.The unentitled page is of no concern to us because as soon as you have RHEL you're entitled. The empty page state will go away with blueprints (since those don't expire). Meaning after blueprints we can do the new zerostate.

There's no good ways to distinguish between the 3 states rn, so I'm blocking this work up until the blueprints wizard is there.

PR #1234, opened 2023-07-12
LandingPage: Update the "About" popover

regexowl:
This updates the text on the "About" popover and adds links according to recent [mocks](https://www.sketch.com/s/ac25729e-df9b-4e74-a09c-badb62787d24/a/QDREEpx).

### DIFF
--- a/src/Components/LandingPage/LandingPage.js
+++ b/src/Components/LandingPage/LandingPage.js
@@ -13,7 +13,11 @@ import {
   Text,
   TextContent,
 } from '@patternfly/react-core';
-import { ArrowRightIcon, HelpIcon } from '@patternfly/react-icons';
+import {
+  ArrowRightIcon,
+  ExternalLinkAltIcon,
+  HelpIcon,
+} from '@patternfly/react-icons';
 // eslint-disable-next-line rulesdir/disallow-fec-relative-imports
 import {
   OpenSourceBadge,
@@ -31,7 +35,6 @@ import { resolveRelPath } from '../../Utilities/path';
 import { useGetEnvironment } from '../../Utilities/useGetEnvironment';
 import EdgeImagesTable from '../edge/ImagesTable';
 import ImagesTable from '../ImagesTable/ImagesTable';
-import DocumentationButton from '../sharedComponents/DocumentationButton';
 
 export const LandingPage = () => {
   const [showBetaAlert, setShowBetaAlert] = useState(true);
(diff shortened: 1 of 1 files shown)


regexowl:
I wasn't sure what's the convention for spacing in RPM-DNF (vs. RPM - DNF), went with no spaces around the dash everywhere.

lucasgarfield:
This popover has a lot of text and renders poorly, both at large and small viewport sizes. I don't think there is very much that we can do about it, unfortunately...

For instance, here it is opening to the right instead of down for some reason: 
image

Forcing it to open down doesn't improve the situation (look at the nav bar not properly rendering):
image

Would it be better to split this out into two or three icons and popovers? That's more in line with Patternfly guidelines for using popovers:
https://www.patternfly.org/v4/components/popover/design-guidelines#content-considerations
**Popover copy is longer than tooltip copy, but it should still be concise. Aim for 1–3 sentences.**

regexowl:
Hmm, those are good points. Splitting the popover would make sense. I wonder how many icons we would end up with in the end though as other functionality might get added (we would already be on four now). Other thing that came to mind would be making the popover fixed size and the content scrollable. But that would go against PF guidelines... Maybe opening the information in a modal instead of a popover could work?

I would add this to the agenda for the next UX/dev sync so we can get more opinions on how to solve this, if that make sense

regexowl:
Tried to split the popover into three:
image

image

image

image

I agree it makes the amount of information more manageable. If we're okay with the three (?) then I 'd say let's maybe do it this way. I can also try to check how the scrollable popover would look like, but I'm thinking if I'd be looking for more information about the Traditional vs. Immutable I'd probably wouldn't click on a popover next to the title.

lucasgarfield:
Wow, I actually really like how the 3 popovers looks. Great work @regexowl! If you like it too and want to push then I would be happy to merge.

Thanks for the feedback, @kelseamann!
>I am wondering if- and this is more of a technical question- we can pepper some of this information in throughout the process? For example, we can show them how easy it is to track further updates on the review screen for OSTree images... this is a lot of information for users to remember/reference.

Maybe we want to make quickstarts for some of these workflows, like we've done for Image Builder?

regexowl:
Perfect, thanks a lot for the feedback! I've pushed the three popovers.

lucasgarfield (approved):
Great work! All links are working (for now). We should probably figure out a good way to manage all of these external links, they are starting to become hard to maintain... but for now this is just fine.

PR #1235, opened 2023-07-12
ImagesTable: Update wording and links on empty state

regexowl:
This updates text and links on empty state according to recent mocks.

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.js
+++ b/src/Components/ImagesTable/ImagesTable.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 
 import {
+  Button,
   EmptyState,
   EmptyStateBody,
   EmptyStateIcon,
@@ -8,12 +9,13 @@ import {
   EmptyStateVariant,
   Pagination,
   PaginationVariant,
+  Text,
   Title,
   Toolbar,
   ToolbarContent,
   ToolbarItem,
 } from '@patternfly/react-core';
-import { PlusCircleIcon } from '@patternfly/react-icons';
+import { ExternalLinkAltIcon, PlusCircleIcon } from '@patternfly/react-icons';
 import {
   ActionsColumn,
   ExpandableRowContent,
@@ -42,7 +44,6 @@ import {
   hoursToExpiration,
   timestampToDisplayString,
 } from '../../Utilities/time';
-import DocumentationButton from '../sharedComponents/DocumentationButton';
 
 const ImagesTable = () => {
   const [page, setPage] = useState(1);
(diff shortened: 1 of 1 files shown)


src/Components/ImagesTable/ImagesTable.js
```diff
@@ -175,7 +205,20 @@ const ImagesTable = () => {
...
+              variant="link"
+              icon={<ExternalLinkAltIcon />}
+              iconPosition="right"
+              isInline
+              href={
+                'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/creating_customized_rhel_images_using_the_image_builder_service/index'
+              }
+              className="pf-u-pt-md"
+            >
+              Image builder for RPM - DNF documentation
```
lucasgarfield, line 220: This has spaces but on line 177 it is without spaces. I think without spaces looks better. According to some random grammar websites, in US English an en dash should not have spaces around it. We can remove the spaces from the tabs to be consistent in another PR.

regexowl: Agree, spaces removed! I've tried to remove them everywhere, but missed this

src/Components/ImagesTable/ImagesTable.js
```diff
@@ -175,7 +205,20 @@ const ImagesTable = () => {
...
-            <DocumentationButton />
+            <Button
+              component="a"
+              target="_blank"
+              variant="link"
+              icon={<ExternalLinkAltIcon />}
+              iconPosition="right"
+              isInline
+              href={
+                'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/creating_customized_rhel_images_using_the_image_builder_service/index'
```
lucasgarfield, line 216: This seems to redirect here, I wonder if we should use the redirected URL?
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/creating_customized_images_by_using_insights_image_builder/index

src/Components/ImagesTable/ImagesTable.js
```diff
@@ -160,12 +161,41 @@ const ImagesTable = () => {
...
+            </Text>
+            <br />
+            <Text>
+              With RPM-DNF, you can manage the system software by using the DNF
+              package manager and updated RPM packages. This is a simple and
+              adaptive method of managing and modifying the system over its
+              lifecycle.
+            </Text>
+            <br />
+            <Text>
```
lucasgarfield, line 183: It seems like this `<Text>` component might not be necessary, what do you think?

regexowl: It has effect on the spacing. If we want the double empty space between the link and the "Create image" button, we can add another line break instead of the Text component. Both variants those work for me.

Looks like this without the `<Text>` tags and extra linebreak:
image

src/Components/ImagesTable/ImagesTable.js
regexowl: Good point, the original url behaves a bit strange. Updated!

src/Components/ImagesTable/ImagesTable.js
regexowl: Switched the `Text` component for one extra linebreak

regexowl:
@lucasgarfield Tried to remove the line breaks. It seems the automatic padding for the `Text` components in `<EmptyStateBody>` gets stripped. 

image

We can get the proper spacing if we enclose the `Text` components in a `TextContent`, but then we loose the `EmptyStateBody` formatting.

image

So for now what I'd do is leave the `Text` components with two extra line breaks so the formatting is as it should be. What do you think?

lucasgarfield (approved):
Looks great, matches the designs perfectly, nice work!

PR #1238, opened 2023-07-13
LandingPage: Fixes tabs selection on browser back/forward.

ldjebran:
In the context of HMS-parity with edge management, fixes the current behavior when using browser history back and forward.
- When selecting a tab and then push the browser history back/forward, the tab of the current url is selected.

FIXES: https://issues.redhat.com/browse/THEEDGE-3460

image-builder-tabs-browser-back-forward


regexowl (approved):
Looks good to me, thank you! :tada:

PR #1247, opened 2023-07-18
test/fixtures: Update compose fixture and migrate it to TypeScript

regexowl:
This adds `request` to a reponse from the `composes/{composeId}` endpoint and `meta` to a response from the `/composes/{composeId}/clones` endpoint as per API documentation.

The incomplete responses were causing problems when fetching data from the endpoints mentioned above.

The fixture was also migrated to TypeScript.

### DIFF
--- a/src/test/fixtures/composes.tsx
+++ b/src/test/fixtures/composes.tsx
@@ -1,9 +1,19 @@
 import { RHEL_8 } from '../../constants';
+import {
+  ClonesResponse,
+  ComposeStatus,
+  ComposesResponse,
+  UploadStatus,
+} from '../../store/imageBuilderApi';
 
 // CreateImageWizard mocks
-export const mockComposesEmpty = {
+export const mockComposesEmpty: ComposesResponse = {
   meta: {
     count: 0,
+  },
+  links: {
+    first: '',
+    last: '',
   },
   data: [],
 };
@@ -12,9 +22,13 @@ export const mockComposesEmpty = {
 const currentDate = new Date();
 const currentDateInString = currentDate.toISOString();
 
-export const mockComposes = {
+export const mockComposes: ComposesResponse = {
   meta: {
     count: 12,
+  },
+  links: {
+    first: '',
+    last: '',
   },
   data: [
     {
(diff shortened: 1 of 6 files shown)


regexowl:
These changes will be needed to test https://github.com/RedHatInsights/image-builder-frontend/pull/1219

regexowl:
Maaybe? Let me take a look at it :eyes:

regexowl:
This should do the trick. Skipping my overly ambitious plan to type all the fixtures in one PR as that would make it a bit too much I think.

The only untyped mocks remaining for now are ShareImageModal ones as those mock a state and not an api response.

src/test/fixtures/composes.tsx
```diff
@@ -386,45 +580,67 @@ export const mockClones = (composeId) => {
...
+    return ({
+      options: {
+        ami: 'ami-9fdskj12fdsak1211',
+        region: 'us-west-2',
+      },
+      status: 'failure',
+      type: 'aws',
+    })
+  } else {
+    // uuid: '4a851db1-919f-43ca-a7ef-dd209877a77e'
```
lucasgarfield, line 634: This is a bit of a personal preference, but I feel like the long if/else if/else conditionals in this file are a bit cumbersome. It's very verbose and there are so many returns to keep track of.

It's also a bit of a code smell for me to keep track of which UUID we are returning in a comment as we do here in the else statement... we have to maintain the comment, and there is no guarantee that the comment will be correct in the future.

I would prefer to move back to use an object to keep track of everything, and just have a single return statement at the end of the function similar to `return statuses[uuid]`... but at the very least, I would like to move this UUID into its own else if and do something different with the else statement (we could consider some error handling but it might be overkill).

lucasgarfield (commented):
This is going to make our lives so much easier! :grin: 

Can we change `composes.tsx` to 'composes.ts`? The convention I am planning to follow is to use `.tsx` for files that include React jsx syntax, and `.ts` for "pure" Typescript files.

src/test/fixtures/composes.tsx
```diff
@@ -232,124 +246,298 @@ export const mockComposes = {
...
-  '1579d95b-8f1d-4982-8c53-8c2afa4ab04c': {
-    image_status: {
-      status: 'success',
-      upload_status: {
-        options: {
-          ami: 'ami-0217b81d9be50e44b',
-          region: 'us-east-1',
-        },
+export const mockStatus = (composeId: string): ComposeStatus => {
+  const mockComposes = {
```
lucasgarfield, line 250: We can actually type this object when it is declared, rather than typing the individual keys:
```suggestion
  const mockComposes: { [key: string]: ComposeStatus } = {
```

src/test/fixtures/composes.tsx
```diff
@@ -232,124 +246,298 @@ export const mockComposes = {
...
-    },
-  },
-  'ca03f120-9840-4959-871e-94a5cb49d1f2': {
-    image_status: {
-      status: 'success',
-      upload_status: {
-        options: {
-          image_name: 'composer-api-d446d8cb-7c16-4756-bf7d-706293785b05',
-          project_id: 'red-hat-image-builder',
+    } as ComposeStatus,
```
lucasgarfield, line 278: If we type the object at declaration, `as ComposeStatus` is no longer needed.

src/test/fixtures/composes.tsx
```diff
@@ -232,124 +246,298 @@ export const mockComposes = {
...
+            upload_request: {
+              options: {},
+              type: 'aws.s3',
+            },
+          },
+        ],
+      },
+    } as ComposeStatus,
+  };
+  return mockComposes[composeId as keyof typeof mockComposes];
```
lucasgarfield, line 526: If mockComposes has its types declared at declaration, Typescript will be able to infer this information and we can just write:
```suggestion
return mockComposes[composeId];
```

src/test/fixtures/composes.tsx
```diff
@@ -386,45 +574,64 @@ export const mockClones = (composeId) => {
...
-  '4a851db1-919f-43ca-a7ef-dd209877a77e': {
-    options: {
-      ami: 'ami-9fdskj12fdsak1211',
-      region: 'eu-central-1',
-    },
-    status: 'success',
-    type: 'aws',
-  },
+export const mockCloneStatus = (cloneId: string): UploadStatus => {
+  const mockClones = {
```
lucasgarfield, line 600:
```suggestion
  const mockClones: { [key: string]: UploadStatus } = {
```

src/test/fixtures/composes.tsx
```diff
@@ -386,45 +574,64 @@ export const mockClones = (composeId) => {
...
+    'f9133ec4-7a9e-4fd9-9a9f-9636b82b0a5d': {
+      options: {
+        ami: 'ami-0e778053cd490ad21',
+        region: 'us-west-1',
+      },
+      status: 'success',
+      type: 'aws',
+    } as UploadStatus,
```
lucasgarfield, line 608: We can remove `as UploadStatus` if we type the object.

src/test/fixtures/composes.tsx
```diff
@@ -386,45 +574,64 @@ export const mockClones = (composeId) => {
...
+    '4a851db1-919f-43ca-a7ef-dd209877a77e': {
+      options: {
+        ami: 'ami-9fdskj12fdsak1211',
+        region: 'eu-central-1',
+      },
+      status: 'success',
+      type: 'aws',
+    } as UploadStatus,
+  };
+  return mockClones[cloneId as keyof typeof mockClones];
```
lucasgarfield, line 634: Typing the object will allow TS to infer this info.
```suggestion
return mockClones[cloneId];
```

src/test/fixtures/composes.tsx
regexowl: Ah, got it. Looks much cleaner. Updated

lucasgarfield (approved):
Nice! Thank you!

PR #1252, opened 2023-07-19
msw: Version bump in `mockServiceWorker.js`

regexowl:
Recently msw was bumped from 1.2.1 to 1.2.2 (https://github.com/RedHatInsights/image-builder-frontend/pull/1245). The version gets updated in `mockServiceWorker.js` upon running `npm ci`.

This PR changes the version as it wasn't a part of the original bump PR, but appears after every clean install.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -2,7 +2,7 @@
 /* tslint:disable */
 
 /**
- * Mock Service Worker (1.2.1).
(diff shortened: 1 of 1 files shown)


lucasgarfield:
Nice catch. 

I ran `npx msw init src` to generate this file programmatically and verified that this is the correct diff.

lucasgarfield (approved):
Thank you!!!

PR #1255, opened 2023-07-20
test/fixtures: Migrate some fixtures to TypeScript

regexowl:
This migrates following fixtures to TypeScript:
- architectures
- customizations
- sources

Typing and explicit returns were added.

The extensions of `composes` and `packages` fixtures were also updated from .tsx to .ts


lucasgarfield:
Similar to the comment I left in #1247, let's use a `.ts` extension for "pure" Typescript files (files without React JSX).

regexowl:
The extensions are updated :+1:

src/Components/CreateImageWizard/formComponents/ActivationKeys.js
```diff
@@ -80,7 +80,7 @@ const ActivationKeys = ({ label, isRequired, ...props }) => {
           isDisabled={!isSuccessActivationKeys}
         >
           {isSuccessActivationKeys &&
-            activationKeys.body.map((key, index) => (
```
lucasgarfield, line 83: I don't think that we can remove the body property, it is part of the API response:
https://console.redhat.com/docs/api/rhsm/v2#operations-activationKey-listActivationKeys

I made a mistake when I was adding typing to this endpoint. I just learned that there is an official tool for generating RTKQ query hooks from OpenAPI specifications. I'm going to try that and see how it works, then revisit this PR:
https://redux-toolkit.js.org/rtk-query/usage/code-generation

src/Components/CreateImageWizard/formComponents/ActivationKeyInformation.js
```diff
@@ -60,19 +60,19 @@ const ActivationKeyInformation = () => {
               Role:
             </TextListItem>
             <TextListItem component={TextListItemVariants.dd}>
-              {activationKeyInfo.body?.role || 'Not defined'}
+              {activationKeyInfo.role || 'Not defined'}
```
lucasgarfield, line 63: This is actually still going to need to be `activationKeyInfo.body.role`. It seems a bit clunky, but that's how their API works... and right now, I want to move towards just being 1:1 with the APIs that we consume and not doing any sort of transformations or normalizations. Less decision making that way, which hopefully will mean less errors. :sweat_smile: 

I think it would actually be best to just convert ActivationKeyInfo.js into a .tsx file in this PR. That will help you ensure that all the properties that are accessed actually exist. :wink:

regexowl: Hmm, let's split the migration of the activation key to a different PR then.

lucasgarfield: We could just do it as a commit in this PR, too. I checked and the necessary changes are really pretty minor. It's up to you! :grin:

src/test/fixtures/sources.ts
```diff
@@ -22,35 +28,37 @@ export const mockSourcesByProvider = (provider) => {
...
+    '123': {
       provider: 'aws',
       aws: {
         account_id: '123456789012',
       },
-    };
-  }
+    },
+  };
+  return mockInfo[sourceId as keyof typeof mockInfo];
```
lucasgarfield, line 63: Similar to the previous PR, I think it would be more consistent to type the mockInfo object and remove `as keyof typeof mockInfo`.

src/test/fixtures/sources.ts
```diff
@@ -22,35 +28,37 @@ export const mockSourcesByProvider = (provider) => {
...
+  };
+  return mockSources[provider as keyof typeof mockSources];
 };
 
-export const mockUploadInfo = (sourceId) => {
-  if (sourceId === '666') {
-    return {
+export const mockUploadInfo = (
+  sourceId: string
+): GetSourceUploadInfoApiResponse => {
```
lucasgarfield, line 38: There is some debate on whether inferred or explicit return types are the best but right now my idea is to write as *little* Typescript as possible, and I'm fine with inferred returns. I think if we type the mockInfo object below we don't need to specify the return type. Feel free to disagree, the most important thing for me is that we agree on a consistent style.

src/test/fixtures/sources.ts
```diff
@@ -1,15 +1,21 @@
...
-    return [
+import {
+  GetSourceListApiResponse,
+  GetSourceUploadInfoApiResponse,
+} from '../../store/provisioningApi';
+
+export const mockSourcesByProvider = (
+  provider: string
+): GetSourceListApiResponse => {
+  const mockSources = {
```
lucasgarfield, line 9: Same comment as below, what if we type `mockSources`, get rid of the explicit return type and let Typescript infer it, and remove `as keyof typeof mockSources` in the return statement?

src/test/fixtures/architectures.ts
```diff
@@ -67,6 +109,7 @@ export const mockArchitecturesByDistro = (distro) => {
           },
         ],
       },
-    ];
-  }
+    ],
+  };
+  return mockDistros[distro as keyof typeof mockDistros];
```
lucasgarfield, line 114: Let's try typing mockDistros and removing `as keyof typeof mockDistros`.

src/test/fixtures/architectures.ts
```diff
@@ -1,8 +1,25 @@
-export const mockArchitecturesByDistro = (distro) => {
-  if (distro === 'rhel-92') {
-    return [
+import { Architectures, Distributions } from '../../store/imageBuilderApi';
+
+export const mockArchitecturesByDistro = (
+  distro: Distributions
+): Architectures => {
+  const mockDistros = {
```
lucasgarfield, line 6: Let's try typing mockDistros and removing the explicit `Architectures` return type from the function, letting TS infer it instead.

src/test/fixtures/sources.ts
regexowl: Got it. I think in this case I would also be okay with inferred return as there's very little going on with the data along the way. On the other hand I'm wondering whether explicit returns could help in cases where there's significant transformation of the data. It would make the returns more rigid, but could also safeguard against errors. What do you think?

Do you see some possible problems with explicit returns? 
I can't really see any with the inferred returns other than more wiggle room for us but also maybe some :bug:?

lucasgarfield: As discussed offline, let's go with explicit returns. :+1:

regexowl:
I've split the activationKey in the end, will create a separate PR for that one. Fixtures mentioned above are updated as per the comments and I've also updated extensions of `composes` and `packages` fixtures. Even though I believe I've changed those in the previous PRs those were still .tsx instead of .ts

lucasgarfield (approved):
Very nice!!!

PR #1258, opened 2023-07-21
ImagesTable: Fix bug where failed AWS images showed sharing error

regexowl:
Failed AWS images showed a `Failed to share image to one or more regions.` error even though they weren't shared to other regions.

This was caused by checking for a 'failure' status in an array of `imageStatuses` that included the parent status.

### DIFF
--- a/src/Components/ImagesTable/ImageBuildStatus.js
+++ b/src/Components/ImagesTable/ImageBuildStatus.js
@@ -42,9 +42,7 @@ export const ImageBuildStatus = ({
 
   const cloneErrorMessage = () => {
     let region = '';
-    hasFailedClone.includes(image.id)
-      ? (region = 'one or more regions')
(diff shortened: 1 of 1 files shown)


regexowl:
This solves the problem although it's built on a few assumptions - as far as I know the clones don't carry info about possible errors, is that right? Also is it possible for a clone to fail for a different reason than an unsuccessful sharing?

image
image

croissanne:
Do we show the error details somewhere too? Because those might contain useful info

regexowl:
I should have specified - the screenshots are from msw browser. "Failed to share..." would be the same, but the "Error in depsolve job" is just a mock value we use in our fixtures. The real failure should have more information.

croissanne (approved):
Thank you for the quick fix! Looks good

PR #1262, opened 2023-07-24
Wizard: Add description field to details step

mgold1234:
fixes https://github.com/RedHatInsights/image-builder-frontend/issues/1244
this commit adds the description fields to the Details step when a user creates a new image,
 and when the user recreates the image, this field will exist. 
 this pr depends on pr in the backend side - https://github.com/osbuild/image-builder/pull/792

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -103,6 +103,7 @@ const onSave = (values) => {
     const request = {
       distribution: values.release,
       image_name: values?.['image-name'],
+      image_description: values?.['image-description'],
       image_requests: [
         {
           architecture: 'x86_64',
@@ -138,6 +139,7 @@ const onSave = (values) => {
     const request = {
       distribution: values.release,
       image_name: values?.['image-name'],
+      image_description: values?.['image-description'],
       image_requests: [
         {
           architecture: 'x86_64',
@@ -167,6 +169,7 @@ const onSave = (values) => {
     const request = {
       distribution: values.release,
       image_name: values?.['image-name'],
+      image_description: values?.['image-description'],
       image_requests: [
         {
           architecture: 'x86_64',
@@ -189,6 +192,7 @@ const onSave = (values) => {
     const request = {
       distribution: values.release,
       image_name: values?.['image-name'],
+      image_description: values?.['image-description'],
       image_requests: [
         {
           architecture: 'x86_64',
@@ -208,6 +212,7 @@ const onSave = (values) => {
     const request = {
       distribution: values.release,
       image_name: values?.['image-name'],
+      image_description: values?.['image-description'],
       image_requests: [
         {
           architecture: 'x86_64',
@@ -227,6 +232,7 @@ const onSave = (values) => {
     const request = {
       distribution: values.release,
       image_name: values?.['image-name'],
+      image_description: values?.['image-description'],
       image_requests: [
         {
           architecture: 'x86_64',
@@ -246,6 +252,7 @@ const onSave = (values) => {
     const request = {
       distribution: values.release,
       image_name: values?.['image-name'],
+      image_description: values?.['image-description'],
       image_requests: [
         {
           architecture: 'x86_64',
@@ -290,6 +297,7 @@ const requestToState = (composeRequest, distroInfo, isBeta, isProd) => {
     const formState = {};
 
     formState['image-name'] = composeRequest.image_name;
+    formState['image-description'] = composeRequest.image_description;
 
     formState.release = composeRequest?.distribution;
     // set defaults for target environment first
@@ -472,7 +480,6 @@ const formStepHistory = (composeRequest, isBeta) => {
     } else {
       steps.push('File system configuration', 'packages');
     }
-
     steps.push('details');
 
     return steps;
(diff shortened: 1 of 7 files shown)


src/store/composesSlice.js
```diff
@@ -53,6 +53,7 @@ export const selectComposeById = (state, composeId) => {
       created_at: compose.created_at,
       id: compose.id,
       imageName: compose.image_name || compose.id,
+      ImageDescription: compose.image_description || compose.id,
```
regexowl, line 56: Let's unify the capitalization from ImageDescription -> imageDescription

src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
```diff
@@ -558,6 +558,15 @@ export const ImageDetailsList = () => {
             <TextListItem component={TextListItemVariants.dd}>
               {getState()?.values?.['image-name']}
             </TextListItem>
+            <TextListItem
```
regexowl, line 561: The description should be rendered conditionally so it doesn't show in the review when empty. Because of `{getState()?.values?.['image-name'] && (` it renders when the imageName is filled, so we'll just need to split the condition for name and description.

Screenshot from 2023-07-25 12-40-49

src/Components/CreateImageWizard/steps/imageName.js
```diff
@@ -43,5 +51,23 @@ export default {
         },
       ],
     },
+    {
+      component: componentTypes.TEXT_FIELD,
```
regexowl, line 55: Text area might be better than text input in this case as it provides more space for the text and wraps based on the width of the modal - https://www.data-driven-forms.org/provided-mappers/textarea?mapper=pf4

`component: componentTypes.TEXTAREA`

Also the character count does not seem to work for me :(

Screenshot from 2023-07-25 12-52-33

regexowl (commented):
Could you also add the description to our unit tests to check if it renders properly? You can look at how we test rendering of the image name as this will be very similar. We are testing the rendering of image name in `CreateImageWizard.beta.test.js` and `CreateImageWizard.test.js`

src/Components/CreateImageWizard/steps/imageName.js
mgold1234: changed to TEXTAREA

src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
mgold1234: done, thank you

src/store/composesSlice.js
mgold1234: good catch, thank you

src/store/composesSlice.js
```diff
@@ -53,6 +53,7 @@ export const selectComposeById = (state, composeId) => {
       created_at: compose.created_at,
       id: compose.id,
       imageName: compose.image_name || compose.id,
+      imageDescription: compose.image_description || compose.id,
```
lucasgarfield, line 56: I don't think we want to use the compose id in case there is a no description, maybe it would be better to use an empty string or `none`.

src/Components/CreateImageWizard/steps/imageName.js
```diff
@@ -1,12 +1,20 @@
...
 import validatorTypes from '@data-driven-forms/react-form-renderer/validator-types';
+import { Flex, FlexItem, Text } from '@patternfly/react-core';
 
 import StepTemplate from './stepTemplate';
 
 import CustomButtons from '../formComponents/CustomButtons';
 
+const CharacterCount = () => {
+  const { getState } = useFormApi();
+  const description = getState().values?.description;
```
lucasgarfield, line 14: Small bug here, this should be `image-description`, not `description`.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -474,6 +482,7 @@ const formStepHistory = (composeRequest, isBeta) => {
     }
 
     steps.push('image-name');
+    steps.push('image_description');
```
lucasgarfield, line 485: Should this be `image_description` or `image-description`?

mgold1234: chnaged

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1121,6 +1121,21 @@ describe('Step Details', () => {
     await user.type(nameInput, 'valid-name');
     expect(getNextButton()).not.toHaveClass('pf-m-disabled');
     expect(getNextButton()).toBeEnabled();
+
+    // Enter description image
+    const descriptionInput = screen.getByRole('textbox', {
+      name: /description 0\/250/i,
+    });
+    // 101 character name
```
lucasgarfield, line 1129: I think this comment is from the image name, not description... Maybe just delete it, it is obvious what is happening in the line below...

lucasgarfield:
#1267 is going to remove the `/types` directory as the types (at least for the API endpoints) will be found in `/src/store` instead. 

Since none of the files are Typescript files or using the added type, can you just remove `imageBuilderSchema.ts` from this PR?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
mgold1234: removed. thank you

lucasgarfield (approved):
Amazing work, Michal! And so awesome you've already got the required changes merged into the back-end!

PR #1275, opened 2023-07-30
Fix missing aria label in launch wizard  modal console error

amirfefer:
When no aria label is given to a modal component, an error is thrown in console:
```
Modal: Specify at least one of: title, aria-label, aria-labelledby. 
[Error] Modal: When using hasNoBodyWrapper or setting a custom header, ensure you assign an accessible name to the the modal container with aria-label or aria-labelledby.
```


lucasgarfield (approved):
Nice catch, thank you Amir!

PR #1279, opened 2023-07-31
src/App.js: Remove obsolotete APP_NAVIGATION listener

croissanne:
v6 of react-router-dom is now supported everywhere, and leaving the event listener can cause navigation bugs.

See gkarat/ocp-advisor-frontend#101 as an example.


regexowl (approved):
LGTM

PR #1280, opened 2023-07-31
wizard/registration: cleaning of a useEffect

lavocatt:
A useEffect was used without conditions to update the value of a state variable depending on the form state. That's apparent to a way to initialize the state. Instead, what can be done, is to immediately initialize the state at the proper value.


regexowl (approved):
This seems to work really well. Thanks for pointing out that [we might not need the useEffect](https://react.dev/learn/you-might-not-need-an-effect) here! :sparkles:

croissanne (approved):

PR #1282, opened 2023-08-01
wizard/activation key: remove a useEffect

lavocatt:
Since the propagation of the subscription serer url and base url was
performed with a useEffect on every redraw, it's equivalent as calling
directly change to initialize the form directly.


lavocatt:
well this doesn't work for now :sweat_smile:

lucasgarfield:
Would you mind converting this to a `.tsx` in this PR? It might have some implications on your useEffect refactor, but will give us more confidence in the refactor and now is a good time since you are familiar with the code.

PR #1283, opened 2023-08-01
wizard/size_unit: remove a useEffect

lavocatt:
A useEffect was used to propagate the new form values to the parent. However it's possible to use the `onChange` and `onSelect` properties of the form inputs. Thus removing the need of a useEffect to propagate the update.


lucasgarfield:
Would you mind adding a commit to convert this to a `.tsx` extension in this commit? It should be pretty easy and if we just convert our .js files to .tsx files every time we are working on them we will really speed up our Typescript migration.

PR #1284, opened 2023-08-01
eslint: Add support for Typescript

lucasgarfield:
This commit adds eslint support for .ts and .tsx files.

The recommended Typescript rules are applied only to .ts and .tsx files and not to existing .js or .jsx files. This is accomplished by creating a separate .eslintrc-typescript.yml file and pointing to it in the .eslintrc.yml overrides parameter.

A .eslintignore file was added. This file has syntax similiar to .gitignore and is used to ignore the programatically generated API slices so that we do not have to deal with a massive diff whenever we update one of them.


regexowl (approved):
LGTM!

PR #1288, opened 2023-08-02
Wizard: Disable non-valid repositories and add a Status column

regexowl:
This disables adding of repositories that do not have 'Valid' status.

Status column was also added to the Repositories table to make the reason why the checkbox is disabled visible.


src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+import {
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+  ExclamationTriangleIcon,
+  InProgressIcon,
+} from '@patternfly/react-icons';
+import { Alert, Button, Popover } from '@patternfly/react-core';
+
+type RepositoryStatusProps = {
```
lucasgarfield, line 11: I'm still trying to work out best practices for Typescript types, but I think that here it might be good future proofing to use the types provided by the API.

`ApiRepositoryResponse[status]` and `ApiRepositoryResponse[url]`...

Those are both just strings in the current API spec... the url will always be a string but I could imagine the status changing to something like `'valid' | 'failure'`, etc...

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
```diff
@@ -0,0 +1,82 @@
...
+        >
+          <Button variant="link" className="pf-u-p-0 pf-u-font-size-sm">
+            <div className="failure-button">
+              <ExclamationTriangleIcon className="expiring" /> {repoStatus}
+            </div>
+          </Button>
+        </Popover>
+      </>
+    );
+  } else {
```
lucasgarfield, line 73: This is Javascript but I'm taking inspiration from PEP 20 - The Zen of Python, "Explicit is better than implicit". 

I think it would be more explicit to use an `else if` here where repoStatus === 'Pending', or whatever the exact value is.

The exact values don't seem to show up in our generated API types, which seems to be a bit of an oversight in their API spec... unless I'm missing something?

lucasgarfield:
Seems to be a minor issue with the dotted underline, it is underlining the space in-between the icon and the text. I know we ran into this in the past with the statuses on the Images Table and figured out a good solution there.

image

lucasgarfield:
Wow, you're fast!

Would it be a good idea to update our MSW fixture as well so we show all statuses, preferably on the first page of results in the Wizard? 

So the first one could be 'Valid', the second 'Invalid', the third 'Unavailable', the fourth 'Pending'. We can rename the filler repos to something like z-filler-repos so that they appear *after* all the other results, and then just change the status on some of the repos in the repos object.

image

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
regexowl: Agree, but I'm also unable to find the exact value. Am I looking in the right place? The status seems to be just a string in their API spec.

regexowl: Ok se the last value would be probably "Pending"

image

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
regexowl: Completely on board, just still forgetting :no_mouth:

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
regexowl: Updated

regexowl:
It would absolutely be a good idea! I've added the mock repos to the fixture.

lucasgarfield:
It's so beautiful...

One last nit-pick... At least on my machine, at max modal width, the Packages column isn't wide enough to display the word "Packages". Can we take a bit of width away from the repo name column to make room for it to display the full word "Packages"?

image

regexowl:
Sure can! Should be better now

image

lucasgarfield (approved):
Great job!

PR #1291, opened 2023-08-03
LandingPage: fix the button in a button error

lavocatt:
Previously the Popover was used within the title of the tab. In addition the fact that the Popover had a button to get triggered, it created a button in a button issue. The web browser was reporting this as abnormal and the UI behavior was not ideal: while clicking on the interrogation mark the user was also moved to the corresponding tab.

This commit fixes that issue by using a TabAction to get the Popover working. Inspired by the patternfly documentation:
- http://v4-archive.patternfly.org/v4/components/tabs#help-action

### DIFF
--- a/src/Components/LandingPage/LandingPage.js
+++ b/src/Components/LandingPage/LandingPage.js
@@ -12,6 +12,7 @@ import {
   TabTitleText,
   Text,
   TextContent,
+  TabAction,
 } from '@patternfly/react-core';
 import {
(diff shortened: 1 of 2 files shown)


lucasgarfield (dismissed):
This is great! I love that not only is the warning gone, clicking the (?) doesn't cause the tab to change anymore.

regexowl (approved):
This looks great, thank you!

PR #1293, opened 2023-08-07
test/fixtures: Update timestamps of mock composes

regexowl:
The format of the timestamps of mock composes was causing an error in msw browser after merging https://github.com/RedHatInsights/image-builder-frontend/pull/1137

This updates the mock timestamps to the format of the actual ones fetched from the API.

### DIFF
--- a/src/Utilities/time.js
+++ b/src/Utilities/time.js
@@ -1,5 +1,5 @@
 export const timestampToDisplayString = (ts) => {
-  // timestamp has format 2021-04-27 12:31:12.794809 +0000 UTC
+  // timestamp has format 2021-04-27T12:31:12Z
   // must be converted to ms timestamp and then reformatted to Apr 27, 2021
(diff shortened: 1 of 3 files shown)


lucasgarfield:
Would it make sense to update `timeUtilities` to Typescript? Would that have helped catch this bug? Could be in a follow-up PR, of course.

regexowl:
Actually thinking about it more - what would be the scenario when we would need to check for the right format of the date? I mean the dates we are fetching from API will be always what we need and the only problem are our fixtures. Maybe we could put together validator just for those? But at the same time, how often do we change them?

Sorry, not sure if that makes sense. Just thinking about real usage of the validator.

Also weird that the tests weren't screaming even though the mocked data crashed the page in browser.

lucasgarfield (approved):
Hmm, good questions all around. For now I agree that we should just get the fix in.

PR #1294, opened 2023-08-07
Wizard: Fix visibility of the Image Details expandable

regexowl:
The Image Details expandable is now visible even if no name or description were added for the image. This fixes the issue.


lucasgarfield (approved):
Nice catch!

PR #1295, opened 2023-08-07
ImagesTable: Fix bug where vsphere-ova details are not displayed

lucasgarfield:
Fixes HMS-2266.

The conditional in the image details component needed to have a check added for vsphere-ova type images.

A mock compose and corresponding status for a vsphere-ova type image have also been added.

### DIFF
--- a/src/test/fixtures/composes.tsx
+++ b/src/test/fixtures/composes.tsx
@@ -24,7 +24,7 @@ const currentDateInString = currentDate.toISOString();
 
 export const mockComposes: ComposesResponse = {
   meta: {
-    count: 12,
+    count: 13,
   },
   links: {
     first: '',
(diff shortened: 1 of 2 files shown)


src/test/fixtures/composes.tsx
```diff
@@ -472,6 +489,31 @@ export const mockStatus = (composeId: string): ComposeStatus => {
         ],
       },
     },
+    'hyk93673-8dcc-4a61-ac30-e9f4940d8346': {
+      image_status: {
+        status: 'success',
+        upload_status: {
+          options: {
+            url: 'https://s3.amazonaws.com/b7193673-8dcc-4a5f-ac30-e9f4940d8346-disk.vmdk',
```
regexowl, line 497: This might be mostly irrelevant, but can you update the url to reflect the uuid of the image? Could prevent some confusion later even though I can't really recall if I have ever needed to check an url of a mock compose.

lucasgarfield: Nice catch, I think it could definitely cause problems in the future.

regexowl:
And I think the last nit pick - `meta.count` should be also updated for `mockComposes` from 12 to 13

regexowl (approved):
Works great, thanks!

PR #1297, opened 2023-08-08
fixing the tests from their errors and warnings

lavocatt:
This pull request aims at addressing the current issue regarding the errors thrown while executing the tests.
For now it fixes issues related with:
* scalprum
* act
* null access
* (and also one no-op)

For the errors that were not fixed in this PR, they are converted into warnings, and we will have to take care of them in the future. The conversion is made with a new lib https://github.com/ValentinH/jest-fail-on-console that @regexowl found out.
The lib also makes the tests actually fail if other errors are presents, so hopefully we can't add new ones.

### DIFF
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.beta.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.beta.test.js
@@ -106,7 +106,9 @@ describe('Step Upload to AWS', () => {
 
     // select aws as upload destination
     const awsTile = await screen.findByTestId('upload-aws');
-    awsTile.click();
+    await act(async () => {
+      awsTile.click();
+    });
 
     await clickNext();
 
@@ -176,14 +178,20 @@ describe('Step Upload to AWS', () => {
     });
     // Wait for isSuccess === true, dropdown is disabled while isSuccess === false
     await waitFor(() => expect(sourceDropdown).toBeEnabled());
-    sourceDropdown.click();
+    await act(async () => {
+      sourceDropdown.click();
+    });
 
     const source = await screen.findByRole('option', {
       name: /my_source/i,
     });
-    source.click();
+    await act(async () => {
+      source.click();
+    });
 
-    await clickNext();
+    await act(async () => {
+      await clickNext();
+    });
 
     // registration
     await screen.findByRole('textbox', {
@@ -191,14 +199,18 @@ describe('Step Upload to AWS', () => {
     });
 
     const registerLaterRadio = screen.getByLabelText('Register later');
-    await user.click(registerLaterRadio);
+    await act(async () => {
+      await user.click(registerLaterRadio);
+    });
 
     // click through to review step
-    await clickNext();
-    await clickNext();
-    await clickNext();
-    await clickNext();
-    await clickNext();
+    await act(async () => {
+      await clickNext();
+      await clickNext();
+      await clickNext();
+      await clickNext();
+      await clickNext();
+    });
 
     const composeImage = jest
       .spyOn(api, 'composeImage')
@@ -227,7 +239,9 @@ describe('Step Upload to AWS', () => {
       });
 
     const create = screen.getByRole('button', { name: /Create/ });
-    create.click();
+    await act(async () => {
+      create.click();
+    });
 
     // API request sent to backend
     expect(composeImage).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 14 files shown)


src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -481,17 +512,33 @@ describe('Step Registration', () => {
...
+    await act(async () => {
+      await user.click(activationKey);
+    });
     screen.getByDisplayValue('name0');
 
-    screen.getByRole('button', { name: /Next/ }).click();
-    screen.getByRole('button', { name: /Next/ }).click();
-    screen.getByRole('button', { name: /Next/ }).click();
-    screen.getByRole('button', { name: /Next/ }).click();
+    const n1 = screen.getByRole('button', { name: /Next/ });
```
regexowl, line 526: tl;dr nothing needs to be fixed here from my point of view, just wanted to share my thoughts

For a moment I was thinking if there should be some naming scheme that could be used for these constants so the names are consistent throughout the tests. But to be completely honest, I don't think that's a massive problem right now. 

I agree that the `act`s are mostly a quick patch to get us to the point where we have clean test outputs and then we can go from there. The names and `act`s themselves will be a part of the refactoring so for now I'm perfectly god with those.

src/test/Components/CreateImageWizard/CreateImageWizard.beta.test.js
```diff
@@ -919,5 +1021,5 @@ describe('Click through all steps', () => {
     );
     expect(store.getState().composes.allIds).toEqual(ids);
     // set test timeout of 10 seconds
```
regexowl, line 1023: Let's update the comment to 20 seconds as well

regexowl (commented):
This already looks great, thanks a lot for taking this on Thomas!

I've tried to run the tests as few times. The output looks much cleaner, but I can still see a few console errors and jest does not bail on those. Checked once more how is `bail` supposed to work and it seems it would end the run of the tests only for failing suites. Which I think is still a neat thing that we should keep.

When talking about ending the run on encountering a console error in the logs, maybe this could help? https://github.com/ValentinH/jest-fail-on-console

lavocatt:
> This already looks great, thanks a lot for taking this on Thomas!
> 
> I've tried to run the tests as few times. The output looks much cleaner, but I can still see a few console errors and jest does not bail on those. Checked once more how is `bail` supposed to work and it seems it would end the run of the tests only for failing suites. Which I think is still a neat thing that we should keep.
> 
> When talking about ending the run on encountering a console error in the logs, maybe this could help? https://github.com/ValentinH/jest-fail-on-console

are these errors or warnings ? can you paste them here please ?

regexowl:
> are these errors or warnings ? can you paste them here please ?

Errors starting with word "Warning" :sweat_smile: 
```
  console.error
    Warning: Cannot update a component (`TextField`) while rendering a different component (`TextField`). To locate the bad setState() call inside `TextField`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
```
```
  console.error
    Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```
```
  console.error
    Warning: Cannot update a component (`ReactFinalForm`) while rendering a different component (`TargetEnvironment`). To locate the bad setState() call inside `TargetEnvironment`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
```

Nine errors in total, all should be multiples of the ones mentioned above. You're not getting these? Maybe I'm doing something wrong. Just fetched changes and ran clean install to be sure. Did I miss anything?

lavocatt:
https://github.com/ValentinH/jest-fail-on-console works well, but the current warnings being outputed as `console.errors` it makes the tests failing :thinking:

lavocatt:
> > are these errors or warnings ? can you paste them here please ?
> 
> Errors starting with word "Warning" sweat_smile
> 
> ```
>   console.error
>     Warning: Cannot update a component (`TextField`) while rendering a different component (`TextField`). To locate the bad setState() call inside `TextField`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
> ```
> 
> ```
>   console.error
>     Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
> ```
> 
> ```
>   console.error
>     Warning: Cannot update a component (`ReactFinalForm`) while rendering a different component (`TargetEnvironment`). To locate the bad setState() call inside `TargetEnvironment`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
> ```
> 
> Nine errors in total, all should be multiples of the ones mentioned above. You're not getting these? Maybe I'm doing something wrong. Just fetched changes and ran clean install to be sure. Did I miss anything?

Yeah I'm getting the same  warningerrors :sweat_smile:

regexowl (approved):
Thanks a lot Thomas! This is amazing, the logs are so clean!! :broom: :sparkles:

PR #1303, opened 2023-08-09
eslint: enforce react rules and fix errors

lavocatt:
This PR enforces react linting rules for eslint and fixes the errors.

React doc: https://react.dev/learn/editor-setup#linting

(Warnings are staying untouched because I think that's next level to fix them all :sweat_smile:)

### DIFF
--- a/src/Components/ImagesTable/ImageDetails.js
+++ b/src/Components/ImagesTable/ImageDetails.js
@@ -20,7 +20,7 @@ import ClonesTable from './ClonesTable';
 import { extractProvisioningList } from '../../store/helpers';
 import { useGetSourceListQuery } from '../../store/provisioningApi';
 
-const sourceNotFoundPopover = () => {
+const SourceNotFoundPopover = () => {
   return (
     <Popover
       position="bottom"
@@ -59,7 +59,7 @@ const sourceNotFoundPopover = () => {
   );
 };
 
-const getAzureSourceName = (id) => {
+const AzureSourceName = ({ id }) => {
   const { data: rawSources, isSuccess } = useGetSourceListQuery({
     provider: 'azure',
   });
@@ -68,16 +68,16 @@ const getAzureSourceName = (id) => {
   if (isSuccess) {
     const sourcename = sources.find((source) => source.id === id);
     if (sourcename) {
-      return sourcename.name;
+      return <p>{sourcename.name}</p>;
     } else {
-      return sourceNotFoundPopover();
+      return <SourceNotFoundPopover />;
     }
   } else {
     return <Spinner isSVG size="md" />;
   }
 };
 
-const getAWSSourceName = (id) => {
+const AwsSourceName = ({ id }) => {
   const { data: rawSources, isSuccess } = useGetSourceListQuery({
     provider: 'aws',
   });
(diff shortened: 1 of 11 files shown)


src/Components/ImagesTable/ImageDetails.js
```diff
@@ -17,7 +17,7 @@ import { useSelector } from 'react-redux';
 
 import ClonesTable from './ClonesTable';
 
-import { useGetSourceListQuery } from '../../store/provisioningApi';
+import { useGetSourceListQuery as UseGetSourceListQuery } from '../../store/provisioningApi';
```
regexowl, line 20: On a first quick run through the changes - this error is weird. It's a hook, maybe not a react one, but an RTK Query one. Still there are several similar hooks in the code and it seem the linter doesn't have any problem with those. 

Looks like `ReviewStepTextLists.js` uses the very same hook without any errors or warnings. :thinking:

lavocatt: Ah! Thanks! I'll double check what's happening here

lavocatt: Good catch, thanks for double checking. I completely misunderstood the issue yesterday evening. That commit is now patched up!

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1106,7 +1106,7 @@ describe('Step Packages', () => {
     searchboxChosen.click();
     await user.type(searchboxChosen, 'asdf');
 
-    expect(screen.getAllByText('No packages found').length === 2);
+    expect(screen.getByText('No packages found')).toBeInTheDocument();
```
lucasgarfield, line 1109: I'm not sure this is testing the same thing... "No packages found" should appear twice, that's why we check whether length === 2 on getAllByText.

src/test/Components/CreateImageWizard/CreateImageWizard.beta.test.js
```diff
@@ -454,7 +454,7 @@ describe('Step Packages', () => {
     searchboxChosen.click();
     await user.type(searchboxChosen, 'asdf');
 
-    expect(screen.getAllByText('No packages found').length === 2);
+    expect(screen.getByText('No packages found')).toBeInTheDocument();
```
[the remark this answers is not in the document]
lucasgarfield: Hm, let's try emulating the test by hand and verifying what should be there...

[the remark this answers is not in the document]
lucasgarfield: It doesn't even display "No packages found"... what is going on?
image

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
lucasgarfield: Disregard above, it should only appear once.

lucasgarfield:
It's looking really good, one last thing... on the last commit, the warnings are:
```
/home/lgarfield/dev/image-builder-frontend/src/Components/ImagesTable/ImageDetails.js
  63:43  error  React Hook "useGetSourceListQuery" is called in function "getAzureSourceName" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"  react-hooks/rules-of-hooks
  81:43  error  React Hook "useGetSourceListQuery" is called in function "getAWSSourceName" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"    react-hooks/rules-of-hooks
  ```

Can we refactor this again, and instead just convert `getAzureSourceName` to be a React function component named `AzureSourceName`? Same for `getAWSSourceName`.

Also, please call the new component `getAwsSourceName`... note that camelCase on the acronym.

lucasgarfield (approved):
Amazing!!!

PR #1310, opened 2023-08-13
Wizard: update popup with repositories name on Review step

mgold1234:
fixes https://github.com/RedHatInsights/image-builder-frontend/issues/1305 
This commit update popup of custom repositories with name of repository instead of url of repository
Screenshot 2023-08-13 at 12 29 21

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
+++ b/src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { useFormApi } from '@data-driven-forms/react-form-renderer';
-import { Panel, PanelMain } from '@patternfly/react-core';
+import { Alert, Panel, PanelMain, Spinner } from '@patternfly/react-core';
 import {
   TableComposable,
   Tbody,
@@ -10,8 +10,31 @@ import {
   Thead,
   Tr,
 } from '@patternfly/react-table';
+import PropTypes from 'prop-types';
 
 import { UNIT_GIB, UNIT_MIB } from '../../../constants';
+import { useListRepositoriesQuery } from '../../../store/contentSourcesApi';
+
+const RepoName = ({ repoUrl }) => {
+  const { data, isSuccess, isFetching, isError } = useListRepositoriesQuery({
+    url: repoUrl,
+  });
+
+  return (
+    <>
+      {isSuccess && <p>{data.data?.[0].name}</p>}
+      {isFetching && <Spinner isSVG size="md" />}
+      {isError && (
+        <Alert
+          variant="danger"
+          isInline
+          isPlain
+          title="Error loading repository name"
+        />
+      )}
+    </>
+  );
+};
 
 export const FSReviewTable = () => {
   const { getState } = useFormApi();
@@ -95,7 +118,9 @@ export const RepositoriesTable = () => {
           <Tbody data-testid="repositories-tbody-review">
             {repositories.map((repo, repoIndex) => (
               <Tr key={repoIndex}>
-                <Td className="pf-m-width-60">{repo.baseurl}</Td>
+                <Td className="pf-m-width-60">
+                  <RepoName repoUrl={repo.baseurl} />
+                </Td>
               </Tr>
             ))}
           </Tbody>
@@ -103,4 +128,8 @@ export const RepositoriesTable = () => {
       </PanelMain>
     </Panel>
   );
+};
+
+RepoName.propTypes = {
+  repoUrl: PropTypes.string,
 };


src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
```diff
@@ -13,9 +13,14 @@ import { ApiRepositoryResponse } from '../../../store/contentSourcesApi';
...
   repoStatus: ApiRepositoryResponse['status'];
   repoUrl: ApiRepositoryResponse['url'];
+  Name: ApiRepositoryResponse['name'];
 };
 
-const RepositoriesStatus = ({ repoStatus, repoUrl }: RepositoryStatusProps) => {
+const RepositoriesStatus = ({
+  repoStatus,
+  repoUrl,
+  Name,
```
kingsleyzissou, line 22: For consistency, I would change this to `repoName`

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
```diff
@@ -31,7 +36,7 @@ const RepositoriesStatus = ({ repoStatus, repoUrl }: RepositoryStatusProps) => {
           bodyContent={
             <>
               <Alert variant="danger" title="Invalid" isInline isPlain />
-              Cannot fetch {repoUrl}
+              Cannot fetch {Name}
```
kingsleyzissou, line 39: This would also change to `repoName`

mgold1234: changed, thank you

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
mgold1234: I removed the name from this file, solve it in a different way

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
mgold1234: I removed the name from this file, solve it in a different way

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -93,14 +113,22 @@ export const RepositoriesTable = () => {
             </Tr>
           </Thead>
           <Tbody data-testid="repositories-tbody-review">
-            {repositories.map((repo, repoIndex) => (
-              <Tr key={repoIndex}>
-                <Td className="pf-m-width-60">{repo.baseurl}</Td>
-              </Tr>
-            ))}
+            {repositories.map((repo, repoIndex) => {
```
regexowl, line 116: This could be cleaned up a bit. You don't a nested return nor the jsx brackets here:
```javascript
            {repositories.map((repo, repoIndex) => (
              <Tr key={repoIndex}>
                <Td className="pf-m-width-60">
                  <RepoName repoUrl={repo.baseurl} />
                </Td>
              </Tr>
            ))}
```

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -10,8 +10,28 @@ import {
...
 
 import { UNIT_GIB, UNIT_MIB } from '../../../constants';
+import { useListRepositoriesQuery } from '../../../store/contentSourcesApi';
+
+const RepoName = ({ repoUrl }) => {
+  const { data, isSuccess, isLoading } = useListRepositoriesQuery({
+    url: repoUrl,
+  });
+  if (!isSuccess || isLoading) {
+    return null;
```
regexowl, line 23: I'm not really in favor of returning null. React sees null as a valid value and it will try to render it which can cause unexpected behaviour along the way. 

What we could probably do here is remove the `if` conditional, check for success in the second return and based on that return data or spinner?

Maybe something like this:
```javascript
  return (
    <>
      {isSuccess ? <p>{data.data?.[0].name}</p> : <Spinner isSVG size="md" />}
    </>
  );
```

What do you think?

regexowl (commented):
Added two comments. Also I'm wondering - would it be possible to migrate the file to Typescript as well? Not sure how big of a change would be needed though.

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
lucasgarfield: The only issue I have with this is that if there is ever an error, we will show an infinite spinner. Of course, that is very unlikely, but generally I would prefer to display the spinner when `isFetching` is true. (RTKQ hooks provide both `isLoading` and `isFetching`, `isLoading` is true when the FIRST request is in flight, `isFetching` is true when ANY request is in flight).

What about something like this? Both specifically for this PR, and as a general convention for us to use moving forward?

```Typescript
return (
    <>
      {isSuccess ? <p>{data.data?.[0].name}</p>} 
      {isFetching ? <Spinner isSVG size="md" />}
      {isError ? <Alert variant="danger" isInline isPlain title="Error loading repository name" />}
    </>
  );
```

mgold1234: its great idea, thank you @regexowl and @lucasgarfield

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
mgold1234: done, thank you @regexowl

regexowl (approved):
LGTM!

PR #1311, opened 2023-08-13
feat(HMS-1909): enable a custom host in LOCAL_API env var

amirfefer:
This PR adds the ability to customize the host (default to localhost to prevent any regression)  in the `LOCAL_API` env var.
This change  will allow us to run containerized  image-builder-frontend and provisioning (in federated mode) apps locally, as we do  in https://github.com/RHEnVision/provisioning-compose/pull/9


lucasgarfield (approved):
Looks good to me!

PR #1315, opened 2023-08-16
HMS-2322: Wizard: Update registration step activation key link URL

lucasgarfield:
Fixes HMS-2322: On the Register step of the image build wizard, the link to "Create and manage activation keys here" should point to https://console.redhat.com/insights/connector/activation-keys.


ondrejbudai (approved):
Splendid!

PR #1317, opened 2023-08-17
router: lazy loading requires a suspense tag

lavocatt:
According to the React documentation[0], a suspense tag is needed to
encapsulate a lazy loaded component. Some bad behavior have been fix in
the UI thanks to that tag (i.e notification popup unclosable after
wizard creation).

[0] https://react.dev/reference/react/lazy#suspense-for-code-splitting

The notification bug began in 1b01cfba.

### DIFF
--- a/src/Router.js
+++ b/src/Router.js
@@ -1,4 +1,4 @@
-import React, { lazy } from 'react';
+import React, { lazy, Suspense } from 'react';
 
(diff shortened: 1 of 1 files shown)


lucasgarfield (approved):
This is amazing, this bug has been driving me crazy. So glad you figured out a fix. :heart:

PR #1320, opened 2023-08-21
ImagesTable: Convert ImagesTable to Typescript & RTK Query

lucasgarfield:
This is an enormous pull request, my apologies. Converting all API calls to RTK Query was much easier to do with the safety of Typescript, so these two changes were combined.

Main commit message:
This commit converts the Images Table to Typescript and converts all API
calls to image-builder to use RTK Query hooks.

This should increase the performance of the app significantly.
Previously our calls to the image-builder API were made in series. They
are now made in parallel. We may want to investigate the possibility of
hitting rate limiting now that we will be issuing requests in much more
rapid succession.

In the tests, moving to RTK Query hooks has allowed us to remove
virtually all Jest mocking. However, this means that some of our
previous tests which tested against implementation details were broken.
Most notably, we no longer check the Redux store to verify that clones
have been added correctly and we no longer check that compose requests
were issued successfully. Test coverage will be restored in a follow-up
PR where the dev-dependency @msw/data is added. Adding a persistent data
layer to the tests using @msw/data will allow us to verify that our POST
requests (creating composes and cloning them) are working by testing
that the Images Table has been updated.


src/Components/ImagesTable/Instance.tsx
```diff
@@ -0,0 +1,209 @@
...
+    );
+  } else {
+    return (
+      <Button isDisabled variant="link" isInline>
+        Download ({fileExtensions[compose.request.image_requests[0].image_type]}
+        )
+      </Button>
+    );
+  }
+};
```
lavocatt, line 209: Nitpick:

I would prefer

```
if(status === 'success'){
  if(isExpired){
    return ...
  }
  return ...
}
return ...
```

lavocatt (changes requested):
That's an amazing work you've put together! I've got a bunch of comments sorry :sweat_smile: Glad to discuss every one of them with you ofc!

On a more general note, I would really love to see more docstring, Typescript is great, having types helps to know what we're dealing with in the code but I think we should go the extra mile and provide good dosctring for the helper functions and the components.

With:
* What the thing is doing
* What is the returned value if any
* What are the parameters
* What are the known limitations

Having these information helps for any new comers to the code base, but also to spot any mismatch between what you want to achieve and what the code actually does. The second is what matters the most for a review.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -0,0 +1,570 @@
...
+  status: ComposeStatus | undefined,
+  navigate: NavigateFunction
+) => [
+  {
+    title: 'Share to new region',
+    onClick: () => navigate(resolveRelPath(`share/${compose.id}`)),
+    isDisabled: status?.image_status.status === 'success' ? false : true,
+  },
+  ...defaultActions(compose, navigate),
+];
```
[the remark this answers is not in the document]
lavocatt: As far as I understand this, defaultActions is an array of stuff and awsActions is an array of stuff containing the stuff of defaultActions :sweat_smile:  and somehow it's making the recreate image possible

src/Components/ImagesTable/ClonesTable.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <Row
+          ami={<Ami status={status} />}
+          region={<CloneRegion region={clone.request.region} />}
+          status={<StatusClone clone={clone} status={status} />}
+        />
+      );
+    default:
+      return <LoadingRow clone={clone} />;
+  }
+};
```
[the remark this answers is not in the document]
lucasgarfield: This is a standard pattern to replace if/else chains in Javascript, I'm using it here because I find it improves readability. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#an_alternative_to_if...else_chains

In this case, if isSuccess === true, we render the `<Row>`, otherwise we render `<LoadingRow>`.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -566,27 +565,19 @@ const CreateImageWizard = () => {
     <ImageCreator
       onClose={handleClose}
       onSubmit={({ values, setIsSaving }) => {
-        setIsSaving(() => true);
+        setIsSaving(true);
         const requests = onSave(values);
+        navigate(resolveRelPath(''));
```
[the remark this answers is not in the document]
lucasgarfield: There is some strange interaction with Data Driven Forms and the wizard's state that I wasn't able to debug.

After clicking "Create image", the wizard would flash back the first step before closing.

Changing the order prevents this behavior from happening. As we will be removing our dependency on DDF soon, I don't feel like troubleshooting the issue is worth the effort.

config/dev.webpack.config.js
```diff
@@ -39,6 +39,8 @@ const webpackProxy = {
 };
 
 const { config: webpackConfig, plugins } = config({
+  localChrome: '/home/lgarfield/dev/insights-chrome/build/',
```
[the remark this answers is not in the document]
lucasgarfield: Yes, that's right... :+1:

src/Components/ImagesTable/ImageBuildErrorDetails.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+// TODO what is this?
```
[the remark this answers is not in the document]
lucasgarfield: This file should actually just be deleted, the functionality has been replaced in the  new <Status> components.

src/Components/ImagesTable/ClonesTable.tsx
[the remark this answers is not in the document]
lavocatt: Alright, I didn't know it was a standard pattern, I'm not used to it so it makes me not liking it, but that's just syntactic sugar

src/Components/ImagesTable/ImageDetails.tsx
```diff
@@ -0,0 +1,383 @@
...
+  compose: ComposesResponseItem;
+  clones: ClonesResponseItem[];
+};
+
+export const AwsDetails = ({ compose, clones }: AwsDetailsPropTypes) => {
+  const options = compose.request.image_requests[0].upload_request.options;
+
+  if (!isAwsUploadRequestOptions(options)) {
+    throw TypeError(
+      `Error: options must be of type AwsUploadRequestOptions, not ${typeof options}.`
```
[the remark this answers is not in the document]
lucasgarfield: Basically, options is a union of several different types... object | GcpUploadRequestOptions | AwsUploadRequestOptions | AzureUploadRequestOptions.

Later, we need to access properties specific to an AwsUploadRequestOptions. In order to satisfy Typescript, we verify that we have the correct type (an AwsUploadRequestOptions) with the type guard. We should never be passing any other type into AwsDetails, so we throw an error if the type guard fails.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -0,0 +1,570 @@
...
+  if (!isSuccess) {
+    return <></>;
+  }
+
+  const composes = data.data;
+  const itemCount = data.meta.count;
+
+  return (
+    <>
+      {!isSuccess && <></>}
```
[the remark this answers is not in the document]
lucasgarfield: Done!

src/Components/ImagesTable/ImagesTable.tsx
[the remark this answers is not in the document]
lucasgarfield: Check the Patternfly docs:
Thttps://www.patternfly.org/components/table/#actionscolumn

src/Components/ImagesTable/Instance.tsx
lucasgarfield: Good point, I fixed it but in a slightly different way to avoid nesting:

```
  if (status !== 'success') {
    ...
  } else if (!isExpired) {
    ...
  } else if (isExpired) {
    ...
  }

```

src/test/Components/ShareImageModal/ShareImageModal.test.js
```diff
@@ -101,107 +99,24 @@ describe('Create Share To Regions Modal', () => {
...
 
   test('select options disabled correctly based on status and region', async () => {
-    renderCustomRoutesWithReduxRouter(`share/${composeId}`, mockState, routes);
+    renderCustomRoutesWithReduxRouter(`share/${composeId}`, {}, routes);
 
-    const selectToggle = screen.getByRole('button', { name: /options menu/i });
+    const selectToggle = await screen.findByRole('button', {
+      name: /options menu/i,
+    });
     // eslint-disable-next-line testing-library/no-unnecessary-act
```
regexowl, line 107: We can get rid of most (maybe all) `// eslint-disable-next-line testing-library/no-unnecessary-act` lint, but that can be done in a separate PR.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -48,9 +45,9 @@ let router = undefined;
 // to navigate to the images table (via useNavigate hook), which will in turn
 // result in a call to getComposes. If it is not mocked, tests fail due to MSW
 // being unable to resolve that endpoint.
-jest
-  .spyOn(api, 'getComposes')
-  .mockImplementation(() => Promise.resolve(mockComposesEmpty));
+// jest
```
regexowl, line 48: Why removing the lines in CreateImageWizard.content.test.js and ImagesTable.test.js, but only commenting them here?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -399,52 +400,50 @@ describe('Step Upload to AWS', () => {
...
-                  share_with_sources: ['123'],
-                },
-              },
-            },
-          ],
-        });
-        const id = 'edbae1c2-62bc-42c1-ae0c-3110ab718f5a';
-        return Promise.resolve({ id });
-      });
+    // const composeImage = jest
```
regexowl, line 403: Same question - why are we only commenting the mockImplementation? Is it to keep an example for the future?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1188,154 +1206,158 @@ describe('Click through all steps', () => {
...
-            customizations: customizations,
-          };
-          id = 'edbae1c2-62bc-42c1-ae0c-3110ab718f5b';
-        }
-        expect(body).toEqual(expectedbody);
-
-        ids.unshift(id);
-        return Promise.resolve({ id });
-      });
+    // // mock the backend API
```
regexowl, line 1209: And same question - y the comment?

lavocatt (approved):
Thanks Lucas for all the work and addressing the issues I mentioned. For the points not fixed we've discussed them on slack and figured out they'll fit well in a follow up.
I've done some more monkey testing and it holds up pretty good.
Amazing work :rocket:

regexowl (commented):
Managed to go through the changes to the tests for today. Added a few questions. Will continue reviewing tomorrow.

src/Components/ImagesTable/Target.tsx
```diff
@@ -0,0 +1,51 @@
...
+  gcp: 'Google Cloud Platform',
+  'guest-image': 'Virtualization - Guest image',
+  'image-installer': 'Bare metal - Installer',
+  vsphere: 'VMWare vSphere',
+  'vsphere-ova': 'VMWare vSphere',
+  wsl: 'Windows Subsystem for Linux',
+  ami: 'Amazon Web Services',
+  'rhel-edge-commit': 'RHEL Edge Commit',
+  'rhel-edge-installer': 'RHEL Edge Installer',
+  vhd: '',
```
regexowl, line 25: Is the string empty, because we don't use vhd? If so, then it's perfectly fine.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
lucasgarfield: Oversight on my part I'm sure :sweat_smile: nice catch!

src/Components/ImagesTable/Instance.tsx
```diff
@@ -0,0 +1,209 @@
...
+    let accountIds = undefined;
+
+    if (isGcpUploadRequestOptions(options)) {
+      accountIds = options.share_with_accounts;
+    }
+
+    if (isAwsUploadRequestOptions(options)) {
+      accountIds = options.share_with_accounts;
+      sourceIds = options.share_with_sources;
+    }
```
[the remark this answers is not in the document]
ezr-ondrej: yup :+1: both can be undefined, we'll just show the no source for this image found :+1:

src/Components/ImagesTable/Target.tsx
lucasgarfield: I don't think we use it. I suppose we could have it read 'Virtual Hard Disk'? That's not a bad idea... I just added it to satisfy Typescript's requirement that there is a key for every image type.

regexowl: Right, we don't as far as I know. We could fill it in, but I'm also not opposed to leaving it blank to signify that we don't really work with vhd.

regexowl (commented):
This looks so... clean :broom: Love it! Added a few comments, but overall this is super nice. Thanks a lot! :star2:

regexowl:
Ugh, didn't know the review comments will be torn away from the thread they were supposed to be a reaction to. Sorry about those.

src/Components/ImagesTable/ClonesTable.tsx
```diff
@@ -0,0 +1,158 @@
...
+      status={<StatusClone clone={clone} status={status} />}
+    />
+  );
+};
+
+type ComposeRowPropTypes = {
+  compose: ComposesResponseItem;
+};
+
+const ComposeRow = ({ compose }: ComposeRowPropTypes) => {
```
regexowl, line 114: As there's no polling, it seems that the ami of parent compose gets stuck in the skeleton phase on failure. :skull: 

image

lavocatt (approved):
LGTM! :partying_face:

PR #1325, opened 2023-08-23
lint: Fix part of the warnings

regexowl:
This fixes the following warnings outputted when running `npm run lint`:
- "Expected a default case"
- "img elements must have an alt prop, either with meaningful text, or an empty string for decorative images"
- "Assign object/arrow function to a variable before exporting as module default"


croissanne (approved):
Thank you!

PR #1327, opened 2023-08-23
Wizard not in modal

lavocatt:
This pull requests makes the wizard appear in its own page when the user clic on the "create an image" button.
However, there's a bug with the height of the wizard still being limited, see the bug opened on data driven forms https://github.com/data-driven-forms/react-forms/issues/1402

To overcome the bug a custom CSS rule deactivate the height limitation.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -641,7 +641,7 @@ const CreateImageWizard = () => {
             name: 'image-builder-wizard',
             className: 'imageBuilder',
             isDynamic: true,
-            inModal: true,
+            inModal: false,
             onKeyDown: (e) => {
               handleKeyDown(e, handleClose);
             },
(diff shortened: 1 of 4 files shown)


lavocatt:
Still "blocked" by the fact that the height of the wizard is limited. But I don't know, maybe we can merge this one anyway ? I think the benefit of having access to the quickstart outweighs the downside of this bug. What's your opinion @lucasgarfield ?

lavocatt:
I think I can also fix this with two simple CSS rules.

lavocatt:
DDF making the calls to PF4 it turns out we're quite limited in the options we have to style the wizard.

lavocatt:
We've decided with @lucasgarfield to get the wizard in page without waiting for DDF to be removed and to add custom CSS rules in the meantime to cover for the things that are blocking us. The main reason why it's important is that the current wizard in modal suffers from flickering, and going in page solves this issue.

regexowl (approved):
Looks good! Also no more re-renders on opening the modal :star_struck:

PR #1328, opened 2023-08-23
Switch to Content Sources search

ezr-ondrej:
Enable searching for RH packages through Content Sources service. Gates the search by Content Sources feature flag, to keep the easy kill switch.

I've renamed the `.beta.test` -> `.content.test` and moved all the Content related tests there.
So now we are testing stable, but with both switch on and switch off, so there is at least little bit of coverage for the flag being turned off :)
At the same time, it's isolated, so easy to remove once we want to remove the fallback :)


croissanne (dismissed):

croissanne:
ergh I think it needs a test tweak

ezr-ondrej:
> ergh I think it needs a test tweak

right, just a linter.... it's never happy with me :( xD

croissanne (approved):

croissanne (approved):

PR #1329, opened 2023-08-23
Extract quickstarts banner into its own component

ezr-ondrej:
Promissed refactoring :)

Looks much cleaner, good thinking @lucasgarfield ! :+1:


croissanne (approved):
Thanks!

croissanne (approved):

PR #1333, opened 2023-08-28
deps/test: Bump to React 18

regexowl:
This bumps dependencies to React 18.2.0 and testing-library to 14.0.0

With this bump comes update to the tests:
- all clicking on elements is now done via `userEvent`, invoking the `.click()` method like so: `user.click(<element>)`
- the await/async logic was revised
- `act()` calls were replaced by `waitFor()` methods and `findBy` queries where possible


lucasgarfield (approved):
Very nice! Thank you!

PR #1334, opened 2023-08-28
CreateImageWizard: add WSL image type in preview

croissanne:


src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -112,6 +112,14 @@ const ReviewStep = () => {
             <TargetEnvOtherList />
           </TextContent>
         )}
+        {getState()?.values?.['target-environment']?.wsl && (
+          <TextContent>
+            <Text component={TextVariants.h3}>
+              Windows Subsystem for Linux (WSL)
```
regexowl, line 118: Should we put an extension in the parentheses so it's consistent with the other options?

```suggestion
              WSL - Windows Subsystem for Linux (.tar.gz)
```

regexowl (commented):
One question/nit pick.

'wsl' will also need to be added to `ImageDetails` as the UUID in the expandable row is missing: 

Screenshot from 2023-08-29 09-14-33

src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
```diff
@@ -237,6 +247,18 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
           name="Bare metal installer"
           data-testid="checkbox-image-installer"
         />
+        {[RHEL_8, CENTOS_8].includes(getState()?.values?.release) &&
+          isBeta() && (
+            <Checkbox
+              label="Windows Subsystem for Linux (WSL)"
```
regexowl, line 253: Should probably also have extension in the parentheses.

```suggestion
              label="WSL - Windows Subsystem for Linux (.tar.gz)"
```

regexowl (commented):
One more tiny nit pick, overlooked that one in the first review. Otherwise looks great!

regexowl (approved):
Looks great, thank you! :tada:

PR #1335, opened 2023-08-29
openapi: import from yaml instead of JSON

lavocatt:
Use the yaml representation of the openapi used in Image Builder to load generate the RTKQ code.


croissanne:
What's the advantage of yaml over json? Also would it be useful to extend the `api.sh` script to pull the latest specs from upstream?

lavocatt:
The advantage would be to be able to pull directly from upstream instead of getting the api representation from [console.redhat.com](http://console.redhat.com/). But that opens a few more questions about compatibility. I'm trying to wrap my head around these questions :sweat_smile:

lavocatt:
Alright so changing the representation from json to yaml is the first step at automating the task for a github action that would live in IB. This action would check that if on the latest main commit API changes induce a change in the generated code from the frontend and if so, open a PR onto this repo.

PR #1343, opened 2023-09-03
feat(HMS-1454): add RBAC permissions to launch wizard

amirfefer:
This PR adds RBAC implementation for provisioning/launch wizard.

https://github.com/RedHatInsights/image-builder-frontend/assets/11807069/f9a7c3a1-8b21-43c3-8365-90d62c2673fe

### DIFF
--- a/src/Utilities/useProvisioningPermissions.js
+++ b/src/Utilities/useProvisioningPermissions.js
@@ -1,0 +1,36 @@
+import { usePermissions } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
+
+const GLOBAL_PERMISSIONS = ['provisioning:reservation:write'];
+const APP = 'provisioning';
+
+const useProvisioningPermissions = () => {
+  const { hasAccess: hasAWSAccess, isLoading: awsLoading } = usePermissions(
+    APP,
+    [...GLOBAL_PERMISSIONS, 'provisioning:reservation.aws:write'],
+    false,
+    true
+  );
+  const { hasAccess: hasAzureAccess, isLoading: azureLoading } = usePermissions(
+    APP,
+    [...GLOBAL_PERMISSIONS, 'provisioning:reservation.azure:write'],
+    false,
+    true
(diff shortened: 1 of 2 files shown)


src/Utilities/useProvisioningPermissions.js
```diff
@@ -0,0 +1,27 @@
...
+    'provisioning',
+    ['provisioning:*:*', 'provisioning:reservation.aws:write']
+  );
+  const { hasAccess: hasAzureAccess, isLoading: azureLoading } = usePermissions(
+    'provisioning',
+    ['provisioning:*:*', 'provisioning:reservation.azure:write']
+  );
+  const { hasAccess: hasGCPAccess, isLoading: gcpLoading } = usePermissions(
+    'provisioning',
+    ['provisioning:*:*', 'provisioning:reservation.gcp:write']
```
ezr-ondrej, line 14: There is also `provisioning:reservation:write` I'm guessing that most likely means, the user has all permissions, but I'd better be safe here :)

amirfefer: Interesting, is it a  bug?
Looks like `provisioning:reservation:write` is not a general permission.
There are `Launch on AWS/Azure/GCP` roles that add the `provisioning:reservation:*` permission,  therefore if I added `provisioning:reservation:write` as a general permission in the array, and I have only `Launch on AWS` role, I can also launch GCP / Azure instances.

amirfefer:
https://github.com/RedHatInsights/image-builder-frontend/assets/11807069/a42962fd-f38b-4c3d-8f28-cad86755523d


After a discussion offline, we are adding a warning in the images table, and in the launch wizard, there will be guidance for launching directly via the cloud provider.

amirfefer:
Following the discussion, I removed the warning and tooltip for now, clicking `Launch` opens a modal
Screen Shot 2023-09-06 at 21 27 48

lucasgarfield (approved):

PR #1345, opened 2023-09-04
API: Add mutations for creating composes and clones

lucasgarfield:
This commit adds mutations for the /compose and
/composes/:composeId/clone POST end points to the Image Builder API slice.


lavocatt:
Sorry for the dumb question, but what are mutations about/doing ?

regexowl (approved):
LGTM!

PR #1347, opened 2023-09-04
Wizard: Update the Repositories step

regexowl:
This updates the Repositories and Review step as per [mocks](https://www.sketch.com/s/d7aa6d29-fca0-4283-a846-09cc5fd10612/a/MyEbDz7).

Repositories with the unavailable or invalid status have a popover that allows for further inspection. The time of the last introspection and the counter of failed attempts was added to the popover, together with the "Go to Repositories" button.

On Recreate the payload repositories are checked against "freshly" fetched list of repositories. In case any of the previously checked repositories is no longer available in content sources an Alert is rendered on both Repositories and Review steps. The unavailable repository is checked, but the checkbox is disabled and the information is dashed out. Since the information about the repository is stored in the Repository type, the only information available to be rendered is the baseurl.

Create image button is also disabled when recreating an image with unavailable repositories.


regexowl:
How can this be tested:
1. Add EPEL 9 repository in Repositories app
2. Create an image with the EPEL 9 repo
3. Remove the EPEL 9 repo from Repositories app
4. Recreate and check the Review and Repositories step

regexowl:
Oh and the repositories mock fixtures had a wrong time format and it took me ages to figure out why I can't make our utilities time functions work :no_mouth:

regexowl:
> Cool PR! I'm wondering if that's possible to add new unit tests to cover for that functionality ? What's your opinion ?

Can't see why it wouldn't and totally agree it's the right thing to do! :sweat_smile: 
I'll see what I can do :eyes:

src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -439,7 +441,9 @@ const Repositories = (props) => {
                               }}
                             />
                             <Td dataLabel={'Name'}>
-                              {repo.name}
+                              {repo.name
```
lavocatt, line 444: nitpick: it's a bit confusing down the line where you have `repo.name ? repo.distribution : '-'` maybe you can add a new constant up top called `repoExists` and have that to true is the repo is set ?

src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -454,16 +458,24 @@ const Repositories = (props) => {
...
                               <RepositoriesStatus
-                                repoStatus={repo.status}
+                                repoStatus={
+                                  repo.name ? repo.status : 'Unavailable'
+                                }
                                 repoUrl={repo.url}
+                                repoIntrospections={
+                                  repo.last_introspection_time
+                                }
+                                repoFailCount={repo.failed_introspections_count}
```
lavocatt, line 478: Is there the check for repo existing missing for the fail count ?

src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -519,10 +530,10 @@ const Empty = ({ isFetching, refetch }) => {
         href={isBeta() ? '/preview/settings/content' : '/settings/content'}
         className="pf-u-mr-sm"
       >
-        Repositories
+        Go to repositories
       </Button>
       <Button
-        variant="primary"
+        variant="secondary"
```
lavocatt, line 536: Sorry if it's a dumb question, but why the variant had to change ?

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
```diff
@@ -1,62 +1,139 @@
...
         <Popover
           position="bottom"
           minWidth="30rem"
           bodyContent={
             <>
-              <Alert variant="warning" title="Unavailable" isInline isPlain />
-              Cannot fetch {repoUrl}
+              <Alert
+                variant={repoStatus === 'Invalid' ? 'danger' : 'warning'}
+                title={repoStatus === 'Invalid' ? 'Invalid' : 'Unavailable'}
```
lavocatt, line 77: Because we check up to that that `repoStatus` can be either `Invalid` or `Unavailable` then you can just let the title be `repoStatus`

src/Components/CreateImageWizard/formComponents/RepositoryUnavailable.js
```diff
@@ -0,0 +1,42 @@
...
+import { Alert, Button } from '@patternfly/react-core';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+
+import { useCheckRepositoriesAvailability } from '../../../Utilities/checkRepositoriesAvailability';
+import { useGetEnvironment } from '../../../Utilities/useGetEnvironment';
+
+const RepositoryUnavailable = () => {
+  const { isBeta } = useGetEnvironment();
+
+  if (useCheckRepositoriesAvailability()) {
```
lavocatt, line 12: There might be an issue there, because `useCheckRepositoriesAvailability` only fetches the 100 first repos for x86:

```
  const { data, isSuccess } = useListRepositoriesQuery({
    availableForArch: 'x86_64',
    availableForVersion: version,
    limit: 100,
    offset: 0,
  });
```

So if there is a lot of repos, we could endip missing some unavailable repos.

src/Components/CreateImageWizard/formComponents/ReviewStepTables.js
```diff
@@ -20,18 +20,23 @@ const RepoName = ({ repoUrl }) => {
...
+        isPlain
+        title="Error loading repository name"
+      />
+    );
+  };
+
   return (
     <>
-      {isSuccess && <p>{data.data?.[0].name}</p>}
+      {isSuccess && data.data?.[0]?.name && <p>{data.data?.[0].name}</p>}
```
lavocatt, line 36: The succeeding blocs are a bit confusing for me to understand sorry.
Can you comment each of these lines please ?

src/Utilities/checkRepositoriesAvailability.js
```diff
@@ -0,0 +1,35 @@
...
+    offset: 0,
+  });
+
+  const unavailableRepos = [];
+  const payloadRepositories = getState()?.values?.['payload-repositories'];
+  // payloadRepositories existing === we came here from Recreate
+  if (isSuccess && payloadRepositories) {
+    payloadRepositories.forEach((repo) => {
+      if (!data.data.find((repoFromApi) => repoFromApi.url === repo.baseurl)) {
+        unavailableRepos.push(repo);
```
lavocatt, line 26: You can early return `true` here :slightly_smiling_face:

src/Utilities/checkRepositoriesAvailability.js
```diff
@@ -0,0 +1,35 @@
+import { useFormApi } from '@data-driven-forms/react-form-renderer';
+
+import { releaseToVersion } from './releaseToVersion';
+
+import { useListRepositoriesQuery } from '../store/contentSourcesApi';
+
+export const useCheckRepositoriesAvailability = () => {
```
lavocatt, line 7: I'd love to have a comment for this function and what it does

src/test/Components/CreateImageWizard/CreateImageWizard.content.test.js
```diff
@@ -868,3 +873,53 @@ describe('Step Custom repositories', () => {
...
+      'imagewizard/b7193673-8dcc-4a5f-ac30-e9f4940d8346',
+      state
+    ));
+  };
+
+  test('selected repositories stored in and retrieved from form state', async () => {
+    await setUp();
+
+    screen.getByRole('heading', { name: /review/i });
+    await screen.findByText('Previously added custom repository unavailable');
```
lavocatt, line 896: Awesome! Could be also nice to have the opposite test where you don't have an expired custom repo. That way we are protected from future misschanges.

src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -404,9 +406,9 @@ const Repositories = (props) => {
                     <Tr>
                       <Th />
                       <Th width={45}>Name</Th>
-                      <Th>Architecture</Th>
-                      <Th>Versions</Th>
-                      <Th>Packages</Th>
+                      <Th>Arch</Th>
```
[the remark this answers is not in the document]
regexowl: I've tried to use the shorter names that are used in the mocks. With the addition of the Status some of the column names can be too long for the width of the column: 

image

The possible solution is either shortening the headers or widening the minimum width of the columns which will shrink the name and the url a bit. Not sure which is better.

Or leaving it as is, but I'm a bit bothered by the "Architec..." :sweat_smile:

src/Components/CreateImageWizard/formComponents/Repositories.js
regexowl: To bring the step closer to the mocks below. I think the idea is to differentiate between the buttons a bit and also [design guidelines](https://www.patternfly.org/components/button/design-guidelines) seem to not put two primary buttons right next to each other as well.

image

src/Components/CreateImageWizard/formComponents/Repositories.js
regexowl: `RepositoriesStatus` component checks whether `repoFailCount` is defined and renders the count conditionaly based on that

```
{repoFailCount && (
  <DescriptionListGroup>
  <DescriptionListTerm>
```

src/Components/CreateImageWizard/formComponents/RepositoriesStatus.tsx
regexowl: Yup :no_mouth:, thanks for pointing that out :sweat_smile:

src/Components/CreateImageWizard/formComponents/RepositoryUnavailable.js
regexowl: Oof, you're right. I'm wondering what the best solution would be. The limit is 100 by default so either we can set an artificial higher number that we hope wouldn't be exceeded... or we do the whole repo shebang with initial request, followed up by a request with the count of the repos from the initial one :thinking:

regexowl: For now chose the first route with a to do note

src/Components/CreateImageWizard/formComponents/Repositories.js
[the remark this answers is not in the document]
lavocatt: IC, that's not an easy choice, let's ask others opinion ?

src/Components/CreateImageWizard/formComponents/RepositoryUnavailable.js
lavocatt: I think we should keep a low page count but ask for next pages if needed. You can also mix that with the early return, so you check the first page, if something unavailable is found in it, you return true immediately, if not you continue checking the next page.

src/Components/CreateImageWizard/formComponents/Repositories.js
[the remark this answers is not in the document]
regexowl: Managed to unsmush the architecture header while keeping the full names. Will update in the next push.

src/Components/CreateImageWizard/formComponents/RepositoryUnavailable.js
regexowl: Can't think of a way to check by page and not use a hook in a cycle. Re-used a part of a code we already utilise in `Repositories.js`. Making an initial request just to get the total count of the repositories and if that is higher than 100 making a follow up request with limit set to count to fetch all of them. `data` and `isSuccess` is then populated based on the count.

Will need to invest in AC for next year, my brain is melting...

lavocatt: Nice pair programming session :rocket:

lavocatt (approved):
Good work! And thanks for all the fixes :rocket:

PR #1351, opened 2023-09-06
api: ci checks that the api is not changed by hand + doc

lavocatt:
This PR adds a check to ensure nobody is updating the API by hand.
Also adds documentation on how to update the endpoints and apis in the README.

Manual changes on the api are detected by running `npm run api` and then `git status`, when there are some, travis will fail:

image

And this check let the rest of the CI execute when there's no manual changes on the API files:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ node_js:
   - '16'
 install:
   - npm ci
+before_script: |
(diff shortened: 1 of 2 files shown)


lavocatt:
I'll add manual changes to the API in the PR to see if the check fails on that

croissanne (approved):
lgtm, thanks!

croissanne (approved):

PR #1355, opened 2023-09-10
Add GCP image sharing options

adiabramovitch:
Depends on https://github.com/osbuild/image-builder/issues/829
Adding GCP image sharing option according to the discussion on [slack](https://redhat-internal.slack.com/archives/C03AZ0264LW/p1692789579814619) and [mocks](https://issues.redhat.com/browse/HMS-2352). 

In summary, within our GCP sources, we store the project ID. Images cannot be shared directly with a project ID, but they can be shared with a service account or a Google account. Consequently, to launch instances in GCP, users are not required to provide their Google account; instead, the images should be shared with the provisioning service account. This ensures that the provisioning team has access to the necessary images, as sharing them with individual users would not allow that.

After a thorough discussion, we have collectively decided to introduce an option. This option allows users to exclusively utilize the Launch service without the need to share the image with a Google account. We are providing an opt-out feature, as depicted in the screenshot below.

image

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -122,20 +122,22 @@ const onSave = (values) => {
 
   if (values['target-environment']?.gcp) {
     let share = '';
-    switch (values['google-account-type']) {
-      case 'googleAccount':
-        share = `user:${values['google-email']}`;
-        break;
-      case 'serviceAccount':
-        share = `serviceAccount:${values['google-email']}`;
-        break;
-      case 'googleGroup':
-        share = `group:${values['google-email']}`;
-        break;
-      case 'domain':
-        share = `domain:${values['google-domain']}`;
-        break;
-      // no default
+    if (values['image_sharing'] === 'gcp-account') {
+      switch (values['google-account-type']) {
+        case 'googleAccount':
+          share = `user:${values['google-email']}`;
+          break;
+        case 'serviceAccount':
+          share = `serviceAccount:${values['google-email']}`;
+          break;
+        case 'googleGroup':
+          share = `group:${values['google-email']}`;
+          break;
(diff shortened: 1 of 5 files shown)


lucasgarfield:
Nice work Adi! Just one thing, I'd like for `Share Image with Red Hat Insights Only` to be the default option for now. Right now there is no default option selected, and I think that will cause a lot of users to drop off. `Share Image with Red Hat Insights Only` requires the least amount of work, so I'd like to try that for now.

adiabramovitch:
@lucasgarfield Tests are passing locally, can you restart the pipline?

croissanne:
> Can we change the wording a bit? How about something like this:
> 
> `Select how to share your image. The image you create can be used to launch instances on GCP, regardless of which method you select.`
> 
> `Your image will be uploaded to GCP and shared with Red Hat Insights. The image expires in 14 days.`
> 
> `Your image will be uploaded to GCP and shared with the account you provide below. The image expires in 14 days. To keep permanent access to your image, copy it to your GCP project.`
> 
> Thoughts, @croissanne ?

Sounds good!

lavocatt:
Can you to include a summary of the slack discussion and the link to the mocks in the commit message please ?
Also, can you update the title of the commit message to be `wizard: add GCP image sharing options` please ?
Thanks.

lavocatt:
> > Can you to include a summary of the slack discussion and the link to the mocks in the commit message please ? Also, can you update the title of the commit message to be `wizard: add GCP image sharing options` please ? Thanks.
> 
> The links to the mocks and slack discussion are linked above, and I have added more detailed description.

Yes the PR description is great. But I meant specifically the commit message, can you `git commit --amend` it please ? Like adding the description you have for the pr would be a great start, and also the title should be `wizard: add GCP image sharing options`.

Thanks.

PR #1356, opened 2023-09-11
api: add new edge endpoints

mgold1234:
This commit adds 2 more RTQ Query in preparation of porting edge source code to image-builder-frontend


lavocatt:
Thanks for this. Can you change the format of your commit please ? We tend to use a short title followed by a description.

In that case, the title should be `api: add new edge endpoints` and the content something inspired by what you wrote in the PR description.

lavocatt (approved):
LGTM!

PR #1358, opened 2023-09-15
Filter content-sources repos by origin and content_type

jlsherrill:
These params will be added to content-sources soon.  And in the short future there will be other origin's available (Red Hat for example).  In the far future there will be other content_types (most likely). 

I haven't tested this change so tread lightly :)

Content sources PR: https://github.com/content-services/content-sources-backend/pull/383

But this shouldn't cause any problem going in earlier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ foobarApi.ts
 7. run api generation
 
 ```
-rpn run api
+npm run api
 ```
 
 And voilà!
--- a/api/schema/contentSources.json
+++ b/api/schema/contentSources.json
@@ -277,6 +277,10 @@
                         "readOnly": true,
                         "type": "string"
                     },
+                    "content_type": {
+                        "description": "Content Type (rpm) of the repository",
+                        "type": "string"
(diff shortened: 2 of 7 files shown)


croissanne (approved):

jlsherrill:
oh i see its generated, working on it

README.md
```diff
@@ -147,7 +147,7 @@ foobarApi.ts
 7. run api generation
 
 ```
-rpn run api
+npm run api
```
jlsherrill, line 150: i assume this was a typo?

lucasgarfield:
Glad you found the instructions! We want to make this fully automated with `npm run api` (and we also want to catch changes to the APIs we consume via a Github action) but we aren't quite there yet...

lucasgarfield:
We will also need to grep for every time we call `useListRepositoriesQuery` and add the new params there, too.

lucasgarfield:
Thank you!!!

lucasgarfield (approved):

PR #1360, opened 2023-09-18
LaunchWizard: Restore Launch wizard functionality

lucasgarfield:
This commit is a hotfix that restores functionality to the Launch wizard.

The `appendTo` variable must be wrapped in a useMemo() hook to work properly. Additionally, the `hasAccess`  property was missing from the `<ProvisioningWizard>` component.

To ensure full backwards compatibility, the component has also been wrapped in a `<Suspense>` as it was previously.


lavocatt (approved):
That's all good to me.

PR #1361, opened 2023-09-18
LaunchWizard: Correct architecture prop

lucasgarfield:
The `<LaunchWizard>` architecture prop was specified incorrectly, which caused the Instance dropdown in the Launch wizard to fail to load instance types.


adiabramovitch:
Tested and work ad Expected! 
Thanks @lucasgarfield

croissanne (approved):

PR #1366, opened 2023-09-27
ImageTable: THEEDGE-3450 - add ignoreImageType to getComposes endpoint

mgold1234:
This commit introduces a filter for the "image_type" to the "getComposes" endpoint, thereby improving the functionality of the Image Builder frontend table. This table consists of two tabs: RPM and OSTree. With this enhancement, users gain the ability to filter out OSTree images from the RPM tab, thus avoiding duplication.
this change is the impact of changes in the backend side in this pr https://github.com/osbuild/image-builder/pull/825

### DIFF
--- a/src/store/emptyImageBuilderApi.ts
+++ b/src/store/emptyImageBuilderApi.ts
@@ -5,6 +5,35 @@ import { IMAGE_BUILDER_API } from '../constants';
 // initialize an empty api service that we'll inject endpoints into later as needed
 export const emptyImageBuilderApi = createApi({
   reducerPath: 'imageBuilderApi',
-  baseQuery: fetchBaseQuery({ baseUrl: IMAGE_BUILDER_API }),
+  baseQuery: fetchBaseQuery({
+    baseUrl: IMAGE_BUILDER_API,
+    paramsSerializer: (params) => {
+      /*
+       * Image builder backend requires the arrays in get requests to be
+       * exploded, see the default behavior for swagger in the documentation:
+       * https://swagger.io/docs/specification/serialization/
+       *
+       * To accommodate to that, make sure that when the request is sent with
+       * an array we do explode properly unlike the default behavior of
+       * URLSearchParams.
+       */
+      const searchParams = new URLSearchParams();
+      for (const [key, value] of Object.entries(params)) {
+        if (value === undefined) {
+          // avoid sending undefined parameters
+          continue;
+        }
+        if (Array.isArray(value)) {
+          for (const v of value) {
+            searchParams.append(key, v);
+          }
+        } else {
+          searchParams.append(key, value);
+        }
+      }
+
+      return searchParams.toString();
+    },
+  }),
   endpoints: () => ({}),
 });
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -72,6 +72,12 @@ const ImagesTable = () => {
   const { data, isSuccess } = useGetComposesQuery({
     limit: perPage,
     offset: perPage * (page - 1),
+    ignoreImageTypes: [
+      'rhel-edge-commit',
+      'rhel-edge-installer',
+      'edge-commit',
+      'edge-installer',
+    ],
   });
 
   if (!isSuccess) {


src/store/imageBuilderApi.ts
```diff
@@ -1,5 +1,13 @@
 import { emptyImageBuilderApi as api } from "./emptyImageBuilderApi";
+import { fetchBaseQuery } from '@reduxjs/toolkit/query'
+
 const injectedRtkApi = api.injectEndpoints({
+  baseQuery: fetchBaseQuery({paramsSerializer: (params) => {
```
lavocatt, line 5: Can you try to write this in an other way ? Or maybe comment what's happening, because it's quite packed rn.

src/store/imageBuilderApi.ts
```diff
@@ -1,5 +1,13 @@
...
+      const [multiples, regulars] = Object.entries(params).reduce((acc, [key, value]) => ([{...acc[0], ...Array.isArray(value) && { [key]: value }}, {...acc[1], ...!Array.isArray(value) && { [key]: value }}]), {}),
+          const searchParams = new URLSearchParams(regulars);
+      Object.entries(multiples).forEach(([ key, value ]) => searchParams.append(key, value));
+      return searchParams.toString();
```
lavocatt, line 9: Can you try to write this in another way ? Or if not, can you add comment about what's happening there and why you need this ? Thanks.

regexowl:
The filtering seems to be working great, thank you!

Noticed one weird thing - the target does not render in the case of AWS images. 

image

I think the problem could be connected to the fetching of the clones (and thus not being able to render the number of clones in the parentheses). This is the `/composes/{composeId}/clones` query I can see. It gets the limit and offset appended as undefined and ends up with the 400 status.

image

On main the query seems to end up with the 200 status without any of the params being appended to the query:

image

regexowl:
Tried changing the `ignoreImageTypes` and the filtering seems to be working nicely. But with making manual changes to imageBuilder.yaml we would loose parity between the api schema in image-builder and frontend.

There could be another possible solution that would not require the removal of the params, but I'm wondering if it would be possible to remove the limit and offset for clones in image-builder.

@lucasgarfield  As you did the refactoring of the images table - is it possible to estimate how many clones would we now expect a user can possibly have? Do we allow more clones for the same region? And also am I guessing right that the offset and limit would be used for some kind of pagination of the clones?

lavocatt:
> Tried changing the `ignoreImageTypes` and the filtering seems to be working nicely. But with making manual changes to imageBuilder.yaml we would loose parity between the api schema in image-builder and frontend.

Indeed, it's not possible to change the api schema this way. Sorry for that!

> There could be another possible solution that would not require the removal of the params, but I'm wondering if it would be possible to remove the limit and offset for clones in image-builder.
>
> @lucasgarfield As you did the refactoring of the images table - is it possible to estimate how many clones would we now expect a user can possibly have? Do we allow more clones for the same region? And also am I guessing right that the offset and limit would be used for some kind of pagination of the clones?

Do you mean actually removing them from IB's API?

lucasgarfield:
> Tried changing the `ignoreImageTypes` and the filtering seems to be working nicely. But with making manual changes to imageBuilder.yaml we would loose parity between the api schema in image-builder and frontend.
> 
> There could be another possible solution that would not require the removal of the params, but I'm wondering if it would be possible to remove the limit and offset for clones in image-builder.
> 
> @lucasgarfield As you did the refactoring of the images table - is it possible to estimate how many clones would we now expect a user can possibly have? Do we allow more clones for the same region? And also am I guessing right that the offset and limit would be used for some kind of pagination of the clones?

It should be less than the number of regions, so ~30 as an absolute maximum but in all likelihood far less. Not enough users currently use this feature for us to say with any real confidence one way or another based on data, all we can do is speculate. While currently possible to share to a region more than once in the UI, that will be changed in the near future and it is not very likely to occur in the mean-time.

A better option would probably be to keep the limit and offset and show the clones in a sub-table that supports pagination. That way we can just do things the "right" way and don't need to guess at anything.

api/schema/imageBuilder.yaml
```diff
@@ -214,21 +214,6 @@ paths:
...
-            minimum: 1
-            maximum: 100
-          description: max amount of clones, default 100
-        - in: query
-          name: offset
-          schema:
-            type: integer
-            default: 0
-            minimum: 0
-          description: clones page offset, default 0
```
lavocatt, line 231: The fact that the clone endpoint is having troubles now is because of the change here: https://github.com/RedHatInsights/image-builder-frontend/pull/1366/files#diff-bd7250abbffed5537d68df905da2a5693d884dbfaa60889e01129f86d1df316eR8-R23 Basically you shouldn't endup in a situation where you need to change the API representation. If it happens it's because something else is wrong :slightly_smiling_face:

src/store/emptyImageBuilderApi.ts
```diff
@@ -5,6 +5,21 @@ import { IMAGE_BUILDER_API } from '../constants';
...
+          for (const v of value) {
+            searchParams.append(key, v);
+          }
+        } else {
+          searchParams.append(key, value);
+        }
+      }
+      return searchParams.toString();
+    },
+  }),
```
lavocatt, line 23: Let's unpack a bit what's changed here, so basically this change implies that now, every request sent to the image builder backend will be executing this chunk of code.

I don't think that's a correct approach to doing the thing that you want. Especially since it has already consequences for the cloneCompose endpoint. In the future there might be other endpoints having the same issue and that's going to be super hard to find out what's going on.

So we have to find another way of doing things for this PR.

lavocatt (changes requested):

lavocatt:
@mgold1234 can you add a link to the backend PR in the description please? It's nice to link the front and backend of things :slightly_smiling_face: https://github.com/osbuild/image-builder/pull/825

lavocatt:
@mgold1234 I've done some changes, can you have a quick look at them please?

src/store/emptyImageBuilderApi.ts
```diff
@@ -5,6 +5,32 @@ import { IMAGE_BUILDER_API } from '../constants';
...
+        }
+        if (Array.isArray(value)) {
+          value.forEach((v) => searchParams.append(key, v));
+        } else {
+          searchParams.append(key, value);
+        }
+      }
+      return searchParams.toString();
+    },
+  }),
```
lavocatt, line 34: @lucasgarfield and @regexowl I really want your opinion on this chunk. I think it's all good now, but it's only handling arrays not objects. I don't know if we should support object as get parameters, I don't think it makes much sense tbh. It would also be possible that this breaks in the future. I'm slightly worried :sweat_smile:

src/store/emptyImageBuilderApi.ts
```diff
@@ -5,6 +5,24 @@ import { IMAGE_BUILDER_API } from '../constants';
...
   reducerPath: 'imageBuilderApi',
-  baseQuery: fetchBaseQuery({ baseUrl: IMAGE_BUILDER_API }),
+  baseQuery: fetchBaseQuery({
+    baseUrl: IMAGE_BUILDER_API,
+    paramsSerializer: (params) => {
+      const searchParams = new URLSearchParams();
+      for (const [key, value] of Object.entries(params)) {
+        if (value === undefined) {
+          // avoid sending undefined parameters
+          break;
```
lavocatt, line 15: `break` will exit the loop immediately, which is not what we want if the first element has an undefined value, because the others after that might be valid and must be included. see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break

`continue` on the other hand just jumps to the next loop. see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/continue

What I would do, if using `continue` is not what you want, is simply to change the logic of the if conditions:

```
if (value){
  if (Array.isArray(value)){
  }else{
  }
}
```

src/store/emptyImageBuilderApi.ts
```diff
@@ -5,6 +5,24 @@ import { IMAGE_BUILDER_API } from '../constants';
...
 export const emptyImageBuilderApi = createApi({
   reducerPath: 'imageBuilderApi',
-  baseQuery: fetchBaseQuery({ baseUrl: IMAGE_BUILDER_API }),
+  baseQuery: fetchBaseQuery({
+    baseUrl: IMAGE_BUILDER_API,
+    paramsSerializer: (params) => {
+      const searchParams = new URLSearchParams();
+      for (const [key, value] of Object.entries(params)) {
+        if (value === undefined) {
+          // avoid sending undefined parameters
```
lavocatt, line 14: indeed this comment can be removed

src/store/emptyImageBuilderApi.ts
```diff
@@ -5,6 +5,24 @@ import { IMAGE_BUILDER_API } from '../constants';
 // initialize an empty api service that we'll inject endpoints into later as needed
 export const emptyImageBuilderApi = createApi({
   reducerPath: 'imageBuilderApi',
-  baseQuery: fetchBaseQuery({ baseUrl: IMAGE_BUILDER_API }),
+  baseQuery: fetchBaseQuery({
+    baseUrl: IMAGE_BUILDER_API,
+    paramsSerializer: (params) => {
```
lavocatt, line 10: I'd like to have a note her about why there is a necessity to use the paramSerializer for the image builder API. Look at it as giving a hint for future us about why we're doing things. It reduces the risk of code becoming debt over time.

lavocatt (changes requested):
Thanks for you work on this! I added a few comments, I hope they help! I've also noticed that you have a merge commit in your commits. You should remove it using `git rebase` in interractive mode:

```
git checkout main
git pull
git rebase -i main add_image_type_filter
```

This will open a text editor with the two commits in your PR, delete the line with the merge commit in it. Then you can push again on your branch  using `-f`

If you encounter any issues in that process ping me on slack and I'll give you a hand.

src/store/emptyImageBuilderApi.ts
mgold1234: ok, added your coments that you suggest me in the preview review

lavocatt:
Thanks for making the changes @mgold1234 it's very much appreciated. I will let @lucasgarfield checking that these changes are fine with him too. But from my POV we can ship it.

lavocatt (approved):
LGTM!

PR #1368, opened 2023-09-28
API: Add Compliance API slice

lucasgarfield:
This commit adds the Compliance service's API, which we can query for a list of SCAP security policies.

In a follow-up PR, we will allow the user to optionally select a policy when creating an image in the wizard.


regexowl (approved):
Looks good :+1:

lavocatt (approved):
:rocket: send it!

lavocatt:
ahah, I reviewed after the party :smile:

PR #1371, opened 2023-10-03
Openscap: allow customers to configure an oscap policy in the wizard

lavocatt:
This pull request adds to the wizard a new step to let the user pick an oscap policy.
Applying a policy has for effect to add (depending on the required customizations of the policy):

* Filesystem customizations
* Packages

The chosen policy is also added to the compose request send to IB.
This work is compatible with the recreate functionality as long as the policy is included in the compose request to recreate.

**Some design choices:**
When the user chose a policy, the next steps of FSC and Packages are impacted. And we took the choice to reinitialize these two steps every time the user makes changes in the policy steps. That means that if a user selects a policy, goes to the packages step to add some and then changes back their policy, the packages step will get reinitialized again.
Same goes on for the recreate use case.

Maybe we can add a small note on the Compliance step to make it clear that any changes in that step will trigger a re-initialization of the FSC and Packages steps ?

**Here are a couple of screenshots for the functionality in action:**

image

image

image

image

image

image

image

Big thanks to @regexowl for the pair programming sessions on this PR!

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
+++ b/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
@@ -7,6 +7,7 @@ import {
   Alert,
   Button,
   Popover,
+  Spinner,
   Text,
   TextContent,
   TextVariants,
@@ -28,10 +29,11 @@ import {
 } from '@patternfly/react-table';
 import { v4 as uuidv4 } from 'uuid';
 
-import MountPoint from './MountPoint';
+import MountPoint, { MountPointValidPrefixes } from './MountPoint';
 import SizeUnit from './SizeUnit';
 
-import { UNIT_GIB } from '../../../constants';
+import { UNIT_GIB, UNIT_KIB, UNIT_MIB } from '../../../constants';
+import { useGetOscapCustomizationsQuery } from '../../../store/imageBuilderApi';
 
 const initialRow = {
   id: uuidv4(),
@@ -51,6 +53,56 @@ const FileSystemConfiguration = ({ ...props }) => {
   const [tempItemOrder, setTempItemOrder] = useState([]);
   const bodyref = useRef();
   const [rows, setRows] = useState([initialRow]);
+
+  const oscapPolicy = getState()?.values?.['oscap-policy'];
+  const hasNoOscapPolicy = !oscapPolicy;
+  const hasCustomizations = !(
+    getState()?.values?.['file-system-configuration'] === undefined ||
+    getState().values['file-system-configuration'].length === 1
+  );
+
+  const {
+    data: customizations,
+    isFetching: isFetchingCustomizations,
+    isSuccess,
+  } = useGetOscapCustomizationsQuery(
+    {
+      distribution: getState()?.values?.['release'],
+      profile: oscapPolicy,
+    },
+    {
+      // Don't override the user's data if they made customizations
+      skip: hasNoOscapPolicy || hasCustomizations,
+    }
+  );
+
+  useEffect(() => {
+    if (hasCustomizations || rows.length > 1) {
+      return;
+    }
+    if (customizations && customizations.filesystem && isSuccess) {
+      const newRows = rows;
+
+      // Only keep the required partitions that have a valid mount points
+      const fss = customizations.filesystem.filter((fs) =>
+        MountPointValidPrefixes.includes(fs.mountpoint)
+      );
+
+      // And add them all to the list.
+      for (const fs of fss) {
+        newRows.push({
+          id: uuidv4(),
+          mountpoint: fs.mountpoint,
+          fstype: 'xfs',
+          size: fs.min_size / UNIT_MIB, // the unit from the customizations are in bytes
+          unit: UNIT_MIB, // and using MIB seems sensible here instead
+        });
+      }
+      setRows(newRows);
+      setItemOrder(newRows.map((row) => row.id));
+      change('file-system-config-radio', 'manual');
+    }
+  }, [customizations, isSuccess]);
 
   useEffect(() => {
     const fsc = getState()?.values?.['file-system-configuration'];
@@ -243,6 +295,13 @@ const FileSystemConfiguration = ({ ...props }) => {
 
     setRows(newRows);
   };
+
+  // Don't let the user interact with the partitions while we are getting the
+  // customizations. Having a customizations added by the user first would mess
+  // up the logic.
+  if (isFetchingCustomizations) {
+    return <Spinner isSVG size="lg" />;
+  }
 
   return (
     <FormSpy>
--- a/src/Components/CreateImageWizard/formComponents/Oscap.js
+++ b/src/Components/CreateImageWizard/formComponents/Oscap.js
@@ -14,6 +14,17 @@ import {
 import PropTypes from 'prop-types';
 
 import { useGetOscapProfilesQuery } from '../../../store/imageBuilderApi';
+import { reinitFileSystemConfiguratioStep } from '../steps/fileSystemConfiguration';
+
+/**
+ * Every time there is change on this form step's state, reinitialise the steps
+ * that are depending on it. This will ensure that if the user goes back and
+ * change their mind, going forward again leaves them in a coherent and workable
+ * form state.
+ */
+const reinitDependingSteps = (change) => {
+  reinitFileSystemConfiguratioStep(change);
+};
 
 /**
  * Component for the user to select the policy to apply to their image.
@@ -49,12 +60,15 @@ const PolicySelector = ({ input, showSelector }) => {
   const handleClear = () => {
     selectPolicy(undefined);
     change(input.name, undefined);
+    reinitDependingSteps(change);
   };
 
   const setPolicy = (_, selection) => {
     selectPolicy(selection);
     setIsOpen(false);
     change(input.name, selection);
+    reinitDependingSteps(change);
+    change('file-system-config-radio', 'manual');
   };
 
   return (
@@ -125,6 +139,7 @@ const AddPolicy = ({ input }) => {
           isChecked={wantsPolicy}
           onChange={() => {
             setWantsPolicy(true);
+            reinitDependingSteps(change);
           }}
         />
         <Radio
@@ -137,6 +152,7 @@ const AddPolicy = ({ input }) => {
           onChange={() => {
             setWantsPolicy(false);
             change(input.name, undefined);
+            reinitDependingSteps(change);
           }}
         />
       </FormGroup>
--- a/src/Components/CreateImageWizard/formComponents/Oscap.js
+++ b/src/Components/CreateImageWizard/formComponents/Oscap.js
@@ -1,0 +1,155 @@
+import React, { useState } from 'react';
+
+import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
+import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
+import {
+  Alert,
+  Radio,
+  FormGroup,
+  Select,
+  SelectOption,
+  SelectVariant,
+  Spinner,
+} from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+
+import { useGetOscapProfilesQuery } from '../../../store/imageBuilderApi';
+
+/**
+ * Component for the user to select the policy to apply to their image.
+ * The selected policy will be stored in the `oscap-policy` form state variable.
+ * The Component is shown or not depending on the ShowSelector variable.
+ */
+const PolicySelector = ({ input, showSelector }) => {
+  const { change, getState } = useFormApi();
+  const [policy, selectPolicy] = useState(getState()?.values?.['oscap-policy']);
+  const [isOpen, setIsOpen] = useState(false);
+
+  const { data, isFetching, isSuccess, isError, refetch } =
+    useGetOscapProfilesQuery(
+      {
+        distribution: getState()?.values?.['release'],
+      },
+      {
+        skip: !showSelector,
+      }
+    );
+
+  if (!showSelector) {
+    return undefined;
+  }
+
+  const handleToggle = () => {
+    if (!isOpen) {
+      refetch();
+    }
+    setIsOpen(!isOpen);
+  };
+
+  const handleClear = () => {
+    selectPolicy(undefined);
+    change(input.name, undefined);
+  };
+
+  const setPolicy = (_, selection) => {
+    selectPolicy(selection);
+    setIsOpen(false);
+    change(input.name, selection);
+  };
+
+  return (
+    <FormGroup
+      isRequired={true}
+      label={'Policy to use for this image'}
+      data-testid="policies-form-group"
+    >
+      <Select
+        ouiaId="policySelect"
+        variant={SelectVariant.typeahead}
+        onToggle={handleToggle}
+        onSelect={setPolicy}
+        onClear={handleClear}
+        selections={policy}
+        isOpen={isOpen}
+        placeholderText="Select a policy"
+        typeAheadAriaLabel="Select a policy"
+        isDisabled={!isSuccess}
+      >
+        {isSuccess &&
+          data.map((key, index) => <SelectOption key={index} value={key} />)}
+        {isFetching && (
+          <SelectOption isNoResultsOption={true} data-testid="policies-loading">
+            <Spinner isSVG size="md" />
+          </SelectOption>
+        )}
+      </Select>
+      {isError && (
+        <Alert
+          title="Error fetching the policies"
+          variant="danger"
+          isPlain
+          isInline
+        >
+          Cannot get the list of policies
+        </Alert>
+      )}
+    </FormGroup>
+  );
+};
+
+PolicySelector.propTypes = {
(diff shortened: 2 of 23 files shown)


lavocatt:
Depends on https://github.com/osbuild/image-builder/pull/848

regexowl:
Great work @lavocatt!!

Just a quick note so I don't forget about it - we shortly talked about disabling the option to remove a partition/package in today's sync with the UX. @lucasgarfield please correct me if I remembered it wrong, but the user should be able to remove anything that was added by the policy. 

This would result in some kind of lowered oscap ranking of the final image. What we could possibly do is check the mount points and packages against the oscap profile and show an alert that the changes will affect the rating if there's something amiss. 

This could go in in some kind of follow up polish PR. What do you think?

lucasgarfield:
> Great work @lavocatt!!
> 
> Just a quick note so I don't forget about it - we shortly talked about disabling the option to remove a partition/package in today's sync with the UX. @lucasgarfield please correct me if I remembered it wrong, but the user should be able to remove anything that was added by the policy.
> 
> This would result in some kind of lowered oscap ranking of the final image. What we could possibly do is check the mount points and packages against the oscap profile and show an alert that the changes will affect the rating if there's something amiss.
> 
> This could go in in some kind of follow up polish PR. What do you think?

Additional safeguards in a follow-up PR is a great idea, we can discuss further... maybe even after getting some customer feedback!

For this MVP though, we don't need to worry about safeguards... let's just assume that anyone using this feature in preview understands what they're doing.

lavocatt:
All tests are passing on my machine now! Thanks @regexowl for creating them together this morning :rocket: 
The PR is all ready to get reviewed.

regexowl:
Should we add an expandable with the name of the selected policy to the Review step? We can also add it in a follow up PR

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -0,0 +1,172 @@
+import React, { useState } from 'react';
```
regexowl, line 1: Should we add an expandable with the name of the selected policy to the Review step? We can also add it in a follow up PR

src/test/fixtures/composes.ts
```diff
@@ -275,8 +280,49 @@ export const mockComposes: ComposesResponseItem[] = [
...
+        {
+          architecture: 'x86_64',
+          image_type: 'aws',
+          upload_request: {
+            type: 'aws',
+            options: {
+              share_with_accounts: ['123123123123'],
+            },
+          },
+        },
```
regexowl, line 299: As we need this compose for testing the Recreate functionality it might be nice to change it to any of the aws.s3 targets so we get the Recreate image instead of Launch for easier manual testing.

Screenshot from 2023-10-05 12-31-01


But this also feels like a very opinionated nit pick so feel free to ignore it :sweat_smile:

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -0,0 +1,173 @@
...
+    flag === 'image-builder.enable-content-sources'
+      ? mockContentSourcesEnabled
+      : true
+  ),
+}));
+
+beforeAll(() => {
+  // scrollTo is not defined in jsdom
+  window.HTMLElement.prototype.scrollTo = function () {};
+  mockContentSourcesEnabled = true;
```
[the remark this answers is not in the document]
regexowl: By "this" I mean `mockContentSourcesEnabled = true;` :sweat_smile:

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -0,0 +1,173 @@
...
+});
+
+afterEach(() => {
+  jest.clearAllMocks();
+  mockContentSourcesEnabled = true;
```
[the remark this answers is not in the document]
regexowl: The same here, by "this" I mean `mockContentSourcesEnabled = true;`

regexowl:
Also as there were some changes to the imageBuilderApi schema, there are two newly introduced errors. 

image
This one is easily fixable by adding Fedora 40 to the Release constant in `Release.tsx`

image
And this one could be silenced by an addition of a default value or a type guard. We don't really need to handle `else` cases there as the check is done before calling `parseGcpSharedWith()`. The goal is just to make TypeScript happy.

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -52,6 +54,42 @@ const FileSystemConfiguration = ({ ...props }) => {
...
+          fstype: 'xfs',
+          size: fs.min_size / 1024 / 1024, // the unit from the customizations are in bytes
+          unit: UNIT_MIB, // and using MIB seems sensible here instead
+        });
+      }
+      setRows(newRows);
+      setItemOrder(newRows.map((row) => row.id));
+      change('file-system-config-radio', 'manual');
+    }
+  }, [customizations, isSuccess]);
```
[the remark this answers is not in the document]
lavocatt: I thought I fixed that one :/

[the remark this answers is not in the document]
lavocatt: Does it happen all of the time ?

lavocatt:
Awesome review @regexowl ! I'm taking care of all of that ASAP.

src/Components/CreateImageWizard/formComponents/Oscap.js
lavocatt: I think that this is a good idea! I'll add a commit for that.

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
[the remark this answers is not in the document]
lavocatt: That's because the packages are stored in a set :smile: But hopefully this one is fixed!

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -0,0 +1,173 @@
...
+    });
+    await waitFor(() => expect(createImageButton).toBeEnabled());
+
+    // check that the FSC contains a /tmp partition
+    await clickBack();
+    await user.click(
+      await screen.findByRole('button', {
+        name: /file system configuration/i,
+      })
+    );
```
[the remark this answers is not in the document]
lavocatt: Also, we can use `within` to do that, and I think that would be the proper way to achieve that.

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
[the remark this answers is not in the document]
lavocatt: @regexowl I think it's all good now, can you give it a new test please ?

lavocatt:
For now the recreate function is a bit broken on stage. The reason why: Image builder doesn't send back the `oscap` object in the compose request. (The `oscap` object is well sent in the first place, so that should be easily fixable) I'll make a PR to change that on IB soon.

src/test/fixtures/composes.ts
```diff
@@ -645,6 +689,52 @@ export const mockStatus = (composeId: string): ComposeStatus => {
         ],
       },
     },
+    '1679d95b-8f1d-4982-8c53-8c2afa4ab04c': {
+      image_status: {
+        status: 'success',
+        upload_status: {
+          options: {
+            ami: 'ami-0217b81d9be50e44b',
+            region: 'us-east-1',
```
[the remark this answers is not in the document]
lavocatt: Ooops :sweat_smile:

src/test/fixtures/composes.ts
```diff
@@ -645,6 +689,52 @@ export const mockStatus = (composeId: string): ComposeStatus => {
...
+        distribution: RHEL_8,
+        image_requests: [
+          {
+            architecture: 'x86_64',
+            image_type: 'aws',
+            upload_request: {
+              type: 'aws',
+              options: {
+                share_with_accounts: ['123123123123'],
+              },
```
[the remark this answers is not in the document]
lavocatt: Thanks for that!!

src/test/fixtures/composes.ts
```diff
@@ -645,6 +689,49 @@ export const mockStatus = (composeId: string): ComposeStatus => {
...
+        distribution: RHEL_8,
+        image_requests: [
+          {
+            architecture: 'x86_64',
+            image_type: 'aws',
+            upload_request: {
+              type: 'aws.s3',
+              options: {},
+            },
+          },
```
regexowl, line 713: Last one, I promise :no_mouth: 

```suggestion
          {
            architecture: 'x86_64',
            image_type: 'guest-image',
            upload_request: {
              type: 'aws.s3',
              options: {},
            },
          },
```

lavocatt: ahahah :smile: I'm glad you catching them all!

regexowl (commented):
The changes look great! Found just one teeny tiny nitpick. The filtering works great when tested via msw :magic_wand:

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -0,0 +1,202 @@
...
+  // policies allowed by Compliance and the policies enabled in IB.
+  const {
+    data: dataCompliance,
+    isFetching: isFetchingCompliance,
+    isSuccess: isSuccessCompliance,
+    isError: isErrorCompliance,
+    refetch: refetchCompliance,
+  } = useListProfilesQuery({ limit: 100 }, { skip: !showSelector });
+
+  // Make the IB endpoint fetched only when the Compliance data as arrived,
```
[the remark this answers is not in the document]
lavocatt: This typo got removed, but thanks for catching it!

src/Components/CreateImageWizard/steps/fileSystemConfiguration.js
```diff
@@ -111,7 +125,10 @@ const fileSystemConfigurationStep = {
         </TextContent>
       ),
       condition: {
-        or: [{ when: 'file-system-config-radio', is: 'automatic' }],
+        or: [
+          { when: 'file-system-config-radio', is: 'automatic' },
+          { when: 'oscap-policy', is: undefined },
+        ],
```
regexowl, line 131: It seems that if there are no fs customizations in the policy, the automatic partitioning text shows:

"xccdf_org.ssgproject.content_profile_anssi_bp28_minimal"
Screenshot from 2023-10-06 10-55-37

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -52,6 +54,45 @@ const FileSystemConfiguration = ({ ...props }) => {
...
+      // Don't override the user's data if they made customizations
+      skip: hasNoOscapPolicy || hasCustomizations,
+    }
+  );
+
+  useEffect(() => {
+    if (hasCustomizations || rows.length > 1) {
+      return;
+    }
+    if (customizations && customizations.filesystem && isSuccess) {
```
regexowl, line 79: We get duplicated `/` mount point when it's a part of fs customizations:

"xccdf_org.ssgproject.content_profile_cis_server_l1"
Screenshot from 2023-10-06 10-50-25

The question is which one should we be keeping? Also should there be some kind of size check on change when a minimal size is set?

lavocatt:
So we've decided not to use the Compliance endpoint and to avoid any role base control on the frontend. The reason why:
* the V1 endpoint is requiring a user to have at least a `host` registered
* the v1 endpoint is gonna be replace by a V2 in the near future (see https://issues.redhat.com/browse/RHINENG-1973 and https://issues.redhat.com/browse/RHINENG-244)

We will have to see how it plays out to integrate the V2 endpoint. This endpoint is planed to have a requirement on Images or Openshift clusters. Which means that a user having a ZeroState in both these apps would not be able to create an image with OpenSCAP customization for their first image.

src/Components/CreateImageWizard/formComponents/ReviewStep.js
```diff
@@ -169,6 +173,17 @@ const ReviewStep = () => {
           <ImageDetailsList />
         </ExpandableSection>
       )}
+      {getState()?.values?.['oscap-policy'] && (
```
regexowl, line 176: Not sure if the problem is clear from the gif, it's a tad bit laggy, sorry about that.

review

When choosing a policy and clicking through to the Review step, everything is set up nicely. But if we use the navigation on the left to go back to OpenSCAP Compliance and change a policy the navigation on the left doesn't get disabled and allows us to go directly to the Review step again.

As the FSC and Packages are populated with the customizations on their own steps, those does not get updated and stay empty.

regexowl:
Bad news - when I go through the Wizard with a policy that does not duplicate the `/` mount point the image starts building just fine. But if the mount point is duplicated, it seems that the build does not start at all? :dizzy_face:

src/Components/CreateImageWizard/formComponents/ReviewStep.js
lavocatt: That is bad! Thanks for finding it!

lavocatt: Can we prevent the user to click on the review step up until FSC and packages are setup ?

regexowl: Not sure. I mean the navigation gets disabled when the user changes the target, so it should be somehow doable. But as far as I can remember we don't really disable on anything else besides when the target step gets added :thinking:

src/Components/CreateImageWizard/steps/fileSystemConfiguration.js
regexowl: The automatic partitioning text seems to be there even if I leave the OpenSCAP Compliance set as Do not add a policy and then choose manual partitioning.

src/Components/CreateImageWizard/formComponents/ReviewStep.js
lavocatt: That would be the easier way I think.

lavocatt: Right now It's the FSC and Packages steps that are populating themselves by querying the data from the API if a policy has been selected. Which means that if they aren't executed, we can't get customizations in.
One way to do it, would be for the OpenSCAP step to populate the form state when the user clicks.
The other way to do it is to force the user clicking through the FSC and Packages steps...

The easier is the second one, the better is the first one I guess :sweat_smile: 
I'll figure a way to do the second option after lunch.

But I would be glad to have your thoughts @regexowl and @lucasgarfield

regexowl: I would vote for forcing the user to go through the FSC and Packages again as the values in the steps change with a newly selected policy. The user should be aware of what changes happened and review them before continuing I think  :shrug:

lavocatt: Ohh, btw, note that the first time you are selecting a policy you can't click on review immediately. Because everything is grayed out. But then it's only after the first time you got to the review step that you are allowed to shortcut the FSC and packages steps.

lavocatt: So I guess It's not that much blocking

lavocatt: I guess, the correct thing to do is: if a user goes back to the Compliance step, then we reinitialize the navigation on the left. What's your opinion about this one ?

regexowl: Yup, exactly. The steps are incrementally enabled. The problem only arises when someone wants to revisit the compliance step only and doesn't care about the customizations that are loaded.

Maybe reinitialize on change of the compliance? Could work.

lavocatt: Lemme try that :)

lavocatt:
Last thing to tackle: the left menu bar

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
lavocatt: This is now patched up. Thanks for finding this!

src/Components/CreateImageWizard/steps/fileSystemConfiguration.js
lavocatt: I got this fixed.

regexowl (approved):
:star::star::star::star::star::star::star::star::star::star::star::star::star:
Amazing work @lavocatt!!! 
:star::star::star::star::star::star::star::star::star::star::star::star::star:

lavocatt:
> ⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐ Amazing work @lavocatt!!! ⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐⭐

It's a team effort :mechanical_arm:

lucasgarfield:
Wooo! I love it! Thank you so much to both of you!

PR #1379, opened 2023-10-10
API: Eliminate race condition in Image Builder API slice

lucasgarfield:
The images table uses the useGetComposesQuery hook to fetch composes and implement pagination. When this query is used, args are provided for the offset and limit and a 'Compose' tag is provided for the query.

When a mutation is triggered (causing a POST request to be sent to the `/compose` end point), the 'Compose' tag is invalidated which clears all cached data from useGetComposesQuery hooks, which in turn causes the table to refetch compose information.

If invalidating the `Compose` tag causes a refetch before the new compose is available in image-builder, the result does not contain the new compose and the table is not updated to include it.

This commit eliminates the race condition by waiting for the query to be fulfilled before invalidating tags (and therefore before refetching the data).

All of the above applies equally to the `cloneCompose` mutation, and its race condition has also been eliminated.

This commit is loosely inspired by the RTK Query docs section on pessimistic updates:
https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates#pessimistic-updates

Typescript complains about the type of the tags. It does not recognize the tag types that are defined in the same enhanceEndpoint() function. For now, we simply ignore the Typescript errors. There is some discussion here: https://github.com/reduxjs/redux-toolkit/issues/1510

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -593,20 +593,20 @@ const CreateImageWizard = () => {
   return (
     <ImageCreator
       onClose={handleClose}
-      onSubmit={({ values, setIsSaving }) => {
+      onSubmit={async ({ values, setIsSaving }) => {
         setIsSaving(true);
         const requests = onSave(values);
-        navigate(resolveRelPath(''));
         // https://redux-toolkit.js.org/rtk-query/usage/mutations#frequently-used-mutation-hook-return-values
         // If you want to immediately access the result of a mutation, you need to chain `.unwrap()`
         // if you actually want the payload or to catch the error.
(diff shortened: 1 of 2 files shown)


regexowl (approved):
Looks good and works well :+1:

lavocatt (dismissed):
Amazing!! :rocket:

lucasgarfield:
Want to check a few more things before merging, converting to draft.

lucasgarfield:
I had to move the call to `navigate()` which causes the wizard to reset to step 1 for an instant before the app navigates to the table. This will pass CI and the race condition should be taken care of :crossed_fingers: but I don't know how to fix the bug in the wizard step state, it seems like a Data Driven Forms problem...

ddfmynemesis

jrusz (approved):
LGTM.

PR #1380, opened 2023-10-10
API: Move notification dispatch to Image Builder API slice

lucasgarfield:
This commit moves the notification dispatching for creating composes and
clones into a more sensible location – the Image Builder API slice.

It is more sensible because it separates the logic of the React
component (the wizard or share images modal) from the logic of handling
the request life cycle (which is now handled entirely in the slice).

There is a subtle but significant change – a new request will be
dispatched for every request. This is the correct way to do things as it
is possible that some requests succeed, and that others fail. Insights
causes the notifications to stack on top of each other neatly, so there
is no UI problem.

To facilitate this, we also need to use use Promise.allSettled instead
of Promise.all.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -1,9 +1,7 @@
 import React from 'react';
 
 import componentTypes from '@data-driven-forms/react-form-renderer/component-types';
-import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
 import { useFlag } from '@unleash/proxy-client-react';
-import { useDispatch, useStore } from 'react-redux';
 import { useNavigate, useParams } from 'react-router-dom';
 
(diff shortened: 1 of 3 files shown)


lucasgarfield:
Behold!
bearmetal

lavocatt:
Looks like this PR depends on https://github.com/RedHatInsights/image-builder-frontend/pull/1379 (writing it so we don't merge this one before the other)

lavocatt (approved):
This really makes a lot of sense! Very good idea :rocket:

lavocatt (changes requested):
Falsely requesting changes so we don't merge this one accidentally before #1379

lavocatt (approved):
#1379 got merged, let's have this one in too!

regexowl (approved):
LGTM!

lucasgarfield:
Discussed this PR with @jrusz, IQE test failure not related to this PR, the important bits all pass. Merging.

PR #1386, opened 2023-10-12
fsc: fix the oscap fs customizations

lavocatt:
There was a bug preventing all the partitions required by an oscap policy to show up in the fsc step. That bug was because the FSC customizations were only accepted if they matched exactly the fix set of base partitions directories.

So now there's a bit of string manipulation to avoid that but it's all patched up.

fixes https://github.com/RedHatInsights/image-builder-frontend/issues/1385

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
+++ b/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
@@ -83,9 +83,13 @@ const FileSystemConfiguration = ({ ...props }) => {
     if (customizations && customizations.filesystem && isSuccess) {
       const newRows = rows;
 
-      // Only keep the required partitions that have a valid mount points
+      // The filesystem customizations required by the OpenSCAP profile may
+      // contain some unsupported mountpoints. They need to be filtered out to
+      // prevent undefined behaviour in the frontend and the backend
       const fss = customizations.filesystem.filter((fs) =>
-        MountPointValidPrefixes.includes(fs.mountpoint)
+        MountPointValidPrefixes.includes(
+          '/'.concat(fs.mountpoint.split('/')[1])
+        )
       );
 
       // And add them all to the list.


regexowl (approved):
Works great!

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -83,9 +83,12 @@ const FileSystemConfiguration = ({ ...props }) => {
     if (customizations && customizations.filesystem && isSuccess) {
       const newRows = rows;
 
-      // Only keep the required partitions that have a valid mount points
+      // Only keep the required partitions that have a mountpoint which root
+      // directory belongs to the supported ones.
```
croissanne, line 87: Had to read the comment a few times before understanding it, maybe "that have a mountpoint _of which the_ root directory belongs to the supported ones?"

kingsleyzissou: Maybe something simpler?

Filter out unsupported mountpoints, keeping the root mountpoint

lavocatt: Yeah that comment is wrong :sweat_smile: Thanks for catching that. What about:

`The filesystems customizations required by the OpenSCAP profile may contain some incompatible mountpoints. They need to be filtered out to avoid any UX rendering issues.`

kingsleyzissou: Lol spending a lot of time discussing a comment 

`The filesystem customizations required by the OpenSCAP profile may contain some unsupported mountpoints. They need to be filtered out to prevent undefined behaviour in the frontend and the backend`

Since that could cause failed composes if submitted to the backend too.

lavocatt: I love it! Thanks :smile:

lavocatt: > Lol spending a lot of time discussing a comment

Yeah that's true. Good comments are hard to write though.

kingsleyzissou: I mean it would be equally valid to leave the comment out here since it is pretty clear what the code is doing :wink:

PR #1387, opened 2023-10-13
wizard: allow the creation of aarch64 images

lavocatt:
This commit extends the supported architectures to aarch64. In the image output step (the first one of the wizard) the user is now faced with a new select choices to pickup the architecture they want to build.

Now the set of compatible targets to build is dynamically loaded from the backend and the UX changes what's accessible on the fly depending on what the user has selected.

HMS-1135: done

- [x] Manually tested
- [x] Unit tests

[Screencast from 2023-10-16 12-59-58.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/f200004d-f911-4576-91a0-1e2c70312d82)

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -7,6 +7,8 @@ export const RHEL_8 = 'rhel-88';
 export const RHEL_9 = 'rhel-92';
 export const CENTOS_8 = 'centos-8';
 export const CENTOS_9 = 'centos-9';
+export const X86_64 = 'x86_64';
+export const AARCH64 = 'aarch64';
 
 export const UNIT_KIB = 1024 ** 1;
 export const UNIT_MIB = 1024 ** 2;
@@ -18,6 +20,8 @@ export const RELEASES = new Map([
   [CENTOS_9, 'CentOS Stream 9'],
   [CENTOS_8, 'CentOS Stream 8'],
 ]);
+
+export const ARCHS = [X86_64, AARCH64];
 
 export const DEFAULT_AWS_REGION = 'us-east-1';
 
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.content.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.content.test.js
@@ -511,7 +511,9 @@ describe('Step Packages', () => {
       ));
 
       // select aws as upload destination
-      user.click(screen.getByTestId('upload-aws'));
+      await waitFor(
+        async () => await user.click(await screen.findByTestId('upload-aws'))
+      );
       await clickNext();
 
       // aws step
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
@@ -193,9 +193,9 @@ describe('Step Image output', () => {
   test('expect only RHEL releases before expansion', async () => {
     await setUp();
 
-    const releaseMenu = screen.getByRole('button', {
+    const releaseMenu = screen.getAllByRole('button', {
       name: /options menu/i,
-    });
+    })[0];
     await user.click(releaseMenu);
 
     await screen.findByRole('option', {
@@ -214,9 +214,9 @@ describe('Step Image output', () => {
   test('expect all releases after expansion', async () => {
     await setUp();
 
-    const releaseMenu = screen.getByRole('button', {
+    const releaseMenu = screen.getAllByRole('button', {
       name: /options menu/i,
-    });
+    })[0];
     await user.click(releaseMenu);
 
     const showOptionsButton = screen.getByRole('button', {
@@ -245,9 +245,9 @@ describe('Step Image output', () => {
   test('CentOS acknowledgement appears', async () => {
     await setUp();
 
-    const releaseMenu = screen.getByRole('button', {
+    const releaseMenu = screen.getAllByRole('button', {
       name: /options menu/i,
-    });
+    })[0];
     await user.click(releaseMenu);
 
     const showOptionsButton = screen.getByRole('button', {
@@ -915,9 +915,9 @@ describe('Step Review', () => {
   const setUpCentOS = async () => {
     ({ router } = renderCustomRoutesWithReduxRouter('imagewizard', {}, routes));
 
-    const releaseMenu = screen.getByRole('button', {
+    const releaseMenu = screen.getAllByRole('button', {
       name: /options menu/i,
-    });
+    })[0];
     await user.click(releaseMenu);
 
     const showOptionsButton = screen.getByRole('button', {
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.test.js
@@ -276,7 +276,9 @@ describe('Step Upload to AWS', () => {
     ));
 
     // select aws as upload destination
-    user.click(screen.getByTestId('upload-aws'));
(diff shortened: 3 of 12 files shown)


lavocatt:
Tests are fixed, I will add a new one too to check that changing an arch has the correct behavior

lavocatt:
@jrusz I'm pinging you here because this PR may break your integration testing. It changes the target selection in the first step of the wizard.

jrusz:
> @jrusz I'm pinging you here because this PR may break your integration testing. It changes the target selection in the first step of the wizard.

Thanks! But looks like this is just adding one more option and not changing/removing anything so it should be all good. Let's see if pr_check agrees :)

src/constants.js
```diff
@@ -19,6 +21,11 @@ export const RELEASES = new Map([
   [CENTOS_8, 'CentOS Stream 8'],
 ]);
 
+export const ARCHS = new Map([
+  [X86_64, 'x86'],
```
ondrejbudai, line 25:
```suggestion
  [X86_64, 'x86_64'],
```

I suggest being explicit here. :)

lavocatt: `x86` is conform to Andy's mock, see https://www.sketch.com/s/d7aa6d29-fca0-4283-a846-09cc5fd10612/a/My1W1mm

lavocatt:
> Thanks! But looks like this is just adding one more option and not changing/removing anything so it should be all good. Let's see if pr_check agrees :)

I was worried because it adds up a loading Spinner to select the targets. But it looks like the IQE are robust to that!

src/constants.js
lucasgarfield: Totally makes sense to build it to the mocks, but one thing to keep in mind is that the designers lean on us as subject matter experts... Design is a team effort. :wink: 

Let's change to `x86_64`. `x86_64` also appears in places like the request, which users will encounter if they are using the API or choose to download the request .json file, so this way it will be consistent.

lavocatt:
Creating images works, but launching theme is waiting for a couple of changes in the provisionning app.
- https://github.com/RHEnVision/provisioning-frontend/pull/351
- (another one to be created for the backend)

src/constants.js
lavocatt: Done.

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -141,7 +141,9 @@ describe('Step Image output', () => {
     ));
 
     // select aws as upload destination
-    await user.click(await screen.findByTestId('upload-aws'));
+    await waitFor(
```
regexowl, line 144: `waitFor()` blocks can be also completely removed from this test. I'd personally prefer that as I feel they only clutter up the async logic. In this case the removal would be a bit more involved as there are other user actions that would need to be awaited.

As this is not the main focus of this PR, I'd say let's go with the `waitFor()` blocks for now and I'd later clean them up a bit. Would that work?

src/test/Components/CreateImageWizard/formComponents/TargetEnvironment.test.js
```diff
@@ -0,0 +1,284 @@
+import React from 'react';
```
regexowl, line 1: Looks good and also brings up question about our testing strategy. Do we want to test components of a component individually? Do we want to test scenarios? Happy paths? Again, not really the main aim of this PR, but wanted to put this here as it might be a good idea to have a discussion about it in the not so distant future.

lavocatt: Thanks for bringing that up Klara! So I have an idea for our tests that I'm starting to experiment here.

I think that Ideally, unit tests and integration tests have to coexist together. So the tests we have now are relevant and are not to be replaced, but they can't test all the paths, it's just too much. Having tests that are focusing on smaller areas means that these tests can test more without exploding in complexity. And then the intersection of these tests is what is interesting and provides more confidence. Because if you test extensively two components A and B, and you get an integration test testing A+B, you don't need to test all the combination of integrating A and B together, their behavior is tested to be robust independently, so a few happy path for A+B should be enough.

For instance in this PR I wanted to test the new Arch64 functionality. I could have added a bunch of tests in the integration testing, like doing an end to end test with arch the same way we are doing it with x86. But it would have meant duplicating tests parts that aren't specific to the change of architecture.

I decided instead to rely on the fact that the wizard is tested already with x86. So I just needed to add specific testing for Arch somewhere, and the combination of the happy path for x86 + the combinatorics tests for arch and x86 would bring me confidence that the FSC step is robust.

The question is, do you like it? What's your opinion about it?

src/test/Components/CreateImageWizard/CreateImageWizard.content.test.js
```diff
@@ -111,7 +111,9 @@ describe('Step Packages', () => {
       ));
 
       // select aws as upload destination
-      await user.click(await screen.findByTestId('upload-aws'));
+      await waitFor(
+        async () => await user.click(await screen.findByTestId('upload-aws'))
+      );
```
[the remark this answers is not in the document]
lavocatt: Oooh interesting, so the thing is, the test wasn't working before I added this because there's a Spiner now while the targets are loading. And it made the test crash because the element wasn't immediately available. Lemme double check what I did just to make sure it makes sense :slightly_smiling_face:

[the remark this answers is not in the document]
lavocatt: Ok, I went a bit too hard replacing all of them. Some needs to be kept though. I'll do some cleaning. Thanks!!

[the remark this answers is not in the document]
lavocatt: @regexowl I cleaned just now all the ones I could without breaking the testsuite. Is it better in your opinion?

src/test/Components/CreateImageWizard/formComponents/TargetEnvironment.test.js
regexowl: I see! Completely agree that we can't possibly test every happy path in the integration tests as the complexity would truly be through the roof. I'm trying to think about what could we possibly test outside of the first step in this case. Nothing comes to mind as a lot of things around architecture are tucked in the background.

I do like the idea of more comprehensive tests for smaller parts and then few happy paths to check if everything fits together well! We can probably discuss the specifics about our testing strategy even more later, but this definitely sounds like a good approach to me :+1:

src/test/Components/CreateImageWizard/CreateImageWizard.content.test.js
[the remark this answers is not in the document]
regexowl: @lavocatt It is, thank you!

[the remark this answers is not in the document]
lavocatt: Cool! thanks, I'll resolve all the related issues.

ezr-ondrej:
> Creating images works, but launching theme is waiting for a couple of changes in the provisionning app.

Everything is merged on the Launch side, we should be ready on stage, it will go to prod next Tuesday.

lavocatt:
I've got this rebased on top of the PF5 update. Everything's fine from what I can tell and I'm looking forward to your reviews.

jrusz:
Hi, sorry I need to release the PF5 update to the test suite as well first before pr_check will pass on this, I'll retest once I have it.

jkozol (approved):
Looks good!

PR #1392, opened 2023-10-17
oscap: lock the wizard step behing a feature flag

lavocatt:
The oscap feature have been quite eagerly pushed forward and it seems we went a bit fast with it. To take more time and make sure all the stakeholders agree on the way forward the decision was taken to put this functionality behind a feature flag.

The two feature flags for both stage and prod are there:
* https://insights-stage.unleash.devshift.net/projects/default/features/image-builder.wizard.oscap.enabled (enabled on stage so we can test the feature)
* https://insights.unleash.devshift.net/projects/default/features/image-builder.wizard.oscap.enabled (disabled on prod)


src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -559,24 +563,30 @@ const CreateImageWizard = () => {
 
   const { isBeta, isProd } = useGetEnvironment();
 
+  // Only allow oscap to be used in Beta even if the flag says the feature is
+  // activated.
+  const oscapFeatureFlag =
+    useFlag('image-builder.wizard.oscap.enabled') && isBeta;
```
regexowl, line 569: The feature flag for stage works great! But there will be a small fix needed if we want to make the step available only in beta.

```suggestion
    useFlag('image-builder.wizard.oscap.enabled') && isBeta();
```

lavocatt: :smile: Good catch!!

lavocatt: Fixed!

regexowl (approved):
Works like magic, thanks!

PR #1393, opened 2023-10-17
oscap: rename policy to profile in the UX

lavocatt:
Policy and profiles seemed to be two interchangeable terms for us. But it turns out there might be some difference between the two. We've been asked to not call the Oscap policies policies in the UX. This commit does exactly this.


src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -87,8 +87,8 @@ const PolicySelector = ({ input, showSelector }) => {
         onClear={handleClear}
         selections={policy}
```
regexowl, line 88: Do we want to switch "policy" for "profile" in the names of variables as well? It will take more work, but if the change is here to stay it could be much less confusing in the long run, what do you think?

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -131,13 +131,13 @@ const AddPolicy = ({ input }) => {
   const [wantsPolicy, setWantsPolicy] = useState(oscapPolicy !== undefined);
   return (
     <>
-      <FormGroup label="Compliance policy">
+      <FormGroup label="Compliance profile">
```
croissanne, line 134: Wonder if we should name is 'oscap profile' just to be sure.

src/Components/CreateImageWizard/formComponents/Oscap.js
lavocatt: Yes, in fact it makes no sense that this is called policy. I'm going to change all of that

src/Components/CreateImageWizard/formComponents/Oscap.js
lavocatt: yup!

regexowl:
Looks good! :sparkles: There's just one small linter error.

lavocatt:
> Looks good! ✨ There's just one small linter error.

Thanks for checking on it this morning. It's updated now.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -477,7 +477,7 @@ const requestToState = (composeRequest, distroInfo, isProd, enableOscap) => {
 
     // oscap policy
```
regexowl, line 478: The last "policy" left behind. It sneaked back with the conflict.

lavocatt: All fixed now :slightly_smiling_face:

regexowl (approved):
Looks good to me! Thank you!

PR #1396, opened 2023-10-19
devDeps: Manually bump eslint-plugin-testing-library

regexowl:
This replaces https://github.com/RedHatInsights/image-builder-frontend/pull/1394


lavocatt (dismissed):
LGTM!

.eslintrc-typescript.yml
```diff
@@ -1,4 +1,10 @@
...
+extends: [
+    "plugin:@typescript-eslint/recommended",
+  ]
 
+rules:
+  "@typescript-eslint/ban-ts-comment": off
+  "@typescript-eslint/ban-types": off
+  "@typescript-eslint/no-explicit-any": warn
+  "@typescript-eslint/no-unused-vars": warn
+  explicit-module-boundary-types: off
```
lavocatt, line 10: Why do we have to change our linting rules for bumping the version?

regexowl: Great question! I think our previous setup didn't include all the recommended rules for linting TypeScript. With those added there were some parts of the code the rules did not really agree with. So those were switched off or replaced by warnings for the time being and we can look into getting rid of them later.

Or keeping them, for example the ban-ts-comment, it's not the best practice to rely on those, but as we are not abusing them, I see no point in strictly banning all of those.

lavocatt: all good to me!

regexowl:
Oof, this was no fun to rebase with all the deps changes from previous PRs. Hopefully everything should be fine now.

jkozol (approved):
This is great! Just to make sure I understand, some of these manual rules changes would ideally be removed in the future once we update the codebase to remove the warnings?

regexowl:
> This is great! Just to make sure I understand, some of these manual rules changes would ideally be removed in the future once we update the codebase to remove the warnings?

Exactly! This PR is just the first step. We should aim to remove all the warnings and errors (that are currently masquerading as warnings thanks to the manual rules). Once those are solved, removal of the manual overrides will follow and we should be left with the recommended set of rules.

PR #1397, opened 2023-10-23
wizard: prepare a dev env to create the new wizard

lavocatt:
Developing a new wizard will require many pull requests and some of them might even be worked on in parallel. To allow to develop all of that without interfering with the legacy wizard this commit introduces a new dev environment alongside an unleash flag when we later want to do the switch.

To use the new dev en use `npm run stage-beta:experimental` or `npm run stage-beta:msw+experimental` depending on what you want.

Then unleash flag to toggle in stage and prod is: `image-builder.new-wizard.enabled`

fixes: HMS-2859


regexowl (approved):
LGTM! \o/

lavocatt:
@jrusz IQE tests are stuck, can we admin merge the PR?

jrusz:
> @jrusz IQE tests are stuck, can we admin merge the PR?

It should be fixed now so unless it's urgent it should be green in < 1h and you can merge regularly :) ... They changed some navigation elements in consoledot again without telling anyone so we had to make some adjustments.

PR #1401, opened 2023-10-26
pr_check: change marker and increase timeout

jrusz:
I have set up a new marker "fe_pr_check" to select tests for the pr_check here. I'm also increasing the timeout of the CJI pod.


mgold1234 (approved):
/lgtm

croissanne (approved):

PR #1403, opened 2023-10-31
Update terraform SHA

jrusz:
We need a newer centos-8 image in our CI.


ochosi (approved):
LGTM, thanks for the update!

PR #1404, opened 2023-11-01
ImagesTable: Fix padding of disabled download links

regexowl:
This fixes an uneven padding for disabled "Download (....)" links.

### DIFF
--- a/src/Components/ImagesTable/Instance.tsx
+++ b/src/Components/ImagesTable/Instance.tsx
@@ -211,7 +211,7 @@ export const AwsS3Instance = ({
 
   if (status !== 'success') {
     return (
-      <Button isDisabled variant="link" isInline>
+      <Button component="a" isDisabled variant="link" isInline>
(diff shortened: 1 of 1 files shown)


regexowl:
As this was really bothering me... there was a weird padding going on with this specific button/link:

Screenshot from 2023-11-01 17-03-30

Which gets fixed by this PR as per [PF5 docs on using links as buttons](https://www.patternfly.org/components/button/#links-as-buttons)

Screenshot from 2023-11-01 17-07-49

lucasgarfield:
Nice catch... do we need to do this to any other buttons that are external links?

lucasgarfield (approved):

regexowl:
> Nice catch... do we need to do this to any other buttons that are external links?

Maybe. I was quickly checking after opening the PR and did not find anything obvious, but there are some button links with the `component` missing that have an extra styling which might be working around the issue. I plan to go through all the buttons tomorrow.

PR #1408, opened 2023-11-02
workflows: Add a workflow with dev checks

regexowl:
As Travis CI is getting removed from RedHatInsights this should replace it in what we used to use Travis for.

This workflow is triggered on pull requests to main and runs clean install of dependencies, lint check and the unit test suite.

The main motivation of this workflow being separated from the IQE tests is to allow to quickly check if the basic tests pass during development without the need to wait for the IQE tests to finish their run.

### DIFF
--- a/.github/workflows/dev-checks.yml
+++ b/.github/workflows/dev-checks.yml
@@ -1,0 +1,29 @@
+name: Development checks
+
+on:
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16.x, 18.x]
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+    - name: Install dependencies
+      run: npm ci
+    - name: Run build
+      run: npm run build
+    - name: Run lint check
+      run: npm run lint
+    - name: Run unit tests
+      run: npm run test


.github/workflows/dev-checks.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16.x, 18.x]
```
lucasgarfield, line 13: What about 17.x and 19.x? I'm not sure what the best practice is here, couldn't we also just use the version that will actually be run on the server? That's what we really care about... not sure what the best practice is here, but fewer job runs might be faster or will at least be more efficient.

croissanne: So the build is done with 1.16 currently (https://github.com/RedHatInsights/image-builder-frontend/blob/9be2ceaba7d402c8cac79fd9d55fac465e974551/build_deploy.sh#L12), and the releaser container has 1.18. I think it's good to test those two, but more can't hurt of course.

croissanne (dismissed):

.github/workflows/dev-checks.yml
lucasgarfield: Per @croissanne "Ok so the releaser container has 1.18, the frontend-build-container has 1.16 by default, but changeable by setting NODE_BUILD_VERSION, which is currently set to 1.16 for us. so testing both 1.16 and 1.18 will be fine"

.github/workflows/dev-checks.yml
```diff
@@ -0,0 +1,27 @@
...
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+    - name: Install dependencies
+      run: npm ci
+    - name: Run lint check
+      run: npm run lint
+    - name: Run unit tests
+      run: npm run test
```
croissanne, line 27: shouldn't we just run `    "verify": "npm-run-all build lint test"` ?

lucasgarfield (approved):
Love it!

PR #1413, opened 2023-11-05
fix some react-hooks/exhaustive-deps lint warning message

mgold1234:
this commit fixes some react-hooks/exhaustive-deps linting warning message

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
+++ b/src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
@@ -67,7 +67,7 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
     if (getState()?.values?.[input.name]) {
       setEnvironment(getState().values[input.name]);
     }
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
```diff
@@ -66,7 +66,7 @@ const TargetEnvironment = ({ label, isRequired, ...props }) => {
     if (getState()?.values?.[input.name]) {
       setEnvironment(getState().values[input.name]);
     }
-  }, []);
+  }, [getState, input.name]);
```
jkozol, line 69: I think you can remove the useEffect here and just keep the conditional.

mgold1234: without this useEffect,  a lot of tests fail

regexowl (approved):
Looks good to me, thank you! :sparkles:

PR #1421, opened 2023-11-08
Wizard: add all/selected toggle buttons to custom repository step

### LINKED ISSUE #1201: Add All/Selected toggle on the custom repositories step #
Easily view the repositories that are currently selected while clicking through the pages of repos.

mgold1234:
This commit resolves #1201. 
It adds a toggle to the repositories step that allows users to toggle between all and selected repositories.
Screenshot 2023-11-08 at 14 48 12

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Repositories.js
+++ b/src/Components/CreateImageWizard/formComponents/Repositories.js
@@ -21,6 +21,8 @@ import {
   ToolbarItem,
   EmptyStateHeader,
   EmptyStateFooter,
+  ToggleGroup,
+  ToggleGroupItem,
 } from '@patternfly/react-core';
 import {
   Dropdown,
@@ -195,6 +197,7 @@ const Repositories = (props) => {
   const [filterValue, setFilterValue] = useState('');
   const [perPage, setPerPage] = useState(10);
   const [page, setPage] = useState(1);
+  const [toggleSelected, setToggleSelected] = useState('toggle-group-all');
   const [selected, setSelected] = useState(
     getState()?.values?.['payload-repositories']
       ? getState().values['payload-repositories'].map((repo) => repo.baseurl)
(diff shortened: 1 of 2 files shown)


lucasgarfield:
This is looking really nice! 

I tested on my machine with multiple repo pages and there is a bug. If you select some repos on the first page of the table, then some pages on the second page of the table, and then filtered for 'selected' _while on the second page of the table_ then the repos do not display correctly.

This would be a good chance for you to write a test that checks for this bug and then fix it Test Driven Development style. :smiley:

mgold1234:
@lucasgarfield if user filter 1 repo from first page, and 1 from the second page, if user go back to first page and press on 'Selected' 2 checked repo will display. 
so the filter wroks fine, maybe I should enable the selected button only from the first page?

regexowl:
The tests are currently failing on a partial match in `src/test/Components/CreateImageWizard/CreateImageWizard.content.test.js`

If you check out test "'correct number of repositories is fetched'" you can replace the partial match with a full match, the tests should pass then.

```javascript
  test('correct number of repositories is fetched', async () => {
    await setUp();

    await user.click(
      await screen.findByRole('button', {
        name: 'Select',
      })
    );

    screen.getByText(/select all \(1015 items\)/i);
  });
```

Also looking at the tests, it looks like we were previously doing all testing of the custom repositories step in `CreateImageWizard.content.test.js`, would it be possible to move the new tests there so the step is tested all in one place?

src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -269,14 +278,23 @@ const Repositories = (props) => {
   };
 
   const filteredRepositoryURLs = useMemo(() => {
```
regexowl, line 280: We could remove a tiny bit of duplicity here by taking the `repoUrls` as a base and map it based on the clicked toggle. Something like this:

```javascript
  const filteredRepositoryURLs = useMemo(() => {
    const repoUrls = Object.values(repositories).filter((repo) =>
      repo.name.toLowerCase().includes(filterValue.toLowerCase())
    );
    if (toggleSelected === 'toggle-group-all') {
      return repoUrls.map((repo) => repo.url);
    } else if (toggleSelected === 'toggle-group-selected') {
      return repoUrls
        .filter((repo) => isRepoSelected(repo.url))
        .map((repo) => repo.url);
    }
  }, [filterValue, repositories, toggleSelected]);
```

What do you think?

mgold1234: great idea

regexowl (approved):
Works great, thank you! And also thank you for the tests!

PR #1426, opened 2023-11-10
api/Wizard: Update ImageBuilder api schema and add ClientId to a compose request

regexowl:
Depends on https://github.com/osbuild/image-builder/pull/897

This updates the api schema and adds a ClientId value to the request sent to the api.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -121,6 +121,7 @@ const onSave = (values) => {
         },
       ],
       customizations,
+      client_id: 'ui',
     };
     requests.push(request);
   }
@@ -160,6 +161,7 @@ const onSave = (values) => {
         },
       ],
       customizations,
+      client_id: 'ui',
     };
 
     if (share !== '') {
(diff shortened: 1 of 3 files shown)


regexowl:
Ready for review! But also blocked until https://github.com/osbuild/image-builder/pull/897 is merged.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -121,6 +121,7 @@ const onSave = (values) => {
         },
       ],
       customizations,
+      client_id: 'ui',
```
lavocatt, line 124: Why is the value always ui?

regexowl: The result is to be able to differentiate between requests created via api and requests created via ui, Since all of these requests are coming from ui, the client id for all of them is the same.

lavocatt: thanks!

lavocatt (approved):
LGTM!

jrusz:
> /retest

The images are not showing up in the table without the backend change so we'll have to merge that one first for this to work. I am running it locally though and it seems to be fine.

PR #1429, opened 2023-11-13
Wizard/MountPoint: hotfix for updated mountpoint policy

thozza:
The updated mountpoint policy merged in osbuild-composer upstream does not allow sub-directories of `/usr` to be placed on a separate partition. There are other changes, but this fix prevents users from making an obvious error during image creation, until the updated policy is reflected in the UI.

Related to https://issues.redhat.com/browse/COMPOSER-2030

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/MountPoint.js
+++ b/src/Components/CreateImageWizard/formComponents/MountPoint.js
@@ -19,7 +19,6 @@ export const MountPointValidPrefixes = [
   '/srv',
   '/tmp',
   '/usr',
-  '/usr/local',
   '/var',
 ];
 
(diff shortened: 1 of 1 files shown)


lucasgarfield:
Looking good! Try running `npm run lint:js:fix`.

regexowl (approved):
Ah, so no sub-directories for `/usr`? Got it! Thanks a lot @thozza! :sparkles:

lavocatt (approved):
LGTM!

thozza:
> Hmm, but also - what should we do in a case of previously created images with `/usr` sub-directories that can be recreated?
> 
> Are the sub-directories still supported in backend and will be phased out or is there a sudden cut and there shouldn't be an option to Recreate with `/usr` sub-directories at all?

`/usr` sub-directories are no longer supported by `osbuild-composer` version on `main`. The reason is that we can't reasonably support this. While we could allow arbitrary sub-dir mounts there, we can't do it for the common sub-dirs. Previously, one could place e.g. `/ust/bin` or `/usr/lib` on a separate partition, which would result in unbootable image.

I'm sorry, but I don't know how was the original set of prefixes and allowed mountpoints picked. This time, I (with help of others) did an assessment of what it technically possible and supportable WRT custom mountpoints. The decision was to not allow `/usr` sub-dirs at all.

This PR is just my attempt to at least prevent users from doing an obvious error in the wizard, until the whole MountPoint step is reworked to the new policy (basically move from allow list to deny list).

Moreover, the UI still allows `/var/run` and `/var/lock` to be placed on a separate partition, which is also no longer allowed by the backend. However, there is no existing code in front-end which selectively disallows specific paths, which I would just modify. 

I admit that this change is not nice and introduces backward incompatibility, unfortunately I do not have a good answer to how to handle this gracefully. :disappointed:

regexowl:
Ah, got it! Thank you for all the information! 

This PR is a great first step so I think we can merge it and then issue a follow up PR which will ensure there are no `/usr` sub-directories on Recreate action. And will look at the `/var` paths as well. What do you think @lucasgarfield?

Thanks a lot for putting this together @thozza!

PR #1431, opened 2023-11-13
HMS-2861: add the TargetEnvironment step to the wizardV2

lavocatt:
In this PR we add the support for the second step of the wizard: The TargetEnvironment.

3 sub steps are added:
- AWS
- GCP
- Azure

The behavior is identical to the DDF wizard.

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -20,6 +20,9 @@ import ReviewStep from './steps/Review/ReviewStep';
 import AWSTarget, {
   validateAWSAccountID,
 } from './steps/TargetEnvironment/AWS/AWSTarget';
+import AzureTarget, {
+  validateAzureId,
+} from './steps/TargetEnvironment/Azure/AzureTarget';
 import GCPTarget, {
   GCPAccountTypes,
   validateGCPData,
@@ -147,6 +150,29 @@ const CreateImageWizard = () => {
     useState<GCPAccountTypes>('googleAccount');
   const [gcpAccountEmail, setGcpAccountEmail] = useState('');
   const [gcpDomain, setGcpDomain] = useState('');
+  // Azure
+  const [azureManual, setAzureManual] = useState(false);
+  const [azureSource, setAzureSource] = useState<[number, string]>([0, '']);
+  const {
+    tenantId,
+    subscriptionId,
+    resourceGroups,
+    isError: azureIDError,
+  } = useGetAccountData(azureSource[0], 'azure');
+  const [azureTenantId, setAzureTenantId] = useState(tenantId);
+  const [azureSubId, setAzureSubId] = useState(subscriptionId);
+  const [azureResourceGroup, setAzureResourceGroup] = useState('');
+  const [prevTenantId, setPrevTenantId] = useState(tenantId);
+  const [prevSubscriptionId, setPrevSubscriptionId] = useState(subscriptionId);
+  if (prevTenantId !== tenantId || prevSubscriptionId !== subscriptionId) {
+    setPrevTenantId(tenantId);
+    setPrevSubscriptionId(subscriptionId);
+    setAzureTenantId(tenantId);
+    setAzureSubId(subscriptionId);
+    // when the tenant id or the subscription_id is changing, reset the resource
+    // group as the user will need to update it from a new list.
+    setAzureResourceGroup('');
+  }
   return (
     <>
       <ImageBuilderHeader />
@@ -230,6 +256,40 @@ const CreateImageWizard = () => {
                   setAccountEmail={setGcpAccountEmail}
                   domain={gcpDomain}
                   setDomain={setGcpDomain}
+                />
+              </WizardStep>,
+              <WizardStep
+                name="Microsoft Azure"
+                id="azure-sub-step"
+                key="azure-sub-step"
+                isHidden={
+                  !(environment.azure.selected && environment.azure.authorized)
+                }
+                footer={
+                  <CustomWizardFooter
+                    isNextDisabled={
+                      !(
+                        validateAzureId(azureTenantId) &&
+                        validateAzureId(azureSubId) &&
+                        azureResourceGroup
+                      )
+                    }
+                  />
+                }
+              >
+                <AzureTarget
+                  manual={azureManual}
+                  setManual={setAzureManual}
+                  source={azureSource}
+                  setSource={setAzureSource}
+                  tenantId={azureTenantId}
+                  setTenantId={setAzureTenantId}
+                  subscriptionId={azureSubId}
+                  setSubscriptionId={setAzureSubId}
+                  isErrorFetchingDetails={azureIDError}
+                  resourceGroups={resourceGroups}
+                  resourceGroup={azureResourceGroup}
+                  setResourceGroup={setAzureResourceGroup}
                 />
               </WizardStep>,
             ]}
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -1,186 +1,50 @@
-import React, { useEffect, useState } from 'react';
+import React, { useContext } from 'react';
 
-import {
-  Button,
-  Wizard,
-  WizardFooterWrapper,
-  WizardStep,
-  useWizardContext,
-} from '@patternfly/react-core';
+import { Wizard, WizardStep } from '@patternfly/react-core';
 import { useNavigate } from 'react-router-dom';
 
+import CustomWizardFooter from './CustomWizardFooter';
 import {
-  EnvironmentStateType,
-  filterEnvironment,
-  hasUserSelectedAtLeastOneEnv,
-  useGetAllowedTargets,
-} from './steps/ImageOutput/Environment';
+  ImageWizardContext,
+  useInitializeImageWizardContext,
+} from './ImageWizardContext';
+import { ValidateImageOutputStep } from './steps/ImageOutput/Environment';
 import ImageOutputStep from './steps/ImageOutput/ImageOutput';
 import ReviewStep from './steps/Review/ReviewStep';
 import AWSTarget, {
-  validateAWSAccountID,
+  ValidateAWSStep,
 } from './steps/TargetEnvironment/AWS/AWSTarget';
 import AzureTarget, {
-  validateAzureId,
+  ValidateAzureStep,
 } from './steps/TargetEnvironment/Azure/AzureTarget';
 import GCPTarget, {
-  GCPAccountTypes,
-  validateGCPData,
+  ValidateGCPStep,
 } from './steps/TargetEnvironment/GCP/GCPTarget';
-import { useGetAccountData } from './steps/TargetEnvironment/SourcesSelect';
 
-import { RHEL_9, X86_64 } from '../../constants';
 import './CreateImageWizard.scss';
-import { ArchitectureItem, Distributions } from '../../store/imageBuilderApi';
 import { resolveRelPath } from '../../Utilities/path';
 import { ImageBuilderHeader } from '../sharedComponents/ImageBuilderHeader';
 
-/**
- * @return true if the array in prevAllowedTargets is equivalent to the array
- * allowedTargets, false otherwise
- */
-const isIdenticalToPrev = (
-  prevAllowedTargets: string[],
-  allowedTargets: string[]
-) => {
-  let identicalToPrev = true;
-  if (allowedTargets.length === prevAllowedTargets.length) {
-    allowedTargets.forEach((elem) => {
-      if (!prevAllowedTargets.includes(elem)) {
-        identicalToPrev = false;
-      }
-    });
-  } else {
-    identicalToPrev = false;
-  }
-  return identicalToPrev;
-};
-
-type CustomWizardFooterPropType = {
-  isNextDisabled?: boolean;
-  isBackDisabled?: boolean;
-};
-/**
- * The custom wizard footer is only switching the order of the buttons compared
- * to the default wizard footer from the PF5 library.
- */
-const CustomWizardFooter = ({
-  isNextDisabled = false,
-  isBackDisabled = false,
-}: CustomWizardFooterPropType) => {
-  const { goToNextStep, goToPrevStep, close } = useWizardContext();
-  return (
-    <WizardFooterWrapper>
-      <Button
-        variant="primary"
-        onClick={goToNextStep}
-        isDisabled={isNextDisabled}
-      >
-        Next
-      </Button>
-      <Button
-        variant="secondary"
-        onClick={goToPrevStep}
-        isDisabled={isBackDisabled}
-      >
-        Back
-      </Button>
-      <Button variant="link" onClick={close}>
-        Cancel
-      </Button>
-    </WizardFooterWrapper>
-  );
-};
-
 const CreateImageWizard = () => {
-  const navigate = useNavigate();
-  //
-  // Image output step states
-  //
-  const [release, setRelease] = useState<Distributions>(RHEL_9);
-  const [arch, setArch] = useState<ArchitectureItem['arch']>(X86_64);
-  const {
-    data: allowedTargets,
-    isFetching,
-    isSuccess,
-    isError,
-  } = useGetAllowedTargets({
-    architecture: arch,
-    release: release,
-  });
-  const [environment, setEnvironment] = useState<EnvironmentStateType>(
-    filterEnvironment(
-      {
-        aws: { selected: false, authorized: false },
-        azure: { selected: false, authorized: false },
-        gcp: { selected: false, authorized: false },
-        oci: { selected: false, authorized: false },
-        'vsphere-ova': { selected: false, authorized: false },
-        vsphere: { selected: false, authorized: false },
-        'guest-image': { selected: false, authorized: false },
-        'image-installer': { selected: false, authorized: false },
-        wsl: { selected: false, authorized: false },
-      },
-      allowedTargets
-    )
+  return (
+    <ImageWizardContext.Provider value={useInitializeImageWizardContext()}>
+      <ContextedImageWizard />
+    </ImageWizardContext.Provider>
   );
-  // Update of the environment when the architecture and release are changed.
-  // This pattern prevents the usage of a useEffect See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
-  const [prevAllowedTargets, setPrevAllowedTargets] = useState(allowedTargets);
-  if (!isIdenticalToPrev(prevAllowedTargets, allowedTargets)) {
-    setPrevAllowedTargets(allowedTargets);
-    setEnvironment(filterEnvironment(environment, allowedTargets));
-  }
+};
 
-  //
-  // Target environment states
-  //
+const ContextedImageWizard = () => {
+  const navigate = useNavigate();
+  const isImageOutputStepValid = ValidateImageOutputStep();
+  const isAwsStepValid = ValidateAWSStep();
+  const isGcpStepValid = ValidateGCPStep();
+  const isAzureStepValid = ValidateAzureStep();
+  const { environmentState } = useContext(ImageWizardContext);
+  const [environment] = environmentState;
   const showTargetEnv =
     (environment.aws.selected && environment.aws.authorized) ||
     (environment.gcp.selected && environment.gcp.authorized) ||
     (environment.azure.selected && environment.azure.authorized);
-  // AWS
-  const [awsManual, setAwsManual] = useState(false);
-  const [awsSource, setAwsSource] = useState<[number, string]>([0, '']);
-  const { accountId: awsID, isError: awsIDError } = useGetAccountData(
-    awsSource[0],
-    'aws'
-  );
-  const [awsAccountId, setAwsAccountId] = useState(awsID);
-  const [prevAwsID, setPrevAwsID] = useState(awsID);
-  if (awsID !== prevAwsID) {
-    setPrevAwsID(awsID);
-    setAwsAccountId(awsID);
-  }
-  // GCP
-  const [shareGoogleAccount, setShareGoogleAccount] = useState(true);
-  const [gcpAccountType, setGcpAccountType] =
-    useState<GCPAccountTypes>('googleAccount');
-  const [gcpAccountEmail, setGcpAccountEmail] = useState('');
-  const [gcpDomain, setGcpDomain] = useState('');
-  // Azure
-  const [azureManual, setAzureManual] = useState(false);
-  const [azureSource, setAzureSource] = useState<[number, string]>([0, '']);
-  const {
-    tenantId,
-    subscriptionId,
-    resourceGroups,
-    isError: azureIDError,
-  } = useGetAccountData(azureSource[0], 'azure');
-  const [azureTenantId, setAzureTenantId] = useState(tenantId);
-  const [azureSubId, setAzureSubId] = useState(subscriptionId);
-  const [azureResourceGroup, setAzureResourceGroup] = useState('');
-  const [prevTenantId, setPrevTenantId] = useState(tenantId);
-  const [prevSubscriptionId, setPrevSubscriptionId] = useState(subscriptionId);
-  if (prevTenantId !== tenantId || prevSubscriptionId !== subscriptionId) {
-    setPrevTenantId(tenantId);
-    setPrevSubscriptionId(subscriptionId);
-    setAzureTenantId(tenantId);
-    setAzureSubId(subscriptionId);
-    // when the tenant id or the subscription_id is changing, reset the resource
-    // group as the user will need to update it from a new list.
-    setAzureResourceGroup('');
-  }
   return (
     <>
       <ImageBuilderHeader />
@@ -191,22 +55,12 @@ const CreateImageWizard = () => {
             id="step-image-output"
             footer={
               <CustomWizardFooter
-                isNextDisabled={!hasUserSelectedAtLeastOneEnv(environment)}
+                isNextDisabled={!isImageOutputStepValid}
                 isBackDisabled
               />
             }
           >
-            <ImageOutputStep
-              release={release}
-              setRelease={setRelease}
-              arch={arch}
-              setArch={setArch}
-              environment={environment}
-              setEnvironment={setEnvironment}
-              isFetching={isFetching}
-              isError={isError}
-              isSuccess={isSuccess}
-            />
+            <ImageOutputStep />
           </WizardStep>
           <WizardStep
             name="Target environment"
@@ -220,21 +74,9 @@ const CreateImageWizard = () => {
                 isHidden={
                   !(environment.aws.selected && environment.aws.authorized)
                 }
-                footer={
-                  <CustomWizardFooter
-                    isNextDisabled={!validateAWSAccountID(awsAccountId)}
-                  />
-                }
+                footer={<CustomWizardFooter isNextDisabled={!isAwsStepValid} />}
               >
-                <AWSTarget
-                  manual={awsManual}
-                  setManual={setAwsManual}
-                  source={awsSource}
-                  setSource={setAwsSource}
-                  isErrorFetchingDetails={awsIDError}
-                  associatedAccountId={awsAccountId}
-                  setAssociatedAccountId={setAwsAccountId}
-                />
+                <AWSTarget />
               </WizardStep>,
               <WizardStep
                 name="Google Cloud Platform"
@@ -243,29 +85,9 @@ const CreateImageWizard = () => {
                 isHidden={
                   !(environment.gcp.selected && environment.gcp.authorized)
                 }
-                footer={
-                  <CustomWizardFooter
-                    isNextDisabled={
-                      !validateGCPData(
-                        shareGoogleAccount,
-                        gcpAccountType,
-                        gcpAccountEmail,
-                        gcpDomain
-                      )
-                    }
-                  />
-                }
+                footer={<CustomWizardFooter isNextDisabled={!isGcpStepValid} />}
               >
-                <GCPTarget
-                  shareGoogleAccount={shareGoogleAccount}
-                  setShareGoogleAccount={setShareGoogleAccount}
-                  accountType={gcpAccountType}
-                  setAccountType={setGcpAccountType}
-                  accountEmail={gcpAccountEmail}
-                  setAccountEmail={setGcpAccountEmail}
-                  domain={gcpDomain}
-                  setDomain={setGcpDomain}
-                />
+                <GCPTarget />
               </WizardStep>,
               <WizardStep
                 name="Microsoft Azure"
@@ -275,31 +97,10 @@ const CreateImageWizard = () => {
                   !(environment.azure.selected && environment.azure.authorized)
                 }
                 footer={
-                  <CustomWizardFooter
-                    isNextDisabled={
-                      !(
-                        validateAzureId(azureTenantId) &&
-                        validateAzureId(azureSubId) &&
-                        azureResourceGroup
-                      )
-                    }
-                  />
+                  <CustomWizardFooter isNextDisabled={!isAzureStepValid} />
                 }
               >
-                <AzureTarget
-                  manual={azureManual}
-                  setManual={setAzureManual}
-                  source={azureSource}
-                  setSource={setAzureSource}
-                  tenantId={azureTenantId}
-                  setTenantId={setAzureTenantId}
-                  subscriptionId={azureSubId}
-                  setSubscriptionId={setAzureSubId}
-                  isErrorFetchingDetails={azureIDError}
-                  resourceGroups={resourceGroups}
-                  resourceGroup={azureResourceGroup}
-                  setResourceGroup={setAzureResourceGroup}
-                />
+                <AzureTarget />
               </WizardStep>,
             ]}
           />
@@ -308,22 +109,7 @@ const CreateImageWizard = () => {
             id="step-review"
             footer={<CustomWizardFooter isNextDisabled={true} />}
           >
-            <ReviewStep
-              release={release}
-              arch={arch}
-              environment={environment}
-              awsManual={awsManual}
-              awsAccountId={awsAccountId}
-              awsSource={awsSource}
-              gcpAccountType={gcpAccountType}
-              gcpAccountEmail={gcpAccountEmail}
-              gcpDomain={gcpDomain}
-              azureManual={azureManual}
-              azureSource={azureSource}
-              azureTenantId={azureTenantId}
-              azureSubscriptionId={azureSubId}
-              azureResourceGroup={azureResourceGroup}
-            />
+            <ReviewStep />
           </WizardStep>
         </Wizard>
       </section>
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import {
   Button,
@@ -17,6 +17,9 @@ import {
 } from './steps/ImageOutput/Environment';
 import ImageOutputStep from './steps/ImageOutput/ImageOutput';
 import ReviewStep from './steps/Review/ReviewStep';
+import AWSTarget, {
+  validateAWSAccountID,
+} from './steps/TargetEnvironment/AWS/AWSTarget';
 
 import { RHEL_9, X86_64 } from '../../constants';
 import './CreateImageWizard.scss';
@@ -75,7 +78,9 @@ const CustomWizardFooter = ({ isNextDisabled }: CustomWizardFooterPropType) => {
 
 const CreateImageWizard = () => {
   const navigate = useNavigate();
+  //
   // Image output step states
+  //
   const [release, setRelease] = useState<Distributions>(RHEL_9);
   const [arch, setArch] = useState<ArchitectureItem['arch']>(X86_64);
   const {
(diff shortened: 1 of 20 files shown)


src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
```diff
@@ -34,6 +79,81 @@ const ReviewStep = ({ release, arch }: ReviewStepPropTypes) => {
...
+              accountId={awsAccountId}
+              sourceId={awsSourceId}
+            />
+          )}
+          {environment.gcp.selected && environment.gcp.authorized && (
+            <TargetEnvGCPList
+              accountType={gcpAccountType}
+              accountEmail={gcpAccountEmail}
+              domain={gcpDomain}
+            />
```
regexowl, line 103: Can't recall when the "Principal" was added. It's empty both in the original wizard and 2.0, we should probably hide it when "Share image with Red Hat Insights only" is selected.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/GCP/GCPTarget.tsx
```diff
@@ -0,0 +1,218 @@
...
+        )}
+        {shareGoogleAccount && accountType !== 'domain' && (
+          <ValidatedTextField
+            aria="GCP account email"
+            label="Principal (e.g e-mail address)"
+            fieldId="gcp-account-email"
+            value={accountEmail}
+            setValue={setAccountEmail}
+            validateFunction={validateEmail}
+            helperText="must be a valid email address"
```
regexowl, line 199: The label of the text field says "e-mail", let's keep it the same in the error.

```suggestion
            helperText="Must be a valid e-mail address"
```

Also similar question as for the AWS validated fields - would it be possible to show the error when there's something wrong going on instead of from the very beginning when the field was not even interacted with?

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+              icon={<ExternalLinkAltIcon />}
+              iconPosition="right"
+              isInline
+              href="https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow"
+            >
+              Learn more about OAuth 2.0
+            </Button>
+          </Text>
+        </TextContent>
+        <FormGroup label="Share method:" isRequired>
```
regexowl, line 167: Share method does not need to be required as there is no way to disable both radios

```suggestion
        <FormGroup label="Share method:">
```

original:
image


current:
image

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+            description={
+              'Use a configured source to launch environments directly from the console.'
+            }
+            checked={!manual}
+            isChecked={!manual}
+            onClick={() => {
+              setManual(false);
+              clearState();
+            }}
+          />
```
regexowl, line 181: A bit more space between the radios would be nice again. We could probably make the gap a bit wider globally?

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+            isChecked={manual}
+            onClick={() => {
+              setManual(true);
+              clearState();
+            }}
+          />
+        </FormGroup>
+        {!manual && (
+          <>
+            <SourcesSelect
```
regexowl, line 196: Should be disabled when Sources are unavailable.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+              provider="azure"
+              selectedSourceId={sourceId}
+              setSelectedSourceId={setSourceId}
+              setFetchingError={setIsErrorFetchingSources}
+            />
+            <SourcesButton />
```
regexowl, line 202: Let's move it after the source select errors

original:
image


current:
image

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+                  Sources cannot be reached, try again later or enter an account
+                  info for upload manually.
+                </Alert>
+              )}
+              {(isErrorFetchingSources || isErrorFetchingDetails) && (
+                <Alert
+                  variant={'danger'}
+                  isPlain
+                  isInline
+                  title={'AWS details unavailable'}
```
regexowl, line 220: Should it be this?

```suggestion
                  title={'Azure details unavailable'}
```

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+                >
+                  Could not fetch Tenant id and Subscription id from Azure for
```
regexowl, line 222: This is the same in the original, but let's update it so it's unified with the names of the fields below.

```suggestion
                  Could not fetch Tenant GUID and Subscription ID from Azure for
```

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+                  isPlain={true}
+                  isInline={true}
+                  title={'Sources unavailable'}
+                >
+                  Sources cannot be reached, try again later or enter an account
+                  info for upload manually.
+                </Alert>
+              )}
+              {(isErrorFetchingSources || isErrorFetchingDetails) && (
+                <Alert
```
regexowl, line 216: Is displayed even when no source is selected. Should be displayed when there was a problem with fetching information about a selected source

image

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+                    type="text"
+                    readOnlyVariant="default"
+                    aria-label="Subscription ID"
+                    isRequired
+                  />
+                </FormGroup>
+              </GalleryItem>
+            </Gallery>
+            <AzureAuthButton tenantId={tenantId} />
+            <FormGroup label="Resource group" isRequired>
```
regexowl, line 262: Does not show "No results found when there are no results. 

original: 
image

current:
image


It could also be disabled when there is no source selected I think?

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+        {manual && (
+          <>
+            <ValidatedTextField
+              aria="Azure tenant id"
+              label="Azure Tenant GUID"
+              fieldId="azure-tenant-id"
+              value={tenantId}
+              setValue={setTenantId}
+              validateFunction={validateAzureId}
+              helperText="Please enter a valid tenant ID"
```
regexowl, line 284:
```suggestion
              helperText="Please enter a valid tenant GUID"
```

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
```diff
@@ -0,0 +1,312 @@
...
+                selected={resourceGroup}
+                setSelected={setResourceGroup}
+                placeholderText="Select a resource group"
+              />
+            </FormGroup>
+          </>
+        )}
+        {manual && (
+          <>
+            <ValidatedTextField
```
regexowl, line 277: Is it possible to show the validation error after user interacted with the inputs in any way instead of from the very beginning?

image

src/Components/CreateImageWizardV2/steps/Review/targetEnvironment.tsx
```diff
@@ -0,0 +1,258 @@
...
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {sourceName}
+            </TextListItem>
+          </>
+        )}
+        {manual && (
+          <>
+            <TextListItem component={TextListItemVariants.dt}>
+              Azure Tenant ID
```
regexowl, line 195: Maybe GUID again? So we keep the labels the same.

```suggestion
              Azure Tenant GUID
```

regexowl (commented):
Ran through the PR for the first time. It's looking really good! :+1: 

Added some comments about the visual things - didn't look through the code and tests yet. Will need to check those out a bit later.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/AWS/AWSTarget.tsx
```diff
@@ -0,0 +1,215 @@
...
+        <p>
+          <b>The shared image will expire within 14 days.</b> To permanently
+          access the image, copy the image, which will be shared to your account
+          by Red Hat, to your own AWS account.
+        </p>
+        <FormGroup label="Share method:">
+          <Radio
+            name="aws-from-sources"
+            id="aws-from-sources"
+            label="Use an account configured from Sources."
```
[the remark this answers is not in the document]
lavocatt: We can add a `<br />`

src/Components/CreateImageWizardV2/steps/TargetEnvironment/AWS/AWSTarget.tsx
```diff
@@ -0,0 +1,215 @@
...
+                  resolved. There might be a problem with the source. Verify
+                  that the source is valid in Sources or select a different
+                  source.
+                </Alert>
+              )}
+            </>
+          </>
+        )}
+        {manual && (
+          <ValidatedTextField
```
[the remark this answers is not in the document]
lavocatt: Yeah, that's definitely possible and easy to do.

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
```diff
@@ -0,0 +1,264 @@
...
+ * component. If the selection of items are changed, the user selection gets
+ * cleared up.
+ *
+ * @param initialSelectOptions a list of options, key and value being the same.
+ * @param selectID the html id for the component.
+ * @param placeholderText the placeholder text inside the textbox.
+ * @param selected the choice the user has made.
+ * @param setSelected a function to update the user choice.
+ */
+const TypeAheadSelect = ({
```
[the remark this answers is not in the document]
lavocatt: Maybe we need to come up to a max-width value in pixels.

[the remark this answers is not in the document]
lavocatt: In percentage, it's not going to work if the windows is too small.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/GCP/GCPTarget.tsx
```diff
@@ -0,0 +1,218 @@
...
+            checked={!shareGoogleAccount}
+            isChecked={!shareGoogleAccount}
+            onClick={() => {
+              setShareGoogleAccount(false);
+              clearState();
+            }}
+          />
+        </FormGroup>
+        {shareGoogleAccount && (
+          <FormGroup label="Account type" isRequired>
```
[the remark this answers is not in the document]
lavocatt: Indeed!

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
lavocatt: I don't know, I can add a `<br />` :shrug:

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
lavocatt: I think the no results found may be a better option, I need to learn how to make that happen.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureTarget.tsx
lavocatt: I could add a minimum size before showing the validation errors. WDYT?

regexowl: Like a min length for the input? That could work. We don't have any inputs anywhere that would take less than 2 chars as far as I know.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/AWS/AWSTarget.tsx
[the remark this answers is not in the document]
lavocatt: the `<br />` doesn't cut it.

src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
lavocatt: yeah, type email and domain shouldn't be displayed when we share with redhatinsights.

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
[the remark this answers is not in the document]
lavocatt: I'm wondering if the text-fields should be impacted too.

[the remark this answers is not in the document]
lavocatt: For instance here, the width isn't that an issue I find. 
image

src/Components/CreateImageWizardV2/steps/TargetEnvironment/AWS/AWSTarget.tsx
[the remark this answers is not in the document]
lavocatt: I added a padding of 20px

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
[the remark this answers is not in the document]
lavocatt: Now it looks like this 
image

What do you think about it?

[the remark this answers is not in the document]
regexowl: I think this looks great! Thanks!

src/Components/CreateImageWizardV2/steps/Review/targetEnvironment.tsx
```diff
@@ -0,0 +1,247 @@
...
+          <ExpirationWarning />
+        </TextListItem>
+        {(accountEmail || domain) && (
+          <>
+            <TextListItem component={TextListItemVariants.dt}>
+              Shared with
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              A Google account
+            </TextListItem>
```
regexowl, line 110: We could leave this out. It makes sense to show the "Shared with Red Hat Insight Only", but when an account is selected, there is other info that can be displayed.

src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
```diff
@@ -1,24 +1,70 @@
...
+  awsAccountId,
+  awsSource,
+  gcpAccountType,
+  gcpAccountEmail,
+  gcpDomain,
+  azureManual,
+  azureSource,
+  azureTenantId,
+  azureSubscriptionId,
+  azureResourceGroup,
```
regexowl, line 59: Before we continue with the following steps - I won't lie, I'm a bit worried about this. We are on the second step and the lists of props are starting to get longer. Not entirely sure how sustainable those will be in the following steps.

The problem is, I can't really think of any solution I would deem "safer" or "better" right now as I didn't look into state management solutions that are available.

On one hand I feel like most of the code is easy to follow and navigate through (also love the comments a lot!), but on the other it's starting to be a lot to review and wrap you mind around when it comes to all the states that come into play.

I'd like to get another opinion about this as I think it could be both perfectly fine in the long run but it could also easily get out of hand with a ton of props needed for every component to work...

@lucasgarfield @jkozol what do you think?

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
```diff
@@ -0,0 +1,390 @@
...
+      isOpen={isOpen}
+      selected={selectedKey}
+      onSelect={onSelect}
+      onOpenChange={() => {
+        setIsOpen(false);
+      }}
+      toggle={toggle}
+      aria-disabled={isError}
+    >
+      <SelectList id={fieldID + '-listbox'}>
```
regexowl, line 360: The select should be disabled when the Sources service is unavailable.

image

regexowl (commented):
Ok, tried to go through all the code, but had to run through some parts as this is getting quite complex. Overall I think the wizard looks good and it seems to be also working very well. So from the user experience point of view no complaints.

What I'm a bit undecided about is the developer experience. I've summed up my thoughts about state management in a comment below and I'd be happy if there were more views or opinions about this.

src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
lavocatt: We can use context for that. But it can be postponed up until the end of the refactor.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -17,6 +17,17 @@ import {
...
+import AWSTarget, {
+  validateAWSAccountID,
+} from './steps/TargetEnvironment/AWS/AWSTarget';
+import AzureTarget, {
+  validateAzureId,
+} from './steps/TargetEnvironment/Azure/AzureTarget';
+import GCPTarget, {
+  GCPAccountTypes,
+  validateGCPData,
+} from './steps/TargetEnvironment/GCP/GCPTarget';
```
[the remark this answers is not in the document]
lavocatt: I kinda prefer to have a sub folder for a substep, but I don't mind too much squashing everything up a folder. If you prefer it strongly that way I can change that.

src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
lavocatt: The thing is you have two possibilities for passing states to a children component, either you list them all in function call, or you store them in a context and the children component load them from the context. That's what every state management libraries are doing under the hood.

Having the state management library doesn't remove the need of having to write every call to the states. At some point they have to be explicitly called when you want to access the data. The advantage that you have by listing all of them in the component call is that you can actually document what is needed for every component. Docstring etc.

I agree that the review step in that case is going to have a ridiculous list of parameters. We can make an exception for it with a context. But I don't think it's that bad to have them all listed. Even if that's a hundred parameters.
Should these parameters keep being forwarded deeply down a stack of component, I would go use a context, but our wizard isn't deep.

My main question would be, do you find it more easy to follow the states in the DDF way? Because in the DDF way we're using the form state to access the states.

For me I'm always lost with the DDF fashion. Especially because I don't know if a state is used somewhere else.
Having the explicit declaration makes it easier to track what is used where. You can follow a state to where it comes from, jump to the parent definition, check it's usages and see every component that has access to it. They are all listed in a top down manner.

The way I represent it in my mind is, by explicitely defining the parameters for every component you work as a Tree. With a context library you work more as a graph. And a tree is easier to grasp (IMO).

lavocatt: For reference, have a look at https://react.dev/learn/passing-data-deeply-with-context#before-you-use-context

src/Components/CreateImageWizardV2/steps/Review/targetEnvironment.tsx
lavocatt: It felt more coherent to add something that the account is shared with a google account. I can remove that if you want.

src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
lavocatt: > What about a structure with all the values? Could we maybe create an object or something like that and start passing and destructuring that? Or is that a stupid idea?

It's a good idea in some places. Another possibility is to pass jsx objects down to the review step instead of props.

But having complex objects means that we might need to start having reducers to update them

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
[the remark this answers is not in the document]
jkozol: I don't think we need to have sub folders here since most of these steps are going to only be a single file. I would suggest just having `/steps/AWSTarget.tsx`

src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
jkozol: I agree that explicitly defining the parameters makes it simpler to follow what is being passed where. But, as the state and steps continue to grow the amount of getters, setters, and params are also going to grow. I think we will quickly hit a point where the passing of props will make our data less clear. 

I think that implementing a Context with Reducers now will save a lot of time later. Even though some of our field's data types are bit complex (filesystem), for the most part each customisation has a value, isValid, and set/get. So, setting up a general reducer pattern should hopefully not get too complicated and would allow us to avoid too much prop passing.

lucasgarfield: The React documentation is great, and in this case I think it actually makes a strong argument for the use of Context+Reducer pattern. From the documentation linked above:

>Managing state: As your app grows, you might end up with a lot of state closer to the top of your app. Many distant components below may want to change it. It is common to [use a reducer together with context](https://react.dev/learn/scaling-up-with-reducer-and-context) to manage complex state and pass it down to distant components without too much hassle.

This is exactly what is happening here. We are ending up with a lot of state at the top of the app, with numerous distant components below it.

I think we should do what the React documentation recommends here, and use a reducer + context pattern to pass the state to children components “without too much hassle”.

lavocatt: Well these technical solutions seems to be quit popular for you all, I won't get in the way and have them implemented then. I do think using contexts for these props is a mistake maintenance wise, but only time will tell who's guess is the correct one.

lucasgarfield: Thanks for being willing to change course. This is a really important design decision... both paradigms have advantages and disadvantages, for sure. I think that right now is the appropriate time to experiment. I'm glad we tried the code using prop drilling first... looking forward to seeing how the code will look with useReducer + useContext now.

lucasgarfield:
General comment that applies to numerous commits: can we please prefix all booleans with `is` or `are`?

I realize that this isn't necessarily accepted as a best practice by all teams, but it is what we currently use in our code base and it is also what Patternfly components and RTKQ hooks use. So I think that for the sake of consistency, it is important we keep following this practice.

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
```diff
@@ -0,0 +1,390 @@
...
+ *
+ * @param isFetching if provided and set to true, the component will show a
+ * loading spinner when the data is on the fly.
+ * @param isError if provided and set to true, the component will disable
+ * itself.
+ * @param refetch if provided, the component will use the refetch function to
+ * ask for a refresh of the data that makes the select options.
+ */
+function TypeAheadSelect<T extends string | number>({
+  inputOptions,
```
lucasgarfield, line 128: I think it would be a bit more "React-y" if the `TypeAheadSelect` took a `children` prop instead of `inputOptions`, where the children were the `<SelectOption>`s.

lucasgarfield, line 128: I think that it would be more idiomatic React for the `TypeAheadSelect` to not take an `inputOptions` prop and instead just render its children, which should be `<SelectOption>`s. That is how the reference component in the PF5 docs works:

```
    <Select
      id="typeahead-select"
      isOpen={isOpen}
      selected={selected}
      onSelect={onSelect}
      onOpenChange={() => {
        setIsOpen(false);
      }}
      toggle={toggle}
    >
      <SelectList id="select-typeahead-listbox">
        {selectOptions.map((option, index) => (
          <SelectOption
            key={option.value || option.children}
            isFocused={focusedItemIndex === index}
            className={option.className}
            onClick={() => setSelected(option.value)}
            id={`select-typeahead-${option.value.replace(' ', '-')}`}
            {...option}
            ref={null}
          />
        ))}
      </SelectList>
    </Select>
```

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
```diff
@@ -0,0 +1,390 @@
...
+      onOpenChange={() => {
+        setIsOpen(false);
+      }}
+      toggle={toggle}
+      aria-disabled={isError}
+    >
+      <SelectList id={fieldID + '-listbox'}>
+        {selectOptions.map((option, index) => (
+          <SelectOption
+            key={option.value}
```
lucasgarfield, line 363: Is `option.value` guaranteed to be unique?

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -136,6 +162,36 @@ const CreateImageWizard = () => {
               isSuccess={isSuccess}
             />
           </WizardStep>
+          <WizardStep
```
lucasgarfield, line 211: I think it might be good to wrap each of these steps in their own component. For instance, we could wrap this one in `<TargetEnvironmentStep>`. 

That would make the list of steps in the wizard much easier to parse.

It will also be necessary in the future when we add the ability to customize the steps (for instance, we may or may not want to include the OpenSCAP step, etc...).

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -147,6 +150,29 @@ const CreateImageWizard = () => {
...
   const [gcpDomain, setGcpDomain] = useState('');
+  // Azure
+  const [azureManual, setAzureManual] = useState(false);
+  const [azureSource, setAzureSource] = useState<[number, string]>([0, '']);
+  const {
+    tenantId,
+    subscriptionId,
+    resourceGroups,
+    isError: azureIDError,
+  } = useGetAccountData(azureSource[0], 'azure');
```
lucasgarfield, line 169: I think that this is the wrong place to call RTKQ hooks. `useGetAccountData` is not from RTKQ directly of course, but it wraps an RTKQ hook so the same argument applies. RTKQ hooks should be called as close as possible to where their return values are needed.

RTKQ hooks cache their values after fetching data, so they can be called wherever their values are needed. That means we don't need to pass down props like `isError` which is hard to keep track of... instead, we can just pass down the query parameters (so azureSource[0] and 'azure' in this case) and call the hook exactly where it is needed. This minimizes the number of props we need to keep track.

Second, calling hooks in the highest level component is potentially inefficient. It means that we front-load all of the data fetching. We may end up not needed to call some RTKQ hooks. Calling them where their data is needed ensures they only fetch data when necessary.

We do of course want the wizard to be fast so we do sometimes want to fetch the data ahead of time... but in that case, we should use prefetching and avoid passing down props:
https://redux-toolkit.js.org/rtk-query/usage/prefetching

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
```diff
@@ -0,0 +1,390 @@
...
+ */
+function TypeAheadSelect<T extends string | number>({
+  inputOptions,
+  selected,
+  setSelected,
+  fieldID,
+  placeholderText,
+  isFetching,
+  isError,
+  refetch,
```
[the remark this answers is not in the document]
lavocatt: I understand your comment. `TypeAheadSelct` in its current form has a couple of useful properties. The first of which is it is entirely self contained and reusable. Changing this aspect would mean that a lot of code would have to be rewritten every time we want a type ahead select. And since we're using it in at least 3 places I think this design holds a great value.

The other design you are proposing wants to have the options and their refetching and the loading spinner extracted. This design would neglect the value of having a TypeAheadSelect in my opinion. Maybe in that case it would be preferable to just use the pf5 pattern as is every time instead of doing some generalization?

Of course my preferred way holds with the designed I proposed, but we can change it ofc.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
lavocatt: Yeah I forgot about this caching. Thanks for bringing it up!

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
lavocatt: Unfortunately, PF5 won't allow it. It checks a child of Wizard is a WizardStep. But I can try again a second time.

lavocatt:
Alright everyone, I did the rework to move to a context here. I'll let you have a look at it and tell me if that's more of your liking.

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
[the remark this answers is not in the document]
lavocatt: Btw that refetch logic is in the legacy wizard. That's why I brought it there. You want to be able to refetch the sources when clicking on the text input of the typeahead. So IMO the typeahead needs to have it. It's a callback function, so i don't see it as breaching the separation of concerns since the typeahead select still don't need to have any clue about what is being refetched or if refetch will actually refetch anything.

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
regexowl: And bumping this one as well :up:

src/Components/CreateImageWizardV2/common/ValidatedTextField.tsx
```diff
@@ -0,0 +1,77 @@
...
+        aria-label={aria}
+        onChange={handleTextInputChange}
+      />
+      <FormHelperText>
+        <HelperTextItem>
+          <HelperTextItem variant={validated}>
+            {validated === 'error' ? helperText : ''}
+          </HelperTextItem>
+        </HelperTextItem>
+      </FormHelperText>
```
regexowl, line 72: The helper text should be red. This should do the trick

```suggestion
      <HelperText>
        <HelperTextItem variant={validated}>
          {validated === 'error' ? helperText : ''}
        </HelperTextItem>
      </HelperText>
```

image

src/Components/CreateImageWizardV2/common/TypeAheadSelect.tsx
lavocatt: I'm not shooting for idiomatic React for this component. I want something reusable with as little code as possible.
The reason why:
* it's a pattern that is quite big and complex from pf5 already
* we're using it in many steps of the wizard, 
* less code, less bugs to fix :slightly_smiling_face:

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
lucasgarfield: Could we do something like:
```
<WizardStep>
<TargetEnvironmentStep />
</WizardStep>
```
?

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+    React.Dispatch<React.SetStateAction<string>>
+  ];
+  azureResourceGroupState: [
+    string,
+    React.Dispatch<React.SetStateAction<string>>
+  ];
+};
+
+export const ImageWizardContext =
+  React.createContext<ImageWizardContextPropType>({
```
lucasgarfield, line 63: Would turning this into a nested object help organize things a bit? For instance, one level of nesting per wizard step... in other words:
```
{
  firstStep: {},
  secondStep: {},
}
```

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+    releaseState: [RHEL_9, () => {}],
+    architectureState: [X86_64, () => {}],
+    environmentState: [
```
lucasgarfield, line 66: We usually call aws/gcp/azure/etc... the 'target', don't we? Could we change 'environment' to 'target'?

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+  ];
+  azureResourceGroupState: [
+    string,
+    React.Dispatch<React.SetStateAction<string>>
+  ];
+};
+
+export const ImageWizardContext =
+  React.createContext<ImageWizardContextPropType>({
+    releaseState: [RHEL_9, () => {}],
```
lucasgarfield, line 64: The `-State` suffix on these variables seems redundant to me... we know it is the state, don't we? We could cut down on verbosity a bit by removing it.

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+        vsphere: { selected: false, authorized: false },
+        'guest-image': { selected: false, authorized: false },
+        'image-installer': { selected: false, authorized: false },
+        wsl: { selected: false, authorized: false },
+      },
+      () => {},
+    ],
+    isAwsManualState: [false, () => {}],
+    awsSourceState: [[0, ''], () => {}],
+    associatedAwsAccountIdState: ['', () => {}],
```
lucasgarfield, line 82: Do we want to initialize things to empty strings? Wouldn't `undefined` be more accurate? In the type, you can use a union of `string | undefined`.

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
```diff
@@ -0,0 +1,208 @@
...
+    setPrevTenantId(tenantId);
+    setPrevSubscriptionId(subscriptionId);
+    setAzureTenantId(tenantId);
+    setAzureSubId(subscriptionId);
+    // when the tenant id or the subscription_id is changing, reset the resource
+    // group as the user will need to update it from a new list.
+    setAzureResourceGroup('');
+  }
+  const context: ImageWizardContextPropType = {
+    releaseState: [release, setRelease],
```
lucasgarfield, line 169: I'm not quite sold on this. Using an array to pass the state value and setters around seems like an anti-pattern to me. When people read the code who aren't familiar with this, it will be unclear what we are doing when accessing releaseState[0] and releaseState[1]. It also means we aren't taking advantage of the type safety Typescript is giving us now. Would the linter pick up on someone trying to access releaseState[2]? Also, I don't think that the Typescript language server is going to be able to make meaningful inferences/suggestions. 

I'll try to look into some alternate means of handling this.

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lucasgarfield: Oh I'm dumb, it is the 'target environment'. Still, it confused me so maybe there is some inconsistency. Could we call it the `targetEnvironment` (dropping the `-State` suffix)?

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lavocatt: That's a big question, having undefined means we'll have to deal with protection everywhere no?

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lavocatt: It's not an array (well it looks like it lol) but a tupple for typescript. It considers it an array that has two members an not more with a specific order in which the values must be passed in. But yeah, if you want to try something else that's fine too.

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lucasgarfield: It's a very typical pattern in our code base. We check for whether something is undefined or not using optional chaining (`thing?.otherthing`) and can take advantage of short circuiting if it is undefined (`thing?.otherthing && ...`). Our linter will ensure that we never attempt to access something that is undefined.

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lavocatt: Ah sorry I wasn't on the right one. Forget my comment.

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lavocatt: That's an interesting idea. Have you considered that the fact that we split in steps may not be relevant within the review step ? And when wanting to access a prop to have to search for it?

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lavocatt: Yeah, atm the empty string only needs to be checked to validate the step, which makes it convenient for everywhere else it needs to be used. But as you prefer!

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lavocatt: The reason why there's `State` everywhere is because the state and its setter are packed together. And then you unpack them the same way you would with `useState`. But I see that you already don't like this packing, so why not have just the state and its setter together in the context.

src/Components/CreateImageWizardV2/ImageWizardContext.tsx
lavocatt: environment was the naming used in the previous wizard. If you prefer targetEnvironment I'm all for it.

PR #1434, opened 2023-11-14
fix some lint warning message

mgold1234:
this commit fix some linting warnings that user get when running npm run lint:js:fix command,
some worning comes after big changes to TypeScript

### DIFF
--- a/src/store/enhancedImageBuilderApi.ts
+++ b/src/store/enhancedImageBuilderApi.ts
@@ -52,10 +52,7 @@ const enhancedApi = imageBuilderApi.enhanceEndpoints({
       },
     },
     composeImage: {
-      onQueryStarted: async (
-        { composeRequest },
-        { dispatch, queryFulfilled }
-      ) => {
+      onQueryStarted: async (_, { dispatch, queryFulfilled }) => {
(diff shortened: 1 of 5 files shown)


src/store/enhancedImageBuilderApi.ts
```diff
@@ -53,7 +53,8 @@ const enhancedApi = imageBuilderApi.enhanceEndpoints({
     },
     composeImage: {
       onQueryStarted: async (
-        { composeRequest },
+        // eslint-disable-next-line no-empty-pattern
+        {},
```
regexowl, line 57: Not super sure about this one. If we need to disable the linting on the line I'd rather disable the original warning than disabling it on a newly introduced change. What do you think?

We could probably do something like this
```suggestion
        // eslint-disable-next-line @typescript-eslint/no-unused-vars
        { composeRequest },
```

Or we could indicate that we don't care about the first argument and use
```suggestion
      onQueryStarted: async (_, { dispatch, queryFulfilled }) => {
```

which should hopefully solve both of the warning at once.

regexowl (approved):
Solving the warnings one PR after another, thank you!

PR #1435, opened 2023-11-14
Wizard: Disable `/usr` sub-directories on Recreate

regexowl:
Paths with `/usr` sub-directories are no longer supported in file system configuration. This meant that an attempt to create an image with the erroneous path ended in an image build not starting but there was also no error displayed. 

The option to add a sub-directory to the `/usr` path was removed in https://github.com/RedHatInsights/image-builder-frontend/pull/1429. This is a follow up PR, that handles the same problem specifically for the Recreate action.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -465,7 +465,9 @@ const requestToState = (composeRequest, distroInfo, isProd, enableOscap) => {
       for (const fsc of fs) {
         const [size, unit] = parseSizeUnit(fsc.min_size);
         fileSystemConfiguration.push({
-          mountpoint: fsc.mountpoint,
+          mountpoint: fsc.mountpoint.includes('/usr/')
+            ? '/usr'
+            : fsc.mountpoint,
           size,
           unit,
         });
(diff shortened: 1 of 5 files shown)


lavocatt:
Should we display a specific message to the user in that case?

regexowl:
> Should we display a specific message to the user in that case?

Do you mean a notification with the error on the build? In this specific case sending an erroneous request is no longer possible.

But generally we should review our on build notifications as the error that should warn about the problem was missing.

lavocatt:
> > Should we display a specific message to the user in that case?
> 
> Do you mean a notification with the error on the build? In this specific case sending an erroneous request is no longer possible.

I meant, if someone wants to recreate an image that can't be recreated, should we have a new notification for that? I think we should at least display something.

regexowl:
Oh, I see what you mean. The path with a sub-directory is now automatically changed from `/usr/whatever-sub-dir` to just `/usr`, but the user wouldn't know that if they don't click on the File system configuration step. What if I added the alert that's displayed on the FSC step to the Review step as well?

lavocatt:
> Oh, I see what you mean. The path with a sub-directory is now automatically changed from `/usr/whatever-sub-dir` to just `/usr`, but the user wouldn't know that if they don't click on the File system configuration step. What if I added the alert that's displayed on the FSC step to the Review step as well?

Yeah that works for me.

regexowl:
Added the alert to the Review step as well. Also the alert should now be displayed only when there' s a mount point with the `/usr` prefix.

lavocatt (approved):
All good :) Thanks for adding the extra message

croissanne (approved):

PR #1440, opened 2023-11-16
Fix HMS-3031

lucasgarfield:
The first commit cleans up some of the warnings related to the exhaustive dependencies rule.

The second commit fixes the bug in the packages search where hitting 'enter' causes a re-render and return to the first page of the wizard.


lucasgarfield:
Just realized it's still broken, converting to draft temporarily...

lucasgarfield:
Okay should be good to go. We need to clean up these useEffect() hooks in the refactor for Wizard V2, there are way too many...

croissanne (approved):

PR #1448, opened 2023-11-21
CreateImageWizard: package search through correct architecture

croissanne:
Instead of always searching through x86 repos, search through the repos of the selected architecture. This fixes HMS-3066.

---

- [ ] https://github.com/content-services/content-sources-backend/pull/479


regexowl (approved):
Thanks a lot for catching this!

PR #1450, opened 2023-11-21
wizard/oscap: update the oscap step to latest specifications

mgold1234:
Before, in the Oscap step, the drop down select menu was only listing profile_ids
Now, it shows the profiles names. When the user selects one, the details about the
profile are displayed under the selection, improving user experience.

FIXES # - HMS-3087
Screenshot 2023-11-21 at 15 40 23

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Oscap.js
+++ b/src/Components/CreateImageWizard/formComponents/Oscap.js
@@ -1,16 +1,30 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
 import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
-import { Alert, Radio, FormGroup, Spinner } from '@patternfly/react-core';
+import {
+  Alert,
+  FormGroup,
+  Spinner,
+  Popover,
+  TextContent,
+  Text,
+  Button,
+} from '@patternfly/react-core';
 import {
   Select,
   SelectOption,
   SelectVariant,
 } from '@patternfly/react-core/deprecated';
+import { HelpIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
 
-import { useGetOscapProfilesQuery } from '../../../store/imageBuilderApi';
+import OscapProfileInformation from './OscapProfileInformation';
+
+import {
+  useGetOscapCustomizationsQuery,
+  useGetOscapProfilesQuery,
+} from '../../../store/imageBuilderApi';
 import { reinitFileSystemConfiguratioStep } from '../steps/fileSystemConfiguration';
 import { reinitPackagesStep } from '../steps/packages';
 
@@ -30,26 +44,35 @@ const reinitDependingSteps = (change) => {
  * The selected profile will be stored in the `oscap-profile` form state variable.
  * The Component is shown or not depending on the ShowSelector variable.
  */
-const ProfileSelector = ({ input, showSelector }) => {
+const ProfileSelector = ({ input }) => {
   const { change, getState } = useFormApi();
-  const [profile, selectProfile] = useState(
-    getState()?.values?.['oscap-profile']
-  );
+  const [profile, setProfile] = useState(getState()?.values?.['oscap-profile']);
+  const [profileName, setProfileName] = useState('');
   const [isOpen, setIsOpen] = useState(false);
+  const {
+    data: profiles,
+    isFetching,
+    isSuccess,
+    isError,
+    refetch,
+  } = useGetOscapProfilesQuery({
+    distribution: getState()?.values?.['release'],
+  });
 
-  const { data, isFetching, isSuccess, isError, refetch } =
-    useGetOscapProfilesQuery(
-      {
-        distribution: getState()?.values?.['release'],
-      },
-      {
-        skip: !showSelector,
-      }
-    );
-
-  if (!showSelector) {
-    return undefined;
-  }
+  const { data } = useGetOscapCustomizationsQuery(
+    {
+      distribution: getState()?.values?.['release'],
+      profile: profile,
+    },
+    {
+      skip: !profile,
+    }
+  );
+  useEffect(() => {
+    if (data) {
+      setProfileName(data?.openscap?.profile_name);
+    }
+  }, [data]);
 
   const handleToggle = () => {
     if (!isOpen) {
@@ -59,13 +82,14 @@ const ProfileSelector = ({ input, showSelector }) => {
   };
 
   const handleClear = () => {
-    selectProfile(undefined);
+    setProfile(undefined);
     change(input.name, undefined);
+    setProfileName(undefined);
     reinitDependingSteps(change);
   };
 
-  const setProfile = (_, selection) => {
-    selectProfile(selection);
+  const handleSelect = (_, selection) => {
+    setProfile(selection);
     setIsOpen(false);
     change(input.name, selection);
     reinitDependingSteps(change);
@@ -75,23 +99,72 @@ const ProfileSelector = ({ input, showSelector }) => {
   return (
     <FormGroup
       isRequired={true}
-      label={'Profile to use for this image'}
       data-testid="profiles-form-group"
+      label={
+        <>
+          OpenSCAP profile
+          <Popover
+            maxWidth="30rem"
+            position="left"
+            bodyContent={
+              <TextContent>
+                <Text>
+                  To run a manual compliance scan in OpenSCAP, download this
+                  image.
+                </Text>
+              </TextContent>
+            }
+          >
+            <Button variant="plain" aria-label="About OpenSCAP" isInline>
+              <HelpIcon />
+            </Button>
+          </Popover>
+        </>
+      }
     >
       <Select
         ouiaId="profileSelect"
         variant={SelectVariant.typeahead}
         onToggle={handleToggle}
-        onSelect={setProfile}
+        onSelect={handleSelect}
         onClear={handleClear}
-        selections={profile}
+        selections={profileName}
         isOpen={isOpen}
         placeholderText="Select a profile"
         typeAheadAriaLabel="Select a profile"
         isDisabled={!isSuccess}
+        onFilter={(_event, value) => {
+          return [
+            <OScapNoneOption setProfileName={setProfileName} key={profiles} />,
+          ].concat(
+            profiles.map((profile_id, index) => {
+              return (
+                <OScapSelectOption
+                  key={index}
+                  profile_id={profile_id}
+                  setProfileName={setProfileName}
+                  input={value}
+                />
+              );
+            })
+          );
+        }}
       >
         {isSuccess &&
-          data.map((key, index) => <SelectOption key={index} value={key} />)}
+          [
+            <OScapNoneOption setProfileName={setProfileName} key={profiles} />,
+          ].concat(
+            profiles.map((profile_id) => {
+              return (
+                <OScapSelectOption
+                  key={profile_id}
+                  profile_id={profile_id}
+                  setProfileName={setProfileName}
+                />
+              );
+            })
+          )}
+
         {isFetching && (
           <SelectOption isNoResultsOption={true} data-testid="policies-loading">
             <Spinner size="md" />
@@ -112,6 +185,55 @@ const ProfileSelector = ({ input, showSelector }) => {
   );
 };
 
+const OScapNoneOption = ({ setProfileName }) => {
+  return (
+    <SelectOption
+      value={undefined}
+      onClick={() => {
+        setProfileName('None');
+      }}
+    >
+      <p>{'None'}</p>
+    </SelectOption>
+  );
+};
+
+OScapNoneOption.propTypes = {
+  setProfileName: PropTypes.any,
+};
+
+const OScapSelectOption = ({ profile_id, setProfileName, input }) => {
+  const { getState } = useFormApi();
+  const { data } = useGetOscapCustomizationsQuery({
+    distribution: getState()?.values?.['release'],
+    profile: profile_id,
+  });
+  if (
+    input &&
+    !data?.openscap?.profile_name.toLowerCase().includes(input.toLowerCase())
+  ) {
+    return null;
+  }
+
+  return (
+    <SelectOption
+      key={profile_id}
+      value={profile_id}
+      onClick={() => {
+        setProfileName(data?.openscap?.profile_name);
+      }}
+    >
+      <p>{data?.openscap?.profile_name}</p>
+    </SelectOption>
+  );
+};
+
+OScapSelectOption.propTypes = {
+  profile_id: PropTypes.string,
+  setProfileName: PropTypes.any,
+  input: PropTypes.string,
+};
+
 ProfileSelector.propTypes = {
   input: PropTypes.any,
   showSelector: PropTypes.bool,
@@ -125,39 +247,10 @@ ProfileSelector.propTypes = {
  *   undefined.
  */
 const AddProfile = ({ input }) => {
-  const { change, getState } = useFormApi();
-  const oscapProfile = getState()?.values?.['oscap-profile'];
-  const [wantsProfile, setWantsProfile] = useState(oscapProfile !== undefined);
   return (
     <>
-      <FormGroup label="Compliance profile">
-        <Radio
-          name="add-a-profile"
-          className="pf-u-mt-md"
-          data-testid="add-a-profile-radio"
-          id="add-a-profile"
-          label="Add a profile"
-          isChecked={wantsProfile}
-          onChange={() => {
-            setWantsProfile(true);
-            reinitDependingSteps(change);
-          }}
-        />
-        <Radio
-          name="dont-add-a-profile"
-          className="pf-u-mt-md"
-          data-testid="dont-add-a-profile-radio"
-          id="dont-add-a-profile"
-          label="Do not add a profile"
-          isChecked={!wantsProfile}
-          onChange={() => {
-            setWantsProfile(false);
-            change(input.name, undefined);
-            reinitDependingSteps(change);
-          }}
-        />
-      </FormGroup>
-      <ProfileSelector input={input} showSelector={wantsProfile} />
+      <ProfileSelector input={input} />
+      <OscapProfileInformation />
     </>
   );
 };
--- a/src/Components/CreateImageWizard/formComponents/OscapProfileInformation.tsx
+++ b/src/Components/CreateImageWizard/formComponents/OscapProfileInformation.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+
+import { useFormApi } from '@data-driven-forms/react-form-renderer';
+import {
+  Spinner,
+  TextContent,
+  TextList,
+  TextListItem,
+  TextListItemVariants,
+  TextListVariants,
+} from '@patternfly/react-core';
+
+import { RELEASES } from '../../../constants';
+import { useGetOscapCustomizationsQuery } from '../../../store/imageBuilderApi';
+
+const OscapProfileInformation = (): JSX.Element => {
+  const { getState } = useFormApi();
+
+  const oscapProfile = getState()?.values?.['oscap-profile'];
+
+  const {
+    data: oscapProfileInfo,
+    isFetching: isFetchingOscapProfileInfo,
+    isSuccess: isSuccessOscapProfileInfo,
+  } = useGetOscapCustomizationsQuery(
+    { distribution: getState()?.values?.['release'], profile: oscapProfile },
+    {
+      skip: !oscapProfile,
+    }
+  );
+
+  return (
+    <>
+      {isFetchingOscapProfileInfo && <Spinner size="lg" />}
+      {isSuccessOscapProfileInfo && (
+        <TextContent>
+          <br />
+          <TextList component={TextListVariants.dl}>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-u-min-width"
+            >
+              Profile description:
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {oscapProfileInfo.openscap?.profile_description}
+            </TextListItem>
+          </TextList>
+          <TextList component={TextListVariants.dl}>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-u-min-width"
+            >
+              Operating system:
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {RELEASES.get(getState()?.values?.['release'])}
+            </TextListItem>
+          </TextList>
+          <TextList component={TextListVariants.dl}>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-u-min-width"
+            >
+              Reference ID:
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {oscapProfileInfo.openscap?.profile_id}
+            </TextListItem>
+          </TextList>
+        </TextContent>
+      )}
+    </>
+  );
+};
+
+export default OscapProfileInformation;
--- a/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
+++ b/src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
@@ -28,6 +28,7 @@ import {
 
 import { RELEASES, UNIT_GIB } from '../../../constants';
 import { extractProvisioningList } from '../../../store/helpers';
+import { useGetOscapCustomizationsQuery } from '../../../store/imageBuilderApi';
 import { useGetSourceListQuery } from '../../../store/provisioningApi';
 import { useShowActivationKeyQuery } from '../../../store/rhsmApi';
 import { useGetEnvironment } from '../../../Utilities/useGetEnvironment';
@@ -618,6 +619,11 @@ export const ImageDetailsList = () => {
 export const OscapList = () => {
   const { getState } = useFormApi();
   const oscapProfile = getState()?.values?.['oscap-profile'];
+  const { data } = useGetOscapCustomizationsQuery({
+    distribution: getState()?.values?.['release'],
+    profile: oscapProfile,
+  });
+
   return (
     <TextContent>
       <TextList component={TextListVariants.dl}>
@@ -625,7 +631,29 @@ export const OscapList = () => {
           component={TextListItemVariants.dt}
           className="pf-u-min-width"
         >
-          Profile
+          Profile name:
+        </TextListItem>
+        <TextListItem component={TextListItemVariants.dd}>
+          {data?.openscap?.profile_name}
+        </TextListItem>
+      </TextList>
+      <TextList component={TextListVariants.dl}>
+        <TextListItem
+          component={TextListItemVariants.dt}
+          className="pf-u-min-width"
+        >
+          Profile description:
+        </TextListItem>
+        <TextListItem component={TextListItemVariants.dd}>
+          {data?.openscap?.profile_description}
+        </TextListItem>
+      </TextList>
+      <TextList component={TextListVariants.dl}>
+        <TextListItem
+          component={TextListItemVariants.dt}
+          className="pf-u-min-width"
+        >
+          Reference ID:
         </TextListItem>
         <TextListItem component={TextListItemVariants.dd}>
           {oscapProfile}
--- a/src/Components/CreateImageWizard/steps/oscap.js
+++ b/src/Components/CreateImageWizard/steps/oscap.js
@@ -1,17 +1,23 @@
 import React from 'react';
 
 import componentTypes from '@data-driven-forms/react-form-renderer/component-types';
-import { Text } from '@patternfly/react-core';
+import { Text, Title } from '@patternfly/react-core';
 
 import StepTemplate from './stepTemplate';
 
+import DocumentationButton from '../../sharedComponents/DocumentationButton';
 import CustomButtons from '../formComponents/CustomButtons';
 
 const oscapStep = {
   StepTemplate,
   id: 'wizard-systemconfiguration-oscap',
-  title: 'OpenSCAP Compliance',
+  title: 'OpenSCAP',
   name: 'Compliance',
+  customTitle: (
+    <Title headingLevel="h1" size="xl">
+      OpenSCAP profile
+    </Title>
+  ),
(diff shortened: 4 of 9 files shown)


regexowl (approved):
Looks good! One teeny tiny suggestion as I'd say that the Documentation button looks better on the same line as the text, but I'll leave the final decision to you :)

src/Components/CreateImageWizard/steps/oscap.js
```diff
@@ -20,8 +26,9 @@ const oscapStep = {
       name: 'oscap-text-component',
       label: (
         <Text>
-          Monitor regulatory compliance profiles of registered RHEL systems you
-          must adhere to via OpenSCAP.
+          Use OpenSCAP to monitor the adherence of your registered RHEL systems
+          to a selected regulatory compliance profile.
+          <DocumentationButton />
```
regexowl, line 31: The documentation buttons needs to be on the same line as the text if we want to keep the space between the button and the text:

Same line:
image

Separate line: 
image


```suggestion
          Use OpenSCAP to monitor the adherence of your registered RHEL systems
          to a selected regulatory compliance profile. <DocumentationButton />
```

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -6,6 +6,10 @@
     overflow-y: unset;
 }
 
+.normal-text {
+  font-weight: normal;
```
regexowl, line 10: This won't be needed when the description list is implemented.

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -91,13 +110,52 @@ const ProfileSelector = ({ input, showSelector }) => {
...
+        </TextList>
+        <TextList component={TextListVariants.dl}>
+          <TextListItem
+            component={TextListItemVariants.dt}
+            className="pf-u-min-width"
+          >
+            Operating system:
+          </TextListItem>
+          <TextListItem
+            component={TextListItemVariants.dt}
```
regexowl, line 143: The TextLists work on a principal of term / description pair. So you'll have one `TextListItemVariants.dt` for the term and `TextListItemVariants.dd` for the description. Docs for this component are here: https://www.patternfly.org/components/text/#description-list

```suggestion
            component={TextListItemVariants.dd}
```

The `className="normal-text"` can also be removed as the `dd` formatting will keep the text not bold. :)

regexowl (commented):
Added a few more comments

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -73,11 +81,22 @@ const ProfileSelector = ({ input, showSelector }) => {
...
+        bodyContent={
+          <TextContent>
+            <Text>
+              Required selection for OpenSCAP, in order to make this work, user
+              cannot clear this field, the default value is None.
+            </Text>
+          </TextContent>
+        }
+      >
+        <HelpIcon className="pf-u-ml-sm" />
```
lucasgarfield, line 98: We should always avoid adding css whenever possible... which is actually most of the time, fortunately. :smile: 

```
<Button variant="plain" aria-label="About OpenSCAP">
    <HelpIcon />
</Button>
```

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -73,11 +78,24 @@ const ProfileSelector = ({ input, showSelector }) => {
...
 
   return (
-    <FormGroup
-      isRequired={true}
-      label={'Profile to use for this image'}
-      data-testid="profiles-form-group"
-    >
+    <FormGroup isRequired={true} data-testid="profiles-form-group">
+      {'OpenSCAP profile'}
+      <Popover
```
lucasgarfield, line 83: Let's actually just delete this popover for now, since we don't have good copy for it.

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -73,11 +78,24 @@ const ProfileSelector = ({ input, showSelector }) => {
   };
 
   return (
-    <FormGroup
-      isRequired={true}
-      label={'Profile to use for this image'}
-      data-testid="profiles-form-group"
-    >
+    <FormGroup isRequired={true} data-testid="profiles-form-group">
+      {'OpenSCAP profile'}
```
lucasgarfield, line 82: This should actually be passed into `<FormGroup>` as a `label` prop, like so: `<FormGroup label={'OpenSCAP profile'} ... />`.

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -91,13 +109,61 @@ const ProfileSelector = ({ input, showSelector }) => {
...
+              <SelectOption key={index + 1} value={key} />
+            ))
+          )}
         {isFetching && (
           <SelectOption isNoResultsOption={true} data-testid="policies-loading">
             <Spinner size="md" />
           </SelectOption>
         )}
       </Select>
+      <TextContent>
```
lucasgarfield, line 123: I think we should only show this `<TextContent>` if a profile has been selected.

lucasgarfield:
We've merged the back-end changes and I have created a PR to update the API... but it isn't merged yet, and probably won't be until Monday... however you can work on this on Sunday by rebasing onto #1471. Then you should be able to get the data you need to populate the Description and Profile ID fields!

src/Components/CreateImageWizard/formComponents/Oscap.js
mgold1234: I asked Kelsea about the content of OpenScap Profile when we go over it and she gave me this content    'To run a manual compliance scan in OpenSCAP, download this image.' i will change it

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -112,6 +136,24 @@ const ProfileSelector = ({ input, showSelector }) => {
...
 
+const OScapSelectOption = ({ key, profile_id, setProfileName }) => {
+  const { getState } = useFormApi();
+  const { data } = useGetOscapCustomizationsQuery({
+    distribution: getState()?.values?.['release'],
+    profile: profile_id,
+  });
+
+  return (
+    <SelectOption key={key} value={data?.openscap.profile_id}
```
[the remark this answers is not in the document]
mgold1234: thank you @lavocatt , I tried to add the spinner, but its seems that we need different solution here.
I added the onMouseEnter, thank you for mention that

src/Components/CreateImageWizard/formComponents/Oscap.js
mgold1234: if I added the label, the question mark moved under the label

lavocatt:
Thanks for your changes, with the code linted and the unit tests passing it's way better to dig into your contribution. I'm now running the PR on my machine with the `msw` and I'm noticing several issues:

* there is no data showing up in the dropdown menu
* the label `initial Profile Name` won't clear up upon clicking on the cross
* I would expect that the label `initial Profile Name` would disappear when I'm clicking on the text input to type and filter what I want to see
* I would expect that the label `initial Profile Name` would be in a different color than solid black.

image

To help you figure out how to change the behavior of the component you can take inspiration on some other type ahead select that have been implemented in the wizard. For instance:

https://github.com/RedHatInsights/image-builder-frontend/blob/fabd3e015a220dbfdb7aef0b30a43f9faa5bc9c8/src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js#L91-L154

src/Components/CreateImageWizard/formComponents/Oscap.js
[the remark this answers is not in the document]
lavocatt: I invite to you have a look on how it's done for other type ahead selects in the code base, for instance:

https://github.com/RedHatInsights/image-builder-frontend/blob/fabd3e015a220dbfdb7aef0b30a43f9faa5bc9c8/src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js#L91-L154

It's very important to check for loading and error states as it provides visual clues to the user that the application isn't stuck and is actually doing something.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -516,6 +518,8 @@ const requestToState = (composeRequest, distroInfo, isProd, enableOscap) => {
     if (enableOscap) {
       formState['oscap-profile'] =
         composeRequest?.customizations?.openscap?.profile_id;
+      composeRequest?.customizations?.openscap?.profile_name;
```
regexowl, line 521: Correct me if I'm wrong, but this is not really forming a state, right?

Shouldn't there be an assignment to the state?
```suggestion
      formState['oscap-profile-name'] = composeRequest?.customizations?.openscap?.profile_name;
```

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -516,6 +518,8 @@ const requestToState = (composeRequest, distroInfo, isProd, enableOscap) => {
...
+      composeRequest?.customizations?.openscap?.profile_description;
```
regexowl, line 522: Same here, I'd expect something like:
```suggestion
      formState['oscap-profile-description'] = composeRequest?.customizations?.openscap?.profile_description;
```

src/Components/CreateImageWizard/formComponents/Oscap.js
regexowl: In this case you'll need to make the popover button a part of the label, then it should be both on the same line and be styled correctly with the little star for a required field.

So something like this:
```javascript
    <FormGroup
      isRequired={true}
      data-testid="profiles-form-group"
      label={
        <>
          OpenSCAP profile
          <Popover
            maxWidth="30rem"
            position="left"
            bodyContent={
              <TextContent>
                <Text>
                  To run a manual compliance scan in OpenSCAP, download this
                  image.
                </Text>
              </TextContent>
            }
          >
            <Button variant="plain" aria-label="About OpenSCAP" isInline>
              <HelpIcon />
            </Button>
          </Popover>
        </>
      }
    >
```

will result in this:
image

You could also add `className="pf-u-p-0"` to the button to make it a bit more compact, but as we have similar spacing in few other places in the Wizard I wouldn't be bothered with it in this PR. :)

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -59,30 +69,42 @@ const ProfileSelector = ({ input, showSelector }) => {
...
+              To run a manual compliance scan in OpenSCAP, download this image.
+            </Text>
+          </TextContent>
+        }
+      >
+        <Button variant="plain" aria-label="About OpenSCAP">
+          <HelpIcon />
+        </Button>
+      </Popover>
       <Select
```
regexowl, line 103: Should the select work as a typeahead select? Typing into it doesn't seem to filter the results. Also when you type something and hit enter to possible trigger the filtering the whole wizard jumps back to the first step.

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -59,30 +69,42 @@ const ProfileSelector = ({ input, showSelector }) => {
...
         ouiaId="profileSelect"
         variant={SelectVariant.typeahead}
         onToggle={handleToggle}
-        onSelect={setProfile}
+        onSelect={handleSelect}
```
regexowl, line 107: When a profile is selected the select contains a reference ID instead of the profile name.

regexowl (commented):
Thank you for all the work, the step is looking really neat! :sparkles: 

I've added some comments. Also wanted to ask how we are thinking about the step now. I'm not really up to date with this, but I've noticed "None" is no longer there as an profile option. Instead we have the "Select a profile" placeholder which works for me, but then we maybe don't want to mark the field as required?

The thing is, with it being required there's a situation when a profile is not selected, but the Next button is not disabled which goes against the mandatory field I think.

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -516,6 +516,10 @@ const requestToState = (composeRequest, distroInfo, isProd, enableOscap) => {
     if (enableOscap) {
       formState['oscap-profile'] =
         composeRequest?.customizations?.openscap?.profile_id;
+      formState['oscap-profile-name'] =
+        composeRequest?.customizations?.openscap?.profile_name;
+      formState['oscap-profile-description'] =
+        composeRequest?.customizations?.openscap?.profile_description;
```
lavocatt, line 522: In this function we are loading a `composeRequest` into a state for the wizard. However these two values are not going to endup in the request, so they will never get to be loaded. It is thus safe to remove these lines.

But the fact that the data can't be loaded from a request will have side effects for the wizard.

It means that, upon recreating a request, when entering the `oscap` step, we will have the issue where the `profileName` is initialized to an empty string. 

Let's fix this :slightly_smiling_face:

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -30,26 +50,22 @@ const reinitDependingSteps = (change) => {
...
  * The Component is shown or not depending on the ShowSelector variable.
  */
-const ProfileSelector = ({ input, showSelector }) => {
+const ProfileSelector = ({ input }) => {
   const { change, getState } = useFormApi();
-  const [profile, selectProfile] = useState(
-    getState()?.values?.['oscap-profile']
+  const [profile, setProfile] = useState(getState()?.values?.['oscap-profile']);
+  const [profileName, setProfileName] = useState(
+    getState()?.values?.['openscap-profile-name']
```
lavocatt, line 57: In order to fix the issue, let's not use a form state here (sorry for pointing you in the wrong direction earlier). We will work with the  `useGetOscapCustomizationsQuery` hook and a `useEffect`.

The main idea is:
1. the `profileName` gets initialized with an empty string all the time.
2. `useGetOscapCustomizationsQuery` is called with the selected `profile` and distribution (same as you did at line 185)
3. then we will use a `useEffect` to propagate the data to the `profileName`

I've tested this solution on my local copy and it works well. Clicking back and forth in the wizard keeps the value, the clear button clear the field and recreating an image works fine.

If you need any help with this I'll be back on Monday to give you a hand.

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -59,13 +75,15 @@ const ProfileSelector = ({ input, showSelector }) => {
   };
 
   const handleClear = () => {
-    selectProfile(undefined);
+    setProfile(undefined);
     change(input.name, undefined);
+    setProfileName(undefined);
+    change('openscap-profile-name', '');
```
lavocatt, line 81: We will also need to remove this call to `change`

src/Components/CreateImageWizard/formComponents/Oscap.js
```diff
@@ -112,6 +163,48 @@ const ProfileSelector = ({ input, showSelector }) => {
...
+    distribution: getState()?.values?.['release'],
+    profile: profile_id,
+  });
+
+  return (
+    <SelectOption
+      value={data?.openscap?.profile_id}
+      onClick={() => {
+        setProfileName(data?.openscap?.profile_name);
+        change('openscap-profile-name', data?.openscap?.profile_name);
```
lavocatt, line 195: And this other one too.

lavocatt (commented):
Thanks for the last changes! I realized that, that during our debugging session I misguided you into using a form state, only to realize later by looking a your PR that it wouldn't work for image recreation. (you'll find more details about why in the comments below)

So we need to change the approach and use the rtk query + a useEffect to make this work correctly.

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -84,36 +87,66 @@ describe('Step Compliance', () => {
...
+        name: /select a profile/i,
       })
     );
 
+    await user.click(
+      await screen.findByText(
+        /cis red hat enterprise linux 8 benchmark for level 1 - workstation/i
+      )
+    );
     // check that the FSC contains a /tmp partition
```
regexowl, line 140: The tests don't check the `/tmp` mount point. Which they probably should in this place. 

The structure of the test could be following:
1. get to the OpenSCAP step
2. select a profile
3. click Next
4. check if there is a mount point which is defined in the profile
5. click Next
6. check if there is/are package/s added based on the profile

src/test/fixtures/composes.ts
```diff
@@ -317,6 +317,10 @@ export const mockComposes: ComposesResponseItem[] = [
         ],
         openscap: {
           profile_id: 'xccdf_org.ssgproject.content_profile_cis_workstation_l1',
+          profile_name:
+            'CIS Red Hat Enterprise Linux 8 Benchmark for Level 1 - Workstation',
+          profile_description:
+            'This profile defines a baseline that aligns to the "Level 1 - Workstation" configuration from the Center for Internet Security® Red Hat Enterprise Linux 8 Benchmark™, v2.0.0, released 2022-02-23.  This profile includes Center for Internet Security® Red Hat Enterprise Linux 8 CIS Benchmarks™ content.',
```
regexowl, line 323: I'm wondering whether the mocks should use real data or data that re obviously mocked. I'm mentioning the same thing in a comment for `oscap.ts` fixture. Will start a thread about it.

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -84,36 +87,66 @@ describe('Step Compliance', () => {
...
     await clickNext();
     await screen.findByRole('heading', { name: /File system configuration/i });
-    await screen.findByText('/tmp');
 
     // check that the Packages contain a nftable package
```
regexowl, line 144: The comment does not reflect what happens in the tests.

lavocatt (approved):
Everything works very well. Great Job Michal!

lavocatt:
@jrusz I think this PR is breaking the IQE tests by changing how the OpenScap step is working. Can you check it out please?

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -124,8 +157,9 @@ describe('On Recreate', () => {
   test('with oscap profile', async () => {
     const user = userEvent.setup();
     await setup();
-
-    await screen.findByRole('heading', { name: /review/i });
+    screen.getByRole('button', {
+      name: /review/i,
+    });
```
regexowl, line 162: As most of the testing runs asynchronously the first option should be in my opinion be using the combination of `await` + `findBy` query.

The difference between `getBy` and `findBy` is that `getBy` tries to get an element instantaneously and fails if the element haven't been rendered yet. This could cause a certain level of flakiness in this case.

The combination of `await` and `findBy` is able to wait for the element to be rendered and therefore doesn't throw an error in case that didn't happen immediately.

Why did the two get switched here? Was the async `findBy` causing any problems in the tests?

src/store/imageBuilderApi.ts
```diff
@@ -295,8 +295,8 @@ export type CustomRepository = {
 };
 export type OpenScap = {
   profile_id: string;
-  profile_name?: string;
-  profile_description?: string;
+  profile_name: string;
+  profile_description: string;
```
regexowl, line 299: This looks like a manual change to the generated code which means that when the API schema gets updated and the `npm run api` is run the change will get overwritten and lost.

If the change is needed it should be done directly in the image-builder API so it gets propagated consistently. Do the name and description need to be required?

src/Components/CreateImageWizard/formComponents/ReviewStepTextLists.js
```diff
@@ -618,14 +619,41 @@ export const ImageDetailsList = () => {
...
+
   return (
     <TextContent>
       <TextList component={TextListVariants.dl}>
         <TextListItem
           component={TextListItemVariants.dt}
           className="pf-u-min-width"
         >
-          Profile
+          Profile Name:
```
regexowl, line 632: For consistency:

```suggestion
          Profile name:
```

jrusz:
> @jrusz I think this PR is breaking the IQE tests by changing how the OpenScap step is working. Can you check it out please?

We don't build OpenScap images in the tests though so it should be fine. I'll check out the `pr_check` once it finishes if there is anything else broken.

regexowl (changes requested):
This looks amazing. Thanks for all the work you've done on this! Still have some code to go through, but wanted to submit a first batch of comments.

lavocatt:
> This looks amazing. Thanks for all the work you've done on this! Still have some code to go through, but wanted to submit a first batch of comments.

Good idea! Thanks for preventing a too enthusiastic merge.

src/store/imageBuilderApi.ts
lavocatt: Is it a manual change? Travis should catch it then! But I think Travis is broken :sob:

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
lavocatt: Oh yes! Good catch! This is super important to do.

src/Components/CreateImageWizard/formComponents/OscapProfileInformation.tsx
```diff
@@ -0,0 +1,78 @@
...
+const OscapProfileInformation = (): JSX.Element => {
+  const { getState } = useFormApi();
+
+  const oscapProfile = getState()?.values?.['oscap-profile'];
+
+  const {
+    data: oscapProfileInfo,
+    isFetching: isFetchingOscapProfileInfo,
+    isSuccess: isSuccessOscapProfileInfo,
+    isError: isErrorOscapProfileInfo,
```
regexowl, line 25: The `isErrorOscapProfileInfo` seems to be unused so it can be removed.

src/store/imageBuilderApi.ts
regexowl: Great point! As we are planning to no longer depend on Travis anyway, we should probably add the check to our dev check actions. I'll make a note and take a look at it.

regexowl: Actually - general question about the name and description in the API - do we even use it in the code? What I mean is, I've tried to add an OSCAP profile to a mock to see what I get on Recreate.

Updated a compose fixture with:

image

But what I see on Recreate is:

image

And now I'm a tad bit confused :sweat_smile:

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
mgold1234: if I dont change it, the test fail    await screen.findByText('button', {
          |                  ^
      168 |       name: /review/i,
      169 |     });

src/store/imageBuilderApi.ts
mgold1234: I run the npm run api and its adding those rows automatically, I didnt add it manually .

regexowl: Yup, the rows should be added automatically. I was wondering about the removed question marks in one of the previous commits. But that could have also been a mismatch between schema and API version before :+1:  :)

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
regexowl: Hmm, it does seem to pass for me with this:

```suggestion
    await screen.findByRole('button', {
      name: /review/i,
    });
```

Wonder what the difference is. But also this shouldn't be a blocker I'd say. There should be some check and update of the tests anyway in the not so distant future as we are still using `getBy` in few other places.

jrusz:
The whole step is not visible in ephemeral environment. Is it hidden behind a feature flag?

regexowl:
> The whole step is not visible in ephemeral environment. Is it hidden behind a feature flag?

Yes, it should be gated behind `image-builder.wizard.oscap.enabled` (https://github.com/RedHatInsights/image-builder-frontend/pull/1392). Enabled on stage, disabled on prod I think?

src/store/imageBuilderApi.ts
lavocatt: Regarding the check that's done in travis, you can move it to to a github action easily, it's just a bash oneliner.

regexowl: Tadaa :tada: https://github.com/RedHatInsights/image-builder-frontend/pull/1510

jrusz:
Ok I've updated the IQE plugin with the new step name. We don't do do anything else on that step besides clicking next currently..

mgold1234:
thank you @jrusz

src/test/fixtures/composes.ts
regexowl, line 323: Can we make this a mocked description as well?

```suggestion
            'This is a mocked profile description. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean posuere velit enim, tincidunt porttitor nisl elementum eu.',
```

src/test/fixtures/composes.ts
```diff
@@ -814,6 +818,10 @@ export const mockStatus = (composeId: string): ComposeStatus => {
           openscap: {
             profile_id:
               'xccdf_org.ssgproject.content_profile_cis_workstation_l1',
+            profile_name:
+              ' CIS Red Hat Enterprise Linux 8 Benchmark for Level 1 - Workstation',
+            profile_description:
+              'This profile defines a baseline that aligns to the "Level 1 - Workstation" configuration from the Center for Internet Security® Red Hat Enterprise Linux 8 Benchmark™, v2.0.0, released 2022-02-23.  This profile includes Center for Internet Security® Red Hat Enterprise Linux 8 CIS Benchmarks™ content.',
```
regexowl, line 824: And this also

```suggestion
              'This is a mocked profile description. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean posuere velit enim, tincidunt porttitor nisl elementum eu.',
```

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -84,36 +87,73 @@ describe('Step Compliance', () => {
...
-      await screen.findByRole('radio', { name: 'Add a profile' })
-    ).toBeChecked();
+    await user.click(screen.getByText(/none/i));
+
+    // check that the FSC does not contain a /tmp partition
+    await clickNext();
+    await screen.findByRole('heading', { name: /File system configuration/i });
+    await screen.findByRole('cell', {
+      name: /xfs/i,
+    });
```
regexowl, line 102: I see, in case there would be more partitions the test would fail as there would be multiple cells with "xfs" as a content. I don't really mind it, but it might not be immediately obvious. What you can do is simply query for the tmp mount point and check it's not there when everything renders.

Which can be done like this:

```suggestion
    expect(
      screen.queryByRole('cell', {
        name: /tmp/i,
      })
    ).not.toBeInTheDocument();
```

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -84,36 +87,73 @@ describe('Step Compliance', () => {
...
+      screen.getByRole('radio', { name: /manually enter an account id\./i })
     );
+    await user.type(screen.getByTestId('aws-account-id'), '012345678901');
+
+    await clickNext();
+    // skip registration
+    await user.click(screen.getByLabelText('Register later'));
+    await clickNext();
+
+    // Now we should be in the Compliance step
```
regexowl, line 130:
```suggestion
    // Now we should be at the OpenSCAP step
```

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -84,36 +87,73 @@ describe('Step Compliance', () => {
...
+    await user.click(screen.getByText(/none/i));
+
+    // check that the FSC does not contain a /tmp partition
+    await clickNext();
+    await screen.findByRole('heading', { name: /File system configuration/i });
+    await screen.findByRole('cell', {
+      name: /xfs/i,
+    });
+
+    // check that there is no Packages contain by selecting the "None" profile option
```
regexowl, line 104:
```suggestion
    // check that there are no Packages contained when selecting the "None" profile option
```

regexowl (changes requested):
This looks great! Thank you so much :heart: 

I've added a few nitpicks, we can merge after those are in!

regexowl:
Considering the Travis failure - from what I was able to find it could be a problem with the npm cache as the `npm ci` command seems to be failing only for Travis. Correct me if I'm wrong, but locally it's fine, right? Also there are no changes to `packages.json` so not really sure what glitched on the clean install.

I can see two solutions - we can somehow try to completely re-trigger the run inside Travis. Or since we plan to drop Travis anyway and the same check seems to be passing in dev checks we could maybe just ignore it? What do you think @lucasgarfield? Would you be okay with merging with failed Travis or should we look into it more?

regexowl:
Although - thinking about the failing Travis CI - couldn't we just add flushing the cache and see what it does?

I mean adding a line to clean the npm cache to this PR, checking if it fixes Travis and if it does we can remove the line before merging the PR. This shouldn't affect anything outside then. Does it make sense? Not sure if I'm overlooking something.

package.json
```diff
@@ -62,7 +62,7 @@
     "@babel/preset-react": "7.23.3",
     "@babel/preset-typescript": "7.23.3",
     "@redhat-cloud-services/eslint-config-redhat-cloud-services": "2.0.3",
-    "@redhat-cloud-services/frontend-components-config": "5.0.5",
+    "@redhat-cloud-services/frontend-components-config": "^4.5.12",
```
regexowl, line 65: This looks strange. Is this an intended change @mgold1234?

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
mgold1234: I covered that test in row 149   // check that the FSC contains a /tmp partition
    await clickNext();
    await screen.findByRole('heading', { name: /File system configuration/i });
    await screen.findByText(/tmp/i);

regexowl (approved):
This looks great! Thank you for all your work on this!! :heart:

regexowl:
As the CI was passing when the ephemeral cluster was up and the changes since then were minimal I think it should be safe to merge this. Is that okay @jrusz?

jrusz:
> As the CI was passing when the ephemeral cluster was up and the changes since then were minimal I think it should be safe to merge this. Is that okay @jrusz?

Yeah, I can trigger a manual test run in stage afterwards to make sure.

PR #1456, opened 2023-11-23
Wizard: Remove hardcoded architecture

regexowl:
This updates requests that contained hardcoded value `x86_64` as an architecture.


jkozol (approved):
Nice to get rid of the hardcoding :)

PR #1461, opened 2023-11-24
[WIP] Migrate from Jest framework to vitest

regexowl:
~~This is a work in progress, the main setup should be finished, but part of the tests is currently not running because of the following issue:~~

```javascript
SyntaxError: Cannot use import statement outside a module
 ❯ Object.<anonymous> node_modules/@scalprum/react-core/src/scalprum-provider.tsx:4:1

Module /home/ksimicko/Projects/image-builder-frontend/node_modules/@scalprum/react-core/node_modules/@openshift/dynamic-plugin-sdk/dist/index.js:13 seems to be an ES Module but shipped in a CommonJS package. You might want to create an issue to the package "@openshift/dynamic-plugin-sdk" asking them to ship the file in .mjs extension or add "type": "module" in their package.json.

As a temporary workaround you can try to inline the package by updating your config:

// vitest.config.js
export default {
  test: {
    server: {
      deps: {
        inline: [
          "@openshift/dynamic-plugin-sdk"
        ]
      }
    }
  }
}
```

Everything's still broken, but at least all the tests are trying to run now.


regexowl:
Re-created the PR https://github.com/RedHatInsights/image-builder-frontend/pull/1526

PR #1464, opened 2023-11-27
Wizard: Fix HMS-2036, bug in file system customization steps

lucasgarfield:
Information about automatic partioning was also being displayed when manual partioning was selected. This commit fixes the bug.

### DIFF
--- a/src/Components/CreateImageWizard/steps/fileSystemConfiguration.js
+++ b/src/Components/CreateImageWizard/steps/fileSystemConfiguration.js
@@ -125,7 +125,8 @@ const fileSystemConfigurationStep = {
         </TextContent>
       ),
       condition: {
-        or: [{ when: 'file-system-config-radio', is: 'automatic' }],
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/steps/fileSystemConfiguration.js
```diff
@@ -125,10 +125,7 @@ const fileSystemConfigurationStep = {
         </TextContent>
       ),
       condition: {
-        or: [
-          { when: 'file-system-config-radio', is: 'automatic' },
-          { when: 'oscap-profile', is: undefined },
-        ],
+        or: [{ when: 'file-system-config-radio', is: 'automatic' }],
```
regexowl, line 128: We could also remove the `or` as there is only one condition now.

```suggestion
        when: 'file-system-config-radio',
        is: 'automatic',
```

regexowl (approved):
LGTM!

PR #1473, opened 2023-12-04
test: Await asynchronous `setUp()`

regexowl:
This adds await to call of asynchronous `setUp()` function. Previously this wasn't awaited which sometimes caused the tests to continue without being properly set up first.


lucasgarfield (approved):
lgtm

PR #1482, opened 2023-12-11
dependabot: Bump open PR limit

regexowl:
This bumps a maximum number of opened dependabot PRs as there are currently three open PRs in need of manual revision. Those are blocking any other bumps from being even opened.

The bump should be temporary.


lucasgarfield (approved):
lgtm

PR #1487, opened 2023-12-12
store: Remove `actions.js`

regexowl:
This removes the `actions.js` file and its mocking from the LandingPage test as these actions are no longer used.


croissanne (approved):
red diffs are best diffs, thank you!

PR #1489, opened 2023-12-13
Blueprints: add blueprint feature

jkozol:
Blueprints can now be created through the Create Blueprint wizard. This is almost identical to the Create Image wizard except for a couple blueprint specific fields.

Created blueprints are shown in the blueprints sidebar where they can be filtered or selected.

This commit also includes improvements to the general UI per the recent design updates. The header is now Images, filtering has been added to the images table, and the Create Image button is moved to the header.

Given that in image-builder blueprints are still experimental these new features can be seen when running stage-beta:msw+experimental. The fixtures mimic blueprints even though the cannot currently be fetched from the service.

Screenshot 2023-12-13 at 02 46 17

### DIFF
--- a/src/Components/Blueprints/BlueprintCard.tsx
+++ b/src/Components/Blueprints/BlueprintCard.tsx
@@ -66,7 +66,7 @@ const BlueprintCard = (props: {
   );
 
   return (
-    <Card>
+    <Card className="pf-v5-u-mb-md">
       <CardHeader actions={{ actions: headerActions }}>
         <CardTitle>{props.blueprint.name}</CardTitle>
       </CardHeader>
--- a/src/Components/Blueprints/BlueprintCard.tsx
+++ b/src/Components/Blueprints/BlueprintCard.tsx
@@ -14,7 +14,7 @@ import {
 } from '@patternfly/react-core';
 import { EllipsisVIcon } from '@patternfly/react-icons';
 
-import { Blueprint } from '../../store/imageBuilderApi';
+import { Blueprint } from '../../store/imageBuilderApiExperimental';
 
 const BlueprintCard = (props: {
   blueprint: Blueprint;
--- a/src/Components/Blueprints/BlueprintCard.tsx
+++ b/src/Components/Blueprints/BlueprintCard.tsx
@@ -1,0 +1,78 @@
+import React from 'react';
+
+import {
+  Card,
+  CardHeader,
+  CardTitle,
+  CardBody,
+  Radio,
+  Dropdown,
+  DropdownList,
+  MenuToggle,
+  MenuToggleElement,
+  DropdownItem,
+} from '@patternfly/react-core';
+import { EllipsisVIcon } from '@patternfly/react-icons';
+
+import { Blueprint } from '../../store/imageBuilderApi';
+
+const BlueprintCard = (props: {
+  blueprint: Blueprint;
+  setSelectedBlueprint: React.Dispatch<React.SetStateAction<string>>;
+  selectedBlueprint: string;
+}) => {
+  const [isOpen, setIsOpen] = React.useState<boolean>(false);
+  const isChecked = props.blueprint.id === props.selectedBlueprint;
+  const onSelect = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const handleChange = () => {
+    props.setSelectedBlueprint(props.blueprint.id);
+  };
+
+  const headerActions = (
+    <>
+      <Dropdown
+        onSelect={onSelect}
+        toggle={(toggleRef: React.Ref<MenuToggleElement>) => (
+          <MenuToggle
+            ref={toggleRef}
+            isExpanded={isOpen}
+            onClick={() => setIsOpen(!isOpen)}
+            variant="plain"
+            aria-label="Card expandable toggle"
+          >
+            <EllipsisVIcon aria-hidden="true" />
+          </MenuToggle>
+        )}
+        isOpen={isOpen}
+        onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
+      >
+        <DropdownList>
+          <DropdownItem key="disabled action" isDisabled>
+            Delete
+          </DropdownItem>
+        </DropdownList>
+      </Dropdown>
+      <Radio
+        isChecked={isChecked}
+        onChange={handleChange}
+        aria-label="blueprint checkbox"
+        id={props.blueprint.id}
+        name={props.blueprint.id}
+      />
+    </>
+  );
+
+  return (
+    <Card>
+      <CardHeader actions={{ actions: headerActions }}>
+        <CardTitle>{props.blueprint.name}</CardTitle>
+      </CardHeader>
+      <CardBody>{props.blueprint.description}</CardBody>
+    </Card>
+  );
+};
+
+export default BlueprintCard;
--- a/src/Components/Blueprints/BlueprintsSidebar.tsx
+++ b/src/Components/Blueprints/BlueprintsSidebar.tsx
@@ -67,14 +67,12 @@ const BlueprintsSidebar = (props: blueprintProps) => {
         className="pf-v5-u-mb-md"
       />
       {filteredBlueprints?.map((blueprint: Blueprint) => (
-        <>
-          <BlueprintCard
-            key={blueprint.id}
-            blueprint={blueprint}
-            selectedBlueprint={props.selectedBlueprint}
-            setSelectedBlueprint={props.setSelectedBlueprint}
-          />
-        </>
+        <BlueprintCard
+          key={blueprint.id}
+          blueprint={blueprint}
+          selectedBlueprint={props.selectedBlueprint}
+          setSelectedBlueprint={props.setSelectedBlueprint}
+        />
       ))}
     </>
   ) : (
--- a/src/Components/Blueprints/BlueprintsSidebar.tsx
+++ b/src/Components/Blueprints/BlueprintsSidebar.tsx
@@ -1,10 +1,22 @@
 import React, { useState, Dispatch, SetStateAction } from 'react';
 
-import { SearchInput } from '@patternfly/react-core';
+import {
+  Button,
+  EmptyState,
+  EmptyStateActions,
+  EmptyStateBody,
+  EmptyStateFooter,
+  EmptyStateHeader,
+  EmptyStateIcon,
+  SearchInput,
+} from '@patternfly/react-core';
+import { CubesIcon } from '@patternfly/react-icons';
 
 import BlueprintCard from './BlueprintCard';
 
 import { Blueprint } from '../../store/imageBuilderApiExperimental';
+import { Link } from 'react-router-dom';
+import { resolveRelPath } from '../../Utilities/path';
 
 interface blueprintProps {
   blueprints: Blueprint[];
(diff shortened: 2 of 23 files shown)


src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -63,6 +72,74 @@ import {
...
+  const [isTypeSelectOpen, setIsTypeSelectOpen] = useState(false);
+  const [isVersionSelectOpen, setIsVersionSelectOpen] = useState(false);
+  const [filters, setFilters] = useState<{
+    imageType: string[];
+    version: string[];
+  }>({
+    imageType: [],
+    version: [],
+  });
+  const experimentalFlag = process.env.EXPERIMENTAL;
```
amirfefer, line 84: do we want to use feature flag here?

jkozol: I do want to but have been having lots of issues testing against stage the past week to verify that the flag actually works.

lucasgarfield:
There are couple of discrepancies I noticed compared to the designs (which can be found here: https://www.sketch.com/s/669c80df-098c-4296-b689-b4e0702d8261/prototype/8DD94999-9DAC-4C66-9C95-3AAF67C2281E/a/8DD94999-9DAC-4C66-9C95-3AAF67C2281E).

In the images table:

1. The buttons at the top should be “Create” and “Build images”. The “Build images” button should only be enabled when the user has selected a blueprint to use, and should rebuild images from it.

This PR:
image

The designs:
image

2. Selecting one of the blueprint cards should apply a filter to the table and show only images associated with that blueprint. No filter is applied in this PR. Additionally, the card selectors are currently using radio selectors instead of check boxes as shown in the designs. I think this is actually an important distinction, as with the radio selectors there is no way to deselect a blueprint in order to remove the filter.

This PR:
blueprint review

3. There should only be one wizard. In this PR there are two routes, ‘/blueprintwizard’ and ‘/imagewizard’. There should only be one ‘/imagewizard’ route, as before. On the review step of the wizard, the user will have two options: “Save” and “Save and build”. “Save” will create the blueprint, “Save and build” will create the blueprint and . This isn’t in the designs, unfortunately, but has been discussed and agreed to in the UX meetings.

I think that we can just put this functionality into the V2 wizard, and leave the original DDF wizard alone. That way everything related to blueprints stays in the experimental build, making development easier for us.

And then there are a few things where I think we should deviate from the designs, as they may have been oversights…

1. The table should have some padding so that it appears as a white box floating against a grey background, which is standard for Patternfly tables. I don’t think that adding the sidebar to the table should “undo” this default styling.
2. The “Create” and “Build images” buttons shouldn’t appear as they do now when the wizard is open.
image
3. Since we don’t have everything that we need to support ostree images with blueprints yet, we should probably remove that filter from the table. It isn’t functional now so removing it temporarily and adding it later shouldn’t be too much work.
4. We should also remove the ability to filter on version, since versions won’t be supported in the first iteration of the blueprint API. As with the ostree filter, it isn’t functional now so removing it temporarily and adding it later shouldn’t be too much work.

jkozol:
Thank you for the review. Before addressing the individual points I'd like to clarify that my goal with this PR was to get in the framework for supporting blueprints in the table with both mocking and with actual functionality by calling the POST request to create a blueprint.

>The buttons at the top should be “Create” and “Build images”. The “Build images” button should only be enabled when the user has selected a blueprint to use, and should rebuild images from it.

I will change it to "Build images" but I think that until the blueprint migration is complete "Create blueprint" is more descriptive. Also, creating a compose from an existing blueprint was still being changed and updated when submitting this PR so there was no clear way to support this at the time. I can add it either in this PR or a follow up.

> Selecting one of the blueprint cards should apply a filter to the table and show only images associated with that blueprint. No filter is applied in this PR. Additionally, the card selectors are currently using radio selectors instead of check boxes as shown in the designs. I think this is actually an important distinction, as with the radio selectors there is no way to deselect a blueprint in order to remove the filter.

This filter requires the incoming `blueprint_id` and/or `blueprint_version` fields. These are shown in [this PR](https://github.com/osbuild/image-builder/pull/938/files#diff-7fca9ba0424b3a6e43a52da6d8e2889451066b3e08291576af523bb05801d028). Mocking this data id would require creating experimental mocks for many of our existing fixtures and creates collisions between the mocked functionality and the actual functionality. So, I've left it out for now.

The radio selectors was a result of conversations in the UI/UX call a few weeks ago. The rebuilding of blueprint images can only be done on one blueprint. For removing the filter I can add a chip at the top that is clearable.

> There should only be one wizard. In this PR there are two routes, ‘/blueprintwizard’ and ‘/imagewizard’. There should only be one ‘/imagewizard’ route, as before. On the review step of the wizard, the user will have two options: “Save” and “Save and build”. “Save” will create the blueprint, “Save and build” will create the blueprint and and build the images.

Blueprints can be updated with just a save but creating a blueprint must build an image currently. If we want to change that we should bring it up in a blueprints meeting.

> I think that we can just put this functionality into the V2 wizard, and leave the original DDF wizard alone. That way everything related to blueprints stays in the experimental build, making development easier for us.

I agree. I was just trying to keep image builds functional so I could check that blueprint images are compatible with the existing images table. Once the we decide on a data structure for the ddf wizard I can add the blueprint create to it. 

> The table should have some padding so that it appears as a white box floating against a grey background, which is standard for Patternfly tables. I don’t think that adding the sidebar to the table should “undo” this default styling.
>
> The “Create” and “Build images” buttons shouldn’t appear as they do now when the wizard is open.

Will do.

> Since we don’t have everything that we need to support ostree images with blueprints yet, we should probably remove that filter from the table. It isn’t functional now so removing it temporarily and adding it later shouldn’t be too much work.

Ya this doesn't take much work although I don't see the harm in displaying the filter for "conventional" image type. But, I can strip it without anything changing.

> We should also remove the ability to filter on version, since versions won’t be supported in the first iteration of the blueprint API. As with the ostree filter, it isn’t functional now so removing it temporarily and adding it later shouldn’t be too much work.

I think we should keep this. The versioning will be added when the blueprint_id is added which is required to do any filtering anyway.

lucasgarfield:
Originally I had hoped that we would take advantage of MSW to develop this against what we felt would be the ideal API for the front end, and use our findings to inform the design of the image builder API. But at this stage, we’re getting really close to having full CRUD ready in image builder so I think a change in strategy is in order. Let’s go back to our normal way of working – where we implement up to what is available from the image builder API.

This is how I think we can get this PR merged:

1. Get this PR to match the current designs. If something isn’t ready yet in the API, then just put in ‘futures’ for it. For instance, regarding the `Save` vs. `Save and Build` button… just let the `Save` handler callback be `() => {}` for now if the API doesn’t support it yet. Same goes for the `Create` and `Build` buttons. This is all under experimental, so it’s fine if some things aren’t functional. This also means using only the V2 wizard. And then please communicate what needs to be done to oezr’s team.

2. Regarding the radio select vs. check boxes, the PF5 guidelines state: `Use checkboxes if the user can select multiple options. If the user can only select 1 option, then use radio buttons.`

I don’t recall these discussions in the UX meetings… perhaps it was related to rate limiting concerns? I could definitely see some problems occurring if someone checked and try to rebuild all the blueprints. Let’s talk to UX and get them to update the designs and go with the radios for now. Please add the chip to clear the filter.

3. Regarding the filters for the version and ostree vs rpm, that’s fine, we can keep them for now.  

4. Regarding the fixtures, I don’t understand the collision issue. Let’s just get the experimental fixtures set up correctly so that some of the images are associated with blueprint1, some are associated with blueprint 2, and some are not associated with a blueprint at all. We’ll have to do that eventually anyway so let’s just do it now.

5. There’s no test coverage. I know it’s really tempting to say “oh we’ll do that in a future PR”. I’ve said/thought that myself many times. But my new year’s resolution is to be more disciplined about this. Let’s write tests for the functionality that we have in place now – or will have in place for this PR. At a minimum, we should check that filtering for blueprints works.

src/Components/Blueprints/BlueprintCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <>
+      <Dropdown
+        onSelect={onSelect}
+        toggle={(toggleRef: React.Ref<MenuToggleElement>) => (
+          <MenuToggle
+            ref={toggleRef}
+            isExpanded={isOpen}
+            onClick={() => setIsOpen(!isOpen)}
+            variant="plain"
+            aria-label="Card expandable toggle"
```
amirfefer, line 44: the aria label should be informative  and referring the object.

src/Components/Blueprints/BlueprintCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+          >
+            <EllipsisVIcon aria-hidden="true" />
+          </MenuToggle>
+        )}
+        isOpen={isOpen}
+        onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
+      >
+        <DropdownList>
```
amirfefer, line 52: `Edit` is missing

src/Components/Blueprints/BlueprintCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+      />
+    </>
+  );
+
+  return (
+    <Card>
+      <CardHeader actions={{ actions: headerActions }}>
+        <CardTitle>{props.blueprint.name}</CardTitle>
+      </CardHeader>
+      <CardBody>{props.blueprint.description}</CardBody>
```
amirfefer, line 73: No `Version`?

src/Components/Blueprints/BlueprintsSidebar.tsx
```diff
@@ -0,0 +1,48 @@
...
+      />
+      {filteredBlueprints?.map((blueprint: Blueprint) => (
+        <>
+          <BlueprintCard
+            key={blueprint.id}
+            blueprint={blueprint}
+            selectedBlueprint={selectedBlueprint}
+            setSelectedBlueprint={setSelectedBlueprint}
+          />
+          <br />
```
amirfefer, line 41: nit: `<br>` is for line breaks, for space between objects it is better to use css

src/Components/Blueprints/BlueprintCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const BlueprintCard = (props: {
+  blueprint: Blueprint;
+  setSelectedBlueprint: React.Dispatch<React.SetStateAction<string>>;
+  selectedBlueprint: string;
+}) => {
+  const [isOpen, setIsOpen] = React.useState<boolean>(false);
+  const isChecked = props.blueprint.id === props.selectedBlueprint;
+  const onSelect = () => {
+    setIsOpen(!isOpen);
```
amirfefer, line 27: I guess this should impact the images table, and renders only the corresponding images. right?

amirfefer (commented):
Thanks @jkozol, looks great. I left a few comments

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -89,22 +166,128 @@ const ImagesTable = () => {
...
+                      onSelect={onTypeSelect}
+                      selected={filters.imageType}
+                      isOpen={isTypeSelectOpen}
+                      onOpenChange={(isOpen) => setIsTypeSelectOpen(isOpen)}
+                    >
+                      {typeMenuItems}
+                    </Select>
+                  </ToolbarFilter>
+                  <ToolbarFilter
+                    categoryName="version"
```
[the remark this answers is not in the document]
jkozol: Currently no.

src/Components/LandingPage/LandingPage.tsx
```diff
@@ -47,13 +53,41 @@ export const LandingPage = () => {
...
+      <PageSection>
+        <Quickstarts />
+      </PageSection>
+      <PageSection variant={PageSectionVariants.light}>
+        <Sidebar hasBorder hasGutter>
+          <SidebarPanel>
+            <BlueprintsSidebar />
+          </SidebarPanel>
+          <SidebarContent>
+            <ImagesTable />
```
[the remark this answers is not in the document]
jkozol: Updated now with the new api schema.

lucasgarfield:
Superseded by https://github.com/RedHatInsights/image-builder-frontend/pull/1571

PR #1495, opened 2023-12-14
API/test: Remove getVersion() Axios call

regexowl:
This removes `getVersion()` API call previously used in Axios.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -37,16 +37,9 @@ async function getPackagesContentSources(repoUrls, search) {
   return packages;
 }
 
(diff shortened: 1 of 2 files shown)


regexowl:
> Do we actually use getVersion() anywhere in the app? I don't think we do... maybe we should just remove it entirely? If we need it again it would be easy enough to add back.

Good point. It seemed to be used only for mocking in the tests. I've removed the RTK Query part.

lucasgarfield (approved):
Nice! Thank you!

PR #1496, opened 2023-12-14
api/imageBuilder: fix the alert in case of failure

lavocatt:
When a compose request was failing, the promises callback function was crashing. The reason was because of a misuse of the error object. This commit simply puts back the correct `path` to access the values within the error object to fix it.

image


regexowl (approved):
Error notifications are back! Thanks Thomas!!

PR #1501, opened 2023-12-18
Wizard/test: Remove unused variables

regexowl:
This removes unused variable from the code.


lucasgarfield (approved):
lgtm

lucasgarfield:
Hm, it is odd that the Travis test flaked. It is the "Click through all steps with valid values" test that failed, it received an empty object instead of the proper requests.

regexowl:
> Hm, it is odd that the Travis test flaked. It is the "Click through all steps with valid values" test that failed, it received an empty object instead of the proper requests.

Weird. The re-run worked, so could have been really just a flake. I'll merge this PR, but will keep an eye on similar errors in the future.

PR #1504, opened 2023-12-19
Wizard: Indicate disabled packages

regexowl:
When a package is added to the chosen packages it stays in the list of available packages, but gets disabled as an option.

This adds a visual indication in a form of greyed out name of the package.

Before:
Screenshot from 2023-12-19 14-28-13

After:
Screenshot from 2023-12-19 14-28-38

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Packages.js
+++ b/src/Components/CreateImageWizard/formComponents/Packages.js
@@ -405,7 +405,11 @@ const Packages = ({ getAllPackages, isSuccess }) => {
                   onOptionSelect={(e) => handleSelectAvailable(e, pkg.name)}
                 >
                   <TextContent key={`${pkg.name}`}>
-                    <span className="pf-c-dual-list-selector__item-text">
+                    <span
+                      className={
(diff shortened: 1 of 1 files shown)


lucasgarfield:
I'm sure we used to have this, I wonder what broke it?

Since it did break, it could break again... let's add a test that verifies this behavior.

Can we do this without CSS? I haven't checked so I'm not 100% sure but I imagine that in PF4 we had it working by passing an `isDisabled` prop into the select option component...

regexowl:
The formatting broke during migration to PF5. It seems the problem is with the structure of our options now. `<TextContent>` which wraps the name and the description of the package overrides any formatting for `isDisabled`. 

Will try to find a way around it.

regexowl:
Alright, took another look at it. This seems like a regression between the PatternFly versions. Opened an issue in patternfly-react repo: https://github.com/patternfly/patternfly-react/issues/9900

regexowl:
@lucasgarfield Would you mind merging this as a temporary fix? The bug wasn't addressed from the PF side yet, but as we plan to drop the dual list selector this wouldn't stay in the code base for long.

lavocatt (approved):

PR #1509, opened 2023-12-21
store/test: Remove unused code

regexowl:
This cleans up the code base by removing actions and slices which are no longer in use.


lucasgarfield:
Love it! I know there are also a bunch of un-used functions in `api.js`, but we can look at that in a follow-up PR.

lucasgarfield (approved):
lgtm

regexowl:
> Love it! I know there are also a bunch of un-used functions in api.js, but we can look at that in a follow-up PR.

All the un-used functions are gone (https://github.com/RedHatInsights/image-builder-frontend/pull/1488). Now there are just two calls which are still used in `Packages.js`, but not for long :broom:

PR #1510, opened 2023-12-21
workflows: Add check for manual changes to API

regexowl:
This adds a check to ensure that the code generated automatically by `npm run api` has no manual changes.


lucasgarfield (approved):
Love it! See ya later, Travis...

PR #1514, opened 2023-12-25
app: useEffect warning fixes

mgold1234:
This commit addresses warnings related to the react-hooks/exhaustive-deps rule.
It resolves these warnings by adding the missing dependencies to the useEffect hook.
This ensures that all necessary dependencies are correctly specified, aligning with the recommended usage patterns for React hooks.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js
+++ b/src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js
@@ -55,7 +55,12 @@ export const AWSSourcesSelect = ({
   useEffect(() => {
     if (isFetchingDetails || !isSuccessDetails) return;
     change('aws-associated-account-id', sourceDetails?.aws?.account_id);
-  }, [isFetchingDetails, isSuccessDetails]);
+  }, [
+    isFetchingDetails,
+    isSuccessDetails,
+    change,
+    sourceDetails?.aws?.account_id,
+  ]);
 
   const onFormChange = ({ values }) => {
     if (
--- a/src/Components/CreateImageWizard/formComponents/ActivationKeys.js
+++ b/src/Components/CreateImageWizard/formComponents/ActivationKeys.js
@@ -91,7 +91,7 @@ const ActivationKeys = ({ label, isRequired, ...props }) => {
       change('subscription-server-url', 'subscription.rhsm.stage.redhat.com');
       change('subscription-base-url', 'https://cdn.stage.redhat.com/');
     }
-  }, []);
+  }, [isProd, change]);
 
   const setActivationKey = (_, selection) => {
     selectActivationKey(selection);
--- a/src/Components/CreateImageWizard/formComponents/AzureSourcesSelect.js
+++ b/src/Components/CreateImageWizard/formComponents/AzureSourcesSelect.js
@@ -49,7 +49,13 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
     if (isFetchingDetails || !isSuccessDetails) return;
     change('azure-tenant-id', sourceDetails?.azure?.tenant_id);
     change('azure-subscription-id', sourceDetails?.azure?.subscription_id);
-  }, [isFetchingDetails, isSuccessDetails]);
+  }, [
+    isFetchingDetails,
+    isSuccessDetails,
+    sourceDetails?.azure?.subscription_id,
+    sourceDetails?.azure?.tenant_id,
+    change,
+  ]);
 
   const onFormChange = ({ values }) => {
     if (
--- a/src/Components/CreateImageWizard/formComponents/FileSystemConfigButtons.js
+++ b/src/Components/CreateImageWizard/formComponents/FileSystemConfigButtons.js
@@ -16,16 +16,16 @@ const FileSystemConfigButtons = ({ handleNext, handlePrev, nextStep }) => {
   );
   const [nextHasBeenClicked, setNextHasBeenClicked] = useState(false);
   const prefetchArchitectures = imageBuilderApi.usePrefetch('getArchitectures');
+  const errors = getState()?.errors?.['file-system-configuration'];
 
   useEffect(() => {
-    const errors = getState()?.errors?.['file-system-configuration'];
     errors ? setHasErrors(true) : setHasErrors(false);
 
     if (!errors) {
       setNextHasBeenClicked(false);
       change('file-system-config-show-errors', false);
     }
-  });
+  }, [errors, change]);
 
   const handleClick = () => {
     if (!hasErrors) {
--- a/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
+++ b/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
@@ -100,7 +100,7 @@ const FileSystemConfiguration = ({ ...props }) => {
       setItemOrder(newRows.map((row) => row.id));
       change('file-system-config-radio', 'manual');
     }
-  }, [customizations, isSuccess]);
+  }, [customizations, isSuccess, change, hasCustomizations, rows]);
 
   useEffect(() => {
     const fsc = getState()?.values?.['file-system-configuration'];
@@ -123,7 +123,7 @@ const FileSystemConfiguration = ({ ...props }) => {
     });
     setRows(newRows);
     setItemOrder(newOrder);
-  }, []);
+  }, [getState]);
 
   const showErrors = () =>
     getState()?.values?.['file-system-config-show-errors'];
@@ -144,7 +144,7 @@ const FileSystemConfiguration = ({ ...props }) => {
         return null;
       })
     );
-  }, [rows, itemOrder]);
+  }, [rows, itemOrder, change, input.name]);
 
   const addRow = () => {
     const id = uuidv4();
--- a/src/Components/CreateImageWizard/formComponents/MountPoint.js
+++ b/src/Components/CreateImageWizard/formComponents/MountPoint.js
@@ -36,7 +36,7 @@ const MountPoint = ({ ...props }) => {
         return;
       }
     }
-  }, []);
+  }, [props.mountpoint]);
 
   useEffect(() => {
     let suf = suffix;
--- a/src/App.js
+++ b/src/App.js
@@ -14,7 +14,7 @@ const App = (props) => {
   useEffect(() => {
     updateDocumentTitle('Image Builder | Red Hat Insights');
     hideGlobalFilter();
-  }, []);
+  }, [hideGlobalFilter, updateDocumentTitle]);
 
   return (
     <React.Fragment>


src/Components/CreateImageWizard/formComponents/ActivationKeys.js
```diff
@@ -91,7 +91,7 @@ const ActivationKeys = ({ label, isRequired, ...props }) => {
       change('subscription-server-url', 'subscription.rhsm.stage.redhat.com');
       change('subscription-base-url', 'https://cdn.stage.redhat.com/');
     }
-  }, []);
+  }, [change, isProd]);
```
lavocatt, line 94: Same kind of comment here for the `useEffect` dependencies:

`change` shouldn't endup in the list, only `isProd()` is making sense here.
The way we're using it is a sign of an antipattern however, I'm not sure with this fix, I would rather let it be in the form it was before.

src/Components/CreateImageWizard/formComponents/AzureSourcesSelect.js
```diff
@@ -49,7 +49,13 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
...
     change('azure-tenant-id', sourceDetails?.azure?.tenant_id);
     change('azure-subscription-id', sourceDetails?.azure?.subscription_id);
-  }, [isFetchingDetails, isSuccessDetails]);
+  }, [
+    isFetchingDetails,
+    isSuccessDetails,
+    sourceDetails?.azure?.subscription_id,
+    sourceDetails?.azure?.tenant_id,
+    change,
+  ]);
```
lavocatt, line 58: Use this pattern instead:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

It's not a good approach to make it dependent to the `change` function and the other you added.

src/Components/CreateImageWizard/formComponents/FileSystemConfigButtons.js
```diff
@@ -16,16 +16,16 @@ const FileSystemConfigButtons = ({ handleNext, handlePrev, nextStep }) => {
...
   useEffect(() => {
-    const errors = getState()?.errors?.['file-system-configuration'];
     errors ? setHasErrors(true) : setHasErrors(false);
 
     if (!errors) {
       setNextHasBeenClicked(false);
       change('file-system-config-show-errors', false);
     }
-  });
+  }, [errors, change]);
```
lavocatt, line 28: Use this pattern:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -100,7 +100,7 @@ const FileSystemConfiguration = ({ ...props }) => {
       setItemOrder(newRows.map((row) => row.id));
       change('file-system-config-radio', 'manual');
     }
-  }, [customizations, isSuccess]);
+  }, [customizations, isSuccess, hasCustomizations, change, rows]);
```
lavocatt, line 103: Use this pattern instead:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -123,7 +123,7 @@ const FileSystemConfiguration = ({ ...props }) => {
     });
     setRows(newRows);
     setItemOrder(newOrder);
-  }, []);
+  }, [getState]);
```
lavocatt, line 126: Use this pattern instead:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -143,7 +143,7 @@ const FileSystemConfiguration = ({ ...props }) => {
         }
       })
     );
-  }, [rows, itemOrder]);
+  }, [rows, itemOrder, change, input.name]);
```
lavocatt, line 146: Use this pattern instead:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

src/Components/CreateImageWizard/formComponents/MountPoint.js
```diff
@@ -36,7 +36,7 @@ const MountPoint = ({ ...props }) => {
         return;
       }
     }
-  }, []);
+  }, [props.mountpoint]);
```
lavocatt, line 39: Use this pattern instead:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -159,7 +159,7 @@ const Packages = ({ getAllPackages, isSuccess }) => {
       }
       setChosenPackages(newChosenPackages);
     }
-  }, []);
+  }, [selectedPackages]);
```
lavocatt, line 162: Use this pattern instead:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

src/Components/CreateImageWizard/formComponents/Packages.js
```diff
@@ -222,7 +222,7 @@ const Packages = ({ getAllPackages, isSuccess }) => {
   }, [chosenPackages, filterChosen, searchResultsComparator]);
 
   // call api to list available packages
-  const handleAvailablePackagesSearch = async () => {
+  const handleAvailablePackagesSearch = useCallback(async () => {
```
lavocatt, line 225: I don't understand this change, can you explain it to me please?

src/Components/CreateImageWizard/formComponents/Registration.js
```diff
@@ -140,13 +140,20 @@ const Registration = ({ label, ...props }) => {
...
       registerSystem === 'register-now'
   );
+  const radioRef = useRef(null);
+
+  useEffect(() => {
+    // Set focus when the component mounts
+    if (radioRef.current) {
+      radioRef.current.focus();
+    }
+  }, []);
```
lavocatt, line 150: Why adding a `useEffect` change the focus here ? At first glance it looks unrelated to linting warning fixes. Can you develop why this is needed please?

src/Components/CreateImageWizard/formComponents/Repositories.js
```diff
@@ -155,42 +155,46 @@ const convertSchemaToContentSources = (repo) => {
...
-  const initializeRepositories = (contentSourcesReposList) => {
-    // Convert list of repositories into an object where key is repo URL
-    const contentSourcesRepos = contentSourcesReposList.reduce(
-      (accumulator, currentValue) => {
-        accumulator[currentValue.url] = currentValue;
-        return accumulator;
-      },
-      {}
-    );
+  const initializeRepositories = useCallback(
```
lavocatt, line 158: Why are you using a `useCallback` here?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -760,7 +760,9 @@ describe('Step File system configuration', () => {
     );
     await user.click(manuallyConfigurePartitions);
 
-    const addPartition = await screen.findByTestId('file-system-add-partition');
+    const addPartition = screen.getByRole('button', {
+      name: /add partition/i,
+    });
```
lavocatt, line 765: Tests shouldn't use `getBy` but only `findBy` why did you need to revert this back?

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
```diff
@@ -1528,7 +1530,7 @@ describe('Keyboard accessibility', () => {
       'Automatically register and enable advanced capabilities'
     );
     const registerRadio = screen.getByTestId('registration-radio-now');
-    expect(registerRadio).toHaveFocus();
+    expect(registerRadio);
```
lavocatt, line 1533: The tests shouldn't get changed to accommodate for changes in the sources. Especially a test that checks that the radio button has the focus. Why did you need to change this line? Was this because it was incompatible with the change you made in the [Registration.js](https://github.com/RedHatInsights/image-builder-frontend/pull/1514/files#diff-86a6e0c27e139603b172a79bafc527d737a7ad666cbd3e264ea4dd6c43d7e817) file?

If that's the case, you shouldn't change the test and instead make the code passes them :slightly_smiling_face:

lavocatt (changes requested):
Thanks for this attempt at fixing the pending warnings in the code base.

However I have several comments, notably on how the `useEffect` are addressed. I don't think that this is the correct approach. You'll find my comments down below, happy hacking!

src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js
```diff
@@ -55,7 +55,7 @@ export const AWSSourcesSelect = ({
   useEffect(() => {
     if (isFetchingDetails || !isSuccessDetails) return;
     change('aws-associated-account-id', sourceDetails?.aws?.account_id);
-  }, [isFetchingDetails, isSuccessDetails]);
+  });
```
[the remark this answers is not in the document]
mgold1234: I tried to change this part of the code to   const [isFetch, setIsFetch] = useState(isFetchingDetails);
  const [isFinish, SetIsFinish] = useState(isSuccessDetails);
  if (isFetchingDetails || !isSuccessDetails) return;
  if (isFetchingDetails !== isFetch || isSuccessDetails !== isFinish) {
    setIsFetch(isFetchingDetails);
    SetIsFinish(isSuccessDetails);
    change('aws-associated-account-id', sourceDetails?.aws?.account_id);
  }
  but its look like a lot of tests fail. 
  can you tell me if this is the direction I should go?
  @lavocatt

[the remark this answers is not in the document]
mgold1234: I tried this solution useEffect(() => {
    if (isFetchingDetails || !isSuccessDetails) return;
    change('aws-associated-account-id', sourceDetails?.aws?.account_id);
  }, [
    change,
    sourceDetails?.aws?.account_id,
    isFetchingDetails,
    isSuccessDetails,
  ]);
  and its look like its working good

src/Components/CreateImageWizard/formComponents/ActivationKeys.js
mgold1234: but without the 'change' I still get this warning message - /Users/michalgold/git/edge-management/image-builder-project/image-builder-frontend/src/Components/CreateImageWizard/formComponents/ActivationKeys.js
  94:6  warning  React Hook useEffect has a missing dependency: 'change'. Either include it or remove the dependency array  react-hooks/exhaustive-deps

src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js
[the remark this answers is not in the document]
lavocatt: It's not that this code chunk won't work, it's more that it is addressing the issue incorrectly.

We have to ask ourselves, why is the linter producing a warning when not all the dependencies of a `useEffect` are listed?

The reason why the react devs have produced these warnings is to alert us on bad patterns. They have a list of situations where people tend to use `useEffects` to solve problems where in reality they shouldn't. And we're on one of these cases

**First let's break down what's happening here:**

```
  useEffect(() => {
    if (isFetchingDetails || !isSuccessDetails) return;
    change('aws-associated-account-id', sourceDetails?.aws?.account_id);
  }, [isFetchingDetails, isSuccessDetails]);
  });
```

What this piece of code does: it will execute the function whenever the value `isFetchingDetails` or  `isSuccessDetails` are getting updated.

```
() => {
    if (isFetchingDetails || !isSuccessDetails) return;
    change('aws-associated-account-id', sourceDetails?.aws?.account_id);
  }
```

The function itself updates the value of the form state `aws-associated-account-id` with what's inside `sourceDetails?.aws?.account_id`. And that's it.

**Now, if we change the `useEffect` to the solution you were proposing:**

```
  useEffect(() => {
    if (isFetchingDetails || !isSuccessDetails) return;
    change('aws-associated-account-id', sourceDetails?.aws?.account_id);
  }, []);
  });
```

What's happening there, is that after every render of the component an extra render will get performed to call the inner function and update the form value `aws-associated-account-id` with the content of `sourceDetails?.aws?.account_id`.

So why the two approaches are working:
* In the first scenario the function is called only in certain conditions, but these conditions are the one we are interested in, so no issues
* In the second scenario the function is always called, which means that it necesserarily includes the conditions we are inserted in

**But what about the efficiency impact of the solution you are proposing?**

Clearly the solution before was less intrusive because it is called in less cases than every time.

Granted that we shouldn't seek performance optimizations too early in the development process, but this is an example of a performance regression. Not a bad one, but we can do better.


**So what should we do about it?**

There's a couple of possibilities for us:
* Not using a `useEffect`, there's a pattern to use instead for this https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
* Silence the warning with an eslint comment
* Do nothing

**first solution: Not using the `useEffect`:**
```
  const [previsFetchingDetails, setPrevisFetchingDetails] = useState(isFetchingDetails);
  const [previsSuccessDetails, setPrevisSuccessDetails] = useState(isSuccessDetails);
  if (isFetchingDetails !== previsFetchingDetails || isSuccessDetails !== previsSuccessDetails) {
    setPrevisFetchingDetails(isFetchingDetails);
    setPrevisSuccessDetails(isSuccessDetails);
    if (!(isFetchingDetails || !isSuccessDetails)) {
      change('aws-associated-account-id', sourceDetails?.aws?.account_id);
    }
  }
```

I quickly wrote this by following the link https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

So this will behave exactly like a `useEffect` except now we're compliant with the react guidelines.

**second solution: silence the warning with an esling comment:**

So the previous solution is bringing a new way of doing things in the code base, and we need a team discussion about if that's what we want to do or not. Because guidelines are one thing but what the team wants is another.
A simpler solution to address the issue would be to simply silence the warning:

```
  useEffect(() => {
    if (isFetchingDetails || !isSuccessDetails) return;
    change('aws-associated-account-id', sourceDetails?.aws?.account_id);
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [isFetchingDetails, isSuccessDetails]);
  });
```

**third solution: do nothing:**

But maybe, silencing warnings will just make us forget we're using wrong patterns, and what if we want to fix them in the future? Maybe it's good to have them listed somewhere so that we can address them once and for all?


**conclusion**

My own opinion is that we should address the `useEffect` by following the guidelines on the React page. But I'll let @regexowl and @lucasgarfield give their opinions too here.

Lemme know if you have any questions.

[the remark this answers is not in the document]
mgold1234: with your suggestion change without ussing useEffect const [previousFetchingDetails, setPreviousFetchingDetails] = useState(isFetchingDetails);
  const [previousSuccessDetails, setPreviousSuccessDetails] = useState(isSuccessDetails);
  if (isFetchingDetails !== previousFetchingDetails || isSuccessDetails !== previousSuccessDetails) {
    setPreviousFetchingDetails(isFetchingDetails);
    setPreviousSuccessDetails(isSuccessDetails);
    if (!(isFetchingDetails || !isSuccessDetails)) {
      change('aws-associated-account-id', sourceDetails?.aws?.account_id);
    }
  }
  I get this error   ● Step Upload to AWS › validation works

    Expected test not to call console.error().

    If the error is expected, test for it explicitly by mocking it out using jest.spyOn(console, 'error').mockImplementation() and test that the warning occurs.

    Warning: Cannot update a component (`TextField`) while rendering a different component (`AWSSourcesSelect`). To locate the bad setState() call inside `AWSSourcesSelect`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render

      20 |
      21 | export const AWSSourcesSelect = ({
    > 22 |   label,
         |   ^
      23 |   isRequired,
      24 |   className,
      25 |   ...props

src/Components/CreateImageWizard/formComponents/Packages.js
mgold1234: the usecallback is for better performace, React will reuse the memoized version of the callback and will ensure that only when the dependencis changed it should trigger the recreation of the callback

src/Components/CreateImageWizard/formComponents/Registration.js
mgold1234: becasue I remove the autofocus that bring this warning:
  150:13  warning  The autoFocus prop should not be used, as it can reduce usability and accessibility for users  jsx-a11y/no-autofocus

  const radioRef = useRef(null);

  useEffect(() => {
    // Set focus when the component mounts
    if (radioRef.current) {
      radioRef.current.focus();
    }
  }, []);
  and one test was fail because the we still need that focus

src/Components/CreateImageWizard/formComponents/Repositories.js
mgold1234: same reason as previous one

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
mgold1234: that's why I added this  const radioRef = useRef(null);
  useEffect(() => {
    // Set focus when the component mounts
    if (radioRef.current) {
      radioRef.current.focus();
    }
  }, []);
  to the registration.js file

src/Components/CreateImageWizard/formComponents/Registration.js
regexowl: Quick note on the auto focus - I had a short discussion about it with a few people some months ago and the outcome was that a11y's "auto focus bad" approach is not really always true as it's quite heavily contextual. 

Accessibility is one of the areas we should and want to look into more as it was not a focus until now. But generally I'd say let's keep the warning for now as a reminder and deal with it when we get to the topic. Adding a11y rules was just a beginning and the rule was switched from error to warning for this purpose. :)

src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js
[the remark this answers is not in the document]
lucasgarfield: In general I agree that it’s a good idea to work on removing unneeded useEffect() hooks. But I think we ought to do that in a follow-up PR, and focus on the low hanging fruit in this PR… in this specific case, we can fix the linter warning by simply adding the missing dependencies to the dependencies array as suggested by the linter (just add `change` and `sourceDetails` to the array):

image

Having said that, I think that this is actually one of those edge cases where we do need the useEffect().

If I try first solution proposed, I get the same warning as Michal:

```Warning: Cannot update a component (`TextField`) while rendering a different component (`AWSSourcesSelect`).```

I did some experiments and I believe the problem is the call to change() with `aws-associated-account-id`, as it causes a re-render of <TextField> responsible for rendering the Associated Account ID  while rendering <AWSSourcesSelect>. Screenshot for reference:

image

That warning was introduced in React v16, and they added some notes about it:
https://legacy.reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render

>This warning will help you find application bugs caused by unintentional state changes. In the rare case that you intentionally want to change the state of another component as a result of rendering, you can wrap the setState call into useEffect.

From the “You might not need an Effect” article from the React docs:

>Don’t rush to add Effects to your components. Keep in mind that Effects are typically used to “step out” of your React code and synchronize with some external system. This includes browser APIs, third-party widgets, network, and so on. If your Effect only adjusts some state based on other state, you might not need an Effect.

In this case, the ‘external system’ we need to synchronize with is the state managed by Data Driven Forms. So my interpretation is that it is appropriate here – and within the React guidelines – for us to use the useEffect hook in this case.

src/Components/CreateImageWizard/formComponents/AzureSourcesSelect.js
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

src/Components/CreateImageWizard/formComponents/FileSystemConfigButtons.js
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

src/Components/CreateImageWizard/formComponents/Packages.js
mgold1234: I remove this useCallback from that pr

src/Components/CreateImageWizard/formComponents/Registration.js
mgold1234: remove the usefref from that pr for now

src/test/Components/CreateImageWizard/CreateImageWizard.test.js
mgold1234: change it back to findby

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

src/Components/CreateImageWizard/formComponents/MountPoint.js
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

src/Components/CreateImageWizard/formComponents/Packages.js
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

src/Components/CreateImageWizard/formComponents/AWSSourcesSelect.js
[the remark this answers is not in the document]
mgold1234: as we discussed in the group channel, we agreed to add those dependencies
and in the future maybe we will refector the code and remove the useEffect hook were ever we can

lavocatt (approved):
It's all very clean! good job.

lavocatt (changes requested):
One last thing to fix: the commit message :wink:

regexowl:
Is it just me or does the page not render with the changes? Trying on stage-beta and all I'm getting is a spinner.

lavocatt:
> Is it just me or does the page not render with the changes? Trying on stage-beta and all I'm getting is a spinner.

I haven't tried on my local machine, I was trusting the tests, but if that's broken they should even not pass right?
If it's broken it might be because of the first useEffect that got removed. And it might cause a forever loop.

mgold1234:
> Is it just me or does the page not render with the changes? Trying on stage-beta and all I'm getting is a spinner.

mgold1234:
> > Is it just me or does the page not render with the changes? Trying on stage-beta and all I'm getting is a spinner.
> 
> I haven't tried on my local machine, I was trusting the tests, but if that's broken they should even not pass right? If it's broken it might be because of the first useEffect that got removed. And it might cause a forever loop.

its very strange that all tests pass :(
I fixed that and return the useEffect to app.js

regexowl:
Agree, that the tests shouldn't pass in this case, although not super sure why it happened. Some of the re-rendering got the tests stuck in the past. Not sure if this looked like a clean run in the tests are the new run rewrote the logs.

@mgold1234 Another bug I've noticed - when you add some packages on the RH packages step and continue to the custom packages step, the added packages will disappear from the added. Can you check if it works the same for you?

mgold1234:
I tried the scenario and see the same beavior at stage.
or maybe I didnt understand the exact problem @regexowl 
Uploading Screen Recording 2024-01-08 at 13.06.05.mov…

regexowl:
> I tried the scenario and see the same beavior at stage. or maybe I didnt understand the exact problem @regexowl

Ha, it seems the reproduction is a bit more complicated.

So the steps to reproduce the bug would be:
1. get to the OSCAP step
2. choose an OSCAP profile
3. get to the packages step
4. add any package
5. go to the Repositories step
6. add any custom Repository
7. go to the custom packages step
8. compare the list of selected packages with the list on packages step

What I see is that the packages I've added in the first packages step disappeared in the second packages step. I've tried on stage-beta main and the list of packages persisted.

mgold1234:
@regexowl , I think its not related to that pr ,
I see the same problem if I do that: 
get to the OSCAP step
choose an OSCAP profile
get to the packages step
add any package
go to the Repositories step
press Next 
press back
press back
press Next
-> the added packages disappear

regexowl (approved):
I'd say this looks good now! Also clicked around and the parts with the changes seem to be working well still. 

@mgold1234 Thank you for working on this even though it turned out that the `useEffect` fixes can be much more tangled :dizzy:

Also, could you please rebase the PR? :)

regexowl:
@lavocatt Hey hey, is it okay if I dismiss your "change requested" review? Alternatively can you approve? :)

lavocatt:
> @lavocatt Hey hey, is it okay if I dismiss your "change requested" review? Alternatively can you approve? :) 

Yeah sure. Sorry I wasn't meaning to block the PR I thought your approval would suffice for the merge. I'm approving rn.

lavocatt (approved):
Lgtm!

regexowl:
> Yeah sure. Sorry I wasn't meaning to block the PR I thought your approval would suffice for the merge. I'm approving rn.

No worries, I also though one approval should be enough :sweat_smile:

PR #1515, opened 2023-12-26
fsc: fix array-callback-return linting warnings

mgold1234:
this comit fixes the array-callback warning by changing map to foreach
and return null to the second map

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
+++ b/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
@@ -110,7 +110,7 @@ const FileSystemConfiguration = ({ ...props }) => {
 
     const newRows = [];
     const newOrder = [];
-    fsc.map((r) => {
+    fsc.forEach((r) => {
       const id = uuidv4();
       newRows.push({
         id,
@@ -141,6 +141,7 @@ const FileSystemConfiguration = ({ ...props }) => {
             };
           }
         }
+        return null;
       })
     );
   }, [rows, itemOrder]);


src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -120,6 +120,7 @@ const FileSystemConfiguration = ({ ...props }) => {
         unit: r.unit,
       });
       newOrder.push(id);
+      return null;
```
lavocatt, line 123: I think this fix should be addressed a bit differently.

See, map is supposed to be used to transform a list into another one (check out the documentation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).

However here the return value of the result of the map function at line 113 is ignored. Meaning that what we want to have instead is a loop over all the entries of `fsc`.

So a better fix is to change `map` to `foreach` (check out the documentation there https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)

Does my explication makes sense?

lavocatt (changes requested):
Thanks for working on cleaning the warnings, it's very much appreciated. Here's my review, happy hacking.

lavocatt:
Also, please update your commit title and description to match the required specification.

```
fsc: fix eslint warnings

Previously [...] this commit fixes this by [...]
```

I'll let you fill in the blanks :rocket:

src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
```diff
@@ -141,6 +141,7 @@ const FileSystemConfiguration = ({ ...props }) => {
             };
           }
         }
+        return null;
```
lavocatt, line 144: you don't need to return null in the `foreach`

lavocatt (dismissed):
LGTM thanks.

PR #1518, opened 2024-01-02
ImagesTable: fix grey box under AMI when image build is in progress

mgold1234:
this commit fix the fix grey box under AMI when image build is in progress
FIXES # https://github.com/RedHatInsights/image-builder-frontend/issues/1362

### DIFF
--- a/src/Components/ImagesTable/ClonesTable.tsx
+++ b/src/Components/ImagesTable/ClonesTable.tsx
@@ -37,11 +37,8 @@ const Ami = ({ status }: AmiPropTypes) => {
         </ClipboardCopy>
       );
 
-    case 'failure':
-      return undefined;
-
     default:
-      return <Skeleton width="12rem" />;
+      return undefined;
   }
 };
 


src/Components/ImagesTable/ClonesTable.tsx
```diff
@@ -41,7 +41,7 @@ const Ami = ({ status }: AmiPropTypes) => {
       return undefined;
 
     default:
-      return <Skeleton width="12rem" />;
+      return;
```
regexowl, line 44: Nice! The code could be cleaned up a bit more. As we expect to get an AMI for success, and undefined for failure and by default we could use:

```javascript
    case 'failure':
    default:
      return undefined;
  }
```

or even 
```javascript
    default:
      return undefined;
  }
```

if we don't mind including failure with all the other cases when we don't display AMI. What do you think?

regexowl:
Could you also please add a component to the name of the commit and a short description? Maybe something like this:

> _ImagesTable: fix grey box under AMI when image build is in progress_
>
> _Previously a Skeleton component was used as a placeholder for an AMI when the build was still pending. This removes the Skeleton and returns undefined rendering the AMI as empty._

src/Components/ImagesTable/ClonesTable.tsx
mgold1234: good idea thank you @regexowl

regexowl:
One more tiny nit pick - in the PR name "progrss" -> "progress" and the commit message looks unfinished, it seems to end with "rendering the AMI as" now.

regexowl (approved):
Looks great, thank you!

PR #1521, opened 2024-01-03
ESLint: Remove temporary rule overrides and fix a lint problem

regexowl:
This removes some of the temporary rule overrides and fixes one occurrence of no-unused-vars.


regexowl:
As the rules are set to "error" by default this will make them stricter:

Screenshot from 2024-01-04 10-30-45

lucasgarfield (approved):
lgtm! stricter is nice :)

PR #1525, opened 2024-01-04
Travis: Disable npm cache

regexowl:
This disables npm caching for all Travis runs. The effect on the duration of the run is not significant and it should prevent problems with dirty cache in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ notifications:
   email: false
 node_js:
   - '16'
+cache:
+  npm: false
 install:
   - npm ci
 before_script: |


regexowl:
This is a reaction to a problem with dirty npm cache in https://github.com/RedHatInsights/image-builder-frontend/pull/1450

The good news is - when cache cleaning was added the run was successful. But when the cleaning got removed again, Travis decided to just cancel the run completely.

Let's disable npm caching completely. The time increase does not seem to be super high and it should prevent any similar problems in the future hopefully.

croissanne (approved):

PR #1526, opened 2024-01-04
[WIP] Migrate from Jest framework to vitest

regexowl:
Test suites that work:
- [ ] CreateImageWizard
- [x] CreateImageWizardV2
- [x] edge
- [x] ImagesTable
- [x] LandingPage
- [x] ShareImageModal


regexowl:
The setup seem to be working, unit tests can be run with `npm run test` as vitest should replace jest.

Running the command goes through the tests, but vitest stays in a watch mode after all the tests are done:

image

While the watch mode is running any saved changes to a test or tested file should re trigger tests connected to that part of the code. The watch mode can be exited by pressing `q` (running the tests in the watch mode can be also turned off in case we'd like an option to just do a single run of the tests).

Currently there are two test files passing, both edge tests.

image

When trying to look into the failing tests there might be a common denominator - routing. Not sure if there is some persisting state that does not get cleaned properly between the tests or if the routing simply does not route, but when trying to log the components via Testing Playground the result seems to be empty.

regexowl:
Lol, making progress - the unhandled errors seem to be caused by missing node-fetch polyfills. Adding those seems to solve the problem for some of the test files.

regexowl:
Okay, everything but CreateImageWizard tests work. For those I'm getting a completely new error `TypeError: formOptions.getState is not a function` which seems to be connected to DDF's schema somehow maybe? The only place where we use it is the CustomButtons component and since WizardV2 test suite seems to be working well with vitest I'm trying to look in the way of DDF.

regexowl:
This PR has been stale for quite some time. It depends on migration to V2Wizard and will need to be mostly rewritten to accommodate all the changes in code and dependencies.

PR #1527, opened 2024-01-04
Repositories: fix lint warning react-hooks/exhaustive-deps with adding useCallback hook

mgold1234:
this commit fix the exhaustive-deps by adding the useCallback hook.
after adding the missing dependencies to the useEffect,
I fix this problem that we have  -'The 'initializeRepositories' function makes the
dependencies of useMemo Hook (at line 256) change on every render.'
Move it inside the useMemo callback. Alternatively,
wrap the definition of 'initializeRepositories' in its own useCallback() Hook  react-hooks/exhaustive-deps'

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Repositories.js
+++ b/src/Components/CreateImageWizard/formComponents/Repositories.js
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 
 import {
   useFieldApi,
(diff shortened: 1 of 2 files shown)


regexowl:
The Repositories step PR https://github.com/RedHatInsights/image-builder-frontend/pull/1569 will simplify the structure of the step and some of the code in this PR may no longer be applicable. 

My proposal would be to set this one as blocked for now and after https://github.com/RedHatInsights/image-builder-frontend/pull/1569 gets merged let's rebase and see what could be updated.

regexowl:
Ok, the blocking PR is merged. Would you like to rebase this one so we can see if anything needs updating? :)

regexowl:
The code looks good, the changes follow the linter suggestions, but the suggestions seem to be unreliable specifically for the exhaustive-deps rules :thinking: We should keep an eye on those.

I've ran into a bug, could you try if the behaviour is the same for you?

1. in stage-beta open the old Wizard
2. click through to the Repositories step (what you choose on the other steps shouldn't matter)
3. select any repository
4. the Next button should get immediately disabled
5. unselect the repository
6. the button is still disabled even if you try to click back through the steps and the Wizard needs to be refreshed to enable it back

mgold1234:
Repositories unavailable now so I chant check this but
The Content service cannot be reached, please check back later.

regexowl:
https://github.com/osbuild/image-builder-frontend/pull/1662 was merged, unblocking

lucasgarfield:
Closing this as so much has changed since this was originally authored.

PR #1531, opened 2024-01-05
V2 Wizard - Foundations

lucasgarfield:
The commits in this PR set the foundation for using RTK to manage the state of the V2 wizard.

image


regexowl (approved):
Looks good :+1: I like the uncommenting/reenabling tests strategy! 

One question - do we want to keep all the styling in `CreateImageWizard.scss` from the get go or would it be better to add it when it's really needed?

lucasgarfield:
> Looks good 👍 I like the uncommenting/reenabling tests strategy!
> 
> One question - do we want to keep all the styling in `CreateImageWizard.scss` from the get go or would it be better to add it when it's really needed?

I'd love to get rid of as much scss as possible, but I don't want to get blocked on it. But we can do a follow-up PR, could be a good one for @mgold1234 to look at.

PR #1532, opened 2024-01-06
V2 Wizard - Image Output

lucasgarfield:
This PR adds the Image Output step to the V2 wizard.

All relevant tests are passing.

The wizard's state is now managed using RTK query and is visible using the Redux devtools.

Note that for the most part, our existing components can easily be converted from using DDF to RTK. Calls to `change()` in DDF are updated to calls to `dispatch()`, and calls to `getState()` are update to calls to the relevant Redux selector.

RTK's listeners are used to ensure that the wizard's state is always "correct" (i.e., "infeasible" states are automatically corrected by the listeners):

rtk listeners


croissanne (approved):
:heart_eyes:

PR #1533, opened 2024-01-07
V2 Wizard - AWS

lucasgarfield:
This PR adds the AWS substep to the Image Output step. 

awsstep

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import {
   Button,
@@ -10,10 +10,17 @@ import {
 import { useNavigate } from 'react-router-dom';
 
 import ImageOutputStep from './steps/ImageOutput';
+import Aws, { AwsShareMethod } from './steps/TargetEnvironment/Aws';
+import { isAwsAccountIdValid } from './validators';
 
 import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import './CreateImageWizard.scss';
-import { initializeWizard, selectImageTypes } from '../../store/wizardSlice';
+import {
+  initializeWizard,
+  selectAwsAccount,
+  selectAwsSource,
+  selectImageTypes,
+} from '../../store/wizardSlice';
 import { resolveRelPath } from '../../Utilities/path';
 import { ImageBuilderHeader } from '../sharedComponents/ImageBuilderHeader';
 
@@ -49,6 +56,11 @@ const CreateImageWizard = () => {
 
   const targetEnvironments = useAppSelector((state) => selectImageTypes(state));
 
+  const [awsShareMethod, setAwsShareMethod] =
+    useState<AwsShareMethod>('sources');
+  const awsAccountId = useAppSelector((state) => selectAwsAccount(state));
+  const awsSourceId = useAppSelector((state) => selectAwsSource(state));
+
   return (
     <>
       <ImageBuilderHeader />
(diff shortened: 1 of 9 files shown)


lucasgarfield (commented):
Seems to be a bug related to the state of the source selector. If you select a source, go to the next step, then return back to the AWS step, the source name gets dropped. (The account id is still there, because it comes from the store, but something is causing the select state to get reset.)

regexowl:
> Seems to be a bug related to the state of the source selector. If you select a source, go to the next step, then return back to the AWS step, the source name gets dropped. (The account id is still there, because it comes from the store, but something is causing the select state to get reset.)

Something similar is happening for the manual option. If you choose it, input an ID, click Next and Back you'll be put on empty "sources" variant of the step.

lucasgarfield:
:man_facepalming: I completely forgot about how the API schema for AWS works. We don't need to keep track of the AWS account ID for both cases. We need to keep track of the AWS account ID fr the share_with_accounts field *or* the source ID for the share_with_sources field. Will patch this up tomorrow, don't bother reviewing until then.

lucasgarfield:
Okay, this has been refactored with the actual Image Builder API in mind and is ready for review again, @regexowl & @croissanne.

Couple of notes on this implementation: in our API the aws account ids and sources are a list of strings. That doesn't really map to the actual front end, where we have a single account id or a single source. So I have set the selectors and actions up such that the need to access the [0] index for these arrays is abstracted and hidden in the wizard slice.

Note that we _do_ want the Redux store's schema itself to map well to the API, so it _does_ define these as arrays of strings. That way when we finally create our handleSubmit function for the wizard, we have as close to 1:1 between the redux store data and the image request as possible.

I also want to make sure that we cannot have both a source id and account id set at the same time. So I've added some listeners. Setting an aws account id will clear the source id if it exists, and vice versa.

The listeners list is growing but still seems manageable at this point... I could have also just done this in the actions (so for instance, dispatching `changeAwsAccount` could also reset the `awsSource`). That probably would be a bit simpler, but to stay consistent with what we're doing for the targets I went ahead and created listeners. I'm open to getting rid of these in favor of a simpler solution in the future, but I'd suggest for now we just try it this way... maybe let's see how bad it gets by the time the wizard is done and re-evaulate then. :sweat_smile:

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -57,15 +59,27 @@ const CreateImageWizard = () => {
...
-                  useAppSelector((state) => selectImageTypes(state)).length ===
-                  0
-                }
+                disableNext={targetEnvironments.length === 0}
               />
             }
           >
             <ImageOutputStep />
           </WizardStep>
+          <WizardStep
```
croissanne, line 80: So on the existing wizard we don't show the target environment step, not even in the sidebar, as long as they haven't selected anything that needs additional sharing details.

Don't think it's a blocker but we might ned to change that at some point.

croissanne (approved):
thank you!

croissanne (dismissed):

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
croissanne: nvm I just saw the first isHidden :facepalm:

regexowl:
I'll add my comment here for transparency - the source ID does not persists when clicking through the steps. Otherwise looks great!

lucasgarfield:
Ready for review again. 3rd time’s a charm I hope… I put all of my changes into a new commit so it is easy to review.

I have changed my philosophy on the RTK store schema. Previously I said I would like to keep it close to the API schema so it is easy to generate the requests. I now think it would be better to have the schema map directly to the wizard's steps and relevant step state.

The reason is that we should classify state as one of two types: 
1. ephemeral state (e.g. is a drop down open or closed)
2. persistent state that we need to remember even when its related component is not being rendered

The persistent state includes both data that will make it into the final request, and data that we just need to keep track of for later (for instance, for use in the Review step or for when you go back and forth between steps). 

I think it might become very confusing if we handle persistent state with a mixture of RTK and lifted useState. So for now, let’s rely on RTK for all persistent state and make the store’s schema match the wizard's layout.

Our eventual `handleSubmit` function will be a bit more complicated, but I suspect it will still be a huge improvement over what we have now... especially because we will have Typescript's safety.

regexowl (approved):
Looks good, thank you! :mage: I'm curious to see what the store will look like with the complete wizard. :eyes:

PR #1534, opened 2024-01-07
Oscap: change oscap file to typeSript file

mgold1234:
this commit convert oscap file from JavaScript to TypeScript


mgold1234:
we will not convert to patternFly 5

PR #1537, opened 2024-01-08
wizard/size_unit: fix linting warning

mgold1234:
This commit addresses the react-hooks/exhaustive-deps warnings in the SizeUnit file. The solution involves removing the useEffect hook from the component and introducing the handleSizeChange function as a replacement.


lavocatt (commented):
There is unfortunately a bug. If you set the unit, go next and back (or back and next) the unit isn't properly setup.

[Screencast from 2024-01-09 11-51-19.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/3761c8bc-2bd8-43c3-a804-c953e53349df)

mgold1234:
I dont find other solution in this case, closing this pr

PR #1541, opened 2024-01-09
Oscap: convert oscap javaScript file to typeScript file

mgold1234:
this commit convert oscap file from JavaScript to TypeScript

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Oscap.tsx
+++ b/src/Components/CreateImageWizard/formComponents/Oscap.tsx
@@ -1,6 +1,8 @@
 import React, { useEffect, useState } from 'react';
 
-import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
+import useFieldApi, {
+  UseFieldApiConfig,
+} from '@data-driven-forms/react-form-renderer/use-field-api';
 import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
 import {
   Alert,
@@ -17,16 +19,21 @@ import {
   SelectVariant,
 } from '@patternfly/react-core/deprecated';
 import { HelpIcon } from '@patternfly/react-icons';
-import PropTypes from 'prop-types';
 
 import OscapProfileInformation from './OscapProfileInformation';
 
 import {
+  DistributionProfileItem,
   useGetOscapCustomizationsQuery,
   useGetOscapProfilesQuery,
 } from '../../../store/imageBuilderApi';
 import { reinitFileSystemConfiguratioStep } from '../steps/fileSystemConfiguration';
 import { reinitPackagesStep } from '../steps/packages';
+
+type ChangeType = <F extends keyof Record<string, string | undefined>>(
+  name: F,
+  value?: Record<string, string | undefined>[F]
+) => void;
 
 /**
  * Every time there is change on this form step's state, reinitialise the steps
@@ -34,20 +41,23 @@ import { reinitPackagesStep } from '../steps/packages';
  * change their mind, going forward again leaves them in a coherent and workable
  * form state.
  */
-const reinitDependingSteps = (change) => {
+const reinitDependingSteps = (change: ChangeType) => {
   reinitFileSystemConfiguratioStep(change);
   reinitPackagesStep(change);
 };
-
+type ProfileSelectorProps = {
+  input: { name: string };
+};
 /**
  * Component for the user to select the profile to apply to their image.
  * The selected profile will be stored in the `oscap-profile` form state variable.
  * The Component is shown or not depending on the ShowSelector variable.
  */
-const ProfileSelector = ({ input }) => {
+
+const ProfileSelector = ({ input }: ProfileSelectorProps) => {
   const { change, getState } = useFormApi();
+  const [profileName, setProfileName] = useState<string>('None');
   const [profile, setProfile] = useState(getState()?.values?.['oscap-profile']);
-  const [profileName, setProfileName] = useState('');
   const [isOpen, setIsOpen] = useState(false);
   const {
     data: profiles,
@@ -68,9 +78,14 @@ const ProfileSelector = ({ input }) => {
       skip: !profile,
     }
   );
+
   useEffect(() => {
-    if (data) {
-      setProfileName(data?.openscap?.profile_name);
+    if (
+      data &&
+      data.openscap &&
+      typeof data.openscap.profile_name === 'string'
+    ) {
+      setProfileName(data.openscap.profile_name);
     }
   }, [data]);
 
@@ -84,17 +99,47 @@ const ProfileSelector = ({ input }) => {
   const handleClear = () => {
     setProfile(undefined);
     change(input.name, undefined);
+    // @ts-ignore
     setProfileName(undefined);
     reinitDependingSteps(change);
   };
 
-  const handleSelect = (_, selection) => {
+  const handleSelect = (_: React.MouseEvent, selection: string) => {
     setProfile(selection);
     setIsOpen(false);
     change(input.name, selection);
     reinitDependingSteps(change);
     change('file-system-config-radio', 'manual');
   };
+
+  const options = [
+    <OScapNoneOption setProfileName={setProfileName} key="oscap-none-option" />,
+  ];
+  if (isSuccess) {
+    options.concat(
+      profiles.map((profile_id) => {
+        return (
+          <OScapSelectOption
+            key={profile_id}
+            profile_id={profile_id}
+            setProfileName={setProfileName}
+          />
+        );
+      })
+    );
+  }
+
+  if (isFetching) {
+    options.push(
+      <SelectOption
+        isNoResultsOption={true}
+        data-testid="policies-loading"
+        key={'None'}
+      >
+        <Spinner size="md" />
+      </SelectOption>
+    );
+  }
 
   return (
     <FormGroup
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -135,9 +183,12 @@ const ProfileSelector = ({ input }) => {
...
         onFilter={(_event, value) => {
           return [
-            <OScapNoneOption setProfileName={setProfileName} key={profiles} />,
+            <OScapNoneOption
+              setProfileName={setProfileName}
+              key="oscap-none-option"
+            />,
           ].concat(
-            profiles.map((profile_id, index) => {
+            (profiles || []).map((profile_id, index) => {
```
lavocatt, line 191: That's a pretty original way of solving this!
Have you thought about making the profile not null if empty in the first step instead?

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -17,37 +19,47 @@ import {
...
  * Component for the user to select the profile to apply to their image.
  * The selected profile will be stored in the `oscap-profile` form state variable.
  * The Component is shown or not depending on the ShowSelector variable.
  */
-const ProfileSelector = ({ input }) => {
+
+const ProfileSelector = ({ input }: ProfileSelectorProps) => {
   const { change, getState } = useFormApi();
+  const [profileName, setProfileName] = useState<string | undefined>('');
   const [profile, setProfile] = useState(getState()?.values?.['oscap-profile']);
```
lavocatt, line 62: Could be   `const [profile, setProfile] = useState(getState()?.values?.['oscap-profile'] ? getState()?.values?.['oscap-profile'] : []);` instead

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -17,37 +19,47 @@ import {
...
 
 import {
+  DistributionProfileItem,
   useGetOscapCustomizationsQuery,
   useGetOscapProfilesQuery,
 } from '../../../store/imageBuilderApi';
 import { reinitFileSystemConfiguratioStep } from '../steps/fileSystemConfiguration';
 import { reinitPackagesStep } from '../steps/packages';
 
+// eslint-disable-next-line
```
lavocatt, line 33: Why do you need to disable the linter here?

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -17,37 +19,47 @@ import {
...
+type ChangeType = <F extends keyof Record<string, any>>(
+  name: F,
+  // eslint-disable-next-line
```
lavocatt, line 36: Why do you need to disable the linter here?

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -68,9 +80,16 @@ const ProfileSelector = ({ input }) => {
...
-    if (data) {
-      setProfileName(data?.openscap?.profile_name);
+    if (
+      data &&
+      data.openscap &&
+      typeof data.openscap.profile_name === 'string'
+    ) {
+      setProfileName(data.openscap.profile_name);
+    } else {
+      setProfileName('None');
```
lavocatt, line 92: I don't think setting the profile name to litteral string `None` is necessary to convert to typescript? Why choosing this particular value?
There were no else case in the `js` version, why adding one was necessary here?

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -198,45 +234,46 @@ const OScapNoneOption = ({ setProfileName }) => {
...
+    }
   }
 
   return (
     <SelectOption
       key={profile_id}
       value={profile_id}
       onClick={() => {
-        setProfileName(data?.openscap?.profile_name);
+        setProfileName(data?.openscap?.profile_name || 'Default Value');
```
lavocatt, line 267: I don't think adding a `Default value` literal string is the correct approach to solve this issue. As for the `None` literal string you were adding up there, it changes the logic of the component. Which is not what you want in a conversion from `js` to `tsx`. The goal being to make sure that behavior is identical.

You need to find another way to solve this one.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -198,45 +234,46 @@ const OScapNoneOption = ({ setProfileName }) => {
...
-const OScapSelectOption = ({ profile_id, setProfileName, input }) => {
+const OScapSelectOption = ({
+  profile_id,
+  setProfileName,
+  input,
+}: {
+  profile_id: DistributionProfileItem;
+  setProfileName: (name: string) => void;
+  input?: string;
+}) => {
```
lavocatt, line 245: We usually define a type that holds the info and reference that instead of typing directly.
Could be cool to keep the pattern that is used in the rest of code base. 

```
type OScapSelectOptionPropType = {
  profile_id: DistributionProfileItem;
  setProfileName: (name: string) => void;
  input?: string;
}
```

I agree that's practically equivalent as the other option you wrote, but let's have all the code having the same vibe.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -198,45 +234,46 @@ const OScapNoneOption = ({ setProfileName }) => {
...
-    return null;
+
+  if (input) {
+    const profileName = data?.openscap?.profile_name;
+    if (
+      profileName &&
+      !profileName.toLowerCase().includes(input.toLowerCase())
+    ) {
+      return null;
+    }
```
lavocatt, line 259: Would just adding an interrogation mark would have sufficed here?

```
  if (
    input &&
    !data?.openscap?.profile_name?.toLowerCase().includes(input.toLowerCase())
  )
```

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -185,7 +217,11 @@ const ProfileSelector = ({ input }) => {
   );
 };
 
-const OScapNoneOption = ({ setProfileName }) => {
+const OScapNoneOption = ({
+  setProfileName,
+}: {
+  setProfileName: (name: string) => void;
+}) => {
```
lavocatt, line 224: We usually define a type that holds the info and reference that instead of typing directly.
Could be cool to keep the pattern that is used in the rest of code base.

```
type OScapNoneOptionPropType = {
  setProfileName: (name: string) => void;
}
```

I agree that's practically equivalent as the other option you wrote, but let's have all the code having the same vibe.

lavocatt (changes requested):
Thanks for working on this. That's a good start.
Keep in mind that in a `js` to `tsx` transformation, we need to make sure the behavior of the component stays exactly the same. I've pointed out two places in your PR where changes in the logic have occurred. It's important that these two get written differently.

Happy hacking :rocket:

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -88,14 +107,43 @@ const ProfileSelector = ({ input }) => {
...
+      <SelectOption
+        isNoResultsOption={true}
+        data-testid="policies-loading"
+        key={'None'}
+      >
+        <Spinner size="md" />
+      </SelectOption>
+    );
+  }
+
```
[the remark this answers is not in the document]
mgold1234: @lavocatt , I moved this part outside of Select, becuase children of select should be React.ReactElement[] and if its in the middle of the select it can return boolean

src/Components/CreateImageWizard/formComponents/Oscap.tsx
mgold1234: because there was error  when runnig the lint: 33:51  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  35:26  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

src/Components/CreateImageWizard/formComponents/Oscap.tsx
mgold1234: because there was error  when runnig the lint: 33:51  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  35:26  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

src/Components/CreateImageWizard/formComponents/Oscap.tsx
lavocatt: Thanks for the explanation. Unfortunately you can't disable the linter in that case. You need to find the correct type to replace `any`.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
lavocatt: Thanks for the explanation. Unfortunately you can't disable the linter in that case. You need to find the correct type to replace `any`.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
mgold1234: I define it with new variable const safeProfiles = useMemo(() => profiles || [], [profiles]);
but I am not sure why is it better then  it is now

src/Components/CreateImageWizard/formComponents/Oscap.tsx
lavocatt: The main problem is that it is changing the logic of the component. Which could be okay in another context. Right now the goal is to transform a js to a ts. It's better for trackability if we don't add other more than necessary changes during that step.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
mgold1234: We utilize the 'change' function from useFormApi, which is sourced from data-driven-forms. This necessitates the temporary use of 'any' as the type. We have plans to phase out data-driven-forms in the near future, making this a provisional solution.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
mgold1234: agree

src/Components/CreateImageWizard/formComponents/Oscap.tsx
mgold1234: its enough, thank you @lavocatt

src/Components/CreateImageWizard/formComponents/Oscap.tsx
mgold1234: Screenshot 2024-01-10 at 16 59 40

lavocatt:
```
type ChangeType = <F extends keyof Record<string, string | undefined>>(
  name: F,
  value?: Record<string, string | undefined>[F]
) => void;
```

When we call `change` we pass it either an undefined value or a string. So let's use that to type the `any` part.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -17,37 +19,47 @@ import {
...
 /**
  * Component for the user to select the profile to apply to their image.
  * The selected profile will be stored in the `oscap-profile` form state variable.
  * The Component is shown or not depending on the ShowSelector variable.
  */
-const ProfileSelector = ({ input }) => {
+
+const ProfileSelector = ({ input }: ProfileSelectorProps) => {
   const { change, getState } = useFormApi();
+  const [profileName, setProfileName] = useState<string | undefined>('');
```
lavocatt, line 61: I would prefer to `profileName` to just be typed as a string. and then when you need to empty it, to just pass an empty string instead of undefined.

lucasgarfield: Agreed, that is consistent with the `<Select>`'s `selections` prop typing:
image

Just want to add one thing, shouldn't the profileName be initialized to `'None'`? We're still working through the final UX but one intention we have for now is it should be very clear to the user that they do not _need_ to select a profile...

lavocatt (approved):
Thanks for all the work you've put into this one! All got to me now. (You just need a final rebase to merge it)

PR #1548, opened 2024-01-11
V2Wizard: Add `<Gcp>` step

lucasgarfield:
`<Gcp>` state is now managed with RTK instead of DDF.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
@@ -130,7 +130,7 @@ const Gcp = () => {
           >
             <ValidatedTextInput
               ariaLabel="google principal"
-              data-testid="principal"
+              dataTestId="principal"
               value={gcpEmail || ''}
               validator={isGcpEmailValid}
               onChange={(_event, value) => dispatch(changeGcpEmail(value))}
--- a/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
@@ -15,7 +15,12 @@ import { ValidatedTextInput } from '../../../ValidatedTextInput';
 import { isGcpEmailValid } from '../../../validators';
 
 export type GcpShareMethod = 'withGoogle' | 'withInsights';
-export type GcpAccountTypes = 'google' | 'service' | 'group' | 'domain';
+export type GcpAccountType =
+  | 'google'
+  | 'service'
+  | 'group'
+  | 'domain'
+  | undefined;
 
 const Gcp = () => {
   const dispatch = useAppDispatch();
--- a/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
@@ -1,0 +1,134 @@
+import React from 'react';
+
+import { Radio, Text, Form, Title, FormGroup } from '@patternfly/react-core';
+
+import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
+import {
+  changeGcpAccountType,
+  changeGcpEmail,
+  changeGcpShareMethod,
+  selectGcpAccountType,
+  selectGcpEmail,
+  selectGcpShareMethod,
+} from '../../../../../store/wizardSlice';
+import { ValidatedTextInput } from '../../../ValidatedTextInput';
+import { isGcpEmailValid } from '../../../validators';
+
+export type GcpShareMethod = 'withGoogle' | 'withInsights';
+export type GcpAccountTypes = 'google' | 'service' | 'group' | 'domain';
+
+const Gcp = () => {
+  const dispatch = useAppDispatch();
+
+  const accountType = useAppSelector((state) => selectGcpAccountType(state));
+  const shareMethod = useAppSelector((state) => selectGcpShareMethod(state));
+  const gcpEmail = useAppSelector((state) => selectGcpEmail(state));
+
+  return (
+    <Form>
+      <Title headingLevel="h2">
+        Target environment - Google Cloud Platform
+      </Title>
+      <Text>
+        Select how to share your image. The image you create can be used to
+        launch instances on GCP, regardless of which method you select.
+      </Text>
+      <FormGroup label="Select image sharing" isRequired>
+        <Radio
+          id="radio-with-description"
+          label="Share image with a Google acount"
+          name="radio-7"
+          description={
+            <Text>
+              Your image will be uploaded to GCP and shared with the account you
+              provide below.
+              <b>The image expires in 14 days.</b> To keep permanent access to
+              your image, copy it to your GCP project.
+            </Text>
+          }
+          isChecked={shareMethod === 'withGoogle'}
+          onChange={() => {
+            dispatch(changeGcpShareMethod('withGoogle'));
+          }}
+        />
+        <Radio
+          id="radio"
+          label="Share image with Red Hat Insights only"
+          name="radio-8"
+          description={
+            <Text>
+              Your image will be uploaded to GCP and shared with Red Hat
+              Insights.
+              <b> The image expires in 14 days.</b> You cannot access or
+              recreate this image in your GCP project.
+            </Text>
+          }
+          isChecked={shareMethod === 'withInsights'}
+          onChange={() => {
+            dispatch(changeGcpShareMethod('withInsights'));
+          }}
+        />
+      </FormGroup>
+      {shareMethod === 'withGoogle' && (
+        <>
+          <FormGroup label="Account type" isRequired>
+            <Radio
+              id="google"
+              label="Google account"
+              name="radio-9"
+              isChecked={accountType === 'google'}
+              onChange={() => {
+                dispatch(changeGcpAccountType('google'));
+              }}
+            />
+            <Radio
+              id="service"
+              label="Service account"
+              name="radio-10"
+              isChecked={accountType === 'service'}
+              onChange={() => {
+                dispatch(changeGcpAccountType('service'));
+              }}
+            />
+            <Radio
+              id="group"
+              label="Google group"
+              name="radio-11"
+              isChecked={accountType === 'group'}
+              onChange={() => {
+                dispatch(changeGcpAccountType('group'));
+              }}
+            />
+            <Radio
+              id="domain"
+              label="Google Workspace domain or Cloud Identity domain"
+              name="radio-12"
+              isChecked={accountType === 'domain'}
+              onChange={() => {
+                dispatch(changeGcpAccountType('domain'));
+              }}
+            />
+          </FormGroup>
+          <FormGroup
+            label={
+              accountType === 'domain'
+                ? 'Domain'
+                : 'Principal (e.g. e-mail address)'
+            }
+            isRequired
+          >
+            <ValidatedTextInput
+              ariaLabel="principal"
+              value={gcpEmail || ''}
+              validator={isGcpEmailValid}
+              onChange={(_event, value) => dispatch(changeGcpEmail(value))}
+              helperText="Please enter a valid e-mail address."
+            />
+          </FormGroup>
+        </>
+      )}
+    </Form>
+  );
+};
+
+export default Gcp;
--- a/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
@@ -40,9 +40,10 @@ const Gcp = () => {
       </Text>
       <FormGroup label="Select image sharing" isRequired>
         <Radio
-          id="radio-with-description"
+          id="share-with-google"
+          data-testid="share-with-google"
           label="Share image with a Google acount"
-          name="radio-7"
+          name="radio-1"
           description={
             <Text>
               Your image will be uploaded to GCP and shared with the account you
@@ -57,9 +58,10 @@ const Gcp = () => {
           }}
         />
         <Radio
-          id="radio"
+          id="share-with-insights"
+          data-testid="share-with-insights"
           label="Share image with Red Hat Insights only"
-          name="radio-8"
+          name="radio-2"
           description={
             <Text>
               Your image will be uploaded to GCP and shared with Red Hat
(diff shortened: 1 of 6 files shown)


lucasgarfield:
I left a note about how the validator allows `asdf@asdf` for the Google principal. This behavior is actually also present in production, it is the same regex expression as the V1 wizard. I don't know enough about GCP to say for sure but I assume this is a bug, any ideas @croissanne?

croissanne:
> I left a note about how the validator allows `asdf@asdf` for the Google principal. This behavior is actually also present in production, it is the same regex expression as the V1 wizard. I don't know enough about GCP to say for sure but I assume this is a bug, any ideas @croissanne?

Wasn't really bug, I think the idea was that some validation is better than no validation. Ig we should just check for a valid email.

regexowl:
The question is what should the Google Workspace domain / Cloud Identity domain look like? From a (very) quick search it seems that those would be e-mails as well. If so we can use the same validation pattern as we do for the other principals. Or do we need a different pattern to check? I'm still not super familiar with GCP :sweat_smile:

regexowl:
The regex is mostly fine, it's just the last period that needs to be escaped and then it should work for emails. So like this: `^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$`

image

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Gcp/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+              Your image will be uploaded to GCP and shared with Red Hat
+              Insights.
+              <b> The image expires in 14 days.</b> You cannot access or
+              recreate this image in your GCP project.
+            </Text>
+          }
+          isChecked={shareMethod === 'withInsights'}
+          onChange={() => {
+            dispatch(changeGcpShareMethod('withInsights'));
+          }}
```
regexowl, line 69: `accountType` and principal should probably get reset here.

image

lucasgarfield: So I think there are 3 options:
1. Don't reset them. In the submitHandler we use a conditional based on the shareMethod to generate the request, if the shareMethod is 'withInsights' we just ignore accountType and email. (lowest maintenance burden but also the least 'safe' solution)
2. Use the RTK listeners (consistent with how we are managing the architecture/release/target situation, but adds more maintenance burden).
3. Just set those to undefined when the shareMethod is set to 'withInsights'. (easy to understand and maintain, just not completely consistent with using RTK listeners we already have in place)

I don't have strong feelings either way, given the pros and cons of the above options which is your favorite?

regexowl: My vote would go to setting them to undefined so the store is not cluttered with things we won't use and go against the other choices.

lucasgarfield:
Current code ready for review but don't merge yet I need to enable the GCP tests in a new commit, forgot about that! :man_facepalming:

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -9,11 +9,10 @@ export const isAwsAccountIdValid = (awsAccountId: string | undefined) => {
   return false;
 };
 
-// TODO: this validator thinks asdf@asdf is a valid e-mail address, is that intentional or a bug?
 export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
   if (
     gcpShareWithAccount !== undefined &&
-    /^[a-z0-9._%+-]+@[a-z0-9.-]+.[a-z]{2,}$/.test(gcpShareWithAccount)
+    /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,253}$/.test(gcpShareWithAccount)
```
regexowl, line 15:
```suggestion
    /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,12}$/.test(gcpShareWithAccount) &&
    gcpShareWithAccount.length <= 253
```

This boundary applies only the to the last character set `[a-z]` so for example to the **.com** in `test@test.com`. I'd say we can make the upper limit way smaller. 

We could probably also add a max limit to the entire e-mail address, but that could be done with adding `.length` check to the validator which could be easier to read than adding the char limits into the regex.

src/store/wizardSlice.ts
```diff
@@ -126,6 +126,11 @@ export const wizardSlice = createSlice({
       state.aws.source = action.payload;
     },
     changeGcpShareMethod: (state, action: PayloadAction<GcpShareMethod>) => {
+      if (action.payload === 'withInsights') {
+        // Reset 'withGoogle' options to initial state
+        state.gcp.accountType = 'google';
```
regexowl, line 131: More of a question - do we need to keep the accountType set to "google"?

regexowl (commented):
Looks great!! I've added one more nit pick and a question

src/store/wizardSlice.ts
lucasgarfield: @lavocatt does your :+1: mean you agree it is a good question or that you agree we need to keep the accountType set to "google"? :sweat_smile: 

My thinking was that it is the default state. But I am open to changing it to `undefined` - that would help keep the store state cleaner as you mentioned and that will really help us reduce our bug surface area for the submitHandler (and more, I suspect). So let's try it!

lavocatt: :+1: to: wondering the same ;)

lavocatt (commented):
All looks good to me except one behavior that I would qualify as bug-ish, but you be the judge.

If the user selects a target env, fills up the details, remove the target, and finally select the same again (hard to read sorry!) then the details are the ones the user already filled up the first time.

I think I would be better to reinitialize the state of a target step when the user selects it, regardless of the fact that it was already filled up in a past interaction.

I don't know if my description is super clear, don't hesitate to ask me to clarify.

Here's a video:
[Screencast from 2024-01-12 16-36-21.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/fd5b857f-a2e9-4ec8-a7b9-ea043362d017)

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -8,3 +8,13 @@ export const isAwsAccountIdValid = (awsAccountId: string | undefined) => {
...
 };
+
+export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
+  if (
+    gcpShareWithAccount !== undefined &&
+    /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,253}$/.test(gcpShareWithAccount)
+  ) {
+    return true;
+  }
+  return false;
```
lavocatt, line 19: Slight code improvement:

```
return gcpShareWithAccount !== undefined && 
    /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,253}$/.test(gcpShareWithAccount)
```

does the same as:
```
  if (
    gcpShareWithAccount !== undefined &&
    /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,253}$/.test(gcpShareWithAccount)
  ) {
    return true;
  }
  return false;
```

but shorter :slightly_smiling_face:

src/Components/CreateImageWizardV2/validators.ts
lavocatt: My 2 cents here, but there's actually an RFC that gives the proper regex to use to validate email addresses:
https://pdw.ex-parrot.com/Mail-RFC822-Address.html
there's a JS implementation apparently here https://gist.github.com/badsyntax/719800

It's worth having a look I think :slightly_smiling_face:

lavocatt: https://stackoverflow.com/questions/20771794/mailrfc822address-regex has a good answer on the topic too.

regexowl: Ugh, I've just realized - is the validation case sensitive? Because I think it is and we don't really take this into account. 

```suggestion
    /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,12}$/.test(gcpShareWithAccount) &&
    gcpShareWithAccount.length <= 253
```

lucasgarfield:
> All looks good to me except one behavior that I would qualify as bug-ish, but you be the judge.
> 
> If the user selects a target env, fills up the details, remove the target, and finally select the same again (hard to read sorry!) then the details are the ones the user already filled up the first time.
> 
> I think I would be better to reinitialize the state of a target step when the user selects it, regardless of the fact that it was already filled up in a past interaction.
> 
> I don't know if my description is super clear, don't hesitate to ask me to clarify.
> 
> Here's a video: [Screencast from 2024-01-12 16-36-21.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/fd5b857f-a2e9-4ec8-a7b9-ea043362d017)

That makes sense to me!

src/Components/CreateImageWizardV2/validators.ts
lucasgarfield: I agree we should dig into this at some point, but perhaps not now... so much depends on shipping the V2 Wizard, I think we should avoid getting bogged down in details like this. I opened an issue and we can figure this all out in a follow-up PR. For now, would you all mind calling our minor improvement on what we already have in production a win and moving on?

https://github.com/RedHatInsights/image-builder-frontend/issues/1554

src/store/wizardSlice.ts
lucasgarfield: Everything gets set to `undefined` now!

lucasgarfield:
> All looks good to me except one behavior that I would qualify as bug-ish, but you be the judge.
> 
> If the user selects a target env, fills up the details, remove the target, and finally select the same again (hard to read sorry!) then the details are the ones the user already filled up the first time.
> 
> I think I would be better to reinitialize the state of a target step when the user selects it, regardless of the fact that it was already filled up in a past interaction.
> 
> I don't know if my description is super clear, don't hesitate to ask me to clarify.
> 
> Here's a video: [Screencast from 2024-01-12 16-36-21.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/fd5b857f-a2e9-4ec8-a7b9-ea043362d017)



> All looks good to me except one behavior that I would qualify as bug-ish, but you be the judge.
> 
> If the user selects a target env, fills up the details, remove the target, and finally select the same again (hard to read sorry!) then the details are the ones the user already filled up the first time.
> 
> I think I would be better to reinitialize the state of a target step when the user selects it, regardless of the fact that it was already filled up in a past interaction.
> 
> I don't know if my description is super clear, don't hesitate to ask me to clarify.
> 
> Here's a video: [Screencast from 2024-01-12 16-36-21.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/fd5b857f-a2e9-4ec8-a7b9-ea043362d017)

Agreed 100%. Can we do this in a follow-up PR so this PR doesn't get too big? This behavior could be a bit tricky to implement (will we use listeners? etc...).

I created an issue to track it:
https://github.com/RedHatInsights/image-builder-frontend/issues/1555

lucasgarfield:
Ready for review again @regexowl & @lavocatt.

I would like to have follow-up PRs for (1) the e-mail validation and (2) the deselecting target bug if you all agree. I've created issues for both of these so we don't lose track of them.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -8,3 +8,14 @@ export const isAwsAccountIdValid = (awsAccountId: string | undefined) => {
...
+export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
+  if (
+    gcpShareWithAccount !== undefined &&
+    /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,12}$/.test(gcpShareWithAccount) &&
+    gcpShareWithAccount.length <= 253
+  ) {
+    return true;
+  }
+  return false;
+};
```
lavocatt, line 21: Near nitpick, but I would prefer to have:
```
export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
  return gcpShareWithAccount !== undefined &&
    /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,12}$/.test(gcpShareWithAccount) &&
    gcpShareWithAccount.length <= 253;
};
```

src/Components/CreateImageWizardV2/validators.ts
regexowl: Thanks for the issue, let's look at it later.

lavocatt:
> Ready for review again @regexowl & @lavocatt.
> 
> I would like to have follow-up PRs for (1) the e-mail validation and (2) the deselecting target bug if you all agree. I've created issues for both of these so we don't lose track of them.

It looks fine to me to have the email validation in a followup since that is qualifying as an improvement compared to the legacy wizard.

However I prefer if the deselecting target bug would be handled in this PR. The reason why I believe that should be done here is that, follow up PRs (GCP target for instance) will suffer from the same issue. Finding a solution early to this would  simplify the implementation and review of subsequent steps.

An alternative to fixing the bug in this PR is to tackle it before the development of the next step starts.

wdyt?

regexowl:
Is the bug connected only to GCP substep or does it affect other steps as well? If the later is true then I'd maybe prefer to have the bug sorted out in a separate PR for the purpose of more clarity in git history. 

I feel like that might make our lives with git-blame and searching through closed PRs easier in the future, please correct me if I'm getting something wrong.

But I also agree that it would be good to sort this one out earlier than later.

Talking about the e-mail validation - I'm perfectly fine with leaving that for a follow up PR. The pattern can be fine tuned later. As for how we return the values - if we change the way we do that for both validators at once, then I'd say again a separate PR would be better.  My way of seeing this is that we shouldn't make "random" changes to the code under a PR of which the scope does not contain said part of the code.

Do I make sense? I feel like I may not :sweat_smile:

regexowl:
> I find that the main risk there is with follow up PRs is that they are rarely done. Even when we track issues, they often pile up and more urgent things are quite often getting in the way. Until we discard the issue because it's not relevant anymore. Writing a `if then else` one way or the other is never going to change our lives, and I wouldn't be surprised if no one is spending time opening a PR just for that 😅 In any case, I don't mind at all if the comment is taken into account or not.

@lavocatt Fair enough! I must agree that hoping to follow up with a follow up PR is currently not a reliable way to handle exceptions :( But this also should not block this specific PR. 

For the returns in validators - I volunteer to open up the follow up and we can decide on a way how to handle them there. 

Talking about the bug you've found I feel like that's one of those that we will need to inevitably tackle in the not so distant future as there will not be a way around it. What could be done now is to add an "IMPORTANT" label or something like that to the issue and also look into reviewing our workflows.

From my point of view this PR can be approved and merged. What do you think?

lavocatt:
> But this also should not block this specific PR.

I agree, that it is nice to keep the flow of contribution free from blockers.

> For the returns in validators - I volunteer to open up the follow up and we can decide on a way how to handle them there.

Thanks!

> Talking about the bug you've found I feel like that's one of those that we will need to inevitably tackle in the not so distant future as there will not be a way around it. What could be done now is to add an "IMPORTANT" label or something like that to the issue and also look into reviewing our workflows.
> 
> From my point of view this PR can be approved and merged. What do you think?

:+1:  for the two points.

lavocatt (approved):
:rocket:

regexowl (approved):

PR #1551, opened 2024-01-12
V2Wizard: Registration step

regexowl:
This adds the Registration step to the V2Wizard.


regexowl:
Some notes:

- `ActivationKeys` was renamed to `ActivationKeysList` because it was clashing with `ActivationKeys` item in rhsm API schema. I didn't end up using the schema item here in the end, but I feel like this could save someone the same debugging in the future.
- There are `env` values added to the `wizardSlice`, those come from https://github.com/RedHatInsights/image-builder-frontend/pull/738 and as far as I understand it they're needed outside of the `ActivationKeysList` as well. So that's why they live in the store now.

lavocatt (approved):
:rocket: good one! I've manually monkey tested it (I hope as good as you would have done on one of my PRs :smile:). Everything is working fine.

I have an out-of-the-scope question (regarding the other ones on the topic that have been merged so far). What is the plan regarding testing? Because so far we're only relying on manual testing, I'm getting slightly worried that adding new steps might have an impact on the existing ones and that we might miss that during reviews.

lavocatt (approved):
:rocket:

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
```diff
@@ -396,1203 +398,1216 @@ describe('Step Upload to AWS', () => {
...
+  //     };
+
+  //     await setUp();
+  //     await clickNext();
+
+  //     await waitFor(() => screen.findByTestId('upload-aws'));
+  //     testTile(screen.getByTestId('upload-aws'));
+  //     testTile(screen.getByTestId('upload-google'));
+  //     testTile(screen.getByTestId('upload-azure'));
+  //   });
```
lavocatt, line 1612: these commented lines have been added, can you remove them please?
@regexowl can you confirm that we don't want those?

lavocatt: sorry wrong pr...

PR #1553, opened 2024-01-12
V2Wizard: Fix wizard reset action

### LINKED ISSUE #1552: wizardv2: state issue
I've noticed a small bug on the wizard refactor. The state doesn't reset when closing/reopenning the wizard.To reproduce:

* open the wizard
* select aws
* close the wizard
* open the wizard

aws is still selected.

lucasgarfield:
Fixes #1552.

The wizardSlice's reset wizard action was not working because when it was passed to `dispatch()` as an argument it was missing the `()`... oops!

It also needs to be dispatched inside of a useEffect hook with an empty dependency array so that it is only dispatched once, when the wizard opens.

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import {
   Button,
@@ -56,7 +56,12 @@ const CreateImageWizard = () => {
   const dispatch = useAppDispatch();
 
   // IMPORTANT: Ensure the wizard starts with a fresh initial state
-  dispatch(initializeWizard);
+  useEffect(() => {
+    dispatch(initializeWizard());
+    // This useEffect hook should run *only* on mount and therefore has an empty
+    // dependency array. eslint's exhaustive-deps rule does not support this use.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   /*           *
    * Selectors *


src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -52,8 +52,13 @@ const CreateImageWizard = () => {
...
   const dispatch = useAppDispatch();
 
-  // Ensure the wizard starts with a fresh initial state
-  dispatch(initializeWizard);
+  // IMPORTANT: Ensure the wizard starts with a fresh initial state
+  useEffect(() => {
+    dispatch(initializeWizard());
+    // This useEffect hook should run *only* on mount and therefore has an empty
+    // dependency array. eslint's exhaustive-deps rule does not support this use.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
regexowl, line 60: Hmm... it works, but I must admit I'm not a fan of commenting out rules which we should probably follow. How about we re-initialized (reset would be then probably a better name for the action) the Wizard on the `onClick` of close button? Would that make sense?

So something like 
```javascript
      <Button
        variant="link"
        onClick={() => {
          dispatch(initializeWizard());
          close();
        }}
      >
```

in the Custom Wizard Footer.

Also would this `useEffect` interfere with the Recreate action in the future?

lucasgarfield: That would be okay if those were the only ways to get into the wizard, but we also currently support accessing the wizard from its own route (which is totally appropriate since it is now its own page).

But I think in this case the problem really is with eslint. Adding an empty dependency array is a standard practice outlined in the React docs: https://react.dev/reference/react/useEffect (grep for 'empty').

We could also add `dispatch` to the dependency array... that would probably work. I don't think anything would ever cause dispatch to change, so the effect would be the same.

But I like having the empty array because it communicates to the reader exactly the intention - run only once on initial mount.

regexowl: I see, okay, let's merge it as is. Later we should probably discuss how to approach the `exhaustive-deps` rule in our code base as it seems to be causing most of the linting problems we have now.

regexowl (dismissed):

lavocatt:
OOps I've done it too lol https://github.com/RedHatInsights/image-builder-frontend/issues/1552 I'll close mine

regexowl (approved):

PR #1556, opened 2024-01-14
V2Wizard: Oscap step

mgold1234:
this commit adds the Oscap component to V2Wizrad.

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -10,6 +10,7 @@ import {
 import { useNavigate } from 'react-router-dom';
 
 import ImageOutputStep from './steps/ImageOutput';
+import OscapStep from './steps/Oscap';
 import Aws from './steps/TargetEnvironment/Aws';
 import { isAwsAccountIdValid } from './validators';
 
@@ -21,6 +22,7 @@ import {
   selectAwsShareMethod,
   selectAwsSource,
   selectImageTypes,
+  selectProfile,
 } from '../../store/wizardSlice';
 import { resolveRelPath } from '../../Utilities/path';
 import { ImageBuilderHeader } from '../sharedComponents/ImageBuilderHeader';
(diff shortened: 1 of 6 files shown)


src/store/imageBuilderApi.ts
```diff
@@ -721,6 +721,7 @@ export type PackagesResponse = {
   data: Package[];
 };
 export type DistributionProfileItem =
+  | "None"
```
lavocatt, line 724: Hey, just a quick comment on this one:

One thing to remember is that the api files shouldn't get updated manually. These files are generated by the `npm run api` command. The command is loading the openapi yaml file for the image builder api and generating the types for us. If you change manually the file, the next time someone will update the yaml file, your changes will be erased and lost.

And the other thing to remember is that the openapi yaml file in the frontend *has to* be in sync with image builder's file that lives in the image builder repository too.

If you feel like a None option is necessary, the workflow is the following:
* You need to first issue a PR on image builder to have its API changed.
* Once it's merged, import the new yaml file into the frontend repository.
* Run the `npm run api` command
* And issue a PR on the frontend :slightly_smiling_face:

lavocatt: BTW, I don't think adding a `None` option to the image builder api is a good idea. In the API perspective, not sending a profile means `None` already.

lavocatt:
I've been running the code locally using the mock service workers. There's an issue with in the OScap step, where on `msw` a profile is selected by default 
image
it's not doing the same without the msw, I think it's worth investigating why.

mgold1234:
I run this with npm run stage-beta:experimental.
@lucasgarfield should we be able to run it with msw as well?

src/store/imageBuilderApi.ts
mgold1234: I remove it from the imageBuilderApi. thank you

mgold1234:
I rebase the code and dill with a lot of conflict, so closing this pr and opend new one. 
https://github.com/RedHatInsights/image-builder-frontend/pull/1559

PR #1559, opened 2024-01-15
V2Wizard: add oscap to wizard2

mgold1234:
This adds the OpenSCAP step to the V2Wizard,
adds new values to the wizardSlice and enables relevant tests in the `CreateImageWizard.test.tsx` test suite

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -10,6 +10,7 @@ import {
 import { useNavigate } from 'react-router-dom';
 
 import ImageOutputStep from './steps/ImageOutput';
+import OscapStep from './steps/Oscap';
 import RegistrationStep from './steps/Registration';
 import Aws from './steps/TargetEnvironment/Aws';
 import Gcp from './steps/TargetEnvironment/Gcp';
@@ -158,6 +159,13 @@ const CreateImageWizard = () => {
             }
           >
             <RegistrationStep />
+          </WizardStep>
+          <WizardStep
+            name="OpenSCAP"
+            id="step-oscap"
+            footer={<CustomWizardFooter disableNext={false} />}
+          >
+            <OscapStep />
           </WizardStep>
           <WizardStep
             name="Review"
--- a/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
@@ -1,0 +1,257 @@
+import React, { useEffect, useState } from 'react';
+
+import {
+  Alert,
+  FormGroup,
+  Spinner,
+  Popover,
+  TextContent,
+  Text,
+  Button,
+} from '@patternfly/react-core';
+import {
+  Select,
+  SelectOption,
+  SelectVariant,
+} from '@patternfly/react-core/deprecated';
+import { HelpIcon } from '@patternfly/react-icons';
+
+import OscapProfileInformation from './OscapProfileInformation';
+
+import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
+import {
+  DistributionProfileItem,
+  useGetOscapCustomizationsQuery,
+  useGetOscapProfilesQuery,
+} from '../../../../store/imageBuilderApi';
+import {
+  changeOscapProfile,
+  selectDistribution,
+  selectProfile,
+} from '../../../../store/wizardSlice';
+
+const ProfileSelector = () => {
+  const oscapProfile = useAppSelector((state) => selectProfile(state));
+  const release = useAppSelector((state) => selectDistribution(state));
+  const dispatch = useAppDispatch();
+  const [profileName, setProfileName] = useState<string | undefined>('None');
+  const [isOpen, setIsOpen] = useState(false);
+  const {
+    data: profiles,
+    isFetching,
+    isSuccess,
+    isError,
+    refetch,
+  } = useGetOscapProfilesQuery({
+    distribution: release,
+  });
+
+  const { data } = useGetOscapCustomizationsQuery(
+    {
+      distribution: release,
+      // @ts-ignore if oscapProfile is undefined the query is going to get skipped, so it's safe here to ignore the linter here
+      profile: oscapProfile,
+    },
+    {
+      skip: !oscapProfile,
+    }
+  );
+
+  useEffect(() => {
+    if (
+      data &&
+      data.openscap &&
+      typeof data.openscap.profile_name === 'string'
+    ) {
+      setProfileName(data.openscap.profile_name);
+    }
+  }, [data]);
+
+  const handleToggle = () => {
+    if (!isOpen) {
+      refetch();
+    }
+    setIsOpen(!isOpen);
+  };
+
+  const handleClear = () => {
+    dispatch(changeOscapProfile(undefined));
+    setProfileName(undefined);
+  };
+
+  const handleSelect = (
+    _event: React.MouseEvent<Element, MouseEvent>,
+    selection: DistributionProfileItem
+  ) => {
+    dispatch(changeOscapProfile(selection));
+    setIsOpen(false);
+  };
+
+  const options = [
+    <OScapNoneOption setProfileName={setProfileName} key="oscap-none-option" />,
+  ];
+  if (isSuccess) {
+    options.concat(
+      profiles.map((profile_id) => {
+        return (
+          <OScapSelectOption
+            key={profile_id}
+            profile_id={profile_id}
+            setProfileName={setProfileName}
+          />
+        );
+      })
+    );
+  }
+
+  if (isFetching) {
+    options.push(
+      <SelectOption
+        isNoResultsOption={true}
+        data-testid="policies-loading"
+        key={'None'}
+      >
+        <Spinner size="md" />
+      </SelectOption>
+    );
+  }
+
+  return (
+    <FormGroup
+      isRequired={true}
+      data-testid="profiles-form-group"
+      label={
+        <>
+          OpenSCAP profile
+          <Popover
+            maxWidth="30rem"
+            position="left"
+            bodyContent={
+              <TextContent>
+                <Text>
+                  To run a manual compliance scan in OpenSCAP, download this
+                  image.
+                </Text>
+              </TextContent>
+            }
+          >
+            <Button variant="plain" aria-label="About OpenSCAP" isInline>
+              <HelpIcon />
+            </Button>
+          </Popover>
+        </>
+      }
+    >
+      <Select
+        ouiaId="profileSelect"
+        variant={SelectVariant.typeahead}
+        onToggle={handleToggle}
+        onSelect={handleSelect}
+        onClear={handleClear}
+        selections={profileName}
+        isOpen={isOpen}
+        placeholderText="Select a profile"
+        typeAheadAriaLabel="Select a profile"
+        isDisabled={!isSuccess}
+        onFilter={(_event, value) => {
+          if (profiles) {
+            return [
+              <OScapNoneOption
+                setProfileName={setProfileName}
+                key="oscap-none-option"
+              />,
+            ].concat(
+              profiles.map((profile_id, index) => {
+                return (
+                  <OScapSelectOption
+                    key={index}
+                    profile_id={profile_id}
+                    setProfileName={setProfileName}
+                    input={value}
+                  />
+                );
+              })
+            );
+          }
+        }}
+      >
+        {options}
+      </Select>
+      {isError && (
+        <Alert
+          title="Error fetching the profiles"
+          variant="danger"
+          isPlain
+          isInline
+        >
+          Cannot get the list of profiles
+        </Alert>
+      )}
+    </FormGroup>
+  );
+};
+
+type OScapNoneOptionPropType = {
+  setProfileName: (name: string) => void;
+};
+
+const OScapNoneOption = ({ setProfileName }: OScapNoneOptionPropType) => {
+  return (
+    <SelectOption
+      value={undefined}
+      onClick={() => {
+        setProfileName('None');
+      }}
+    >
+      <p>{'None'}</p>
+    </SelectOption>
+  );
+};
+
+type OScapSelectOptionPropType = {
+  profile_id: DistributionProfileItem;
+  setProfileName: (name: string) => void;
+  input?: string;
+};
+
+const OScapSelectOption = ({
+  profile_id,
+  setProfileName,
+  input,
+}: OScapSelectOptionPropType) => {
+  const release = useAppSelector((state) => selectDistribution(state));
+  const { data } = useGetOscapCustomizationsQuery({
+    distribution: release,
+    profile: profile_id,
+  });
+
+  if (
+    input &&
+    !data?.openscap?.profile_name?.toLowerCase().includes(input.toLowerCase())
+  ) {
+    return null;
+  }
+
+  return (
+    <SelectOption
+      key={profile_id}
+      value={profile_id}
+      onClick={() => {
+        if (data?.openscap?.profile_name) {
+          setProfileName(data?.openscap?.profile_name);
+        }
+      }}
+    >
+      <p>{data?.openscap?.profile_name}</p>
+    </SelectOption>
+  );
+};
+
+export const Oscap = () => {
+  return (
+    <>
+      <ProfileSelector />
+      <OscapProfileInformation />
+    </>
+  );
+};
--- a/src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
@@ -1,0 +1,84 @@
+import React from 'react';
+
+import {
+  Spinner,
+  TextContent,
+  TextList,
+  TextListItem,
+  TextListItemVariants,
+  TextListVariants,
+} from '@patternfly/react-core';
+
+import { RELEASES } from '../../../../constants';
+import { useAppSelector } from '../../../../store/hooks';
+import { useGetOscapCustomizationsQuery } from '../../../../store/imageBuilderApi';
+import {
+  selectDistribution,
+  selectProfile,
+} from '../../../../store/wizardSlice';
+
+const OscapProfileInformation = (): JSX.Element => {
+  const release = useAppSelector((state) => selectDistribution(state));
+  const oscapProfile = useAppSelector((state) => selectProfile(state));
+
+  const {
+    data: oscapProfileInfo,
+    isFetching: isFetchingOscapProfileInfo,
+    isSuccess: isSuccessOscapProfileInfo,
+  } = useGetOscapCustomizationsQuery(
+    {
+      distribution: release,
+      // @ts-ignore if oscapProfile is undefined the query is going to get skipped, so it's safe here to ignore the linter here
+      profile: oscapProfile,
+    },
+    {
+      skip: !oscapProfile,
+    }
+  );
+
+  return (
+    <>
+      {isFetchingOscapProfileInfo && <Spinner size="lg" />}
+      {isSuccessOscapProfileInfo && (
+        <TextContent>
+          <br />
+          <TextList component={TextListVariants.dl}>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-u-min-width"
+            >
+              Profile description:
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {oscapProfileInfo.openscap?.profile_description}
+            </TextListItem>
+          </TextList>
+          <TextList component={TextListVariants.dl}>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-u-min-width"
+            >
+              Operating system:
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {RELEASES.get(release)}
+            </TextListItem>
+          </TextList>
+          <TextList component={TextListVariants.dl}>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-u-min-width"
+            >
+              Reference ID:
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {oscapProfileInfo.openscap?.profile_id}
+            </TextListItem>
+          </TextList>
+        </TextContent>
+      )}
+    </>
+  );
+};
+
+export default OscapProfileInformation;
--- a/src/store/wizardSlice.ts
+++ b/src/store/wizardSlice.ts
@@ -1,6 +1,11 @@
 import { PayloadAction, createSlice } from '@reduxjs/toolkit';
 
-import { Distributions, ImageRequest, ImageTypes } from './imageBuilderApi';
+import {
+  DistributionProfileItem,
+  Distributions,
+  ImageRequest,
+  ImageTypes,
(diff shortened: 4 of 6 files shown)


src/store/wizardSlice.ts
```diff
@@ -28,6 +34,12 @@ type wizardState = {
     accountType: GcpAccountType;
     email: string | undefined;
   };
+  profile: DistributionProfileItem;
```
lavocatt, line 37: This should be  `profile: DistributionProfileItem | undefined;`

src/store/wizardSlice.ts
```diff
@@ -44,6 +56,13 @@ const initialState: wizardState = {
     accountType: 'google',
     email: undefined,
   },
+  // @ts-ignore
```
lavocatt, line 59: you can remove the ts_ignore comment if you fix the type above

src/store/wizardSlice.ts
```diff
@@ -44,6 +56,13 @@ const initialState: wizardState = {
...
+  profile: undefined,
+  openScap: {
+    profileId: 'None',
```
lavocatt, line 62: the `None` profileId shouldn't be sent to the backend, it would be better to have undefined there too.

src/store/wizardSlice.ts
```diff
@@ -37,6 +43,12 @@ type wizardState = {
     registrationType: string;
     activationKey: ActivationKeys['name'];
   };
+  profile: DistributionProfileItem | undefined;
+  openScap: {
+    profileId: OpenScap['profile_id'] | undefined;
+    profileName: OpenScap['profile_name'];
+    profileDescription: OpenScap['profile_description'];
+  };
```
regexowl, line 51: The wizardState is currently split by steps, `aws: {}` contains all values we need for AWS step, `registration: {}` contains everything needed for Registration step. It would be nice to have profile and openScap enclosed in a common object as well.

Also - I think it would make sense to rename `openScap` to `profileInformation` or something similar.

So the structure could then be something like this:
```typescript
openScap: {
    profile: DistributionProfileItem | undefined;
    profileInformation: {
      profileId: OpenScap['profile_id'] | undefined;
      profileName: OpenScap['profile_name'];
      profileDescription: OpenScap['profile_description'];
    };
}
```

What do you think?

src/store/wizardSlice.ts
```diff
@@ -61,6 +73,12 @@ const initialState: wizardState = {
     registrationType: 'register-now-rhc',
     activationKey: '',
   },
+  profile: undefined,
+  openScap: {
+    profileId: undefined,
+    profileName: undefined,
+    profileDescription: undefined,
+  },
```
regexowl, line 81: This would need to get updated as per the comment above. Having an `openScap` as a step, and `profile` and `profileInformation` as a information it contains.

src/store/wizardSlice.ts
```diff
@@ -117,6 +135,22 @@ export const selectActivationKey = (state: RootState) => {
...
+};
+
+export const selectOscapProfileName = (state: RootState) => {
+  return state.wizard.openScap.profileName;
+};
+
+export const selectOscapProfileDescription = (state: RootState) => {
+  return state.wizard.openScap.profileDescription;
+};
+
```
regexowl, line 153: This would just get small update in the structure:
- state.wizard.profile -> state.wizard.openScap.profile
- state.wizard.openScap.profileId -> state.wizard.openScap.profileInformation.profileId

src/store/wizardSlice.ts
```diff
@@ -190,6 +224,24 @@ export const wizardSlice = createSlice({
...
+      action: PayloadAction<string | undefined>
+    ) => {
+      state.openScap.profileName = action.payload;
+    },
+    changeOscapProfileDescription: (
+      state,
+      action: PayloadAction<string | undefined>
+    ) => {
+      state.openScap.profileDescription = action.payload;
+    },
```
regexowl, line 244: Are you using the functions to change the values somewhere? The structure seems to stay empty even when a profile is selected

image

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -159,6 +163,15 @@ const CreateImageWizard = () => {
           >
             <RegistrationStep />
           </WizardStep>
+          <WizardStep
+            name="OpenSCAP"
+            id="step-oscap"
+            footer={
+              <CustomWizardFooter disableNext={oscapProfile === undefined} />
```
regexowl, line 170: I see, so now the user has to choose "None" if they don't want any profile and as "None" is not undefined we can block on undefined.

It seems in the old Wizard we have "None" as the default option, but we also don't block on undefined (which I'm not sure if we should, would have to look deeper into the code).

Would it be possible to set "None" as a default option here as well?

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
```diff
@@ -0,0 +1,274 @@
...
+  changeOscapProfile,
+  selectDistribution,
+  selectProfile,
+} from '../../../../store/wizardSlice';
+
+/**
+ * Component for the user to select the profile to apply to their image.
+ * The selected profile will be stored in the `oscap-profile` form state variable.
+ * The Component is shown or not depending on the ShowSelector variable.
+ */
```
regexowl, line 37: This was true for the DDF version, right? We should either update the comment or remove it.

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+const ProfileSelector = () => {
+  const oscapProfile = useAppSelector((state) => selectProfile(state));
+  const release = useAppSelector((state) => selectDistribution(state));
+  const dispatch = useAppDispatch();
+  const [profileName, setProfileName] = useState<string | undefined>('');
```
regexowl, line 43: Isn't the profileName and oscapProfile a duplicate value?

What do we plan to store in each of those?

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
```diff
@@ -0,0 +1,274 @@
...
+    isError,
+    refetch,
+  } = useGetOscapProfilesQuery({
+    distribution: release,
+  });
+
+  const { data } = useGetOscapCustomizationsQuery(
+    {
+      distribution: release,
+      // @ts-ignore
```
regexowl, line 58: Let's sort out the rest of the comments and see if we can get rid of this ignore.

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
```diff
@@ -0,0 +1,274 @@
...
+};
+
+type OScapNoneOptionPropType = {
+  setProfileName: (name: string) => void;
+};
+
+const OScapNoneOption = ({ setProfileName }: OScapNoneOptionPropType) => {
+  return (
+    <SelectOption
+      value={undefined}
```
regexowl, line 207: Hmm, okay so value of the "None" option is undefined. This is just an idea, but does it mean that when we clear the Select it should display "None" instead of being empty?

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
```diff
@@ -0,0 +1,274 @@
...
+  );
+};
+
+/**
+ * Component to prompt the use with two choices:
+ * - to add a profile, in which case the ProfileSelector will allow the user to
+ *   pick a profile to be stored in the `oscap-profile` variable.
+ * - to not add a profile, in which case the `oscap-profile` form state goes
+ *   undefined.
+ */
```
regexowl, line 262: I think this is also no longer true as we changed the structure of the step. The comment needs to get updated or removed.

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
```diff
@@ -0,0 +1,274 @@
...
+    <>
+      <ProfileSelector />
+      <OscapProfileInformation />
+    </>
+  );
+};
+
+export const Oscap = () => {
+  return <AddProfile />;
+};
```
regexowl, line 274: Do we need to have these separate? Couldn't we just do

```typescript
export const Oscap = () => {
  return (
    <>
      <ProfileSelector />
      <OscapProfileInformation />
    </>
  );
};
```

src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const oscapProfile = useAppSelector((state) => selectProfile(state));
+
+  const {
+    data: oscapProfileInfo,
+    isFetching: isFetchingOscapProfileInfo,
+    isSuccess: isSuccessOscapProfileInfo,
+  } = useGetOscapCustomizationsQuery(
+    {
+      distribution: release,
+      // @ts-ignore
```
regexowl, line 30: Again - it would be good to be able to get rid of this. Let's address the rest of the comments and then we'll revisit this one.

regexowl (commented):
This is looking great! Added some comments, let me know if anything is unclear.

As for the `//@ts-ignore` ones - those are mostly a reminder for me to check out where the problem is when the other comments are addressed so feel free to ignore them for now :)

src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
lavocatt: The thing is, we can't get rid of this `ts-ignore`, the linter doesn't understand that the query is not going to get executed if the profile is undefined, because of the syntax of the skip token. So my take on this is that we should keep the ignore comment but add some description to it so that we have a reminder on why it's there.

regexowl:
Also, could you please update the commit message? Maybe something like

```
V2Wizard: add oscap to wizard2

This adds the OpenSCAP step to the V2Wizard, adds new values to the wizardSlice and enables relevant tests in the `CreateImageWizard.test.tsx` test suite
```

src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
regexowl: Aha, yeah in cases where we truly can't get rid of the ignore I completely agree with adding a comment with a reason why it's there.

src/store/wizardSlice.ts
mgold1234: good idea

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
mgold1234: no, the oscapProfile contains the profile_ id  abd profileName contains the profile_name

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
mgold1234: remove that, thank you

src/store/wizardSlice.ts
mgold1234: I used only changeOscapProfile functions, other data comes from the rtk query

regexowl: Fair enough, would that mean that we don't need 
```typescript
    profileInformation: {
      profileId: OpenScap['profile_id'] | undefined;
      profileName: OpenScap['profile_name'];
      profileDescription: OpenScap['profile_description'];
    };
```
in the wizardSlice at all?

Together with `changeOscapProfileName` and `changeOscapProfileDescription` as I can't see those used outside of the wizardSlice as well.

mgold1234: yes, I thought its good idea to add this inforamtion together with the profileInformation

regexowl: Got it, if we want to keep the information in the Wizard form so we for example don't have to make another API call on the Review step, then we absolutely can do that :) 

My only problem is now that we have the information structure in the store, but we don't fill it in anywhere so it stays empty.

lavocatt: Good catch @regexowl! Yeah, keeping the bare minimum of data in the store is preferable.

mgold1234: removed that part, thank you both

lavocatt:
I find it off that `"None"` is being shown in the dropdown menu by default:
image
Can you change that please?

lavocatt:
Also it's not possible to clear the `"None"` value, even though there's a button to clear out the field. Which makes me feel that the UI is buggy.

regexowl:
> I find it off that `"None"` is being shown in the dropdown menu by default: image Can you change that please?

That would be a result of my comments and might need further discussion if it doesn't feel like a correct functionality
- https://github.com/RedHatInsights/image-builder-frontend/pull/1559#discussion_r1455159762
- https://github.com/RedHatInsights/image-builder-frontend/pull/1559#discussion_r1455171725

What I was wondering about was how to communicate to the user that they don't need to build an image with the OSCAP profile but still manage to work with the complete data, be it profile info or information that the user wishes to not use a profile at all.

Currently we have "None" set as a default value in production which I think communicates the first part clearly. Personally I think it's either that or we should expand the description a bit to let users know that there is a "None" option even if they do not open the select. Might save us some drop off on the step by evading "I wanted to create an image, but I don't want to deal with OSCAP stuff so I might just give up" situation.

But I can see the problem with clearing the select on the "default" undefined option. It will most probably mess with the search, right?

What might be the solution is entering the step on the default "None" option to communicate that option is also a possibility. But we should allow to clear the field (so it's empty and can be used for searching easily), and if it's empty and truly undefined, we would disable the Next button? Or do we want to allow next both on "None" option which is valid and on empty select which is kind off weird... I guess the question is, should the profile be a required field?

Does that make sense? What do you think?

lavocatt:
> What I was wondering about was how to communicate to the user that they don't need to build an image with the OSCAP profile but still manage to work with the complete data, be it profile info or information that the user wishes to not use a profile at all.
> 
> Currently we have "None" set as a default value in production which I think communicates the first part clearly. Personally I think it's either that or we should expand the description a bit to let users know that there is a "None" option even if they do not open the select. Might save us some drop off on the step by evading "I wanted to create an image, but I don't want to deal with OSCAP stuff so I might just give up" situation.
> 
> But I can see the problem with clearing the select on the "default" undefined option. It will most probably mess with the search, right?
> 
> What might be the solution is entering the step on the default "None" option to communicate that option is also a possibility. But we should allow to clear the field (so it's empty and can be used for searching easily), and if it's empty and truly undefined, we would disable the Next button? Or do we want to allow next both on "None" option which is valid and on empty select which is kind off weird... I guess the question is, should the profile be a required field?
> 
> Does that make sense? What do you think?

I had no thought about that this way! What about we gate the drop down under a radio button swicth?
A bit like we do for the filesystem customization?

regexowl:
> I hadn't thought about that this way, thanks! What about we gate the drop down under a radio button swicth? A bit like we do for the filesystem customization?

Do you mean like we did before updating the step? :sweat_smile: (took the screenshot from the very first OSCAP PR https://github.com/RedHatInsights/image-builder-frontend/pull/1371)

image

Let me kick off the discussion in the frontend channel. As I'm not working on OSCAP I'm probably missing some context about the changes.

lavocatt:
> > I hadn't thought about that this way, thanks! What about we gate the drop down under a radio button swicth? A bit like we do for the filesystem customization?
> 
> Do you mean like we did before updating the step? 😅 (took the screenshot from the very first OSCAP PR #1371)
> 
> image
> 
> Let me kick off the discussion in the frontend channel. As I'm not working on OSCAP I'm probably missing some context about the changes.

ahah indeed,t hat :smile:

regexowl:
@mgold1234 We've discussed the step behavior in the frontend channel. I'll mention you in the thread :)

The result is duplicating behavior of the old Wizard 1:1 so:
- "None" as a default/init option (this should be fine)
- empty on clear (this will need updating, sorry about my previous suggestion :disappointed: )
- Next button enabled for both (this should be also fine)

Could you please update the PR as per the second point? :heart:

lavocatt:
The issue now is that we're missing the default `None` value in the field 
image

The behavior should be:
* you've got `None` by default in there
* But when the user goes to clear the field using the button, `None` disappears.

lavocatt (approved):
That is looking good to me. Thanks for improving the UX. You just need a rebase though, I think it'll make the unit tests pass.

lavocatt:
I used the rebase button in the github UI to rebase.

regexowl:
The Travis failure seems like a clear one - the querying isn't awaited. Which made me look at the tests a bit more - I think there might have been some rebasing shenanigans (noticed similar thing in my previous PR) and the commented code got replicated. Currently the file has 1633 lines, but in this PR their number jumped to 2772 and I can see loads of `getBy` queries that were switched to `findBy` in https://github.com/RedHatInsights/image-builder-frontend/pull/1563

It looks like there is only one more test enabled in `CreateImageWizard.test.tsx`, could you please re-do the changes?

Also there is an entire separate testing suite for OSCAP in `test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js` which we should copy over. From my point of view we don't have to do it necessarily in this PR, but we also shouldn't forget about it :)

lavocatt:
> Also there is an entire separate testing suite for OSCAP in `test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js` which we should copy over. From my point of view we don't have to do it necessarily in this PR, but we also shouldn't forget about it :)

Should be done here I think. Too easy to forget otherwise!

lavocatt:
> The Travis failure seems like a clear one - the querying isn't awaited. Which made me look at the tests a bit more - I think there might have been some rebasing shenanigans (noticed similar thing in my previous PR) and the commented code got replicated. Currently the file has 1633 lines, but in this PR their number jumped to 2772 and I can see loads of `getBy` queries that were switched to `findBy` in #1563

Thanks for catching this! Unbelievable that github hides the too large diffs, I completely missed it!

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
```diff
@@ -752,6 +752,57 @@ describe('Step Upload to AWS', () => {
...
   });
 
+  describe('Step Oscap profile', () => {
+    const user = userEvent.setup();
+    const setUp = async () => {
+      ({ router } = await renderCustomRoutesWithReduxRouter(
+        'imagewizard',
+        {},
+        routes
+      ));
```
regexowl, line 762: Hmm, looking at the original tests again. Is this test new? Can't find it in the original file.

mgold1234: the original tests on comliance file on Recreacte describe('Step Compliance', () => {
  const setup = async () => {
    renderCustomRoutesWithReduxRouter('imagewizard', {}, routes);
  };
  test('create an image with None oscap profile', async () => {
    await setup();

src/test/Components/CreateImageWizardV2/CreateImageWizard.compliance.test.tsx
```diff
@@ -0,0 +1,204 @@
+import React from 'react';
+
+import '@testing-library/jest-dom';
+
+import { screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import CreateImageWizard from '../../../Components/CreateImageWizard/CreateImageWizard';
```
regexowl, line 8: The tests are currently testing the old wizard, this will need to be updated like this:
```suggestion
import CreateImageWizard from '../../../Components/CreateImageWizardV2/CreateImageWizard';
```

It should then make some tests fail. Some parts of the tests can be commented out as they are not relevant yet (like checking for the "File system configuration" header and so).

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
regexowl: I see, so is this tests duplicate of this one with changed values? https://github.com/mgold1234/image-builder-frontend/blob/893c26eb5bdd7ed6c6c6c8fc950c3fb336b1726e/src/test/Components/CreateImageWizardV2/CreateImageWizard.compliance.test.tsx#L119-L170

If so I don't think we need to add here again as it's already contained in the `.compliance` test suite.

src/test/Components/CreateImageWizard/CreateImageWizard.compliance.test.js
```diff
@@ -5,7 +5,7 @@ import '@testing-library/jest-dom';
 import { screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
-import CreateImageWizard from '../../../Components/CreateImageWizard/CreateImageWizard';
+import CreateImageWizard from '../../../Components/CreateImageWizardV2/CreateImageWizard';
```
regexowl, line 8: These changes should probably be in the `src/test/Components/CreateImageWizardV2/CreateImageWizard.compliance.test.tsx` file.

mgold1234: I think I chagned it :)

regexowl: Oh I mean that you made the changes in the CreateImageWizard version of the test

image

mgold1234: ye, sure you correct

lavocatt (changes requested):
:rocket: Thanks for all the work!

lavocatt (dismissed):
🚀 Thanks for all the work!

src/test/Components/CreateImageWizardV2/CreateImageWizard.compliance.test.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+    //   await screen.findByRole('heading', {
+    //     name: /Additional Red Hat packages/i,
+    //   });
+    //   await screen.findByText(/nftables/i);
+    //    await screen.findByText(/libselinux/i);
+  });
+});
+
+describe('On Recreate', () => {
```
regexowl, line 177:
```suggestion
// 
// TO DO - check if the correct version of Wizard is tested
//
describe('On Recreate', () => {
```

src/test/Components/CreateImageWizardV2/CreateImageWizard.compliance.test.tsx
```diff
@@ -0,0 +1,209 @@
...
+    await screen.findByText('/tmp');
+
+    // check that the Packages contain a nftable package
+    await clickNext();
+    await screen.findByRole('heading', {
+      name: /Additional Red Hat packages/i,
+    });
+    await screen.findByText('nftables');
+  });
+});
```
regexowl, line 209: This entire test shouldn't pass. Let's comment it out for now and add a comment to check if the right Wizard is tested when we implement the functionality in V2

regexowl (changes requested):
The code looks great! The only last thing that I'm looking at is the compliance test suite - the "On Recreate" test shouldn't pass. There is no recreate functionality in V2Wizard, no FSC step... so I've tried `screen.logTestingPlaygroundURL()` and it looks like the original Wizard gets rendered in this specific test (other tests in the suite look fine).

Could you please comment out the entire "On Recreate" test for now as we don't have anything to test in there yet? I've also added a comment in a suggestion so we know there's some checking to do when we get to the test.

regexowl:
Looks good! Can you please fix the linting errors? I'd say we are good to merge afterwards :)

regexowl (approved):
Looks great! Thanks for all your work on this! :sparkles:

PR #1561, opened 2024-01-16
wizardv2: fix the reinitialisation bug

### LINKED ISSUE #1552: wizardv2: state issue

lavocatt:
The V2 wizard was not reinitializing itself when the user was opting out of the form and back again. Which meant that the old data from a previous compose was still set.

This commit fixes this issue by:
* Making sure the initializing function is actually called
* Making sure the initialization happens only once per rendering of the wizard

Fixes #1552

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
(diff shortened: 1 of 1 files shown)


lavocatt:
Duplicates, my bad!

PR #1569, opened 2024-01-18
V2Wizard: Add Repositories step

regexowl:
This PR adds Repository step to the V2Wizard. Files were migrated to TypeScript, new directory was created for utilities that are used across more steps and `.content` test suite was added and updates so only relevant tests are enabled.

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -13,6 +13,7 @@ import DetailsStep from './steps/Details';
 import ImageOutputStep from './steps/ImageOutput';
 import OscapStep from './steps/Oscap';
 import RegistrationStep from './steps/Registration';
+import RepositoriesStep from './steps/Repositories';
 import Aws from './steps/TargetEnvironment/Aws';
 import Gcp from './steps/TargetEnvironment/Gcp';
 import {
@@ -179,6 +180,20 @@ const CreateImageWizard = () => {
           >
             <OscapStep />
           </WizardStep>
+          <WizardStep
+            name="Content"
+            id="step-content"
+            steps={[
+              <WizardStep
+                name="Custom repositories"
+                id="wizard-custom-repositories"
+                key="wizard-custom-repositories"
+                footer={<CustomWizardFooter disableNext={false} />}
+              >
+                <RepositoriesStep />
+              </WizardStep>,
+            ]}
+          ></WizardStep>
           <WizardStep
             name="Details"
             id="step-details"
--- a/src/Components/CreateImageWizardV2/utilities/checkRepositoriesAvailability.ts
+++ b/src/Components/CreateImageWizardV2/utilities/checkRepositoriesAvailability.ts
@@ -1,24 +1,23 @@
 import { useMemo } from 'react';
 
-import { useFormApi } from '@data-driven-forms/react-form-renderer';
-
-import { releaseToVersion } from './releaseToVersion';
-
-import { useListRepositoriesQuery } from '../store/contentSourcesApi';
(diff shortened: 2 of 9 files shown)


regexowl:
I've tried to clean up the logic in `Repositories.tsx` a bit. There are a few things I'm not super happy with - those being not null assertions. I did my best to eliminate most of them alas some remained. If you have any suggestions on how to deal with them, please send them my way.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -179,6 +180,21 @@ const CreateImageWizard = () => {
           >
             <OscapStep />
           </WizardStep>
+          <WizardStep
+            name="Content"
+            id="step-content"
+            footer={<CustomWizardFooter disableNext={true} />}
```
lucasgarfield, line 186: Since this step will consist of substeps, does it need its own footer? I don't think that the parent step for the target environments has a footer, just the children.

src/Components/CreateImageWizardV2/utilities/checkRepositoriesAvailability.ts
```diff
@@ -0,0 +1,74 @@
...
+import { releaseToVersion } from './releaseToVersion';
+
+import { useListRepositoriesQuery } from '../store/contentSourcesApi';
+
+/**
+ * This checks the list of the payload repositories against a list of repos freshly
+ * fetched from content source API and returns true whether there are some
+ * repositories that are no longer available in the Repositories service.
+ *
+ * (The payload repositories are comming from the useFormApi hook).
```
lucasgarfield: This isn't true anymore since we dropped DDF, right?

regexowl: The part about the `useFormApi` hook? It was dropped in one of the other commits: https://github.com/RedHatInsights/image-builder-frontend/pull/1569/commits/ba35ddeeb9f651fd755c902e0ab0da8d5e12b7b5#diff-89a5b0134c0bbbe72c49676be29f95346ac6852a06ee652ee3fcaea64a58b7b6

I've tried to structure the PR in a way that the files are first copied over in one commit and then updated in another one to show the diff a bit better.

lucasgarfield: Oh super, yeah that totally makes sense, I'm just reviewing sequentially and adding comments immediately. Awesome!

src/store/wizardSlice.ts
```diff
@@ -45,6 +46,9 @@ type wizardState = {
   openScap: {
     profile: DistributionProfileItem | undefined;
   };
+  repositories: {
+    customRepositories: CustomRepository[] | undefined;
```
lucasgarfield, line 50: If we are going to initialize `customRepositories` as `[]` and use that as the empty state, do we need to allow the `undefined` type? Kind of similar to how our controlled inputs are just `''` for the empty state and only strings are allowed.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
regexowl: Oops, great catch. Removed :sweat_smile:

src/store/wizardSlice.ts
regexowl: Good point, I've removed the `undefined`

lucasgarfield (approved):
Wonderful work! I wrote most of this original code and you've done a great job cleaning it up and adapting it to the new wizard... I'd still like to do a bit more refactoring on my old messy code but I think this is perfect for now and will let us keep moving and start working on the packages step! Thank you!

PR #1570, opened 2024-01-18
V2Wizard:  add Details step to wizrd2

mgold1234:
this commit add Details step to v2Wizard

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -9,12 +9,18 @@ import {
 } from '@patternfly/react-core';
 import { useNavigate } from 'react-router-dom';
 
+import DetailsStep from './steps/Details';
 import ImageOutputStep from './steps/ImageOutput';
 import OscapStep from './steps/Oscap';
 import RegistrationStep from './steps/Registration';
 import Aws from './steps/TargetEnvironment/Aws';
 import Gcp from './steps/TargetEnvironment/Gcp';
-import { isAwsAccountIdValid, isGcpEmailValid } from './validators';
+import {
+  isAwsAccountIdValid,
+  isBlueprintDescriptionValid,
+  isBlueprintNameValid,
+  isGcpEmailValid,
+} from './validators';
 
 import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import './CreateImageWizard.scss';
@@ -24,6 +30,8 @@ import {
   selectAwsAccountId,
   selectAwsShareMethod,
   selectAwsSource,
+  selectBlueprintDescription,
+  selectBlueprintName,
   selectGcpEmail,
   selectGcpShareMethod,
   selectImageTypes,
@@ -83,6 +91,10 @@ const CreateImageWizard = () => {
 
   const registrationType = useAppSelector((state) =>
     selectRegistrationType(state)
+  );
+  const blueprintName = useAppSelector((state) => selectBlueprintName(state));
+  const blueprintDescription = useAppSelector((state) =>
+    selectBlueprintDescription(state)
   );
   const activationKey = useAppSelector((state) => selectActivationKey(state));
 
@@ -166,6 +178,20 @@ const CreateImageWizard = () => {
             footer={<CustomWizardFooter disableNext={false} />}
           >
             <OscapStep />
+          </WizardStep>
+          <WizardStep
+            name="Details"
+            id="step-details"
+            footer={
+              <CustomWizardFooter
+                disableNext={
+                  !isBlueprintNameValid(blueprintName) ||
+                  !isBlueprintDescriptionValid(blueprintDescription)
+                }
+              />
+            }
+          >
+            <DetailsStep />
           </WizardStep>
           <WizardStep
             name="Review"
--- a/src/Components/CreateImageWizardV2/steps/Details/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Details/index.tsx
@@ -1,0 +1,92 @@
+import React from 'react';
+
+import {
+  Form,
+  FormGroup,
+  FormHelperText,
+  HelperText,
+  HelperTextItem,
+  Text,
+  Title,
+} from '@patternfly/react-core';
+
+import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
+import {
+  changeBlueprintDescription,
+  changeBlueprintName,
+  selectBlueprintDescription,
+  selectBlueprintName,
+} from '../../../../store/wizardSlice';
+import { ValidatedTextInput } from '../../ValidatedTextInput';
+import {
+  isBlueprintDescriptionValid,
+  isBlueprintNameValid,
+} from '../../validators';
+
+const DetailsStep = () => {
+  const dispatch = useAppDispatch();
+  const blueprintName = useAppSelector((state) => selectBlueprintName(state));
+  const blueprintDescription = useAppSelector((state) =>
+    selectBlueprintDescription(state)
+  );
+  const handleNameChange = (
+    _event: React.FormEvent<HTMLInputElement>,
+    name: string
+  ) => {
+    dispatch(changeBlueprintName(name));
+  };
+
+  const handleDescriptionChange = (
+    _event: React.FormEvent<HTMLInputElement>,
+    description: string
+  ) => {
+    dispatch(changeBlueprintDescription(description));
+  };
+
+  return (
+    <Form>
+      <Title headingLevel="h2">Details</Title>
+      <Text>
+        Optionally enter a name to identify your image later quickly. If you do
+        not provide one, the UUID will be used as the name.
+      </Text>
+      <FormGroup label="Blueprint name" fieldId="blueprint-name">
+        <ValidatedTextInput
+          ariaLabel="blueprint name"
+          dataTestId="blueprint"
+          value={blueprintName}
+          validator={isBlueprintNameValid}
+          onChange={handleNameChange}
+          helperText="Please enter a valid name"
+          placeholder="Blueprint name"
+        />
+        <FormHelperText>
+          <HelperText>
+            <HelperTextItem>
+              The image name can be 3-63 characters long. It can contain
+              lowercase letters, digits and hyphens, has to start with a letter
+              and cannot end with a hyphen.
+            </HelperTextItem>
+          </HelperText>
+        </FormHelperText>
+      </FormGroup>
+
+      <FormGroup
+        label="Blueprint description"
+        fieldId="blueprint-description-name"
+      >
+        <ValidatedTextInput
+          ariaLabel="blueprint description"
+          dataTestId="blueprint description"
+          value={blueprintDescription || ''}
+          validator={isBlueprintDescriptionValid}
+          onChange={handleDescriptionChange}
+          helperText="Please enter a valid description"
+          placeholder="Add description"
+        />
+      </FormGroup>
+    </Form>
+  );
+};
+
+export default DetailsStep;
--- a/src/Components/CreateImageWizardV2/validators.ts
+++ b/src/Components/CreateImageWizardV2/validators.ts
@@ -13,3 +13,17 @@ export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
     gcpShareWithAccount.length <= 253
   );
 };
+
+export const isBlueprintNameValid = (blueprintName: string) => {
+  if (blueprintName === '') {
+    return true;
+  }
+  return (
+    /^[a-z][a-z0-9-]+[a-z0-9]$/.test(blueprintName) &&
+    blueprintName.length <= 63
+  );
+};
+
+export const isBlueprintDescriptionValid = (blueprintDescription: string) => {
+  return blueprintDescription.length <= 250;
+};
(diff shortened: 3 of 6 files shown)


src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -159,6 +160,9 @@ const CreateImageWizard = () => {
           >
             <RegistrationStep />
           </WizardStep>
+            <WizardStep name="Details" id="wizard-details">
+                <DetailsStep />
+            </WizardStep>
```
regexowl, line 165: The `CustomWizardFooter` needs to be added even when we don't need to disable the Next button, because it switches the order of the buttons.

Also suggesting changing the id to "step-details" so it's consistent with the other steps ;) (AWS and GCP are a substep, that's why the name is different)

```suggestion
          <WizardStep
            name="Details"
            id="step-details"
            footer={<CustomWizardFooter disableNext={false} />}
          >
            <DetailsStep />
          </WizardStep>
```

src/Components/CreateImageWizardV2/steps/Details/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    name: string
+  ) => {
+    setName(name);
+  };
+  const handleDescriptionChange = (
+    _event: React.FormEvent<HTMLInputElement>,
+    name: string
+  ) => {
+    setDescription(name);
+  };
```
regexowl, line 34: Here the name and the description get stored in the variables that will not persist when continuing to the Review step. What will need to be done is adding both to the wizardSlice and storing them in the store. 

That would mean adding a structure for the step, maybe something like:
```typescript
  details: {
    blueprintName: ComposeRequest["image_name"] | undefined;
    blueprintDescription: ComposeRequest["image_description"] | undefined;
  };
```

and `changeBlueprintName` and `changeBlueprintDescription` functions will be needed. :)

mgold1234: thank you @regexowl

src/Components/CreateImageWizardV2/steps/Details/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <Form>
+      <Title headingLevel="h2">Details</Title>
+      <Text>
+        Optionally enter a name to identify your image later quickly. If you do
+        not provide one, the UUID will be used as the name.
+      </Text>
+      <FormGroup label="Blueprint Name" fieldId="blueprint-name">
```
regexowl, line 42: Let's use sentence case for the FormGroup labels :)

```suggestion
      <FormGroup label="Blueprint name" fieldId="blueprint-name">
```

src/Components/CreateImageWizardV2/steps/Details/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <ValidatedTextInput
+          ariaLabel="blueprint name"
+          dataTestId="blueprint"
+          value={name || ''}
+          validator={isBlueprintNameValid}
+          onChange={handleNameChange}
+          helperText="Please enter a valid name"
+          placeholder="Blueprint Name"
```
regexowl, line 50:
```suggestion
          placeholder="Blueprint name"
```

src/Components/CreateImageWizardV2/steps/Details/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+              The image name can be 3-63 characters long. It can contain
+              lowercase letters, digits and hyphens, has to start with a letter
+              and cannot end with a hyphen.
+            </HelperTextItem>
+          </HelperText>
+        </FormHelperText>
+      </FormGroup>
+
+      <FormGroup
+        label="Blueprint Description"
```
regexowl, line 64:
```suggestion
        label="Blueprint description"
```

src/Components/CreateImageWizardV2/steps/Details/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        fieldId="blueprint-description-name"
+      >
+        <ValidatedTextInput
+          ariaLabel="blueprint description"
+          dataTestId="blueprint description"
+          value={description || ''}
+          validator={isBlueprintDescriptionValid}
+          onChange={handleDescriptionChange}
+          helperText="Please enter a valid description"
+          placeholder="Add Description"
```
regexowl, line 74:
```suggestion
          placeholder="Add description"
```

lavocatt (changes requested):
Thanks for the work you've put so far. I've got two bugs to report to you, here's a video to display what's going on:

[Screencast from 2024-01-19 09-40-50.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/ecfbb2b2-3728-4105-a578-da5ab891a1b8)

The two issues are:
* you can click next even though the field isn't valid
* when you go back and forth to the next step or the previous step, the data disappears.

Happy hacking :rocket:

lavocatt (approved):
:rocket:

src/store/wizardSlice.ts
```diff
@@ -37,6 +42,10 @@ type wizardState = {
     registrationType: string;
     activationKey: ActivationKeys['name'];
   };
+  details: {
+    blueprintName: ComposeRequest['image_name'] | '';
```
lucasgarfield, line 46: We can remove the  `| ''`, the type of `ComposeRequest['image_name']` is already `string | undefined` so `''` is already taken care of.

src/store/wizardSlice.ts
```diff
@@ -37,6 +42,10 @@ type wizardState = {
...
+    blueprintDescription: ComposeRequest['image_description'] | '';
```
lucasgarfield, line 47: Same comment as above, we can remove the  `| ''`, the type of `ComposeRequest['image_name']` is already `string | undefined` so `''` is already taken care of.

src/store/wizardSlice.ts
lucasgarfield: Actually since `ComposeRequest['image_name']` can be undefined, let's make the type just be `string` for now. So this should be: 
```
blueprintName: string;
blueprintDescription: string;
```

src/store/wizardSlice.ts
```diff
@@ -190,6 +211,18 @@ export const wizardSlice = createSlice({
     ) => {
       state.registration.activationKey = action.payload;
     },
+    changeBlueprintName: (
+      state,
+      action: PayloadAction<ComposeRequest['image_name']>
```
lucasgarfield, line 216: Let's make this `PayloadAction<string>` for now.

src/store/wizardSlice.ts
```diff
@@ -190,6 +211,18 @@ export const wizardSlice = createSlice({
...
+    ) => {
+      state.details.blueprintName = action.payload;
+    },
+    changeBlueprintDescription: (
+      state,
+      action: PayloadAction<ComposeRequest['image_description']>
```
lucasgarfield, line 222: Let's make this `PayloadAction<string>`.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -13,3 +13,22 @@ export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
...
+    /^[a-z][a-z0-9-]+[a-z0-9]$/.test(blueprintName) &&
+    blueprintName.length <= 63
+  );
+};
+
+export const isBlueprintDescriptionValid = (
+  blueprintDescription: string | undefined
+) => {
+  return (
+    blueprintDescription !== undefined && blueprintDescription.length <= 250
```
lucasgarfield, line 32: We won't need to check for undefined here either if blueprintDescription is of type `string`.

src/Components/CreateImageWizardV2/steps/Details/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <Title headingLevel="h2">Details</Title>
+      <Text>
+        Optionally enter a name to identify your image later quickly. If you do
+        not provide one, the UUID will be used as the name.
+      </Text>
+      <FormGroup label="Blueprint name" fieldId="blueprint-name">
+        <ValidatedTextInput
+          ariaLabel="blueprint name"
+          dataTestId="blueprint"
+          value={blueprintName || ''}
```
lucasgarfield, line 57: We won't need the short circuit here if blueprintName has to be of type `string`, so this can just be `value={blueprintName}`.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -159,6 +171,22 @@ const CreateImageWizard = () => {
           >
             <RegistrationStep />
           </WizardStep>
+          <WizardStep
+            name="Details"
+            id="step-details"
+            footer={
+              <CustomWizardFooter
+                disableNext={
+                  (blueprintName !== undefined &&
```
lucasgarfield, line 180: We won't need to check for undefined.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -159,6 +171,22 @@ const CreateImageWizard = () => {
...
+                    !isBlueprintNameValid(blueprintName)) ||
+                  (blueprintDescription !== undefined &&
```
lucasgarfield, line 182: Won't need to check for undefined.

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
```diff
@@ -833,62 +833,80 @@ describe('Step Upload to AWS', () => {
...
-  //     await user.type(
-  //       await screen.findByTestId('aws-account-id'),
-  //       '012345678901'
-  //     );
-  //     await clickNext();
-  //     // skip registration
-  //     await screen.findByRole('textbox', {
-  //       name: 'Select activation key',
-  //     });
+      //     // aws step
```
lucasgarfield, line 851: Let's clean this up just a bit and remove the duplicate `//`.

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
```diff
@@ -833,62 +833,80 @@ describe('Step Upload to AWS', () => {
...
+          name: 'aws account id',
+        }),
+        '012345678901'
+      );
 
-  //     const registerLaterRadio = await screen.findByTestId('registration-radio-later');
-  //     await user.click(registerLaterRadio);
-  //     await clickNext();
+      await clickNext();
+      //     // skip registration
```
lucasgarfield, line 861: Remove duplicate `//`.

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
```diff
@@ -833,62 +833,80 @@ describe('Step Upload to AWS', () => {
...
+      //     await user.type(nameInput, invalidName);
+      //     expect(await getNextButton()).toHaveClass('pf-m-disabled');
+      //     expect(await getNextButton()).toBeDisabled();
+      //     await user.clear(nameInput);
+      //     // Enter image name
+
+      const nameInput = screen.getByRole('textbox', {
+        name: /blueprint name/i,
+      });
+      //     // 64 character name
```
lucasgarfield, line 896: Let's just delete this comment, it is pointless (you can see that it is 'a'.repeat(64) right on the next line).

lucasgarfield:
This looks so great Michal! :heart:

I just want to change the way we handle the typing for now.

I really like your idea of being super specific in the API typing. So this:
```
  details: {
    blueprintName: ComposeRequest['image_name'];
    blueprintDescription: ComposeRequest['image_description'];
  };
```
is really clever. But, for right now, I think it might cause some problems since the type of `ComposeRequest['image_name']` is `string | undefined`. Since it can be undefined, we're going to have to be very careful to avoid allowing an input 

I made a thread about this in our frontend channel, so we can discuss and get consensus tomorrow. But for now, I would propose that we just keep the form state typing simple when it comes to controlled inputs and do this:
```
  details: {
    blueprintName: string;
    blueprintDescription: string;
  };
```

Basically all my comments are just where we would need to do some clean-up to make this happen.

My reasoning is this:
If we are managing the state of controlled inputs using Redux, and we know that the type of the value displayed in those inputs _has_ to be string and that we _have_ to use `''` for the empty state... we should just enforce this using typing, rather than having to handle the `undefined` edge case everywhere. Hope that makes sense.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -13,3 +13,22 @@ export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
     gcpShareWithAccount.length <= 253
   );
 };
+
+export const isBlueprintNameValid = (blueprintName: string | undefined) => {
+  if (blueprintName === '') {
+    return true;
+  }
+  return (
+    blueprintName !== undefined &&
```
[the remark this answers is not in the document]
mgold1234: if I remove the blueprintName !== undefined I have an error:
TS18048: 'blueprintName' is possibly 'undefined'.

src/Components/CreateImageWizardV2/validators.ts
mgold1234: if I remove the blueprintDescription !== undefined I have an error:
TS18048: 'blueprintDescription' is possibly 'undefined'.

src/Components/CreateImageWizardV2/validators.ts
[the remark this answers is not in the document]
lucasgarfield: Take a look at the way the function argument is typed :wink:

lucasgarfield (approved):
Thank you!

PR #1571, opened 2024-01-18
feat(HMS-3386): add blueprints initial layout

amirfefer:
This PR adds an initial layout for the blueprints feature.
It uses the `/experimental` endpoint for fetching real blueprints from stage.

https://github.com/RedHatInsights/image-builder-frontend/assets/11807069/19690de5-81a3-4fb8-9002-6b015c6bf5f5

### DIFF
--- a/src/Components/Blueprints/BlueprintCard.tsx
+++ b/src/Components/Blueprints/BlueprintCard.tsx
@@ -1,0 +1,97 @@
+import React, { useState } from 'react';
+
+import {
+  Badge,
+  Card,
+  CardHeader,
+  CardTitle,
+  CardBody,
+  CardFooter,
+  Dropdown,
+  DropdownList,
+  MenuToggle,
+  MenuToggleElement,
+  DropdownItem,
+} from '@patternfly/react-core';
+import { EllipsisVIcon } from '@patternfly/react-icons';
+
+import { BlueprintItem } from '../../store/imageBuilderApi';
+
+type blueprintProps = {
+  blueprint: BlueprintItem;
+  selectedBlueprint: string;
+  setSelectedBlueprint: React.Dispatch<React.SetStateAction<string>>;
+};
+
+const BlueprintCard = ({
+  blueprint,
+  selectedBlueprint,
+  setSelectedBlueprint,
+}: blueprintProps) => {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const isChecked = blueprint.id === selectedBlueprint;
+  const onSelect = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const onClickHandler = () => {
+    setSelectedBlueprint(blueprint.id);
+  };
+
+  const headerActions = (
+    <>
+      <Dropdown
+        ouiaId={`blueprint-card-${blueprint.id}-dropdown`}
+        onSelect={onSelect}
+        toggle={(toggleRef: React.Ref<MenuToggleElement>) => (
+          <MenuToggle
+            ref={toggleRef}
+            isExpanded={isOpen}
+            onClick={() => setIsOpen(!isOpen)}
+            variant="plain"
+            aria-label="blueprint menu toggle"
+          >
+            <EllipsisVIcon aria-hidden="true" />
+          </MenuToggle>
+        )}
+        isOpen={isOpen}
+        onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
+      >
+        <DropdownList>
+          <DropdownItem>Edit details</DropdownItem>
+          <DropdownItem>Delete blueprint</DropdownItem>
+        </DropdownList>
+      </Dropdown>
+    </>
+  );
+
+  return (
+    <Card
+      ouiaId={`blueprint-card-${blueprint.id}`}
+      isCompact
+      isClickable
+      isSelectable
+      isSelected={isChecked}
+    >
+      <CardHeader
+        selectableActions={{
+          selectableActionId: blueprint.id,
+          selectableActionAriaLabelledby: 'blueprint radio select',
+          name: blueprint.name,
+          variant: 'single',
+          isChecked: isChecked,
+          onChange: onClickHandler,
+        }}
+        actions={{ actions: headerActions }}
+      >
+        <CardTitle>{blueprint.name}</CardTitle>
+      </CardHeader>
+      <CardBody>{blueprint.description}</CardBody>
+      <CardFooter>
+        Version <Badge isRead>{blueprint.version}</Badge>
+      </CardFooter>
+    </Card>
+  );
+};
+
+export default BlueprintCard;
--- a/src/Components/Blueprints/BlueprintsSideBar.tsx
+++ b/src/Components/Blueprints/BlueprintsSideBar.tsx
@@ -1,0 +1,83 @@
+import React, { useState, Dispatch, SetStateAction } from 'react';
+
+import {
+  Button,
+  EmptyState,
+  EmptyStateActions,
+  EmptyStateBody,
+  EmptyStateFooter,
+  EmptyStateHeader,
+  EmptyStateIcon,
+  SearchInput,
+  Stack,
+  StackItem,
+} from '@patternfly/react-core';
+import { PlusCircleIcon } from '@patternfly/react-icons';
+
+import BlueprintCard from './BlueprintCard';
+
+import { BlueprintItem } from '../../store/imageBuilderApi';
+
+type blueprintProps = {
+  blueprints: BlueprintItem[] | undefined;
+  selectedBlueprint: string;
+  setSelectedBlueprint: Dispatch<SetStateAction<string>>;
+};
+
+const BlueprintsSidebar = ({
+  blueprints,
+  selectedBlueprint,
+  setSelectedBlueprint,
+}: blueprintProps) => {
+  const [blueprintFilter, setBlueprintFilter] = useState('');
+
+  const onChange = (value: string) => {
+    setBlueprintFilter(value);
+  };
+
+  const emptyBlueprints = (
+    <EmptyState variant="sm">
+      <EmptyStateHeader
+        titleText="No blueprints yet"
+        headingLevel="h4"
+        icon={<EmptyStateIcon icon={PlusCircleIcon} />}
+      />
+      <EmptyStateBody>To get started, create a blueprint.</EmptyStateBody>
+      <EmptyStateFooter>
+        <EmptyStateActions>
+          <Button>Create</Button>
+        </EmptyStateActions>
+      </EmptyStateFooter>
+    </EmptyState>
+  );
+
+  if (blueprints === undefined || blueprints?.length === 0) {
+    return emptyBlueprints;
+  }
+
+  return (
+    <>
+      <Stack hasGutter>
+        <StackItem>
+          <SearchInput
+            placeholder="Search by name or description"
+            value={blueprintFilter}
+            onChange={(_event, value) => onChange(value)}
+            onClear={() => onChange('')}
+          />
+        </StackItem>
+        {blueprints.map((blueprint: BlueprintItem) => (
+          <StackItem key={blueprint.id}>
+            <BlueprintCard
+              blueprint={blueprint}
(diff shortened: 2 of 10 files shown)


src/Components/sharedComponents/ImageBuilderHeader.tsx
```diff
@@ -16,62 +23,72 @@ export const ImageBuilderHeader = () => {
...
+                aria-label="About image builder"
+                className="pf-u-pl-sm header-button"
+              >
+                <HelpIcon />
+              </Button>
+            </Popover>
+            <OpenSourceBadge repositoriesURL="https://www.osbuild.org/guides/image-builder-service/architecture.html" />
+          </FlexItem>
+          <FlexItem align={{ default: 'alignRight' }}>
+            <Button>New blueprint</Button>
```
amirfefer, line 86: According to the mockups it should be `Create`, but it affect other tests for now. will be change [here](https://issues.redhat.com/browse/HMS-3387)

src/Components/Blueprints/BlueprintCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  DropdownList,
+  MenuToggle,
+  MenuToggleElement,
+  DropdownItem,
+} from '@patternfly/react-core';
+import { EllipsisVIcon } from '@patternfly/react-icons';
+
+import { BlueprintItem } from '../../store/imageBuilderApi';
+
+const BlueprintCard = (props: {
```
lucasgarfield, line 20: The codebase's convention is to just destructure the props so we can use them directly in the component. E.g. we would simply use `blueprint` instead of `props.blueprint`.

It is also our convention to define the prop types in a separate `Type` definition right above the component. So in this case, we should create a `BluePrintCardTypes` type definition above.

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+} from '@patternfly/react-core';
+import { PlusCircleIcon } from '@patternfly/react-icons';
+import { Link } from 'react-router-dom';
+
+import BlueprintCard from './BlueprintCard';
+
+import { BlueprintItem } from '../../store/imageBuilderApi';
+import { resolveRelPath } from '../../Utilities/path';
+
+interface blueprintProps {
```
lucasgarfield, line 22: Convention is to use a `Type` and not `interface`.

lucasgarfield: This should also match the component name, so `type BlueprintsSidebarProps = { ... }`.

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <EmptyState variant="sm">
+      <EmptyStateHeader
+        titleText="No blueprints yet"
+        headingLevel="h4"
+        icon={<EmptyStateIcon icon={PlusCircleIcon} />}
+      />
+      <EmptyStateBody>To get started, create a blueprint.</EmptyStateBody>
+      <EmptyStateFooter>
+        <EmptyStateActions>
+          <Link
```
lucasgarfield, line 44: Let's use a `<Button>` here instead of `<Link>` and specify the `variant` prop as `"link"`. Then we don't need to write any CSS.

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+            data-testid="create-blueprint-action"
+          >
+            Create
+          </Link>
+        </EmptyStateActions>
+      </EmptyStateFooter>
+    </EmptyState>
+  );
+
+  return props.blueprints?.length ? (
```
lucasgarfield, line 56: To be more consistent with the rest of the code base, let's use short circuits here for conditional rendering, e.g.
```typescript
return (
  {blueprints?.length > 0 && (<> ... </>)}
  {blueprints === undefined || blueprints.length === 0 && (<EmptyState>)
)
```

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -0,0 +1,82 @@
...
+    <>
+      <SearchInput
+        placeholder="Search by name or description"
+        value={blueprintFilter}
+        onChange={(_event, value) => onChange(value)}
+        onClear={() => onChange('')}
+        className="pf-v5-u-mb-md"
```
lucasgarfield, line 63: What do we need this for? I would prefer to not need to specify any CSS classes and rely entirely on Patternfly's props.

src/Components/LandingPage/LandingPage.tsx
```diff
@@ -45,15 +53,62 @@ export const LandingPage = () => {
...
+  const experimentalImageList = (
+    <>
+      <PageSection>
+        <Quickstarts />
+      </PageSection>
+      <PageSection>
+        <Sidebar
+          hasBorder
+          hasGutter
+          className="pf-v5-u-background-color-100 pf-v5-u-p-lg"
```
lucasgarfield, line 83: What does this CSS do? Can we eliminate it?

lucasgarfield:
Nice work so far Amir, thank you. I'm going to give this a closer look tomorrow and may have some additional feedback before approving, but so far my only comments are small nitpicks just related to keeping the style consistent with the rest of the code base.

As you can see, I would prefer to avoid CSS and using the `className` prop to the greatest extent possible. :sweat_smile:

src/Components/Blueprints/BlueprintsSideBar.tsx
lucasgarfield: And fwiw both are acceptable and shown in the official React docs, we just tend to use the short circuit method:
https://react.dev/learn/conditional-rendering#logical-and-operator-

src/Components/sharedComponents/ImageBuilderHeader.tsx
lucasgarfield: That makes sense. We're taking this incrementally so it is fine if it isn't 1:1 for this first PR.

src/Components/Blueprints/BlueprintsSideBar.tsx
amirfefer: leftovers from previous PR

src/Components/sharedComponents/ImageBuilderHeader.tsx
```diff
@@ -16,62 +23,72 @@ export const ImageBuilderHeader = () => {
...
-            variant="plain"
-            aria-label="About image builder"
-            className="pf-u-pl-sm header-button"
-          >
-            <HelpIcon />
-          </Button>
-        </Popover>
-        <OpenSourceBadge repositoriesURL="https://www.osbuild.org/guides/image-builder-service/architecture.html" />
+        <Flex>
+          <FlexItem>
```
lucasgarfield, line 27: I'm not necessarily opposed to moving this all inside `<Flex><FlexItem>...</Flex></FlexItem>` but what is the benefit?

lucasgarfield: Oh I see, because the buttons have moved into the header. Makes sense. Nevermind.

src/test/fixtures/blueprints.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  GetBlueprintsApiResponse,
+  CreateBlueprintResponse,
+} from '../../store/imageBuilderApi';
+
+export const mockBlueprintsCreation: CreateBlueprintResponse[] = [
+  {
+    id: '677b010b-e95e-4694-9813-d11d847f1bfc',
```
lucasgarfield, line 8: Could we use (visibly) unique UUIDs for all of these? In this case I think we *haven't* done the best job in our code base so shame on us, but going forward we should probably use something like https://www.uuidgenerator.net/version4 when we need a new UUID.

src/test/Components/Blueprints/Blueprints.test.js
```diff
@@ -0,0 +1,46 @@
...
+  useUnleashContext: () => jest.fn(),
+  useFlag: jest.fn((flag) =>
+    flag === 'image-builder.new-wizard.enabled' ? true : false
+  ),
+}));
+
+describe('Blueprints', () => {
+  test('renders blueprints page', async () => {
+    renderWithReduxRouter('', {});
+    const blueprintName = mockGetBlueprints.data[0].name;
```
lucasgarfield, line 30: I've been thinking a lot about how to handle this in the tests in general recently, and I think we should de-couple the tests from the data that we use to generate the mock API responses.

So here we expect the name to be 'Dark Chocolate'. Let's just do `await screen.findByText('Dark Chocolate');`.

I know that means some additional overhead in the code, but it means we test exactly what we expect to get. It also means that if you look at the tests while `npm run stage-beta:msw` is running as your local dev environment, you'll be able to quickly make a 1:1 correlation between what's in your browser and what is in the test code.

lucasgarfield: @regexowl thoughts?

src/Components/Blueprints/BlueprintsSideBar.tsx
lucasgarfield: I see. Can we remove it?

lucasgarfield: @jkozol can you advise?

src/Components/Blueprints/BlueprintsSideBar.tsx
amirfefer: this example with the && operator in the same return segment,  ends up with unreachable code.

src/Components/Blueprints/BlueprintsSideBar.tsx
amirfefer: removed

src/Components/LandingPage/LandingPage.tsx
amirfefer: I used the `width` prop instead of `pf-v5-u-p-lg` but for the background color it makes sense to use a class.

src/test/Components/Blueprints/Blueprints.test.js
amirfefer: doesn't it make sense to use one source of truth for fixtures?  both for testing and for mocking API?

lucasgarfield: Try thinking about it like this: if we use `await screen.findByText('Dark Chocolate')`, we are effectively testing _both_ the React components *and* the mocking API.

If we use `await screen.findByText(mockGetBlueprints.data[0].name)` and `mockGetBlueprints.data[0].name` is not what we expect it to be for whatever reason, and the page is rendered incorrectly, our tests would still pass.

Is that a reasonable way of looking at things?

src/Components/Blueprints/BlueprintsSideBar.tsx
lucasgarfield: Cool, yeah that was just a quick off the cuff example not supposed to be final code :sweat_smile: thanks for fixing it up.

src/test/Components/Blueprints/Blueprints.test.js
amirfefer: The blueprint's name in the fixture is a constant, in the test we verify that it propagated well from the API to the component, so to me it looks a bit redundant declaring the same name again. but, if it works for you let it be :)

lucasgarfield: It is redundant. Try to remember I come from a traditional engineering background and we like to have lots of redundancy in our safety systems. ;) 

So let's go ahead and use `await screen.findByText('Dark Chocolate')`.

amirfefer: done

src/test/fixtures/blueprints.ts
amirfefer: I think it a different scope, most of the fixtures use hard coded uuid, so this is a refactor for the codebase.

lucasgarfield (approved):
CI is being flaky but I'm ready to merge this. Feel free to start working on the next PR @amirfefer.

PR #1574, opened 2024-01-21
feat(HMS-3391): add blueprint onclick handler

amirfefer:
This PR adds an handler for selecting a blueprint and triggering its corresponding composes

https://github.com/RedHatInsights/image-builder-frontend/assets/11807069/69ebb332-f97c-4be7-88d2-e905637b28c1

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -54,6 +54,7 @@ import {
 import {
   ComposesResponseItem,
   ComposeStatus,
+  useGetBlueprintComposesQuery,
   useGetComposesQuery,
   useGetComposeStatusQuery,
 } from '../../store/imageBuilderApi';
@@ -63,7 +64,11 @@ import {
   timestampToDisplayString,
 } from '../../Utilities/time';
 
-const ImagesTable = () => {
+type ImageTableProps = {
+  selectedBlueprint?: string | undefined;
+};
+
+const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
   const [page, setPage] = useState(1);
   const [perPage, setPerPage] = useState(10);
 
@@ -74,16 +79,53 @@ const ImagesTable = () => {
     setPerPage(perPage);
   };
 
-  const { data, isError, isSuccess } = useGetComposesQuery({
-    limit: perPage,
-    offset: perPage * (page - 1),
-    ignoreImageTypes: [
-      'rhel-edge-commit',
-      'rhel-edge-installer',
-      'edge-commit',
-      'edge-installer',
-    ],
-  });
+  const {
+    data: blueprintsComposes,
+    isSuccess: isBlueprintsSuccess,
+    isFetching: isFetchingBlueprintsCompose,
+    isError: isBlueprintsError,
+  } = useGetBlueprintComposesQuery(
+    {
+      id: selectedBlueprint as string,
+      limit: perPage,
+      offset: perPage * (page - 1),
+    },
+    { skip: !selectedBlueprint }
+  );
+
+  const {
+    data: composesData,
+    isSuccess: isComposesSuccess,
+    isError: isComposesError,
+    isFetching: isFetchingComposes,
+  } = useGetComposesQuery(
+    {
+      limit: perPage,
+      offset: perPage * (page - 1),
+      ignoreImageTypes: [
+        'rhel-edge-commit',
+        'rhel-edge-installer',
+        'edge-commit',
+        'edge-installer',
+      ],
+    },
+    { skip: !!selectedBlueprint }
+  );
+
+  const data = selectedBlueprint ? blueprintsComposes : composesData;
+  const isSuccess = selectedBlueprint ? isBlueprintsSuccess : isComposesSuccess;
+  const isError = selectedBlueprint ? isBlueprintsError : isComposesError;
+  const isFetching = selectedBlueprint
+    ? isFetchingBlueprintsCompose
+    : isFetchingComposes;
+
+  if (isFetching) {
+    return (
+      <Bullseye>
+        <Spinner />
+      </Bullseye>
+    );
+  }
 
   if (!isSuccess) {
     if (isError) {
@@ -103,13 +145,13 @@ const ImagesTable = () => {
     );
   }
 
-  const composes = data.data;
-  const itemCount = data.meta.count;
+  const composes = data?.data;
+  const itemCount = data?.meta.count || 0;
 
   return (
     <>
-      {data.meta.count === 0 && <EmptyImagesTable />}
-      {data.meta.count > 0 && (
+      {itemCount === 0 && <EmptyImagesTable />}
+      {itemCount > 0 && (
         <>
           <Toolbar>
             <ToolbarContent>
@@ -152,7 +194,7 @@ const ImagesTable = () => {
                 <Th />
               </Tr>
             </Thead>
-            {composes.map((compose, rowIndex) => {
+            {composes?.map((compose, rowIndex) => {
               return (
                 <ImagesTableRow
                   compose={compose}
--- a/src/Components/LandingPage/LandingPage.tsx
+++ b/src/Components/LandingPage/LandingPage.tsx
@@ -53,7 +53,9 @@ export const LandingPage = () => {
     }
     setActiveTabKey(tabIndex);
   };
-  const [selectedBlueprint, setSelectedBlueprint] = useState<string>('');
+  const [selectedBlueprint, setSelectedBlueprint] = useState<
+    string | undefined
+  >();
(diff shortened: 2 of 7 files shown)


src/Components/LandingPage/LandingPage.tsx
```diff
@@ -53,7 +53,7 @@ export const LandingPage = () => {
     }
     setActiveTabKey(tabIndex);
   };
-  const [selectedBlueprint, setSelectedBlueprint] = useState<string>('');
+  const [selectedBlueprint, setSelectedBlueprint] = useState<string>();
```
lucasgarfield, line 56: Would it make sense to initialize to `undefined`? I assume at some point we will need to be able to clear the filter, and I imagine we would reset to `undefined` in that case.

lucasgarfield:
Would it make sense to add a chip to clear the filter in this PR, or do you have that planned for a future PR?

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -71,28 +78,56 @@ const ImagesTable = () => {
...
-      'edge-installer',
-    ],
-  });
+  const {
+    data: blueprintsComposes,
+    isSuccess: isBlueprintsSuccess,
+    isFetching: isFetchingBlueprintsCompose,
+  } = useGetBlueprintComposesQuery(
+    {
+      id: selectedBlueprint || '',
```
lucasgarfield, line 87: I assume we need `''` here because `selectedBlueprint` could be undefined and the linter is unhappy about that, because it doesn't realize that the `skip` will prevent it from happening. I've used similar patterns before but I'm not quite happy with them. What if we tried using a lazy query instead?

https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#uselazyquery

We could call the lazy query's trigger function only if a selectedBlueprint was present.

Thoughts?

src/test/fixtures/blueprints.ts
```diff
@@ -35,3 +37,49 @@ export const emptyGetBlueprints: GetBlueprintsApiResponse = {
   meta: { count: 0 },
   data: [],
 };
+
+export const mockBlueprintComposes: GetBlueprintComposesApiResponse = {
```
lucasgarfield, line 41: Let's turn this into a function that returns a different set of images depending on the id of the blueprint.

For 'Dark Chocolate' we can return these two.

For 'Milk Chocolate' we can return a different response, with 0 images. That way we can easily test the state where a blueprint is selected but that blueprint has no images associated with it.

src/test/Components/Blueprints/Blueprints.test.js
```diff
@@ -39,4 +43,17 @@ describe('Blueprints', () => {
     renderWithReduxRouter('', {});
     await screen.findByText('No blueprints yet');
   });
+  test('renders blueprint composes', async () => {
```
lucasgarfield, line 46: Once we've updated the fixture so that 'Dark Chocolate' has two images associated with it and 'Milk Chocolate' has none, we can also update this test to reflect that. It should check for both images associated with 'Dark Chocolate', and then either in this test or another test check that the empty images state is displayed for 'Milk Chocolate'.

src/Components/LandingPage/LandingPage.tsx
amirfefer: it's already initialized to undefined. By default,  empty `useState` values are initiziled to undefined

src/test/mocks/handlers.js
```diff
@@ -111,4 +114,10 @@ export const handlers = [
   rest.get(`${IMAGE_BUILDER_API}/experimental/blueprints`, (req, res, ctx) => {
     return res(ctx.status(200), ctx.json(mockGetBlueprints));
   }),
+  rest.get(
+    `${IMAGE_BUILDER_API}/experimental/blueprint/:id/composes`,
+    (req, res, ctx) => {
+      return res(ctx.status(200), ctx.json(mockBlueprintComposes));
```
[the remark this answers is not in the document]
amirfefer: I'd prefer descriptive named objects as fixtures, so it also would be easier to follow and read,  we can leverage MSW for changing the handler behavior, so  the fixtures would  represent just  the data layer.
something like

```js
rest.get('/experimental/blueprints/:id/composes', (req, res) => {
  const { id } = req.params
  const composes = id === BlueprintA.id ? mockBlueprintAComposes : mockBlueprintBComposes
  res(ctx.status(200), ctx.json(composes));
})
```

src/Components/ImagesTable/ImagesTable.tsx
amirfefer: Agree, I'm not happy with it as well.
The `useLazyQuery` looks interesting, but  it creates a trigger function, so the query won't start once the component renders, so we can trigger it inside an event handler . In ours case, we want to fetch the data once the table renders, so I'd pick `skip` here.  because we use `selectedBlueprint` in the skip, we can be certain that it string, Se can do type assertion:

```js
useGetBlueprintComposesQuery(
    {

      id: selectedBlueprint as string, // thanks to skip - `selectedBlueprint` must be string 
      limit: perPage,
      offset: perPage * (page - 1),
    },
    { skip: !selectedBlueprint }
  );
```

amirfefer:
Thanks @lucasgarfield !

> Would it make sense to add a chip to clear the filter in this PR, or do you have that planned for a future PR?

Totally makes sense to add a reset, to show all the images, but it's missing from the mockups, I think it would be better to ask what is the best practice here to uncheck the selection. My naive choice would be clicking again on a chosen radio button for unselecting it.

src/Components/ImagesTable/ImagesTable.tsx
lucasgarfield: I think that's a lot cleaner and definitely prefer it. Let's go with that for now, I'll think about the implications of the trigger function a bit more on my own.

src/test/mocks/handlers.js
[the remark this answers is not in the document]
lucasgarfield: That's fine. We use the function pattern in some of our other handlers but I'm willing to give your approach of using descriptive named objects a try... we might end up changing our other handlers to match if we like it. :+1:

src/test/fixtures/blueprints.ts
lucasgarfield: Makes sense.

I suppose we will just render an empty table for "Milk Chocolate" for now. We could just verify that the table elements have disappeared in the test, for now.

lucasgarfield:
> Thanks @lucasgarfield !
> 
> > Would it make sense to add a chip to clear the filter in this PR, or do you have that planned for a future PR?
> 
> Totally makes sense to add a reset, to show all the images, but it's missing from the mockups, I think it would be better to ask UX what is the best practice here to uncheck the selection. My naive choice would be clicking again on a chosen radio button for unselecting it.

In my own naive opinion, I don't expect a radio to be something you can 'unselect'. I would use a checkbox for that, which is actually what is in the mocks that I've seen.

In any case, let's discuss it at the next UX/dev sync. 

I'm fine with proceeding without it for now.

src/Components/LandingPage/LandingPage.tsx
lucasgarfield: I'm kind of surprised we don't need to type the useState using `<string | undefined>`... does it infer that?

amirfefer:
According to the  [mockups](https://www.sketch.com/s/669c80df-098c-4296-b689-b4e0702d8261/a/j4eWKxb) it looks like radio buttons 
Screenshot 2024-01-23 at 0 30 41

But, it seems that checkbox would fit more

lucasgarfield:
Ah I see, thanks for posting the most recent mocks. I had seen the iteration before that. We will figure it out with UX, I won't block the PR on this. But for this PR, I _would_ like to have the test fixtures updated so that the `Dark Chocolate` has the two composes associated with it and `Milk Chocolate` has none. Thanks for all the work so far Amir! :smile:

amirfefer:
@lucasgarfield, thanks for your review!

I added a test for choosing a blueprint with no corresponding composes.
Also I added (for now as a temp solution) a reset for showing all images:


https://github.com/RedHatInsights/image-builder-frontend/assets/11807069/be072084-cd7f-4af0-b034-e38458beb2a8

lucasgarfield (dismissed):
Looks great, thank you!

lucasgarfield:
@jrusz can you look at why IQE tests are failing? these changes should only affect experimental (changes only visible in local dev mode) but maybe something leaked again?

jrusz:
> @jrusz can you look at why IQE tests are failing? these changes should only affect experimental (changes only visible in local dev mode) but maybe something leaked again?

Yeah, some deployment timeout. Can you please resolve the conflict so we can try again? I'll try locally in the meantime

lucasgarfield (approved):
lgtm!

PR #1577, opened 2024-01-22
LandingPage: Hotfix to remove exposed blueprint buttons

lucasgarfield:
Commit bae6435fd92797e1879ff6f2798c11819e1dbff0 adds scaffolding for the blueprint table. Blueprint features should be hidden behind experimental flags. The new buttons related to creating blueprints and rebuilding images in the header were not hidden behind the experimental flag, this commit fixes that and does so.

bae6435fd92797e1879ff6f2798c11819e1dbff0 also changed the header title from "Image Builder" to "Images". This change has been a long time coming - it is being left in for now. We may want to revisit the copy in the popovers to align it as the term "Image Builder" is still being used in those.

### DIFF
--- a/src/Components/sharedComponents/ImageBuilderHeader.tsx
+++ b/src/Components/sharedComponents/ImageBuilderHeader.tsx
@@ -18,7 +18,13 @@ import {
 
 import './ImageBuilderHeader.scss';
 
-export const ImageBuilderHeader = () => {
+type ImageBuilderHeaderPropTypes = {
+  experimentalFlag?: string | true | undefined;
+};
+
+export const ImageBuilderHeader = ({
+  experimentalFlag,
+}: ImageBuilderHeaderPropTypes) => {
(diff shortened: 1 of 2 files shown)


src/Components/sharedComponents/ImageBuilderHeader.tsx
```diff
@@ -18,7 +18,13 @@ import {
 
 import './ImageBuilderHeader.scss';
 
-export const ImageBuilderHeader = () => {
+type ImageBuilderHeaderPropTypes = {
+  experimentalFlag?: string | true | undefined;
```
regexowl, line 22: New layout and buttons in experimental only. Seems to be working well! :sparkles: 

The only part that confuses me a bit is `string | true`, but that might be because I didn't look into flags much. Can we get both types?

regexowl (approved):

PR #1578, opened 2024-01-22
api/imageBuilder: update imageBuilder schema for blueprints changes

amirfefer:


regexowl:
There seems to be a bunch of comments removed manually in `imageBuilderApi.ts`, could you please update the file by running `npm run api`? That should re-generate the code based on the schema.

lucasgarfield:
I wonder if the default change to `@rtk-query/codegen-openapi` was changed, because when I run `npm run api` on my machine it removes comments from *all* API files.

amirfefer:
@regexowl, I didn't change anything manually, just copy pasted the new schema and run the openapi generator  script.

lucasgarfield:
superseded by #1580

PR #1580, opened 2024-01-22
API: improve api.sh and update api

lucasgarfield:
First commit modifies `api.sh` so that the image builder api.yaml file is downloaded with curl, this way we don't have to manually download it anymore.

Second commit updates api via `npm run api`.

### DIFF
--- a/api.sh
+++ b/api.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Download the most up-to-date imageBuilder.yaml file and overwrite the existing one
+curl https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml -o ./api/schema/imageBuilder.yaml
+
 # Run commands in the background using & operator
 npx @rtk-query/codegen-openapi ./api/config/imageBuilder.ts &
 npx @rtk-query/codegen-openapi ./api/config/rhsm.ts &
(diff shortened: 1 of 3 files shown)


regexowl (approved):
Amazing, thank you! :star_struck:

ezr-ondrej (approved):
Might be great to download all other specs as well, but it might be a bit slow then and we might want to have separate actions for it and this is a perfect start :+1: 

BTW the failure is "expected" and is being worked on by QE (in Content Sources channel)

ezr-ondrej:
Ah... there is a bit of an issue now, that `npm run api` is run in the CI to verify the API was not tempered with manually, which means that anytime the backend updates api spce, all fronted PR checks fail :)

I'll propose a split of the actions to `npm run api:update` that would download and regen and `npm run api` that would only regen :)
Tomorrow tho, I'm just stopping by today xD

regexowl:
> Ah... there is a bit of an issue now, that `npm run api` is run in the CI to verify the API was not tempered with manually, which means that anytime the backend updates api spce, all fronted PR checks fail :)
> 
> I'll propose a split of the actions to `npm run api:update` that would download and regen and `npm run api` that would only regen :) Tomorrow tho 😛

Oof :dizzy_face: great catch! Splitting the actions into two steps sounds great. :+1:

ezr-ondrej:
> Splitting the actions into two steps sounds great. 👍

Here we go: https://github.com/RedHatInsights/image-builder-frontend/pull/1585 :)

PR #1581, opened 2024-01-22
feat(HMS-3341): pass module_hotfixes flag to compose request

ezr-ondrej:
Content sources now support module_hotfixes flag, which is also supported by image-builder API (https://github.com/osbuild/image-builder/pull/928).
This passes this flag from ContentSources to the Compose request, to adhere to what user had set in Content sources. 

This flag allows for disabling unwanted modularity filtering.

### DIFF
--- a/api/schema/contentSources.json
+++ b/api/schema/contentSources.json
@@ -1,6 +1,33 @@
 {
     "components": {
         "schemas": {
+            "api.ContentUnitSearchRequest": {
+                "properties": {
+                    "limit": {
+                        "description": "Maximum number of records to return for the search",
+                        "type": "integer"
+                    },
+                    "search": {
+                        "description": "Search string to search content unit names",
+                        "type": "string"
+                    },
+                    "urls": {
+                        "description": "URLs of repositories to search",
+                        "items": {
+                            "type": "string"
+                        },
+                        "type": "array"
+                    },
+                    "uuids": {
+                        "description": "List of RepositoryConfig UUIDs to search",
+                        "items": {
+                            "type": "string"
+                        },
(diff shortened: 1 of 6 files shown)


lucasgarfield:
Thanks, @ezr-ondrej. Should I be embarrassed I don't know what this does, though? :sweat_smile: Can you explain what the module hotfixes are used for?

ezr-ondrej:
Ah, sorry I should have realized there are different reviewers, my bad. I've added tiny bit of context in the description, but here goes bit more relevant context from the parent Jira ticket that is not public:

Some [3rd party repositories](https://nginx.org/en/linux_packages.html#RHEL) use module_hotfixes=true. This flag is needed when the repository ships non-modular content that RHEL ships as a modular one. The flag is especially important on RHEL 8, because DNF will always prioritize enabling a default module stream in order to install a required package, even when the NEVR of the 3rd party package is higher. In order to install the 3rd party one, the repository needs to have module_hotfixes=true. This is not an issue on RHEL 9, because there are no default module streams, and IB never enables any streams.

In short my main goal: enable installing nginx from [nginx official repositories](http://nginx.org/packages/centos/). There will be more cases for sure, but nginx is the main one I'm after here, without this, there is no way to install official nginx releases from their repos, because we (RH) ship nginx as modular, so modularity filters the nginx packages out.

ezr-ondrej:
I've rebased, but this should wait until we re-enable custom repos.

croissanne (dismissed):

lucasgarfield (approved):
Thanks, @ezr-ondrej! Repositories is back so this can go in.

PR #1593, opened 2024-01-25
test: Repositories step review

regexowl:
This disables tests connected to the Repositories step while the service is unavailable.

All instances were marked with a comment "Temporarily disable" to make the complete revert easier when the service is reachable again.


lucasgarfield (approved):
Thank you!

PR #1595, opened 2024-01-25
Wizard: Temporarily disable custom packages search

regexowl:
This temporarily disables custom packages search and adds an alert to the step.


regexowl:
> Would users be able to access custom packages if the repos step is disabled?

Yes, on Recreate

PR #1596, opened 2024-01-25
test: Remove debugging statement and change `getBy` query to `findBy`

regexowl:
This removes a leftover debugging statement and changes `getBy` query to an awaited async `findBy` as this was causing problems in the test when the checked component wasn't immediately available.


regexowl:
Also prepared a PR to catch the debugging statements during linting in the future: https://github.com/RedHatInsights/image-builder-frontend/pull/1597

lucasgarfield (approved):
nice catch

PR #1607, opened 2024-01-29
Blueprints: add Blueprints filtering

ezr-ondrej:
Enables Blueprints filtering.
Given this is a server side filtering, it also adds Debounce hook.
This hook enables delay the API request to save server roundtrips.

Refs HMS-3389

[Screencast from 2024-01-30 10-19-03.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/2884324/80030cc9-81ae-4d92-820b-036e5ed45702)

### DIFF
--- a/src/Components/Blueprints/BlueprintsSideBar.tsx
+++ b/src/Components/Blueprints/BlueprintsSideBar.tsx
@@ -1,6 +1,7 @@
-import React, { useState, Dispatch, SetStateAction } from 'react';
+import React, { useState, useCallback } from 'react';
 
 import {
+  Bullseye,
   Button,
   EmptyState,
   EmptyStateActions,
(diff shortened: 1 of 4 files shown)


src/Components/LandingPage/LandingPage.tsx
```diff
@@ -56,7 +57,11 @@ export const LandingPage = () => {
   const [selectedBlueprint, setSelectedBlueprint] = useState<
     string | undefined
   >();
-  const { data: blueprints, isLoading } = useGetBlueprintsQuery({});
+  const [searchQuery, blueprintsFilter, setBlueprintsFilter] =
+    useDebouncedSearch('', 500);
```
amirfefer, line 61: I won't block this custom hook, but lodash has debounce implementation already, why not using it and keep it as regular state?

```js
import _debounce from 'lodash/debounce';

    const debounceFn = useCallback(_debounce(setBlueprintsFilter, 500), []);
        <SearchInput
             placeholder="Search by name or description"
             value={filter}
             onChange={(_event, value) => debounceFn(value)}
             onClear={() => onChange('')}
```

src/Components/LandingPage/LandingPage.tsx
```diff
@@ -56,7 +57,11 @@ export const LandingPage = () => {
...
+  const { data: blueprints, isLoading } = useGetBlueprintsQuery({
+    search: searchQuery,
```
amirfefer, line 63: Can you please change the default search string to undefined? otherwise the [backend](https://github.com/osbuild/image-builder/blob/main/internal/v1/handler_blueprints.go#L191) triggers a redundant  regex search query.

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -61,7 +81,7 @@ const BlueprintsSidebar = ({
         <StackItem>
           <SearchInput
```
amirfefer, line 82: A nice a addition for the search input is adding the `resultsCount` prop for rendering a counter badge, i.e
Screenshot 2024-01-29 at 14 42 54

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -52,7 +54,25 @@ const BlueprintsSidebar = ({
...
+          <Stack hasGutter>
+            <StackItem>
+              <SearchInput
+                placeholder="Search by name or description"
+                value={filter}
+                onChange={(_event, value) => onChange(value)}
+                onClear={() => onChange('')}
+              />
+            </StackItem>
+            <StackItem>No Blueprints found</StackItem>
```
amirfefer, line 71: How about using the empty state design ?
Screenshot 2024-01-29 at 14 46 53

src/Components/LandingPage/LandingPage.tsx
amirfefer: Also, 500 ms feels a bit long to me, would 300 be better? also do we want to add a spinner or another indicator for fetching the data?

amirfefer (changes requested):
Thanks @ezr-ondrej ! looks and works great 👍 
I left a few inline comments, and also I've found a small glitch once clearing the search input:

https://github.com/RedHatInsights/image-builder-frontend/assets/11807069/a061a03c-7475-4a2e-b58b-b7e9fc83145e

src/Components/LandingPage/LandingPage.tsx
ezr-ondrej: because I did not find a way how to use lodash debounce :stuck_out_tongue:
Definitelly on looks better, let me improve it! :+1:

src/Components/LandingPage/LandingPage.tsx
ezr-ondrej: it is undefined - the debounce hook returns undefined if the filter is less then lenght (default 1).
But I agree it's not very straight forward approach.

src/Components/LandingPage/LandingPage.tsx
amirfefer: :)
 I'm thinking about mixing  backend and frontend filtering would work the best together , we already have 100 blueprints records in the store, we can  filter it locally by default for given instant result, and update the store if there were any updates from the debounced query., though I think this enhancement  can  be implemented as a follow up for a general optimistic updating refactor.

ezr-ondrej: That's IMO amazing idea, we should discuss it with Andrew Dewar, he did quite a lot of optimistic updating, we can get some lessons learned :)

ezr-ondrej:
Thanks for the amazing pointers Amir, I've incorporated almost all of them :)
Just the optimistic search is missing, but that would be an amazing followup IMO - as you say, we will have most of the blueprints loaded in majority of cases :)

src/Components/Blueprints/BlueprintsSideBar.tsx
ezr-ondrej: I've used `count blueprints` it feels more user friendly, but kinda long, so I'm not strongly for that, happy to go just with the number if you'd think it's better ;)

src/Components/LandingPage/LandingPage.tsx
ezr-ondrej: Wow... it's amazing, just 200 ms less and it really feels much smoother experience.
So I've read about it because I was amazed and I've learned that smooth feeling is when the interaction computer <-> human is below 400ms, cool

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -22,72 +23,132 @@ type blueprintProps = {
...
 
 const BlueprintsSidebar = ({
   blueprints,
   selectedBlueprint,
   setSelectedBlueprint,
+  filter,
+  setFilter,
 }: blueprintProps) => {
-  const [blueprintFilter, setBlueprintFilter] = useState('');
+  const blueprintsCount = blueprints?.length || 0;
```
amirfefer, line 51: This is just the fetched data length, we can use the meta object for the query search count :)

amirfefer (commented):
Thanks @ezr-ondrej ! works smoothly indeed 👍 
just a minor nitpick

src/Components/Blueprints/BlueprintsSideBar.tsx
amirfefer: agree, feels friendly and verbose, but a bit too long IMHO

src/Components/Blueprints/BlueprintsSideBar.tsx
ezr-ondrej: You're completely right! :)
Adding that.

ezr-ondrej:
hmm while adding the blueprintsTotal, I've realized we can move the query to the sidebar, the only information is the selected blueprint. So I've moved everything else down to the sidebar component.

It looks bit cleaner on the LandingPage component side.
I didn't change anything, just moved the query and added the total. WDYT? :)

amirfefer (approved):
LGTM thanks @ezr-ondrej 👍

lavocatt:
Thanks for working on this! I've got an issue with the filter not filtering on `msw` :
[Screencast from 2024-02-01 10-02-23.webm](https://github.com/RedHatInsights/image-builder-frontend/assets/86971992/a51cfe4d-e558-4259-a036-b51eafe01366)
to reproduce, spin up the frontend with `npm run stage-beta:msw+experimental`

ezr-ondrej:
> to reproduce, spin up the frontend with `npm run stage-beta:msw+experimental`

For me it still loads up data from the server :thinking:

lavocatt (approved):
🚀 This works great! Thanks a lot.

lavocatt:
@jrusz can you have a look at why the tests are failing please? We've been restarting them a bunch of time without any success.

jrusz:
> @jrusz can you have a look at why the tests are failing please? We've been restarting them a bunch of time without any success.

Yeah there is some more widespread issue happening, other consoledot apps are seeing the same issue in their pr_check jobs... Nobody seems to know what's wrong yet... I believe that this should be ok to merge though as this should be behind the feature flag.

lavocatt:
> Yeah there is some more widespread issue happening, other consoledot apps are seeing the same issue in their pr_check jobs... Nobody seems to know what's wrong yet... I believe that this should be ok to merge though as this should be behind the feature flag.

Thanks for the heads up, I'm merging then.

PR #1611, opened 2024-01-30
feat(HMS-3500) V2Wizrd: add kernel and services data

mgold1234:
this commit adds kernel arguments and enabled/disabled services to Oscap step.
and also add to Review step.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
 
 import {
+  Alert,
+  CodeBlock,
+  CodeBlockCode,
   Spinner,
   TextContent,
   TextList,
@@ -17,7 +20,7 @@ import {
   selectProfile,
 } from '../../../../store/wizardSlice';
 
-const OscapProfileInformation = (): JSX.Element => {
+export const OscapProfileInformation = (): JSX.Element => {
   const release = useAppSelector((state) => selectDistribution(state));
   const oscapProfile = useAppSelector((state) => selectProfile(state));
 
@@ -36,46 +39,97 @@ const OscapProfileInformation = (): JSX.Element => {
     }
   );
 
+  const enabledServicesDisplayString =
+    oscapProfileInfo?.services?.enabled?.join(' ');
+  const disableServicesDisplayString =
+    oscapProfileInfo?.services?.disabled?.join(' ');
+
   return (
     <>
       {isFetchingOscapProfileInfo && <Spinner size="lg" />}
       {isSuccessOscapProfileInfo && (
-        <TextContent>
-          <br />
-          <TextList component={TextListVariants.dl}>
-            <TextListItem
-              component={TextListItemVariants.dt}
-              className="pf-u-min-width"
-            >
-              Profile description:
-            </TextListItem>
-            <TextListItem component={TextListItemVariants.dd}>
-              {oscapProfileInfo.openscap?.profile_description}
-            </TextListItem>
-          </TextList>
-          <TextList component={TextListVariants.dl}>
-            <TextListItem
-              component={TextListItemVariants.dt}
-              className="pf-u-min-width"
-            >
-              Operating system:
-            </TextListItem>
-            <TextListItem component={TextListItemVariants.dd}>
-              {RELEASES.get(release)}
-            </TextListItem>
-          </TextList>
-          <TextList component={TextListVariants.dl}>
-            <TextListItem
-              component={TextListItemVariants.dt}
-              className="pf-u-min-width"
-            >
-              Reference ID:
-            </TextListItem>
-            <TextListItem component={TextListItemVariants.dd}>
-              {oscapProfileInfo.openscap?.profile_id}
-            </TextListItem>
-          </TextList>
-        </TextContent>
+        <>
+          <TextContent>
+            <br />
+            <TextList component={TextListVariants.dl}>
+              <TextListItem
+                component={TextListItemVariants.dt}
+                className="pf-u-min-width"
+              >
+                Profile description:
+              </TextListItem>
+              <TextListItem component={TextListItemVariants.dd}>
+                {oscapProfileInfo.openscap?.profile_description}
+              </TextListItem>
+              component={TextListVariants.dl}
+              <TextListItem
+                component={TextListItemVariants.dt}
+                className="pf-u-min-width"
+              >
+                Operating system:
+              </TextListItem>
+              <TextListItem component={TextListItemVariants.dd}>
+                {RELEASES.get(release)}
+              </TextListItem>
+              component={TextListVariants.dl}
+              <TextListItem
+                component={TextListItemVariants.dt}
+                className="pf-u-min-width"
+              >
+                Reference ID:
+              </TextListItem>
+              <TextListItem component={TextListItemVariants.dd}>
+                {oscapProfileInfo.openscap?.profile_id}
+              </TextListItem>
+              <TextListItem
+                component={TextListItemVariants.dt}
+                className="pf-u-min-width"
+              >
+                Kernel arguments:
+              </TextListItem>
+              <TextListItem component={TextListItemVariants.dd}>
+                <CodeBlock>
+                  <CodeBlockCode>
+                    {oscapProfileInfo?.kernel?.append}
+                  </CodeBlockCode>
+                </CodeBlock>
+              </TextListItem>
+              <TextListItem
+                component={TextListItemVariants.dt}
+                className="pf-u-min-width"
+              >
+                Disabled services:
+              </TextListItem>
+              <TextListItem component={TextListItemVariants.dd}>
+                <CodeBlock>
+                  <CodeBlockCode>{disableServicesDisplayString}</CodeBlockCode>
+                </CodeBlock>
+              </TextListItem>
+              <TextListItem
+                component={TextListItemVariants.dt}
+                className="pf-u-min-width"
+              >
+                Enabled services:
+              </TextListItem>
+              <TextListItem component={TextListItemVariants.dd}>
+                <CodeBlock>
+                  <CodeBlockCode>{enabledServicesDisplayString}</CodeBlockCode>
+                </CodeBlock>
+              </TextListItem>
+            </TextList>
+          </TextContent>
+
+          <Alert
+            variant="info"
+            isInline
+            isPlain
+            title="Additional customizations"
+          >
+            Selecting an OpenSCAP profile will cause the appropriate packages,
+            file system configuration, kernel arguments, and services to be
+            added to your image.
+          </Alert>
+        </>
       )}
     </>
   );
--- a/src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
@@ -30,7 +30,6 @@ import {
 } from '../../../../constants';
 import { extractProvisioningList } from '../../../../store/helpers';
 import { useAppSelector } from '../../../../store/hooks';
-import { useGetOscapCustomizationsQuery } from '../../../../store/imageBuilderApi';
 import { useGetSourceListQuery } from '../../../../store/provisioningApi';
 import { useShowActivationKeyQuery } from '../../../../store/rhsmApi';
 import {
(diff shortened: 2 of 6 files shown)


src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -0,0 +1,547 @@
...
+        >
+          Reference ID:
+        </TextListItem>
+        <TextListItem component={TextListItemVariants.dd}>
+          {oscapProfile}
+        </TextListItem>
+        <TextListItem
+          component={TextListItemVariants.dt}
+          className="pf-u-min-width"
+        >
```
regexowl, line 507: Do we want to render an empty code block when there are no kernel arguments / services? I guess we maybe do, right? Just to have a complete overview of what's included in the profile?

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -0,0 +1,547 @@
...
+  const enabledServices = data?.services?.enabled;
+  const enabledServicesDisplayString = enabledServices?.reduce(
+    (accumulator, currentValue) => accumulator + ' ' + currentValue,
+    ''
+  );
+  const disableServices = data?.services?.disabled;
+  const disableServicesDisplayString = disableServices?.reduce(
+    (accumulator, currentValue) => accumulator + ' ' + currentValue,
+    ''
+  );
```
regexowl, line 469: There seems to be an extra space at the beginning of the disabled/enabled services:

image

regexowl: Also the `.reduce` method seems quite heavy duty for this - could you maybe just use [join method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join)? Something like this:

```typescript
  const enabledServicesDisplayString = data?.services?.enabled.join(' ');
  const disableServicesDisplayString = data?.services?.disabled.join(' ');
```

regexowl (commented):
Added a few comments - if I understand it correctly the PR is rebased on the Review step PR so I didn't review any of the Review step commits, which left me just one commit to review (and I'll stop saying the word "review" now :D )

Would it be possible to add a short test for this functionality as well? In the .compliance test file itself it should be easy, just adding something like "check if the kernel and disabled/enabled services are rendered".

And then what would be also needed is expanding the OSCAP mock fixtures with the parameters and services based on the API schema. Could be just some mock values like "service1, service2" and "argument1, argument2", nothing complicated :)

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
mgold1234: this is a very good question, I thought about that and maybe its intersting the user to see the both case - when there is existning data for kernel and services and also when there is no for a specific pofile

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
mgold1234: simple with join, I like it but it seems that it is not fix the problem, maybe do you have another idea?

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
regexowl: Agree, this way the data may render as empty, but the overview doesn't feel incomplete... maybe we could render "None" instead of the empty code block, but that might be something we might want to ask during the UX sync

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
mgold1234: its very strange its look like sometimes there is extra space and sometimes not

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
mgold1234: I spoke with Kelsea and she said in case we dont have any value, keep it empy box

lucasgarfield:
The `<Alert>` component is really wonky when it is embedded inside of the text list. Let's make this easy on ourselves and just put it at the bottom. I also want to change the copy to what I've written below, which I think communicates what we're doing a little more clearly:
image

Additionally, please change `Kernel: ` to `Kernel arguments`, `Disabled Services` to `Disabled services`, and `Enabled Services` to `Enabled services`... just so we are consistent with capitalization. :smile:

regexowl:
> Oops, what happened there?
> 
> image

Oh, I see. Alert nested in the TextList causes some problems. It will need to be mover outside, like this:
```typescript
              <CodeBlock>
                <CodeBlockCode id="code-content">
                  {enabledServicesDisplayString}
                </CodeBlockCode>
              </CodeBlock>
            </TextListItem>
          </TextList>
          <Alert
            variant="info"
            isInline
            isPlain
            title="Additional customizations"
          >
            Selecting an OpenSCAP profile will cause the appropriate packages,
            file system configuration, kernel arguments, and services to be
            added to your image.
          </Alert>
        </TextContent>
      )}
```

src/test/Components/CreateImageWizardV2/CreateImageWizard.compliance.test.tsx
```diff
@@ -154,6 +154,13 @@ describe('Step Compliance', () => {
         /cis red hat enterprise linux 8 benchmark for level 1 - workstation/i
       )
     );
+    screen.getByText(/kernel arguments:/i);
+    screen.getByText(/audit_backlog_limit=8192 audit=1/i);
+    screen.getByText(/disabled services:/i);
+    screen.getByText(/nfs-server/i);
+    screen.getByText(/enabled services:/i);
+    screen.getByText(/enabled services:/i);
```
regexowl, line 162: Should this be a list of the enabled services instead of a duplicate of the previous line?

regexowl, line 162: As the test suite is mostly asynchronous let's use the findBy query where possible to avoid flakes due to timing.

```suggestion
    await screen.findByText(/kernel arguments:/i);
    await screen.findByText(/audit_backlog_limit=8192 audit=1/i);
    await screen.findByText(/disabled services:/i);
    await screen.findByText(/nfs-server/i);
    await screen.findByText(/enabled services:/i);
    await screen.findByText(/enabled services:/i);
```

src/store/wizardSlice.ts
```diff
@@ -80,6 +84,9 @@ const initialState: wizardState = {
   openScap: {
     profile: undefined,
   },
+  kernel: {
+    kernelAppend: '',
+  },
```
regexowl, line 89: This is mostly a question - do we want to include kernel under the openScap state so we're consistent with storing values under a specific step? So like this?
```typescript
  openScap: {
    profile: undefined,
    kernel: {
      kernelAppend: '',
    },
  },
```

or is it its own thing? I might not be up to date with the compliance work so not sure what would the implications be.

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -525,8 +531,56 @@ export const OscapList = () => {
...
         </TextListItem>
+        <TextListItem
+          component={TextListItemVariants.dt}
+          className="pf-u-min-width"
+        >
+          Kernel arguments:
+        </TextListItem>
+        <TextListItem component={TextListItemVariants.dd}>
+          <CodeBlock>
+            <CodeBlockCode id="code-content">
```
regexowl, line 542: Are the ids needed / working like a prop or something? What I mean is if they have a purpose as a variant setter or something similar.

If not we should either drop them or differentiate them so we can use the id for example as a unique identifier in the tests.

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -525,8 +531,56 @@ export const OscapList = () => {
...
+        </TextListItem>
+        <TextListItem
+          component={TextListItemVariants.dt}
+          className="pf-u-min-width"
+        >
+          Disabled services:
+        </TextListItem>
+        <TextListItem component={TextListItemVariants.dd}>
+          <CodeBlock>
+            <CodeBlockCode id="code-content">
```
regexowl, line 555: Same question about the id

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -525,8 +531,56 @@ export const OscapList = () => {
...
+        </TextListItem>
+        <TextListItem
+          component={TextListItemVariants.dt}
+          className="pf-u-min-width"
+        >
+          Enabled services:
+        </TextListItem>
+        <TextListItem component={TextListItemVariants.dd}>
+          <CodeBlock>
+            <CodeBlockCode id="code-content">
```
regexowl, line 568: Same question about the id

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -525,8 +531,56 @@ export const OscapList = () => {
...
+        <Alert
+          variant="info"
+          isInline
+          isPlain
+          title="Additional customizations"
+        >
+          Selecting an OpenSCAP profile will cause the appropriate packages,
+          file system configuration, kernel arguments, and services to be added
+          to your image.
+        </Alert>
```
regexowl, line 582: The Alert will need to be put outside of the TextList otherwise the rendering is chaotic. See https://github.com/osbuild/image-builder-frontend/pull/1611#issuecomment-1933752350

image

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -525,8 +531,56 @@ export const OscapList = () => {
         <TextListItem component={TextListItemVariants.dd}>
           {oscapProfile}
         </TextListItem>
+        <TextListItem
```
regexowl, line 534: Could we somehow use the `<OscapProfileInformation>` component here so we don't have two duplicate lists we'd have to keep in mind when making changes to them?

src/Components/CreateImageWizardV2/steps/Oscap/OscapProfileInformation.tsx
```diff
@@ -74,6 +82,55 @@ const OscapProfileInformation = (): JSX.Element => {
...
+            <Alert
+              variant="info"
+              isInline
+              isPlain
+              title="Additional customizations"
+            >
+              Selecting an OpenSCAP profile will cause the appropriate packages,
+              file system configuration, kernel arguments, and services to be
+              added to your image.
+            </Alert>
```
regexowl, line 133: Alert will need to be put outside of the TextList, see https://github.com/osbuild/image-builder-frontend/pull/1611#issuecomment-1933752350

src/test/Components/CreateImageWizardV2/CreateImageWizard.compliance.test.tsx
mgold1234: sure, you correct :)

regexowl:
Weird behaviour in the store alert - the kernel initializes with an empty string in `kernelAppend` which works for me, no complaints there (value of the store on the very right):

image

 But when I choose a profile it sets the kernel correctly on the step, but the store value changes to an empty object:

image

and when I change the profile again it also gets rendered correctly, but now the store contains the previous value:

image

It seems to be one step behind the current state of the wizard.

src/store/wizardSlice.ts
mgold1234: not sure, @lucasgarfield what is your opinion regarding that?

lucasgarfield: Eventually there will be a step where the kernel arguments can be provided by the user, but we don't know whether it will be grouped with other options (for instance under an `Advanced` step) or by itself...

**For now let's put it under the openScap step as @regexowl suggests.** 

This is one of the nice things about using selectors to access everything from the state, we can easily change where it is stored in the state later and we'll only need to update the selector function.

mgold1234: at the backend side its look sepeare from OpenScap 
Screenshot 2024-02-08 at 14 56 17

are you sure you want it under OpenScap?

mgold1234:
as I understand from Gianluca, there it is possible to improve the way we get the customizations data from the back end side, maybe it will help and solve the wired deily we   see at the frontend side? @lucasgarfield WDYT?
or should we find a solution in our side as well?

lucasgarfield:
Odd it's only failing for Node 16, it passes fine on Node 19.6.0 on my machine locally as well.

lucasgarfield (approved):
Thank you @mgold1234 !

PR #1614, opened 2024-01-31
Fix osbuild pages url

jrusz:
It's now lowercase.


ochosi (approved):
Thank you for fixing my mess!

PR #1617, opened 2024-01-31
feat(HMS-3401): Add pagination for blueprints sidebar

avitova:
So, from the ticket description, we want to have server-side blueprints pagination here. Feel free to review the chosen solution.
<details>
  <summary>The feature</summary>

view-more-blueprints
</details>

### DIFF
--- a/src/Components/Blueprints/BlueprintsPagination.tsx
+++ b/src/Components/Blueprints/BlueprintsPagination.tsx
@@ -1,0 +1,56 @@
+import React from 'react';
+
+import {
+  OnSetPage,
+  Pagination,
+  PaginationVariant,
+} from '@patternfly/react-core';
+
+import {
+  selectBlueprintSearchInput,
+  selectLimit,
+  selectOffset,
+  setBlueprintLimit,
+  setBlueprintsOffset,
+} from '../../store/BlueprintSlice';
+import { useAppDispatch, useAppSelector } from '../../store/hooks';
+import { useGetBlueprintsQuery } from '../../store/imageBuilderApi';
+
+const BlueprintsPagination = () => {
+  const blueprintSearchInput = useAppSelector(selectBlueprintSearchInput);
+  const blueprintsOffset = useAppSelector(selectOffset) || 0;
+  const blueprintsLimit = useAppSelector(selectLimit) || 10;
+  const currPage = Math.floor(blueprintsOffset / blueprintsLimit) + 1;
+  const { data: blueprintsData } = useGetBlueprintsQuery({
+    search: blueprintSearchInput,
+    limit: blueprintsLimit,
+    offset: blueprintsOffset,
+  });
+  const dispatch = useAppDispatch();
+
+  const blueprintsTotal = blueprintsData?.meta?.count || 0;
+  const onSetPage: OnSetPage = (_, page) => {
+    const direction = page > currPage ? 1 : -1; // Calculate offset based on direction of paging
+    const nextOffset = blueprintsOffset + direction * blueprintsLimit;
+    dispatch(setBlueprintsOffset(nextOffset));
+  };
+  const onPerPageSelect: OnSetPage = (_, perPage) => {
+    dispatch(setBlueprintsOffset(0));
+    dispatch(setBlueprintLimit(perPage));
+  };
+  return (
+    <Pagination
+      variant={PaginationVariant.bottom}
+      itemCount={blueprintsTotal}
+      perPage={blueprintsLimit}
+      page={currPage}
+      onSetPage={onSetPage}
+      onPerPageSelect={onPerPageSelect}
+      widgetId="blueprints-pagination-bottom"
+      data-testid="blueprints-pagination-bottom"
+      isCompact
+    />
+  );
+};
+
+export default BlueprintsPagination;
--- a/src/Components/Blueprints/BlueprintsSideBar.tsx
+++ b/src/Components/Blueprints/BlueprintsSideBar.tsx
@@ -20,13 +20,18 @@ import debounce from 'lodash/debounce';
 import { Link } from 'react-router-dom';
 
 import BlueprintCard from './BlueprintCard';
+import BlueprintsPagination from './BlueprintsPagination';
 
 import {
   selectBlueprintSearchInput,
+  selectLimit,
+  selectOffset,
   selectSelectedBlueprintId,
   setBlueprintId,
   setBlueprintSearchInput,
+  setBlueprintsOffset,
 } from '../../store/BlueprintSlice';
+import { imageBuilderApi } from '../../store/enhancedImageBuilderApi';
 import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import {
   useGetBlueprintsQuery,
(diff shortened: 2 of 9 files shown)


amirfefer (commented):
Thanks @avitova, looks great!

From the mocks, it seems that the sidebar should be sticky with specific max height and scrolling enabled, as you can see:
Screenshot 2024-01-31 at 20 25 41

PF5 sidebar supports sticky and static layout - https://www.patternfly.org/components/sidebar#sticky-panel

ezr-ondrej:
I've moved the fetch to the Blueprints sidebar component, so I've probably broken your PR completely, so you'd need to rebase :(
But it should be easier to implement now - you won't need to send all the props down the component tree

avitova:
I have rebased and made the sidebar sticky. :)

ezr-ondrej:
This is what I see, I guess we should make the height of the sidebar bit larger, but it's just a nit, I'm happy to be told we'll clean the UI up in follow-ups
image

lucasgarfield:
@avitova just following up on this, what do you think about adding some additional blueprints to our test fixtures so that we can test the pagination using Mock Service Worker?

amirfefer (commented):
Thanks @avitova! It works but with some glitches, because `merge` breaks the invalid cache.

lucasgarfield:
It sounds like this is pretty tricky using RTKQ. For now should we consider just adding a "Load more blueprints" button at the bottom that loads the next 5 (or however many) blueprints? I would consider infinite scroll a "nice to have", and wouldn't block releasing the MVP on it.

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -64,10 +69,39 @@ const BlueprintsSidebar = ({
...
     search: blueprintsSearchQuery,
+    limit: perPageBlueprint,
+    offset: perPageBlueprint * (pageBlueprint - 1),
   });
-  const blueprints = blueprintsData?.data;
+  React.useEffect(() => {
+    if (pageBlueprint <= 1) {
+      return;
+    }
+    const sidebar = document.querySelector('.pf-v5-c-sidebar__panel');
```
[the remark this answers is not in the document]
lucasgarfield: I consider `useRef` to be a code smell and prefer to avoid using it whenever possible.

amirfefer:
>  For now should we consider just adding a "Load more blueprints" button at the bottom that loads the next 5 (or however many) blueprints?

I think the main issue here is not the scrolling behavior itself, but is the `merge` functionality that would be used in a "Load more"  button  as well, no?

ezr-ondrej:
I'd agree to leave the scrolling out of scope of this PR, so we can do the the button first properly.
As Amir said, it still leaves the main problem that rtk doesn't support infinite pagination an merge is just a hack.

lucasgarfield:
Can we just use Patternfly's compact pagination then? The "Clear selection" button is already taking up a fair amount of space. Maybe we could find a way to fit the compact pagination and "Clear selection" in the same vertical space?

lucasgarfield:
I know we discussed this a lot with UX but if we're blocked because it is too hard to implement, let's just do the easiest thing we can for now to get the MVP out the door.

avitova:
Update: This solution does not implement infinite scroll, but rather a common pagination. This is an MVP to not make things more complicated since RTK does not seem to support infinite paging.
Feel free to comment on the scss updates, and what would be your preference on this. This should be now ready for a review. :)

src/Components/Blueprints/BlueprintsPagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+} from '../../store/BlueprintSlice';
+import { useAppDispatch, useAppSelector } from '../../store/hooks';
+import { useGetBlueprintsQuery } from '../../store/imageBuilderApi';
+
+const BlueprintsPagination = () => {
+  const blueprintSearchInput = useAppSelector(selectBlueprintSearchInput);
+  const blueprintsOffset = useAppSelector(selectOffset);
+  const blueprintsLimit = useAppSelector(selectLimit);
+  const currPage =
+    Math.floor((blueprintsOffset || 0) / (blueprintsLimit || 0)) + 1;
```
ezr-ondrej, line 24: I'd rather default to something that works :P (ideally defaultLimit if we can get it, if not, some reasonable value):
```suggestion
    Math.floor((blueprintsOffset || 0) / (blueprintsLimit || 10)) + 1;
```

ezr-ondrej: TBH I think that expecting there will always be a value is fair too, we do have default in the Slice https://github.com/osbuild/image-builder-frontend/blob/b5c3f12db2139c1d8f32ab1818fb5226b14672b2/src/store/BlueprintSlice.ts#L19

src/Components/Blueprints/BlueprintsPagination.tsx
```diff
@@ -0,0 +1,58 @@
...
+  setBlueprintLimit,
+  setBlueprintsOffset,
+} from '../../store/BlueprintSlice';
+import { useAppDispatch, useAppSelector } from '../../store/hooks';
+import { useGetBlueprintsQuery } from '../../store/imageBuilderApi';
+
+const BlueprintsPagination = () => {
+  const blueprintSearchInput = useAppSelector(selectBlueprintSearchInput);
+  const blueprintsOffset = useAppSelector(selectOffset);
+  const blueprintsLimit = useAppSelector(selectLimit);
```
ezr-ondrej, line 22: we could set the fallbacks here, unless it would break the useQuery of course
```suggestion
  const blueprintsOffset = useAppSelector(selectOffset) || 0;
  const blueprintsLimit = useAppSelector(selectLimit) || 10;
```

src/Components/LandingPage/LandingPage.scss
```diff
@@ -9,3 +9,8 @@
 .expand-section {
     background-color: var(--pf-global--palette--white);
 }
+
+.sidebar-panel {
+    height: 500pt;
```
ezr-ondrej, line 14: should we rather use px here? :)
Dunno about the value, on my monitor 750px worked better, but I do not have a strong opinion :)

ezr-ondrej (approved):
Just small nits, but looks great overall, if we would figure out how to test it, it would be great tho :)

ezr-ondrej (commented):
Linter is unhappy now :thinking:
And I belive we'll need to adjust some tests to pass with the new data.

Here is the gist for making the mock ready for testing pagination: https://github.com/avitova/image-builder-frontend/pull/1

avitova:
Okay, so we have tests for this, and they are green!

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -136,6 +145,8 @@ const BlueprintSearch = ({ blueprintsTotal }: blueprintSearchProps) => {
   const dispatch = useAppDispatch();
   const debouncedSearch = useCallback(
     debounce((filter) => {
+      dispatch(setBlueprintsOffset(0));
+      dispatch(imageBuilderApi.util.invalidateTags([{ type: 'Blueprints' }]));
```
ezr-ondrej, line 149: have you tested ivalidation is needed? with search we have different parameters, so it might not be needed to invalidate...
Might be a good idea anyway, so I'm not opposed to keeping it.

ezr-ondrej (approved):
I belive we're ready :+1:

lucasgarfield (approved):
Works nicely! Thanks, @avitova!

PR #1620, opened 2024-02-01
Blueprints: Add handler for creating new blueprint

amirfefer:
This PR adds a handler for submitting  the blueprint's wizard, by implementing a mapper from redux store (wizard slice) to a valid blueprint creation request.

https://github.com/RedHatInsights/image-builder-frontend/assets/11807069/4264d958-46c7-49c4-a548-383ea900390b

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -15,6 +15,7 @@ import OscapStep from './steps/Oscap';
 import RegistrationStep from './steps/Registration';
 import RepositoriesStep from './steps/Repositories';
 import ReviewStep from './steps/Review';
+import ReviewWizardFooter from './steps/Review/Footer';
 import Aws from './steps/TargetEnvironment/Aws';
 import Gcp from './steps/TargetEnvironment/Gcp';
 import {
@@ -52,13 +53,22 @@ export const CustomWizardFooter = ({
   const { goToNextStep, goToPrevStep, close } = useWizardContext();
   return (
     <WizardFooterWrapper>
-      <Button variant="primary" onClick={goToNextStep} isDisabled={disableNext}>
+      <Button
+        ouiaId="wizard-next-btn"
+        variant="primary"
+        onClick={goToNextStep}
+        isDisabled={disableNext}
+      >
         Next
       </Button>
-      <Button variant="secondary" onClick={goToPrevStep}>
+      <Button
+        ouiaId="wizard-back-btn"
+        variant="secondary"
+        onClick={goToPrevStep}
+      >
         Back
       </Button>
-      <Button variant="link" onClick={close}>
+      <Button ouiaId="wizard-cancel-btn" variant="link" onClick={close}>
         Cancel
       </Button>
     </WizardFooterWrapper>
@@ -212,7 +222,7 @@ const CreateImageWizard = () => {
           <WizardStep
             name="Review"
             id="step-review"
-            footer={<CustomWizardFooter disableNext={true} />}
+            footer={<ReviewWizardFooter />}
           >
             <ReviewStep />
           </WizardStep>
--- a/src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
@@ -1,0 +1,115 @@
+import React, { useState, useEffect } from 'react';
+
+import {
+  Button,
+  Dropdown,
+  DropdownItem,
+  DropdownList,
+  MenuToggle,
+  MenuToggleElement,
+  WizardFooterWrapper,
+  useWizardContext,
+} from '@patternfly/react-core';
+import { SpinnerIcon } from '@patternfly/react-icons';
+import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
+import { useNavigate, useParams } from 'react-router-dom';
+
+import {
+  useComposeBlueprintMutation,
+  useCreateBlueprintMutation,
+} from '../../../../store/imageBuilderApi';
+import { resolveRelPath } from '../../../../Utilities/path';
+import { mapRequestFromState } from '../../utilities/requestMapper';
+
+const ReviewWizardFooter = () => {
+  const { goToPrevStep, close } = useWizardContext();
+  const [
+    createBlueprint,
+    { isLoading: isCreationLoading, isSuccess: isCreationSuccess },
+  ] = useCreateBlueprintMutation();
+  const [buildBlueprint, { isLoading: isBuildLoading }] =
+    useComposeBlueprintMutation();
+  const { auth } = useChrome();
+  const navigate = useNavigate();
+  const { composeId } = useParams();
+  const [isOpen, setIsOpen] = useState(false);
+
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+
+  useEffect(() => {
+    if (isCreationSuccess) {
+      navigate(resolveRelPath(''));
+    }
+  }, [isCreationSuccess, navigate]);
+
+  const getBlueprintPayload = async () => {
+    const userData = await auth?.getUser();
+    const orgId = userData?.identity?.internal?.org_id;
+    const requestBody = orgId && mapRequestFromState(orgId);
+    return requestBody;
+  };
+
+  const onSave = async () => {
+    const requestBody = await getBlueprintPayload();
+    setIsOpen(false);
+    !composeId &&
+      requestBody &&
+      createBlueprint({ createBlueprintRequest: requestBody });
+  };
+
+  const onSaveAndBuild = async () => {
+    const requestBody = await getBlueprintPayload();
+    setIsOpen(false);
+    const blueprint =
+      !composeId &&
+      requestBody &&
+      (await createBlueprint({ createBlueprintRequest: requestBody }).unwrap());
+    blueprint && buildBlueprint({ id: blueprint.id });
+    setIsOpen(false);
+  };
+
+  return (
+    <WizardFooterWrapper>
+      <Dropdown
+        isOpen={isOpen}
+        onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
+        toggle={(toggleRef: React.Ref<MenuToggleElement>) => (
+          <MenuToggle
+            variant="primary"
+            ref={toggleRef}
+            onClick={onToggleClick}
+            isExpanded={isOpen}
+            icon={(isBuildLoading || isCreationLoading) && <SpinnerIcon />}
+          >
+            Save
+          </MenuToggle>
+        )}
+        ouiaId="wizard-finish-dropdown"
+        shouldFocusToggleOnSelect
+      >
+        <DropdownList>
+          <DropdownItem onClick={onSave} ouiaId="wizard-save-btn">
+            Save changes
+          </DropdownItem>
+          <DropdownItem onClick={onSaveAndBuild} ouiaId="wizard-build-btn">
+            Save and build images
+          </DropdownItem>
+        </DropdownList>
+      </Dropdown>
+      <Button
+        ouiaId="wizard-back-btn"
+        variant="secondary"
+        onClick={goToPrevStep}
+      >
+        Back
+      </Button>
+      <Button ouiaId="wizard-cancel-btn" variant="link" onClick={close}>
+        Cancel
+      </Button>
+    </WizardFooterWrapper>
+  );
+};
+
+export default ReviewWizardFooter;
--- a/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
+++ b/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
@@ -1,0 +1,164 @@
+import { RootState, store } from '../../../store';
+import {
+  AwsUploadRequestOptions,
+  CreateBlueprintRequest,
+  Customizations,
+  GcpUploadRequestOptions,
+  ImageRequest,
+  ImageTypes,
+  Subscription,
+  UploadTypes,
+} from '../../../store/imageBuilderApi';
+import {
+  selectActivationKey,
+  selectArchitecture,
+  selectAwsAccountId,
+  selectAwsShareMethod,
+  selectAwsSource,
+  selectBaseUrl,
+  selectBlueprintDescription,
+  selectBlueprintName,
+  selectCustomRepositories,
+  selectDistribution,
+  selectGcpAccountType,
+  selectGcpEmail,
+  selectGcpShareMethod,
+  selectImageTypes,
+  selectRegistrationType,
+  selectServerUrl,
+} from '../../../store/wizardSlice';
+
+/**
+ * This function maps the wizard state to a valid CreateBlueprint request object
+ * @param {string} orgID organization ID
+ * @returns {CreateBlueprintRequest} blueprint creation request payload
+ */
+export const mapRequestFromState = (orgID: string): CreateBlueprintRequest => {
+  const state = store.getState();
+  const imageRequests = getImageRequests(state);
+  const customizations = getCustomizations(state, orgID);
+
+  return {
+    name: selectBlueprintName(state),
+    description: selectBlueprintDescription(state),
+    distribution: selectDistribution(state),
+    image_requests: imageRequests,
+    customizations,
+  };
+};
+
+const getImageRequests = (state: RootState): ImageRequest[] => {
+  const imageTypes = selectImageTypes(state);
+  return imageTypes.map((type) => ({
+    architecture: selectArchitecture(state),
+    image_type: type,
+    upload_request: {
+      type: uploadTypeByTargetEnv(type),
+      options: getImageOptions(type, state),
+    },
+  }));
+};
+
+const uploadTypeByTargetEnv = (imageType: ImageTypes): UploadTypes => {
+  switch (imageType) {
+    case 'aws':
+      return 'aws';
+    case 'gcp':
+      return 'gcp';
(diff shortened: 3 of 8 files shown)


amirfefer:
I created a custom `Footer` for the review step, following the mocks:

Screenshot 2024-02-01 at 17 52 44

src/store/enhancedImageBuilderApi.ts
```diff
@@ -62,7 +87,7 @@ const enhancedApi = imageBuilderApi.enhanceEndpoints({
           .then(() => {
             // Typescript is unaware of tag types being defined concurrently in enhanceEndpoints()
             // @ts-expect-error
-            dispatch(imageBuilderApi.util.invalidateTags(['Compose']));
+            dispatch(imageBuilderApi.util.invalidateTags(['Blueprints']));
```
ezr-ondrej, line 90: not sure, if we want this, but if we do, should we invalidate both of the tags?

src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
```diff
@@ -0,0 +1,92 @@
...
+  useWizardContext,
+} from '@patternfly/react-core';
+
+type FinishedFooterPropType = {
+  onSubmit?: () => void;
+  onBuild?: () => void;
+  isLoading?: boolean;
+};
+
+const FinishedWizardFooter = ({
```
ezr-ondrej, line 20: nitpick
```suggestion
const ReviewStepFooter = ({
```

ezr-ondrej (commented):
Nice, great work here :)

Few founds:
* ~It doesn't save the blueprint for me, with following error:~ EDIT: This is caused by the backend, so this is out of scope of this PR
   ```
    {
      "errors": [
          {
              "detail": "request body has an error: doesn't match schema #/components/schemas/CreateBlueprintRequest: Error at \"/image_requests/1/upload_request/type\": value \"image-installer\" is not one of the allowed values\nSchema:\n  {\n    \"enum\": [\n      \"aws\",\n      \"gcp\",\n      \"azure\",\n      \"aws.s3\",\n      \"oci.objectstorage\"\n    ],\n    \"type\": \"string\"\n  }\n\nValue:\n  \"image-installer\"\n",
              "title": "400"
          }
      ]
  }
  ```
* The image name was required to be lower-cased without spaces, we now impose it on Blueprints, but that's probably not what we want to do, it's fine for now tho, lets have that discussion outside of scope of this PR
* General thought to discuss with UX: Can we make the save button clickable by default (to save the blueprint and require double click only for save and build images)?

Few nitpicks in the code.

amirfefer:
Thanks @ezr-ondrej 👍 
> It doesn't save the blueprint for me, with following error:

What types of images you tried to add in the blueprint request?

ezr-ondrej:
> What types of images you tried to add in the blueprint request?

AWS and baremetal. Baremetal might have been the offender

src/store/enhancedImageBuilderApi.ts
amirfefer: great rebase issues catch! :D

amirfefer:
Thanks @ezr-ondrej! I addressed your comments :)
The wizard is still in progress feature, so some gaps would happen, the handler doesn't handle all cases.
But I found an issue with wrong type casting, which might caused your error

src/store/enhancedImageBuilderApi.ts
```diff
@@ -20,6 +20,31 @@ const enhancedApi = imageBuilderApi.enhanceEndpoints({
         return [{ type: 'Clone', id: arg.composeId }];
       },
     },
+    createBlueprint: {
+      onQueryStarted: async (_, { dispatch, queryFulfilled }) => {
+        queryFulfilled
+          .then(() => {
+            // Typescript is unaware of tag types being defined concurrently in enhanceEndpoints()
+            // @ts-expect-error
```
ezr-ondrej, line 28: linter in the iqe pipeline is saying this is unused (meaning ts would not error?)

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
```diff
@@ -952,6 +952,10 @@ describe('Step Upload to AWS', () => {
       // skip repositories
       await clickNext();
       // skip Details
+      const blueprintName = await screen.findByRole('textbox', {
+        name: /blueprint name/i,
+      });
+      await user.type(blueprintName, 'valid-name');
```
amirfefer, line 958: this block is added because blueprint name is required, a bit long test file :sweat_smile:  later can be refactored to be more dry.

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,150 @@
...
+ * @param {string} orgID organization ID
+ * @returns {CreateBlueprintRequest} creation blueprint request payload
+ */
+export const mapRequestFromState = (orgID: string): CreateBlueprintRequest => {
+  const { wizard: wizardState } = store.getState();
+  const images = getImageRequest(wizardState);
+  const customizations = getCustomizations(wizardState, orgID);
+
+  return {
+    name: wizardState.details.blueprintName,
```
lucasgarfield, line 25: For the name, the description, and distribution can we use a selector function instead? We will just need to define custom selectors in the wizardSlice for this. That way the shape of the wizardState is decoupled from the rest of the application. So something like:
```javascript
export const selectBlueprintName = (state: RootState) => {
  return state.details.blueprintName;
};
```

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,150 @@
...
+    description: wizardState.details.blueprintDescription,
+    distribution: wizardState.distribution,
+    image_requests: images,
+    customizations,
+  };
+};
+
+const getImageRequest = (state: wizardState): ImageRequest[] => {
```
lucasgarfield, line 33: Since this is an array can it be `getImageRequests` (plural)?

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,150 @@
...
+    default: {
+      // eslint-disable-next-line no-console
+      console.warn(`Unknown image type: ${imageType}`);
+      return 'aws';
+    }
+  }
+};
+const getImageOptions = (
+  imageType: string,
+  state: wizardState
```
lucasgarfield, line 72: Let's use selector functions for everything we access from the state here. There are already existing selector functions for a lot of these (if not all).

amirfefer:
@lucasgarfield, Thanks for your review!

Selectors should be used only  in the scope of  react components, for re-rendering optimizations.  I extracted the state-to-request mapper to be outside of a component scope, due to coupling reasons and re-use,  we don't  need re-rendering optimizations in submission handlers, it just add tones of boilerplate with out any benefits.

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,150 @@
...
+      return 'oci.objectstorage';
+    case 'wsl':
+      return 'aws.s3';
+    case 'image-installer':
+      return 'aws.s3';
+    case 'vsphere':
+      return 'aws.s3';
+    case 'ami':
+      return 'aws';
+    default: {
```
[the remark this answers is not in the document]
amirfefer: This might  happen. because edge images haven't implemented yet, returning a string is forced by the image builder API, I'd prefer to return undefined and raise an error.  The wizard is in progress work under experimental flag, there's a TODO for the missing `edge` types, I think throwing an error would be better than just a warning 👍

src/test/Components/CreateImageWizardV2/CreateImageWizard.test.tsx
```diff
@@ -884,7 +884,7 @@ describe('Step Upload to AWS', () => {
         name: /blueprint name/i,
       });
       // 64 character name
```
lucasgarfield, line 886: Can we just delete this silly comment? :sweat_smile:

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,150 @@
...
+  state: wizardState,
+  orgID: string
+): Customizations => {
+  return {
+    containers: undefined,
+    directories: undefined,
+    files: undefined,
+    subscription: getSubscription(state, orgID),
+    packages: undefined,
+    payload_repositories: undefined,
```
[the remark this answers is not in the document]
amirfefer: I guess it hasn't implemented yet? the wizardState has only
```js
 repositories: {
    customRepositories: CustomRepository[];
  };
```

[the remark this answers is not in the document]
lucasgarfield: Hm, you're right. Let's wait for now, sorry for the false alarm. :sweat_smile:

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -44,21 +48,37 @@ import { ImageBuilderHeader } from '../sharedComponents/ImageBuilderHeader';
...
+  isLoading?: boolean;
+  onSubmit?: () => void;
+  nextButtonText?: string;
 };
 
 export const CustomWizardFooter = ({
   disableNext: disableNext,
+  nextButtonText = 'Next',
+  isLoading,
+  onSubmit,
```
lucasgarfield, line 60: I might be missing something, but are any of these changes to `<CustomWizardFooter>` used in this pull request? It kind of looks like it might be from an early iteration before you extracted everything into `<ReviewWizardFooter>` but maybe I'm not seeing something...

lucasgarfield: And FWIW I think having a separate component `<ReviewWizardFooter>` is exactly what we should do, these footers are quite small and simple so it is better to just have separate components.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -212,7 +252,12 @@ const CreateImageWizard = () => {
           <WizardStep
             name="Review"
             id="step-review"
-            footer={<CustomWizardFooter disableNext={true} />}
+            footer={
+              <ReviewWizardFooter
+                isLoading={isCreationLoading}
+                onSubmit={onSubmit}
```
lucasgarfield, line 258: We could get rid of these two props if we just moved the `useMutation` hook into `<ReviewWizardFooter>` and also moved the `onSubmit` handler there as well. 

Generally I've tried to avoid passing down things like `isLoading` as props... since they are cached by RTKQ, it is cheap to just get them by calling the hook in the component where they are being used. The goal being to avoid having to search through the code to figure out exactly _what_ is loading. In this case if you are in the `<ReviewWizardFooter>` component you see that an `isLoading` prop gets passed in, but to figure out exactly _what_ is loading you have to go back to this component (which is already getting quite large and messy). If we just call the `useMutation` hook inside `<ReviewWizardFooter>` then we can more quickly understand where `isLoading` is coming from. Hope that makes sense...

src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const ReviewWizardFooter = ({
+  onBuild,
+  onSubmit,
+  isLoading,
+}: ReviewWizardFooter) => {
+  const { goToPrevStep, close } = useWizardContext();
+  const [isOpen, setIsOpen] = useState(false);
+  const SAVE = 'save';
+  const BUILD = 'build';
```
lucasgarfield, line 28: We could gain a bit of additional type safety by using a type here instead of constants I think:
```typescript
type submitType = 'save' | 'build'
```

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,164 @@
...
+  }
+};
+const getImageOptions = (
+  imageType: ImageTypes,
+  state: RootState
+): AwsUploadRequestOptions | GcpUploadRequestOptions => {
+  switch (imageType) {
+    case 'aws':
+      if (selectAwsShareMethod(state) === 'sources')
+        return { share_with_sources: [selectAwsSource(state)?.id || ''] };
```
lucasgarfield, line 93: I'm not sure about the `|| ''` here... in the state, `aws.source` can be either a `V1ListSourceResponseItem` or `undefined`. And the `id` field is optional on that type. So we do need to handle the case where there is not an id.

But I'm not sure passing an empty string is what we should do. After we build an AWS image, we upload it to an S3 bucket and share it with that AWS account via the id. If there is no id, then we won't be able to share the image, and it will be essentially inaccessible. 

So I think we probably ought to fail in that case. 

I need to think this through a bit more though.

@croissanne can I get a sanity check?

croissanne: yea for AWS if there's nothing to share it with, it's not accessible.

lucasgarfield: Thanks @croissanne!

We will need to ensure that we check that there is an ID on the AWS step (I created an issue for that #1637).  

There are other edge cases to consider related to this but ultimately, we should hopefully never get to this point... if we do, it's a bug. Maybe it would be easier to track down the source of the bug if we just throw an error here, something like:
```typescript
throw (
  new Error('AWS account ID could not be resolved from source')
);
```

What do you think @amirfefer ?

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
amirfefer: Yes, it was a leftover from early iteration, I kept the ouid tho.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
amirfefer: Moved to `<ReviewWizardFooter>`

amirfefer:
Thanks @lucasgarfield!
I added also the `Save and build` implementation to the handler :)

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
amirfefer: How about using `required` validation in the AWS step for the account number (if the sources radio is chosen)

Screenshot 2024-02-08 at 20 18 02

I guess a  backend validation is needed as well?

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,164 @@
...
+    fdo: undefined,
+    ignition: undefined,
+    partitioning_mode: undefined,
+    fips: undefined,
+  };
+};
+
+const getSubscription = (state: RootState, orgID: string): Subscription => {
+  const initialSubscription = {
+    'activation-key': selectActivationKey(state) || '',
```
lucasgarfield, line 151: Similar to the discussion around the AWS source ID, I don't think we want to use an empty string here because if we get to this point and there is no activation key... it is a bug. We should have validation on the activation keys step that prevents that from ever happening, so if it does, I think I would prefer that we just fail catastrophically rather than failing 'silently' by building an image that won't have a working subscription.

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -0,0 +1,164 @@
...
+  }
+  return {};
+};
+
+const getCustomizations = (state: RootState, orgID: string): Customizations => {
+  return {
+    containers: undefined,
+    directories: undefined,
+    files: undefined,
+    subscription: getSubscription(state, orgID),
```
lucasgarfield, line 127: We'll also need to handle the case where a user selects "Register later"... I think we would just keep the subscription as `undefined` in that case.

lucasgarfield:
I think this pull request is coming along nicely. I keep coming back with more and more comments but we support so many possible customizations and states, it is just really really tricky to put the request together. Thanks for bearing with me. :sweat_smile:

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
lucasgarfield: Yes, we'll definitely add that to the AWS step!

lucasgarfield: But we still need to make Typescript happy and I don't think using an empty string is the right way... does throwing an error sound good to you? Like I mentioned below, I'd rather fail catastrophically and immediately rather than build an image that the user won't be able to access later...

src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const onSaveAndBuild = async () => {
+    const requestBody = await getBlueprintPayload();
+    setIsOpen(false);
+    const blueprint =
+      !composeId &&
+      requestBody &&
+      (await createBlueprint({ createBlueprintRequest: requestBody }).unwrap());
+    blueprint && buildBlueprint({ id: blueprint.id });
+    setIsOpen(false);
```
lucasgarfield, line 70: This is also being set to false a few lines above, one of them is probably a duplicate.

ezr-ondrej:
/retest
it was supposed to be fixed! :'(

lucasgarfield (approved):
I'm going to merge this, great work @amirfefer. There are a few things I want to deal with in follow-up PRs but I don't think there's a good reason to delay this.

PR #1623, opened 2024-02-04
feat(HMS-3555) V2Wizard: fix ReleaseLifecycle component in the Review step

mgold1234:
this commit fix https://github.com/RedHatInsights/image-builder-frontend/issues/1615
add the relevant data to release lifeCycle at review step

Screenshot 2024-02-04 at 15 28 57

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
@@ -12,16 +12,22 @@ import {
   TextListItemVariants,
   TextVariants,
   Spinner,
+  FormGroup,
 } from '@patternfly/react-core';
 import { ExclamationTriangleIcon, HelpIcon } from '@patternfly/react-icons';
 import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
 
-import ReleaseLifecycle from './../ImageOutput/ReleaseLifecycle';
 import ActivationKeyInformation from './../Registration/ActivationKeyInformation';
 import { AwsAccountId } from './../TargetEnvironment/Aws/AwsAccountId';
 import { RepositoriesTable } from './ReviewStepTables';
 
-import { RELEASES } from '../../../../constants';
+import {
+  RELEASES,
+  RHEL_8,
+  RHEL_8_FULL_SUPPORT,
+  RHEL_8_MAINTENANCE_SUPPORT,
+  RHEL_9,
+} from '../../../../constants';
 import { extractProvisioningList } from '../../../../store/helpers';
 import { useAppSelector } from '../../../../store/hooks';
 import { useGetOscapCustomizationsQuery } from '../../../../store/imageBuilderApi';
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -57,7 +68,29 @@ export const ImageOutputList = () => {
   const arch = useAppSelector((state) => selectArchitecture(state));
   return (
     <TextContent>
-      <ReleaseLifecycle />
+      {distribution === RHEL_8 && (
+        <>
+          <Text className="pf-v5-u-font-size-sm">
```
lucasgarfield, line 73: I think we can apply DRY (Don’t Repeat Yourself) a bit to this PR so it is easier to maintain in the future. This code is shared with the `<ReleaseLifecycle>` component. Can you try to extract the common logic to a new component that this component and `<ReleaseLifecycle>` can share.

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -57,7 +68,29 @@ export const ImageOutputList = () => {
...
+            {RELEASES.get(distribution)} will be supported through{' '}
+            {toMonthAndYear(RHEL_8_FULL_SUPPORT[0])}, with optional ELS support
+            through {toMonthAndYear(RHEL_8_MAINTENANCE_SUPPORT[0])}. Consider
+            building an image with {RELEASES.get(RHEL_9)} to extend the support
+            period.
+          </Text>
+          <FormGroup label="Release lifecycle">
+            <Panel>
+              <PanelMain maxHeight="10rem">
+                <Bar
```
lucasgarfield, line 83: This code is shared with the `<MajorReleasesLifecyclesChart>` component and can also have DRY principles applied.

lucasgarfield (commented):
Let's DRY things out a bit. :smile:

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
mgold1234: the thing is that this part of the code  {RELEASES.get(release)} will be supported through{' '}
            {toMonthAndYear(RHEL_8_FULL_SUPPORT[0])}, with optional ELS support
            through {toMonthAndYear(RHEL_8_MAINTENANCE_SUPPORT[0])}. Consider
            building an image with {RELEASES.get(RHEL_9)} to extend the support
            period. exist in the old wizard only in case the  (currentStep.name === 'review') 
            how we can handle that with new component?

mgold1234: so for that piece of code I think this is the correct place. becuase its apeear only once in case we are at the review step

lucasgarfield: Oh I see, you're right. I noticed it was being used by the same component that is on the Image Output step last night but didn't notice there was a conditional causing it to only be displayed on the Review step. Sorry for the false alarm. :sweat_smile:

lucasgarfield (approved):
Looks great, thank you!

PR #1626, opened 2024-02-04
feat(HMS-3537) V2Wizard: Create a folder for Azure step and copy needed components

mgold1234:
this commit add the Azure step to the new wizrad

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -17,9 +17,13 @@ import RepositoriesStep from './steps/Repositories';
 import ReviewStep from './steps/Review';
 import ReviewWizardFooter from './steps/Review/Footer';
 import Aws from './steps/TargetEnvironment/Aws';
+import Azure from './steps/TargetEnvironment/Azure';
 import Gcp from './steps/TargetEnvironment/Gcp';
 import {
   isAwsAccountIdValid,
+  isAzureTenantGUIDValid,
+  isAzureSubscriptionIdValid,
+  isAzureResourceGroupValid,
   isBlueprintDescriptionValid,
   isBlueprintNameValid,
   isGcpEmailValid,
@@ -33,6 +37,11 @@ import {
   selectAwsAccountId,
   selectAwsShareMethod,
   selectAwsSource,
+  selectAzureResourceGroup,
+  selectAzureShareMethod,
+  selectAzureSource,
+  selectAzureSubscriptionId,
+  selectAzureTenantId,
   selectBlueprintDescription,
   selectBlueprintName,
   selectGcpEmail,
@@ -100,6 +109,18 @@ const CreateImageWizard = () => {
   // GCP
   const gcpShareMethod = useAppSelector((state) => selectGcpShareMethod(state));
   const gcpEmail = useAppSelector((state) => selectGcpEmail(state));
+  // AZURE
+  const azureShareMethod = useAppSelector((state) =>
+    selectAzureShareMethod(state)
+  );
+  const azureTenantId = useAppSelector((state) => selectAzureTenantId(state));
+  const azureSubscriptionId = useAppSelector((state) =>
+    selectAzureSubscriptionId(state)
+  );
+  const azureResourceGroup = useAppSelector((state) =>
+    selectAzureResourceGroup(state)
+  );
+  const azureSource = useAppSelector((state) => selectAzureSource(state));
 
   const registrationType = useAppSelector((state) =>
     selectRegistrationType(state)
@@ -168,6 +189,29 @@ const CreateImageWizard = () => {
                 isHidden={!targetEnvironments.includes('gcp')}
               >
                 <Gcp />
+              </WizardStep>,
+              <WizardStep
+                name="Azure"
+                id="wizard-target-azure"
+                key="wizard-target-azure"
+                footer={
+                  <CustomWizardFooter
+                    disableNext={
+                      azureShareMethod === 'manual'
+                        ? !isAzureTenantGUIDValid(azureTenantId) ||
+                          !isAzureSubscriptionIdValid(azureSubscriptionId) ||
+                          !isAzureResourceGroupValid(azureResourceGroup)
+                        : azureShareMethod === 'sources'
+                        ? !isAzureTenantGUIDValid(azureTenantId) ||
+                          !isAzureSubscriptionIdValid(azureSubscriptionId) ||
+                          !isAzureResourceGroupValid(azureResourceGroup)
+                        : azureSource === undefined
+                    }
+                  />
+                }
+                isHidden={!targetEnvironments.includes('azure')}
+              >
+                <Azure />
               </WizardStep>,
             ]}
           />
--- a/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
+++ b/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
@@ -1,0 +1,163 @@
+import React, { useState, useEffect } from 'react';
+
+import FormSpy from '@data-driven-forms/react-form-renderer/form-spy';
+import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
+import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
+import { Alert } from '@patternfly/react-core';
+import { FormGroup, Spinner } from '@patternfly/react-core';
+import {
+  Select,
+  SelectOption,
+  SelectVariant,
+} from '@patternfly/react-core/deprecated';
+import PropTypes from 'prop-types';
+
+import { extractProvisioningList } from '../../../store/helpers';
+import {
+  useGetSourceListQuery,
+  useGetSourceUploadInfoQuery,
+} from '../../../store/provisioningApi';
+
+const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
+  const { change } = useFormApi();
+  const { input } = useFieldApi(props);
+  const [isOpen, setIsOpen] = useState(false);
+  const selectedSourceId = input.value;
+
+  const {
+    data: rawSources,
+    isFetching,
+    isSuccess,
+    isError,
+    refetch,
+  } = useGetSourceListQuery({ provider: 'azure' });
+  const sources = extractProvisioningList(rawSources);
+
+  const {
+    data: sourceDetails,
+    isFetching: isFetchingDetails,
+    isSuccess: isSuccessDetails,
+    isError: isErrorDetails,
+  } = useGetSourceUploadInfoQuery(
+    { id: selectedSourceId },
+    {
+      skip: !selectedSourceId,
+    }
+  );
+
+  useEffect(() => {
+    if (isFetchingDetails || !isSuccessDetails) return;
+    change('azure-tenant-id', sourceDetails?.azure?.tenant_id);
+    change('azure-subscription-id', sourceDetails?.azure?.subscription_id);
+  }, [
+    isFetchingDetails,
+    isSuccessDetails,
+    sourceDetails?.azure?.subscription_id,
+    sourceDetails?.azure?.tenant_id,
+    change,
+  ]);
+
+  const onFormChange = ({ values }) => {
+    if (
+      values['azure-type'] !== 'azure-type-source' ||
+      values[input.name] === undefined
+    ) {
+      change(input.name, undefined);
+      change('azure-tenant-id', undefined);
+      change('azure-subscription-id', undefined);
+    }
+  };
+
+  const handleSelect = (_, sourceName) => {
+    const sourceId = sources.find((source) => source.name === sourceName).id;
+    change(input.name, sourceId);
+    setIsOpen(false);
+  };
+
+  const handleClear = () => {
+    change(input.name, undefined);
+  };
+
+  const handleToggle = () => {
+    // Refetch upon opening (but not upon closing)
+    if (!isOpen) {
+      refetch();
+    }
+
+    setIsOpen(!isOpen);
+  };
+
+  return (
+    <>
+      <FormSpy subscription={{ values: true }} onChange={onFormChange} />
+      <FormGroup
+        isRequired={isRequired}
+        label={label}
+        data-testid="azure-sources"
+      >
+        <Select
+          ouiaId="source_select"
+          variant={SelectVariant.typeahead}
+          className={className}
+          onToggle={handleToggle}
+          onSelect={handleSelect}
+          onClear={handleClear}
+          selections={
+            selectedSourceId
+              ? sources.find((source) => source.id === selectedSourceId)?.name
+              : undefined
+          }
+          isOpen={isOpen}
+          placeholderText="Select source"
+          typeAheadAriaLabel="Select source"
+          menuAppendTo="parent"
+          maxHeight="25rem"
+          isDisabled={!isSuccess}
+        >
+          {isSuccess &&
+            sources.map((source) => (
+              <SelectOption key={source.id} value={source.name} />
+            ))}
+          {isFetching && (
+            <SelectOption isNoResultsOption={true}>
+              <Spinner size="lg" />
+            </SelectOption>
+          )}
+        </Select>
+      </FormGroup>
+      <>
+        {isError && (
+          <Alert
+            variant={'danger'}
+            isPlain
+            isInline
+            title={'Sources unavailable'}
+          >
+            Sources cannot be reached, try again later or enter an account info
+            for upload manually.
+          </Alert>
+        )}
+        {!isError && isErrorDetails && (
+          <Alert
+            variant={'danger'}
+            isPlain
+            isInline
+            title={'Azure details unavailable'}
+          >
+            Could not fetch Tenant id and Subscription id from Azure for given
+            Source. Check Sources page for the source availability or select a
+            different Source.
+          </Alert>
+        )}
+      </>
+    </>
+  );
+};
+
+AzureSourcesSelect.propTypes = {
+  className: PropTypes.string,
+  label: PropTypes.node,
+  isRequired: PropTypes.bool,
+};
+
+export default AzureSourcesSelect;
--- a/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
+++ b/src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
@@ -1,8 +1,5 @@
 import React, { useState, useEffect } from 'react';
 
-import FormSpy from '@data-driven-forms/react-form-renderer/form-spy';
-import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
-import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
 import { Alert } from '@patternfly/react-core';
 import { FormGroup, Spinner } from '@patternfly/react-core';
 import {
@@ -10,19 +7,24 @@ import {
   SelectOption,
   SelectVariant,
 } from '@patternfly/react-core/deprecated';
-import PropTypes from 'prop-types';
 
-import { extractProvisioningList } from '../../../store/helpers';
+import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
 import {
   useGetSourceListQuery,
   useGetSourceUploadInfoQuery,
-} from '../../../store/provisioningApi';
+} from '../../../../../store/provisioningApi';
+import {
+  changeAzureResourceGroup,
+  changeAzureSource,
+  changeAzureSubscriptionId,
+  changeAzureTenantId,
+  selectAzureSource,
+} from '../../../../../store/wizardSlice';
 
-const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
-  const { change } = useFormApi();
-  const { input } = useFieldApi(props);
+export const AzureSourcesSelect = () => {
+  const azureSource = useAppSelector((state) => selectAzureSource(state));
+  const dispatch = useAppDispatch();
   const [isOpen, setIsOpen] = useState(false);
-  const selectedSourceId = input.value;
 
   const {
     data: rawSources,
@@ -31,7 +33,6 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
     isError,
     refetch,
   } = useGetSourceListQuery({ provider: 'azure' });
-  const sources = extractProvisioningList(rawSources);
 
   const {
     data: sourceDetails,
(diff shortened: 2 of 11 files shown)


lavocatt:
Thanks for your work on this. Can you also add the corresponding bits in the review step please?

image

lavocatt:
I found out a bug, if you select a group for a source, then you switch to another source, the group from the previous source stays. The group should instead be cleared out when you switch source.
[Screencast from 2024-02-20 10-54-31.webm](https://github.com/osbuild/image-builder-frontend/assets/86971992/f8a743b0-b2cc-4bb4-b6f9-9e4f63e92666)

mgold1234:
@lavocatt I think that the data at review step should be done in a different pr

mgold1234:
not sure, I leave the review task open for that specific reason

regexowl:
> > @lavocatt I think that the data at review step should be done in a different pr
> 
> Well now that the review step exists, my take is that we should update it piece by piece with every new addition to the V2Wizard, wdyt @regexowl & @lucasgarfield ?

I agree that building the review step as we go would be the best way to approach it. At the same time my plan was to tackle the review part of packages PR separately as well :box-of-shame:

So my take would be that both a commit in this PR and a separate PR would could probably work. The Review step will need an extra review either way when we finish up "wiring" all the other parts.

mgold1234:
I think that prs should be simple and short as we can, for those who reviews the prs.
I agreed with @lucasgarfield that it will be in a separate pr - https://github.com/osbuild/image-builder-frontend/pull/1674/commits/badd9b2566b4d2ae98e34e7688092fa31b6383d1

lucasgarfield:
Overall this seems to be working really well! 

Can you please fixup or squash the 3rd and 4th commits into the 2nd commit? It’s a bit difficult to review since so much changes between these three commits… I’ll take a deeper look at the code once that’s done.

For now though, I did notice a bug in the validation for `Resource Group`. It is a required field, but you can click “Next” without selecting a `Resource Group`. 

Regarding the review step, I’m fine with putting that in later personally, I agree we should keep these PRs smaller for easier reviews and this one is already fairly large.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -169,6 +190,25 @@ const CreateImageWizard = () => {
...
+                id="wizard-target-azure"
+                key="wizard-target-azure"
+                footer={
+                  <CustomWizardFooter
+                    disableNext={
+                      azureShareMethod === 'manual'
+                        ? !isAzureTenantGUIDValid(azureTenantId) ||
+                          !isAzureSubscriptionIdValid(azureSubscriptionId) ||
+                          !isAzureResourceGroupValid(azureResourceGroup)
+                        : azureSource === undefined
```
regexowl, line 204: When you get to the Azure step for the first time the Next button is enabled even without filling anything in. This should fix it :)

```suggestion
                        : azureSource === ''
```

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+            <GalleryItem>
+              <FormGroup label="Azure Tenant GUID">
+                <TextInput
+                  aria-label="Azure Tenant GUID"
+                  readOnlyVariant="default"
+                  isRequired
+                  id="tenant id"
+                  value={tenantId}
+                />
+              </FormGroup>
```
regexowl, line 143: Azure Tenant GUID wasn't marked as required next to its label. We'll just have to move the `isRequired` from the TextInput to the FormGroup

image


```suggestion
              <FormGroup label="Azure Tenant GUID" isRequired>
                <TextInput
                  aria-label="Azure Tenant GUID"
                  readOnlyVariant="default"
                  id="tenant id"
                  value={tenantId}
                />
              </FormGroup>
```

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+              <FormGroup label="Subscription ID">
+                <TextInput
+                  aria-label="Subscription ID"
+                  label="Subscription ID"
+                  readOnlyVariant="default"
+                  isRequired
+                  id="subscription id"
+                  value={subscriptionId}
+                />
+              </FormGroup>
```
regexowl, line 155: Same fix for the Subscription ID - this should add the isRequired little star next to the FormGroup label

```suggestion
              <FormGroup label="Subscription ID" isRequired>
                <TextInput
                  aria-label="Subscription ID"
                  label="Subscription ID"
                  readOnlyVariant="default"
                  id="subscription id"
                  value={subscriptionId}
                />
              </FormGroup>
```

src/store/wizardSlice.ts
```diff
@@ -40,6 +42,13 @@ type wizardState = {
     shareMethod: AwsShareMethod;
     source: V1ListSourceResponseItem | undefined;
   };
+  azure: {
+    tenantId: string;
+    shareMethod: AzureShareMethod;
+    subscriptionId: string;
+    source: AzureUploadRequestOptions['source_id'] | undefined;
+    resourceGroup: AzureUploadRequestOptions['resource_group'];
+  };
```
regexowl, line 51: This is very nit picky, but can we please move the shareMethod either to the beginning or to the end of the list? So it's not mixed up with the Azure account info? :sweat_smile: 

```suggestion
  azure: {
    shareMethod: AzureShareMethod;
    tenantId: string;
    subscriptionId: string;
    source: AzureUploadRequestOptions['source_id'] | undefined;
    resourceGroup: AzureUploadRequestOptions['resource_group'];
  };
```

src/store/wizardSlice.ts
```diff
@@ -82,6 +91,13 @@ const initialState: wizardState = {
     shareMethod: 'sources',
     source: undefined,
   },
+  azure: {
+    tenantId: '',
+    shareMethod: 'sources',
+    subscriptionId: '',
+    source: '',
+    resourceGroup: '',
+  },
```
regexowl, line 100: Same here, I think it looks neater in dev tools :sweat_smile: 

```suggestion
  azure: {
    shareMethod: 'sources',
    tenantId: '',
    subscriptionId: '',
    source: '',
    resourceGroup: '',
  },
```

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
mgold1234: https://github.com/osbuild/image-builder-frontend/assets/73419853/e96c2e0c-37be-49e5-b83e-f5727dfe45ff

I see that its disble

regexowl: It is now! My bad :sweat_smile:

mgold1234: no worry :)

src/test/Components/CreateImageWizardV2/CreateImageWizard.azure.test.js
```diff
@@ -0,0 +1,285 @@
...
+  //   );
+
+  //   await setUp();
+
+  //  await screen.findByText(
+  //   /Sources cannot be reached, try again later or enter an account info for upload manually\./i
+  // );
+  //});
+
+  // set test timeout to 15 seconds
```
regexowl, line 284: I wonder why this test doesn't pass. The logic should be implemented, right?

regexowl:
This is looking really nice, thank you!! :heart_eyes_cat: 

Added a few comments. Could you please fix the typo in the name of the second commit:  necussery -> necessary?

It seems the last pushed were added to the test commit: https://github.com/osbuild/image-builder-frontend/pull/1626/commits/a40b07b37c355079e433e04c5689253c19681996 is it possible to keep only test changes in that one?

src/test/Components/CreateImageWizardV2/CreateImageWizard.azure.test.js
regexowl: Looked into it a bit - the problem lies in closing up `isError` inside of a `isSuccess` if in `AzureSourcesSelect.tsx`

You can test if the error gets rendered by breaking a handler and running msw - in `handlers.js` change
```typescript
  rest.get(`${PROVISIONING_API}/sources`, (req, res, ctx) => {
    const provider = req.url.searchParams.get('provider');
    return res(ctx.status(200), ctx.json(mockSourcesByProvider(provider)));
  }),
```
to
```typescript
  rest.get(`${PROVISIONING_API}/souces`, (req, res, ctx) => {
    const provider = req.url.searchParams.get('provider');
    return res(ctx.status(200), ctx.json(mockSourcesByProvider(provider)));
  }),
```
or something similar (just changing the url to a non-existent one will do the trick).

Then in msw browser the error should get rendered.

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -90,12 +95,21 @@ const uploadTypeByTargetEnv = (imageType: ImageTypes): UploadTypes => {
...
   imageType: ImageTypes,
   state: RootState
-): AwsUploadRequestOptions | GcpUploadRequestOptions => {
+): AwsUploadRequestOptions | AzureUploadRequestOptions | GcpUploadRequestOptions => {
   switch (imageType) {
     case 'aws':
       if (selectAwsShareMethod(state) === 'sources')
         return { share_with_sources: [selectAwsSource(state)?.id || ''] };
       else return { share_with_accounts: [selectAwsAccountId(state)] };
+    case 'azure':
```
regexowl, line 104: Looking at the original `onSave` function in `CreateImageWizard/CreateImageWizard.js` it seems that on "sources" the returned values are only `source_id` and `resource_group`. Also source should not be an empty string as it's a required field when creating an image with sources.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+import {
+  isAzureResourceGroupValid,
+  isAzureSubscriptionIdValid,
+  isAzureTenantGUIDValid,
+} from '../../../validators';
+
+export type AzureShareMethod = 'manual' | 'sources';
+
+// The Sources API only defines a V1ListSourceResponseItem[] type
+export type V1ListSourceResponseItem = {
```
lucasgarfield, line 42: This is duplicate code, it's also present in `index.tsx` for the Aws step. Can we extract it from here and the AWS step into a new file? Perhaps `.../CreateImaveWizardV2/types.ts`?

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
```diff
@@ -31,51 +39,45 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
...
-  const sources = extractProvisioningList(rawSources);
 
   const {
     data: sourceDetails,
     isFetching: isFetchingDetails,
     isSuccess: isSuccessDetails,
     isError: isErrorDetails,
   } = useGetSourceUploadInfoQuery(
-    { id: selectedSourceId },
+    // @ts-ignore
```
lucasgarfield, line 49: We shouldn't be using @ts-ignore without a very good reason. In this case, there is a type difference between `id` (which must be a number) and `azureSource` (which is string or undefined).

We should expect azureSource to be castable to number... what if we wrote a helper function, something like `convertAzureSourceIdToNumber()`, that attempts to cast azureSource to Number in a try...catch clause and throws a type error if it fails?

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
```diff
@@ -1,28 +1,36 @@
...
+  selectAzureSource,
+  selectAzureSubscriptionId,
+  selectAzureTenantId,
+} from '../../../../../store/wizardSlice';
 
-const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
-  const { change } = useFormApi();
-  const { input } = useFieldApi(props);
+export const AzureSourcesSelect = () => {
+  let tenantId = useAppSelector((state) => selectAzureTenantId(state));
```
lucasgarfield, line 27: This should be defined using `const`, not `let`. The `useAppSelector` hook creates a subscription - see the docs:

https://react-redux.js.org/api/hooks
>useSelector() will also subscribe to the Redux store, and run your selector whenever an action is dispatched.

If we re-assign the value later as done here, we lose the benefit of having a subscription.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
```diff
@@ -31,51 +39,45 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
...
   } = useGetSourceUploadInfoQuery(
-    { id: selectedSourceId },
+    // @ts-ignore
+    { id: azureSource },
     {
-      skip: !selectedSourceId,
+      skip: !azureSource,
     }
   );
+  tenantId = sourceDetails?.azure?.tenant_id || '';
```
lucasgarfield, line 55: This causes us to lose the subscription created earlier when `tenantId` was initialized using the `useAppSelector` hook. Additionally, assigning a value to `tenantId` like this does not result in an update to the value in the store. Instead, we should be using `useAppDispatch` to update the `tenantId` value in the Redux store.

src/store/wizardSlice.ts
```diff
@@ -40,6 +42,13 @@ type wizardState = {
     shareMethod: AwsShareMethod;
     source: V1ListSourceResponseItem | undefined;
   };
+  azure: {
+    shareMethod: AzureShareMethod;
+    tenantId: string;
+    subscriptionId: string;
+    source: AzureUploadRequestOptions['source_id'] | undefined;
```
lucasgarfield, line 50: Let’s change the type here to just be `string`. 

If you actually look at the type definition for `AzureUploadRequestOptions['source_id']` it is `string | undefined`. So what is written here is actually redundant (as undefined appears twice if you expand out the union) and also, we don’t want this to ever be `undefined` since we will just use `’’` as the empty state.

src/store/wizardSlice.ts
```diff
@@ -40,6 +42,13 @@ type wizardState = {
...
+    resourceGroup: AzureUploadRequestOptions['resource_group'];
```
lucasgarfield, line 51: In this case, if you look at the type of `AzureUploadRequestOptions['source_id']`, it is just `string`. So this would theoretically be okay, but for consistency sake with what I’m suggesting above let’s change this type to just be `string`.

mgold1234: just wander why the source in aws define as aws: {
    source: V1ListSourceResponseItem | undefined;
  };

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -6,6 +6,35 @@ export const isAwsAccountIdValid = (awsAccountId: string | undefined) => {
   );
 };
 
+export const isAzureTenantGUIDValid = (azureTenantGUID: string | undefined) => {
```
lucasgarfield, line 9: Would this ever be undefined? I think it should only ever be a string.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -6,6 +6,35 @@ export const isAwsAccountIdValid = (awsAccountId: string | undefined) => {
...
+  return (
+    azureTenantGUID !== undefined &&
+    /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(
+      azureTenantGUID
+    )
+  );
+};
+
+export const isAzureSubscriptionIdValid = (
+  azureSubscriptionId: string | undefined
```
lucasgarfield, line 19: Would this ever be undefined? I think it should only ever be a string.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -6,6 +6,35 @@ export const isAwsAccountIdValid = (awsAccountId: string | undefined) => {
...
+  return (
+    azureSubscriptionId !== undefined &&
+    /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(
+      azureSubscriptionId
+    )
+  );
+};
+
+export const isAzureResourceGroupValid = (
+  azureResourceGroup: string | undefined
```
lucasgarfield, line 30: Would this ever be undefined? I think it should only ever be a string.

src/store/wizardSlice.ts
lucasgarfield: Oh, that's a good point, I forgot about that. You are right, for AWS we store the entire `V1ListSourceResponseItem` object, not just the id. 

Let's do the same thing here in order to be consistent. So in that case, the type here would not be `string` like I suggested, it would be `V1ListSourceResponseItem | undefined` as it is for AWS.

mgold1234: not sure how you want to handel that for Azure, define only the source or also the resourceGroup?
if I define only the source, I think Iits require more changes in other places as well

mgold1234: I think its more clear to define source and resourceGroup as well

mgold1234: I tried to change the source to source: V1ListSourceResponseItem | undefined;
but its seems that there are some issues. not sure if that is the correct way for azure

lucasgarfield: It works, so we can use this for now and worry about the small inconsistency with the AWS step later.

src/store/wizardSlice.ts
```diff
@@ -40,6 +39,13 @@ type wizardState = {
     shareMethod: AwsShareMethod;
     source: V1ListSourceResponseItem | undefined;
   };
+  azure: {
+    shareMethod: AzureShareMethod;
+    tenantId: string;
+    subscriptionId: string;
+    source: string | undefined;
```
lucasgarfield, line 46: If we are using `''` as the empty state for the source (which I think we should, it is consistent with Patternfly examples for typeahead selects) then the type here should only be `string`.

lucasgarfield (changes requested):

src/store/wizardSlice.ts
```diff
@@ -144,6 +157,26 @@ export const selectAwsShareMethod = (state: RootState) => {
   return state.wizard.aws.shareMethod;
 };
 
+export const selectAzureTenantId = (state: RootState): string => {
```
lucasgarfield, line 160: I don't think we need to specify a return type here, Typescript should be able to infer it. We don't specify return types for any of our other selectors so I would prefer to stay consistent.

src/store/wizardSlice.ts
```diff
@@ -144,6 +157,26 @@ export const selectAwsShareMethod = (state: RootState) => {
...
+
+export const selectAzureShareMethod = (state: RootState) => {
+  return state.wizard.azure.shareMethod;
+};
+
+export const selectAzureSubscriptionId = (state: RootState) => {
+  return state.wizard.azure.subscriptionId;
+};
+
+export const selectAzureSource = (state: RootState): string | undefined => {
```
lucasgarfield, line 172: I don't think we need to specify a return type here, Typescript should be able to infer it. We don't specify return types for any of our other selectors so I would prefer to stay consistent.

src/store/wizardSlice.ts
```diff
@@ -144,6 +157,26 @@ export const selectAwsShareMethod = (state: RootState) => {
...
+  return state.wizard.azure.source;
+};
+
+export const selectAzureResourceGroup = (state: RootState): string => {
```
lucasgarfield, line 176: I don't think we need to specify a return type here, Typescript should be able to infer it. We don't specify return types for any of our other selectors so I would prefer to stay consistent.

lucasgarfield (changes requested):

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+      </Text>
+      <FormGroup label="Share method:">
+        <Radio
+          id="radio-with-description"
+          label="Use an account configured from Sources."
+          name="radio-7"
+          description="Use a configured sources to launch environments directly from the console."
+          isChecked={shareMethod === 'sources'}
+          onChange={() => {
+            dispatch(changeAzureSource(undefined));
```
lucasgarfield, line 103: We'll need to make this `''` for the empty state if we change it in the wizardSlice.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+            dispatch(changeAzureResourceGroup(''));
+          }}
+        />
+        <Radio
+          id="radio"
+          label="Manually enter the account information."
+          name="radio-8"
+          isChecked={shareMethod === 'manual'}
+          onChange={() => {
+            dispatch(changeAzureSource(undefined));
```
lucasgarfield, line 116: We'll need to make this `''` for the empty state if we change it in the wizardSlice.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
```diff
@@ -31,51 +33,54 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
...
-    const sourceId = sources.find((source) => source.name === sourceName).id;
-    change(input.name, sourceId);
+  const handleSelect = (
+    _event: React.MouseEvent<Element, MouseEvent>,
+    sourceName: string
+  ) => {
+    const sourceId = rawSources?.data?.find(
+      (source) => source?.name === sourceName
+    )?.id;
+    dispatch(changeAzureSource(sourceId));
```
lucasgarfield, line 70: I suppose we would need to add `|| ''` here if we change the type of `sourceId` to be `string` only in wizardSlice.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
```diff
@@ -31,51 +33,54 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
...
+    dispatch(changeAzureTenantId(sourceDetails?.azure?.tenant_id || ''));
+    dispatch(
+      changeAzureSubscriptionId(sourceDetails?.azure?.subscription_id || '')
+    );
+    dispatch(changeAzureResourceGroup(''));
```
lucasgarfield, line 75: If you look at the way this code worked previously, we only called `change` on what is now called the `azureSource`. The updates to the `azureSubscriptionId` and `azureTenantId` occur in the `useEffect` hook above this.

Here we effectively end up dispatching changeAzureTenantId and changeAzureSubscriptionId twice. I think we can just remove these two calls to dispatch from the handler.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
```diff
@@ -31,51 +33,54 @@ const AzureSourcesSelect = ({ label, isRequired, className, ...props }) => {
...
     setIsOpen(false);
   };
 
   const handleClear = () => {
-    change(input.name, undefined);
+    dispatch(changeAzureSource(undefined));
```
lucasgarfield, line 80: Let's change this to `''`.

src/Components/CreateImageWizardV2/steps/TargetEnvironment/Azure/AzureSourcesSelect.tsx
mgold1234: right +

mgold1234: I wander why we acctually need it at handleSelect and change it to ''? 
when we clean source name - > resource group also clean   const handleClear = () => {
    dispatch(changeAzureSource(''));
    dispatch(changeAzureTenantId(''));
    dispatch(changeAzureSubscriptionId(''));
    dispatch(changeAzureResourceGroup(''));
    and when we change the source name resource group doesnt change. 
    this is the beavior at stage. is it wrong?

lucasgarfield: The resource groups are linked to an individual source, so if the source changes then the resource group field should be cleared.

lucasgarfield (approved):
Overall this is great, thank you @mgold1234. There's still one minor issue but I want to start clearing out our PR review backlog so I created an issue to track it, #1691, and we can address in a follow-up PR.

PR #1631, opened 2024-02-05
feat(HMS-3515): Change blueprint cards to be clickable

avitova:
I am marking this as draft for now. There are some things that might not have been in the scope of this PR or that I am not so sure of.

1. I have changed the "Show all" button into a card, as I think that Patternfly [deselectable cards](https://www.patternfly.org/components/card/#selectable-cards-deprecated) is deprecated. I tried to make the cards deselectable but did not succeed. I thought that as a solution, we might make the Card component from the "Show all" button. Choosing this card would deselect an already chosen card. Let me know if you think that my assumption about the deprecated deselectable cards is invalid, and if we can keep the button.
2. Issue 3515 states the following: _Move blueprint's kebab dropdown to the right sidebar toolbar, should be enabled only when a blueprint is chosen_. The toolbar is not shown yet when no images are being present for the chosen blueprint. Disabling the kebab works, but in the current design, we would not be able to Edit/Delete any blueprint with no images. That sounds a bit clunky, so we might want to postpone this PR.
<details>
right-toolbar-kebab
</details>
So let's have a conversation about both 1. and 2.

### DIFF
--- a/src/Components/sharedComponents/ImageBuilderHeader.tsx
+++ b/src/Components/sharedComponents/ImageBuilderHeader.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import {
   Button,
@@ -7,6 +7,11 @@ import {
   TextContent,
   Flex,
   FlexItem,
+  Dropdown,
+  DropdownList,
+  MenuToggle,
+  MenuToggleElement,
+  DropdownItem,
 } from '@patternfly/react-core';
 import { ExternalLinkAltIcon, HelpIcon } from '@patternfly/react-icons';
 // eslint-disable-next-line rulesdir/disallow-fec-relative-imports
@@ -17,9 +22,13 @@ import {
 } from '@redhat-cloud-services/frontend-components';
 import { Link } from 'react-router-dom';
 
-import { useComposeBlueprintMutation } from '../../store/imageBuilderApi';
+import {
+  useComposeBlueprintMutation,
+  useDeleteBlueprintMutation,
+} from '../../store/imageBuilderApi';
 import { resolveRelPath } from '../../Utilities/path';
 import './ImageBuilderHeader.scss';
+import { DeleteBlueprintModal } from '../Blueprints/DeleteBlueprintModal';
 
 type ImageBuilderHeaderPropTypes = {
   experimentalFlag?: string | true | undefined;
@@ -36,10 +45,34 @@ export const ImageBuilderHeader = ({
   const onBuildHandler = async () => {
     selectedBlueprint && (await buildBlueprint({ id: selectedBlueprint }));
   };
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const onSelect = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const [showDeleteModal, setShowDeleteModal] = React.useState(false);
+  const [deleteBlueprint] = useDeleteBlueprintMutation({
+    fixedCacheKey: 'delete-blueprint',
+  });
+  const handleDelete = async () => {
+    if (selectedBlueprint) {
+      setShowDeleteModal(false);
+      await deleteBlueprint({ id: selectedBlueprint });
+    }
+  };
+  const onDeleteClose = () => {
+    setShowDeleteModal(false);
+  };
 
   return (
     <>
       {/*@ts-ignore*/}
+      <DeleteBlueprintModal
+        onDelete={handleDelete}
+        selectedBlueprint={selectedBlueprint}
+        isOpen={showDeleteModal}
+        onClose={onDeleteClose}
+      />
       <PageHeader>
         <Flex>
           <FlexItem>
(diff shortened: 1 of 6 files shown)


lucasgarfield:
I think this looks really nice, actually. Nice work!

I suppose we are waiting on feedback from UX at this point?

If we keep the "Clear selection" button, could we make it only appear after a blueprint has been selected?

Another thought... since we can only select one blueprint, would it maybe make sense that clicking it once selects it and clicking it again deselects it?

Quick gif showing current state:
clickable blueprints

src/Components/sharedComponents/ImageBuilderHeader.tsx
```diff
@@ -112,6 +141,34 @@ export const ImageBuilderHeader = ({
                 >
                   Build images
```
amirfefer, line 142: According to the new mocks this should be moved to the images table's toolbar, though not in this PR scope

src/Components/sharedComponents/ImageBuilderHeader.tsx
```diff
@@ -34,10 +41,32 @@ export const ImageBuilderHeader = ({
...
+  const onDeleteClose = () => {
+    setShowDeleteModal(false);
+  };
 
   return (
     <>
       {/*@ts-ignore*/}
+      <DeleteBlueprintModal
+        onDelete={handleDelete}
+        blueprintName={''}
```
amirfefer, line 66: I guess the name is missing because the component has the blueprint id. As we discussed offline, we can get the blueprint's name by creating a selector to select a blueprint's name by its id.

amirfefer (commented):
Thanks @avitova ! I Tested it and it looks great! 👍 
I guess the only thing is missing is the blueprint's name in the delete modal
Plus, tests failures  related.

avitova:
<details>
  <summary>Current state of the PR</summary>
[clickable-card-v3.webm](https://github.com/osbuild/image-builder-frontend/assets/87140865/2825372f-1a62-4648-b7ed-da3da13347ad)
</details>

Rebased. This is the version addressing new mockups. :)

avitova:
Fixed tests in [5fc1f32](https://github.com/osbuild/image-builder-frontend/commit/5fc1f32338fdab2f0b47a7f65b211502ef00ebd2) and rebased in [397fa65](https://github.com/osbuild/image-builder-frontend/commit/397fa650a9c1237208db233c2f2bcb518ee2a141). :)

src/test/Components/Blueprints/Blueprints.test.js
```diff
@@ -50,37 +52,51 @@ describe('Blueprints', () => {
...
-    const blueprintRadioBtn = await screen.findByRole('radio', {
+    const radioButtons = await screen.findAllByRole('radio', {
       name: nameMatcher,
     });
-    await user.click(blueprintRadioBtn);
+    const elementById = radioButtons.find(
+      (button) => button.getAttribute('id') === idMatcher
+    );
+
+    await user.click(elementById);
```
amirfefer, line 99: I was thinking that once we changed the radio buttons to clickable card, it would be easier to access the elements, though, it seems that PF still renders radio buttons. 🤔

amirfefer (approved):
Thanks @avitova ! Looks great 👍 
Tested and works as expected

lucasgarfield:
One thing that worries me a bit about this is that it deviates fairly far from established Patternfly patterns. I like to follow those because it helps ensure a consistent experience across applications on Insights and helps us reduce the amount of decision making we need to do.

I realize we created the "Bueprint actions" button on the upper right because we had created clickable cards and that was interfering with the kebab menu, but there is an established Patternfly pattern for selectable cards with a kebab menu:

https://www.patternfly.org/patterns/card-view/react-demos/card-view/
https://www.patternfly.org/patterns/card-view/design-guidelines#inline-actions

I personally prefer this from a UX perspective as well - I feel it's more discoverable and is less "work" - the blueprint related actions (edit/delete) are co-located with the blueprint which should mean less thinking and less mouse movement is required to find and to use them.

I also think that the "Clear selection" button looks very nice, but again I worry that it doesn't follow established Patternfly conventions and will result in our app being "unique".

Since we only allow selecting one blueprint at a time, I think we could use radio selectors instead of checkboxes and then have a chip group called "Selected blueprint" appear below the search bar once a blueprint is selected.

https://www.patternfly.org/2022.05/components/chip-group

lucasgarfield (dismissed):
Adding a "Request changes" review so this doesn't accidentally get merged.

kelseamann:
I just wanted to weigh in on the UI- confirmed with Margot that this is the clickable card we should be using to be most current. She also said that a clear selection button is a good pattern because sometimes with selectable cards, there is no "all" view and there is always a card selected; I know there is an app with this pattern and I'll try to dig it up. 

For now, this is correct - https://www.patternfly.org/components/card/#

lucasgarfield (approved):
I'm going to merge this so we have the delete blueprint and filter clearing functionality. We'll need to follow-up with another PR to bring us into alignment with the most recent mocks. We have a huge PR review backlog so I don't want to block on the Patternfly bug, I created an issue to track their progress for now, #1687.

avitova:
Thank you!:)

PR #1634, opened 2024-02-06
feat(HMS-3553) V2Wizard: fix source name at review step when target is AWS

mgold1234:
this commit fix source name and shared to account at review step, when user choose to use an account configured from Sources.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
@@ -18,7 +18,6 @@ import { ExclamationTriangleIcon, HelpIcon } from '@patternfly/react-icons';
 import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
 
 import ActivationKeyInformation from './../Registration/ActivationKeyInformation';
-import { AwsAccountId } from './../TargetEnvironment/Aws/AwsAccountId';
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -107,15 +108,21 @@ export const FSCList = () => {
...
-  const awsSources = extractProvisioningList(rawAWSSources);
-  const { isBeta } = useGetEnvironment();
+  const { data } = useGetSourceUploadInfoQuery(
+    {
+      id: parseInt(source?.id as string),
+    },
+    { skip: source === undefined }
+  );
 
+  const getAccountId = data?.aws?.account_id;
```
lucasgarfield, line 125: This is all a bit confusing, and it is entirely my fault. :sweat_smile: 

First, a quick side note... I don't like this variable name. When I see the `get` prefix, I assume this is a getter function... but it is actually a string, not a function.

But the more confusing thing here is that we can't just use the value of `awsAccountId` because in the case that we're using a source, its value is `''`. 

I think that is what we should actually fix.

Would you mind updating this so that on the AWS step we set the value of `awsAccountId` when we are using sources? Then we won't need to call the `useGetSourceUploadInfoQuery` hook at all, we can just use `awsAccountId` for both cases.

lucasgarfield (changes requested):

PR #1640, opened 2024-02-09
V2Wizard: Packages step

regexowl:
This is an initial Packages step PR. Two PRs will follow up:
1. OpenSCAP initialization of the packages included in a profile + onRecreate packages as both are presented as a flat list only, those will need to be rendered in a different way from the package lists fetched from the API
2. Proactive Assistance repositories recommendations

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -12,6 +12,7 @@ import { useNavigate } from 'react-router-dom';
 import DetailsStep from './steps/Details';
 import ImageOutputStep from './steps/ImageOutput';
 import OscapStep from './steps/Oscap';
+import PackagesStep from './steps/Packages';
 import RegistrationStep from './steps/Registration';
 import RepositoriesStep from './steps/Repositories';
 import ReviewStep from './steps/Review';
@@ -246,6 +247,14 @@ const CreateImageWizard = () => {
                 footer={<CustomWizardFooter disableNext={false} />}
               >
                 <RepositoriesStep />
+              </WizardStep>,
+              <WizardStep
+                name="Additional packages"
+                id="wizard-additional-packages"
+                key="wizard-additional-packages"
+                footer={<CustomWizardFooter disableNext={false} />}
+              >
+                <PackagesStep />
               </WizardStep>,
             ]}
           ></WizardStep>
--- a/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
@@ -1,0 +1,545 @@
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
+
+import useFormApi from '@data-driven-forms/react-form-renderer/use-form-api';
+import WizardContext from '@data-driven-forms/react-form-renderer/wizard-context';
+import {
+  Alert,
+  Divider,
+  DualListSelector,
+  DualListSelectorControl,
+  DualListSelectorControlsWrapper,
+  DualListSelectorList,
+  DualListSelectorListItem,
+  DualListSelectorPane,
+  SearchInput,
+  TextContent,
+} from '@patternfly/react-core';
+import {
+  AngleDoubleLeftIcon,
+  AngleDoubleRightIcon,
+  AngleLeftIcon,
+  AngleRightIcon,
+} from '@patternfly/react-icons';
+import PropTypes from 'prop-types';
+
+import api from '../../../api';
+import {
+  useGetArchitecturesQuery,
+  useGetOscapCustomizationsQuery,
+} from '../../../store/imageBuilderApi';
+
+const ExactMatch = ({
+  pkgList,
+  search,
+  chosenPackages,
+  selectedAvailablePackages,
+  handleSelectAvailableFunc,
+}) => {
+  const match = pkgList.find((pkg) => pkg.name === search);
+  return (
+    <DualListSelectorListItem
+      data-testid={`exact-match-${match.name}`}
+      isDisabled={chosenPackages[match.name] ? true : false}
+      isSelected={selectedAvailablePackages.has(match.name)}
+      onOptionSelect={(e) => handleSelectAvailableFunc(e, match.name)}
+    >
+      <TextContent key={`${match.name}`}>
+        <small className="pf-u-mb-sm">Exact match</small>
+        <span className="pf-c-dual-list-selector__item-text">{match.name}</span>
+        <small>{match.summary}</small>
+        <Divider />
+      </TextContent>
+    </DualListSelectorListItem>
+  );
+};
+
+export const RedHatPackages = ({ defaultArch }) => {
+  const { getState } = useFormApi();
+  const distribution = getState()?.values?.release;
+  const arch = getState()?.values?.arch;
+  const { data: distributionInformation, isSuccess: isSuccessDistroInfo } =
+    useGetArchitecturesQuery({ distribution });
+
+  const getAllPackages = async (packagesSearchName) => {
+    // if the env is stage beta then use content-sources api
+    // else use image-builder api
+    if (getState()?.values?.contentSourcesEnabled) {
+      const filteredByArch = distributionInformation.find(
+        (info) => info.arch === arch
+      );
+      const repoUrls = filteredByArch.repositories.map((repo) => repo.baseurl);
+      return await api.getPackagesContentSources(repoUrls, packagesSearchName);
+    } else {
+      const args = [
+        getState()?.values?.release,
+        getState()?.values?.architecture || defaultArch,
+        packagesSearchName,
+      ];
+      const response = await api.getPackages(...args);
+      let { data } = response;
+      const { meta } = response;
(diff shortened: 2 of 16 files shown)


src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -193,6 +194,14 @@ const CreateImageWizard = () => {
               >
                 <RepositoriesStep />
               </WizardStep>,
+              <WizardStep
+                name="Additional packages"
```
mgold1234, line 198: should it be 'Additional Red Hat packages'?

src/Components/CreateImageWizardV2/steps/Packages/index.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import { Text, Form, Title } from '@patternfly/react-core';
+
+const PackagesStep = () => {
+  return (
+    <Form>
+      <Title headingLevel="h2">Additional packages</Title>
```
mgold1234, line 10: same

mgold1234:
I also use UsrSubDirectoriesDisbalnled file, and put it under the CreateImageWizardV2 folder, maye this should be the correct place?

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
regexowl: Not in this case. This step is a bit different from all the other steps as we're merging Red Hat packages and custom packages into one step called "Additional packages" and also migrating from the `DualListSelector` to a table.

src/Components/CreateImageWizardV2/steps/Packages/index.tsx
regexowl: Answered here https://github.com/osbuild/image-builder-frontend/pull/1640#discussion_r1494424651, the "Additional packages" heading is correct in this case :)

regexowl:
> I also use UsrSubDirectoriesDisbalnled file, and put it under the CreateImageWizardV2 folder, maye this should be the correct place?

You're right it shouldn't be in the Packages step folder! I don't think the component is even used on the Packages step. Will check it out and update the PR. Thank you!

As for the shared components/files. We can either put them directly under V2Wizard or we can create a `sharedComponents` folder and put everything in there so it's clear why they are not under a specific step. My vote would go to the folder.

amirfefer (commented):
Thanks @regexowl !
Could you please add the packages state into the customizations object  in [requestMapper](https://github.com/osbuild/image-builder-frontend/blob/main/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx#L37), so these changes will be applied E2E?

regexowl:
> Thanks @regexowl ! Could you please add the packages state into the customizations object in [requestMapper](https://github.com/osbuild/image-builder-frontend/blob/main/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx#L37), so these changes will be applied E2E?

For sure! I've added the changes to a new commit.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -1,545 +1,562 @@
...
-        newChosenPackages[pkg.name] = pkg;
-      }
-      setChosenPackages(newChosenPackages);
-    }
-  }, []);
+  const fetchCustomPackages = async () => {
+    await searchRpms({
+      apiSearchRpmRequest: {
+        search: searchTerm,
+        //@ts-ignore
```
lucasgarfield, line 220: I'd prefer if we can figure out a way to handle the typing issue properly in this PR, I don't want to get into the habit of letting these into the code without a really good reason. :smile:

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -1,545 +1,562 @@
...
+        urls: customRepositories.flatMap((repo) => repo.baseurl),
+      },
+    });
+  };
 
   useEffect(() => {
-    if (isSuccess) {
-      firstInputElement.current?.focus();
+    fetchCustomPackages();
+  }, [searchTerm]);
```
lucasgarfield, line 228: This is missing a dependency, `fetchCustomPackages`, in the dependency array.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
lucasgarfield: Right now Typescript is complaining because urls type is `string[] | undefined`. It should only be `string[]`. 

Casting it using `as string[]` would work but that doesn't seem very safe, either. 

Do we ever expect this `flatMap()` to return `undefined`? If it is impossible (or we believe it to be impossible...) then maybe we can just check for that and throw an error if it is `undefined`?

regexowl: Fair enough, forgot to get back to this one. :see_no_evil: 

I don't think `flatMap` should be returning `undefined` here since `customRepositories` are initialized as an empty array which should get "flattened" to an empty array again if I understand it correctly.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -1,545 +1,563 @@
...
-
-      if (b.startsWith(searchTerm) && !a.startsWith(searchTerm)) {
-        return 1;
-      }
+    const fetchCustomPackages = async () => {
+      await searchRpms({
+        apiSearchRpmRequest: {
+          search: searchTerm,
+          urls: customRepositories.flatMap((repo) =>
+            repo.baseurl ? repo.baseurl : ''
```
lucasgarfield, line 222: This solves the Typescript error but I think that this adds an empty string to the array in the case that `repo.baseurl === undefined`, which I don't think we want.

But what worries me a bit more is that we will be failing silently. We can't properly search without a baseurl, can we? Which means that if a baseurl is missing from a repo there is a big problem. If we fail silently the user simply won't get the results that they are expecting and will be confused as to why. Instead, I think we should just fail loudly so the user knows there is a problem and can report a bug (since this should never happen). Does that sound reasonable?

```typescript
  const fetchCustomPackages = async () => {
    await searchRpms({
      apiSearchRpmRequest: {
        search: searchTerm,
        urls: customRepositories.flatMap((repo) => {
          if (!repo.baseurl) {
            throw new Error(
              `repo (id: ${repo.id}, name: ${repo?.name}) is missing baseurl`
            );
          }
          return repo.baseurl;
        }),
      },
    });
  };
```

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -1,545 +1,563 @@
...
+        <Tbody>
+          {!searchTerm && toggleSelected === 'toggle-available' && (
+            <EmptySearch />
+          )}
+          {searchTerm && transformedPackages.length === 0 && <NoResultsFound />}
+          {searchTerm &&
+            transformedPackages.length >= 100 &&
+            handleExactMatch()}
+          {transformedPackages.length < 100 &&
+            transformedPackages
```
lucasgarfield, line 479: This is such a big code block. Could we extract the sorting logic into a helper function to help improve readability?

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -1,545 +1,563 @@
...
+                }
+                return 0;
               })
-            )}
-          </DualListSelectorList>
-        </DualListSelectorPane>
-      </DualListSelector>
+              .slice(computeStart(), computeEnd())
+              .map((pkg, rowIndex) => (
+                <Tr key={`${pkg.name}-${rowIndex}`} data-testid="package-row">
```
lucasgarfield, line 515: This is also a pretty big code block and it has the conditional rendering too... could we try to improve readability, maybe by extracting this into a `<PackageRow>` component?

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
regexowl: Great point, thank you! I've updated the code.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
regexowl: Updated based on the follow up comment

regexowl:
Rebased and updated the part with silent failing. I'm looking into splitting up the code block now :eyes:

regexowl:
Oh, it seems there have been some changes to the content-sources API in the meantime. Looking into it.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
regexowl: Moved the sorting logic outside of the block.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
regexowl: Tried to extract the logic to a `<PackageRow>` component, but event though everything renders correctly in msw the tests start to fail, showing only one row in the packages table. or completely sure where the problem is. So I've reverted this change for now. Trying to figure this out.

lucasgarfield (approved):
Great work, @regexowl!

PR #1641, opened 2024-02-09
feat(HMS-3567): fix cleanup of Reource Group

mgold1234:
this commit clean the resource group at Azure step when user clean the source field
https://issues.redhat.com/browse/HMS-3567


lucasgarfield (approved):
Great catch, thank you @mgold1234 !

PR #1643, opened 2024-02-09
Wizard: Remove aspect ratio for chart

regexowl:
This removes the aspect ratio for chart and sets `maintainAspectRatio` to false.

The aspect ratio was causing problems with the annotation plugin, rendering the currentDate annotation line in a wrong place for some windows sizes.


mgold1234 (approved):
/lgtm

PR #1644, opened 2024-02-09
 V2Wizard/Registration: Add request assertion tests 

lucasgarfield:
We need to start using `undefined` as the default state for when a value
has not been defined. Previously we had used things like `’’` for string
typed values. But this causes problems later when generating the request
we send to image-builder. Using `undefined` is explicit and will make
generating the requests much easier (as we don’t need to check for `’’`,
determine the intent, and convert it to undefined if necessary).
Explicit is better than implicit.

With that in mind, tests have been added to ensure that the correct
request is sent to the API for every option on the Registration step.
This is facilitated using a new `spyOnRequest()` function. In the
future, we will have similar tests for the rest of the steps.

A few other minor things:

1. We need to get the `store` using `useStore()`
for when we later call `store.getState()` because the tests use a different
store that is configured in the renderer than the one we were importing.

2. In the wizardSlice, a new type RegistrationType is added that provides
additional type safety instead of using `string`.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
@@ -12,6 +12,7 @@ import {
 } from '@patternfly/react-core';
 import { SpinnerIcon } from '@patternfly/react-icons';
 import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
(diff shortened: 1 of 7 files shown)


lucasgarfield:
@croissanne could you please take a very close look at the tests I've written and confirm that they send the expected API requests?

PR #1650, opened 2024-02-14
README: Link to osbuild-getting-started

ochosi:
Let's save our contributors a hop and not link to the (deprecated) devel/README.md anymore, which is only a placeholder pointing to osubild-getting-started. Let's instead direct folks there in the first place.

This will also unblock periodically pulling the frontend docs into the developer guide: https://github.com/osbuild/osbuild.github.io/pull/40
I'd rather not fix the broken link there by pulling in an empty page.


regexowl (approved):
Thank you! :sparkles:

PR #1653, opened 2024-02-15
V2Wizard/Details: Add request assertion tests

lucasgarfield:
Adds tests that verify requests related to blueprint name and details are generated correctly.

### DIFF
--- a/src/test/Components/CreateImageWizardV2/steps/Details/Details.test.tsx
+++ b/src/test/Components/CreateImageWizardV2/steps/Details/Details.test.tsx
@@ -1,0 +1,83 @@
+import { screen } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
+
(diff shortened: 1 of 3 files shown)


regexowl (approved):
Looks good to me!

Two things I'm thinking about - the `goToReviewStep()` function will be probably similar to other move-by-one-step functions. We can see later if we can create something like `goToFollowingStep` or if we could just use `clickNext` directly. But I'd say that depends on how many similar functions will pop up.

The second thing is - currently some of the functions are directly in the step test files and some are in the utils. I see the thought behind it being functions unique to the test file and functions shared between several test files. At the same time the `enterBlueprintName` and `enterBlueprintDescription` confused me for a bit as I couldn't find them in once place. Maybe all the helper functions should live in one place? Not completely sure, but also could be something we'll figure out as the volume of the tests grows.

PR #1654, opened 2024-02-15
Blueprints: Add blueprint editing

amirfefer:
This PR adds editing blueprints capabilities.
When a user clicks on `Edit details` or navigate to `/imagewizard/<BLUEPRINT_ID>`  the wizard opens  in the review step with preloaded blueprint's data, by mapping the blueprint's request to the state.

### DIFF
--- a/src/Components/CreateImageWizardV2/EditImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/EditImageWizard.tsx
@@ -1,0 +1,42 @@
+import React, { useEffect } from 'react';
+
+import { useNavigate } from 'react-router-dom';
+
+import CreateImageWizard from './CreateImageWizard';
+import { mapRequestToState } from './utilities/requestMapper';
+
+import { useAppDispatch } from '../../store/hooks';
+import { useGetBlueprintQuery } from '../../store/imageBuilderApi';
+import { loadWizardState, initializeWizard } from '../../store/wizardSlice';
+import { resolveRelPath } from '../../Utilities/path';
+
+type EditImageWizardProps = {
+  blueprintId: string;
+};
+
+const EditImageWizard = ({ blueprintId }: EditImageWizardProps) => {
+  const dispatch = useAppDispatch();
+  const navigate = useNavigate();
+
+  const { data: blueprintDetails, error } = useGetBlueprintQuery({
+    id: blueprintId,
+  });
+  useEffect(() => {
+    if (blueprintId && blueprintDetails) {
+      const editBlueprintState = mapRequestToState(blueprintDetails);
+      dispatch(loadWizardState(editBlueprintState));
+    } else {
+      dispatch(initializeWizard());
+    }
+  }, [blueprintId, blueprintDetails, dispatch]);
+  useEffect(() => {
+    // redirect to the main page if the composeId is invalid
+
+    if (error) {
+      navigate(resolveRelPath(''));
+    }
+  }, [error, navigate]);
+  return <CreateImageWizard startStepIndex={12} />;
+};
+
+export default EditImageWizard;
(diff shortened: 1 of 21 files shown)


src/Components/CreateImageWizardV2/EditImageWizard.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useAppDispatch } from '../../store/hooks';
+import { useGetBlueprintQuery } from '../../store/imageBuilderApi';
+import { loadWizardState, initializeWizard } from '../../store/wizardSlice';
+import { resolveRelPath } from '../../Utilities/path';
+
+type EditImageWizardProps = {
+  composeId: string;
+};
+
+const EditImageWizard = ({ composeId }: EditImageWizardProps) => {
```
lucasgarfield, line 17: Should this be `blueprintId`?

src/Components/CreateImageWizardV2/EditImageWizard.tsx
```diff
@@ -0,0 +1,39 @@
...
+      const editBlueprintState = mapRequestToState(blueprintDetails);
+      dispatch(loadWizardState(editBlueprintState));
+    } else {
+      dispatch(initializeWizard());
+    }
+  }, [composeId, blueprintDetails, dispatch]);
+  useEffect(() => {
+    // redirect to the main page if the composeId is invalid
+    if (error) navigate(resolveRelPath(''));
+  }, [error, navigate]);
```
lucasgarfield, line 35: Possibly beyond the scope of this PR, but would like to have the conversation here and create an issue to track it if you agree...

Do you think it would be a good idea to issue a notification here letting the user know that we couldn't resolve the `blueprintId`? Right now we fail silently which might be confusing/frustrating.

src/Components/CreateImageWizardV2/steps/Review/Footer.tsx
```diff
@@ -55,20 +57,33 @@ const ReviewWizardFooter = () => {
...
+      requestBody && createBlueprint({ createBlueprintRequest: requestBody });
+    }
   };
 
   const onSaveAndBuild = async () => {
     const requestBody = await getBlueprintPayload();
     setIsOpen(false);
-    const blueprint =
-      !composeId &&
+    if (composeId) {
```
lucasgarfield, line 71: This is getting a bit hard to understand, we have multiple conditionals in series and they have nested conditionals in the form of short circuit statements. Then there’s also the complicated async logic, including the call to `unwrap()` (it is not in this PRs scope but it would be nice to extract that into a function with a clear name to help future us know why we need to do that).

Could we try making this more composable to improve the readability? What if we had two components, `<CreateSubmitButton>` and `<EditSubmitButton>`. We could make the `submitButton` a prop for the ReviewFooter and pass in the appropriate one when creating the wizard. What do you think? Would you mind giving that a try?

amirfefer: Sure thing 👍

amirfefer:
https://github.com/osbuild/image-builder-frontend/assets/11807069/bd1689a9-265c-4f82-98cf-2519b78f3216

lucasgarfield:
I really like the new approach of just storing the AWS source id in the store instead of the entire source object, which results in duplicate data between the original query and the store's state. We'll just need to do the same thing for Azure, too.

src/Components/CreateImageWizardV2/EditImageWizard.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const { data: blueprintDetails, error } = useGetBlueprintQuery({
+    id: blueprintId,
+  });
+  useEffect(() => {
+    if (blueprintId && blueprintDetails) {
+      const editBlueprintState = mapRequestToState(blueprintDetails);
+      dispatch(loadWizardState(editBlueprintState));
+    } else {
+      dispatch(initializeWizard());
```
lucasgarfield, line 29: Wouldn't this mean that the wizard state was incorrect for a split second, before the useGetBlueprintQuery response was received? Would it be better to use a loading state?

lucasgarfield (approved):
Seems to be working nicely!

PR #1658, opened 2024-02-18
V2 Wizard: Add File System Configuration Step (HMS-2781)

mgold1234:
V2Wizrd: Create file System configuration step

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -10,6 +10,7 @@ import {
 import { useNavigate, useSearchParams } from 'react-router-dom';
 
 import DetailsStep from './steps/Details';
+import FileSystemStep from './steps/FileSystem';
 import ImageOutputStep from './steps/ImageOutput';
 import OscapStep from './steps/Oscap';
 import PackagesStep from './steps/Packages';
@@ -131,7 +132,6 @@ const CreateImageWizard = ({ startStepIndex = 1 }: CreateImageWizardProps) => {
     selectAzureResourceGroup(state)
   );
   const azureSource = useAppSelector((state) => selectAzureSource(state));
-
   const registrationType = useAppSelector((state) =>
     selectRegistrationType(state)
   );
@@ -248,6 +248,13 @@ const CreateImageWizard = ({ startStepIndex = 1 }: CreateImageWizardProps) => {
             footer={<CustomWizardFooter disableNext={false} />}
           >
             <OscapStep />
+          </WizardStep>
+          <WizardStep
+            name="File system configuration"
+            id="step-file-system"
+            footer={<CustomWizardFooter disableNext={false} />}
+          >
+            <FileSystemStep />
           </WizardStep>
           <WizardStep
             name="Content"
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
@@ -1,0 +1,329 @@
+import React, { useState } from 'react';
+
+import {
+  Button,
+  Popover,
+  Text,
+  TextContent,
+  TextInput,
+  TextVariants,
+} from '@patternfly/react-core';
+import { Select, SelectOption } from '@patternfly/react-core/deprecated';
+import {
+  HelpIcon,
+  MinusCircleIcon,
+  PlusCircleIcon,
+} from '@patternfly/react-icons';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
+
+import { UNIT_GIB, UNIT_KIB, UNIT_MIB } from '../../../../constants';
+import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
+import {
+  changePartitionMinSize,
+  changePartitionMountpoint,
+  selectPartitions,
+} from '../../../../store/wizardSlice';
+import UsrSubDirectoriesDisabled from '../../UsrSubDirectoriesDisabled';
+import { ValidatedTextInput } from '../../ValidatedTextInput';
+
+export type Partition = {
+  id: string;
+  mountpoint: string;
+  min_size: string;
+};
+
+const FileSystemConfiguration = () => {
+  const partitions = useAppSelector((state) => selectPartitions(state));
+
+  return (
+    <>
+      <TextContent>
+        <Text component={TextVariants.h3}>Configure partitions</Text>
+      </TextContent>
+      {partitions?.find((partition) =>
+        partition?.mountpoint?.includes('/usr')
+      ) && <UsrSubDirectoriesDisabled />}
+      <TextContent>
+        <Text>
+          Create partitions for your image by defining mount points and minimum
+          sizes. Image builder creates partitions with a logical volume (LVM)
+          device type.
+        </Text>
+        <Text>
+          The order of partitions may change when the image is installed in
+          order to conform to best practices and ensure functionality.
+          <br></br>
+          <Button
+            component="a"
+            target="_blank"
+            variant="link"
+            icon={<ExternalLinkAltIcon />}
+            iconPosition="right"
+            href="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/creating_customized_images_by_using_insights_image_builder/customizing-file-systems-during-the-image-creation"
+            className="pf-u-pl-0"
+          >
+            Read more about manual configuration here
+          </Button>
+        </Text>
+      </TextContent>
+      <Table aria-label="File system table" variant="compact">
+        <Thead>
+          <Tr>
+            <Th />
+            <Th>Mount point</Th>
+            <Th></Th>
+            <Th>Type</Th>
+            <Th>
+              Minimum size
+              <Popover
+                hasAutoWidth
+                bodyContent={
+                  <TextContent>
+                    <Text>
+                      Image Builder may extend this size based on requirements,
+                      selected packages, and configurations.
+                    </Text>
+                  </TextContent>
+                }
+              >
+                <Button
+                  variant="plain"
+                  aria-label="File system configuration info"
+                  aria-describedby="file-system-configuration-info"
+                  className="pf-c-form__group-label-help"
+                >
+                  <HelpIcon />
+                </Button>
+              </Popover>
+            </Th>
+            <Th />
+            <Th />
+          </Tr>
+        </Thead>
+        <Tbody data-testid="file-system-configuration-tbody">
+          {partitions &&
+            partitions.map((partition) => (
+              <Row key={partition.id} partition={partition} />
+            ))}
+        </Tbody>
+      </Table>
+      <TextContent>
+        <Button
+          ouiaId="add-partition"
+          data-testid="file-system-add-partition"
+          className="pf-u-text-align-left"
+          variant="link"
+          icon={<PlusCircleIcon />}
+          onClick={() => {}}
+        >
+          Add partition
+        </Button>
+      </TextContent>
+    </>
+  );
+};
+
+type RowPropTypes = {
+  partition: Partition;
+};
+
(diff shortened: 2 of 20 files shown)


amirfefer (commented):
Thanks @mgold1234!
Could you please update the `files` customization in the [requestMapper](https://github.com/osbuild/image-builder-frontend/blob/main/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx#L37) so these changes will be applied E2E?

regexowl:
Following up on our chat yesterday, let me try to sum up what I see as two possible approaches here:

First one would be keeping an array and manipulating it as it is done now, so via useState. That would mean having the `rows`  pre-filled with `initialRow` and adding/removing the mountpoints to it, not dispatching anything. The dispatch would come when the fsc step is finalized and that way we push the changes to the store in bulk. This might make the optional reordering of the mount points easier, but feels very heavy weight.

The second approach I would probably prefer would be to instead split the "add-in-bulk" functionality and create two or maybe three actions in the wizardSlice. One for adding a mountpoint, this one would be called on clicking "Add partition" button and could look something like this:

```typescript
addMountpoint: (state, action) => {
  state.fsc.push(action.payload);
},
```

second would be for removing a mountpoint, would get triggered on clicking the minus button next to a partition and could look something like this:

```typescript
removeMounpoint: (state, action) => {
  state.fsc.splice(
    state.fsc.findIndex((r) => r.mountpoint === action.payload.mountpoint),
    1
  );
},
```

Then we would probably need a third action for updating the mountpoint when the size changes.

All of the three actions can be triggered directly where it matters and the store would be up to date at any point in the code which I think would be really neat. Also the fsc array in the wizardSlice could be initialized with the root partition eliminating possible problems with missing root along the way.

When it comes to the two possible errors - duplicates and missing root - those do not need to persist throughout the steps as causing an error should also disable the next button. Therefore those do not have to be stored in the store. The way to deal with these would be probably again using some kind of custom next button that will, when clicking on next, check if there is "/" somewhere in the array of mountpoints and check if there are any duplicates (by creating a copy, converting it into a set and checking whether the lengths are equal? something like that maybe).

What do you think? The second approach would mean more changes, but I think it could save us some work in the future. The question is whether we have time to go this way or if we should stick with the current/first approach for now.

cc @lucasgarfield

lucasgarfield:
@regexowl I like your suggestion a lot, let's go with your second approach and use specific reducers for adding, removing, and updating mountpoints. I think we need to take a balanced approach to refactoring as we build the V2 Wizard. In this case, as the refactoring is directly related to taking advantage of our new stage management, I think it will be well worth the effort.

regexowl:
Not a review, just noticed a typo in the `FileSystemParition.tsx`, I guess should be Parition -> Partition? I would forget it later :sweat_smile:

src/store/wizardSlice.ts
```diff
@@ -118,6 +125,18 @@ const initialState: wizardState = {
...
+    mode: 'automatic',
+    fileSystemConfiguration: [
+      {
+        id: uuidv4(),
+        mountpoint: '/',
+        fstype: 'xfs',
+        size: 10,
+        unit: UNIT_GIB,
+      },
+    ],
```
regexowl, line 138: This is great for initialization of manual partitioning, but we probably shouldn't use it for automatic. The question is can we / should we put it in a condition here? I mean we could probably put a condition to the requestMapper "if automatic, put only mode in the request". But that would mean that the store has this value in it for automatic partitioning, which could be confusing when debugging.

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -236,6 +239,22 @@ const CreateImageWizard = () => {
...
           </WizardStep>
+          <WizardStep
+            name="File system configuration"
+            id="step-file-system"
+            footer={
+              <CustomWizardFooter
+                disableNext={
+                  (isFileSystemConfigValid(fsc)?.duplicates?.length !== 0 &&
+                    isFileSystemConfigValid(fsc)?.duplicates?.length !==
+                      undefined) ||
```
regexowl, line 250: Could the not zero nor undefined be replaced by `isFileSystemConfigValid(fsc)?.duplicates?.length > 0`

It checks if there are any duplicates, right?

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemPartition.tsx
```diff
@@ -0,0 +1,49 @@
...
+        id="radio-with-description"
+        label={
+          <>
+            <Label isCompact color="blue">
+              Recommended
+            </Label>{' '}
+            Use automatic partitioning
+          </>
+        }
+        name="radio-7"
```
regexowl, line 28: Could the radios get better descriptive names? Here maybe something like "fsc-radio-automatic"?

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemPartition.tsx
```diff
@@ -0,0 +1,49 @@
...
+        description="AutIsDraggingomatically partition your image to what is best, depending on the target environment(s)"
+        isChecked={fileSystemPartition === 'automatic'}
+        onChange={() => {
+          dispatch(changeFileSystemPartitionMode('automatic'));
+        }}
+      />
+      <Radio
+        id="radio"
+        label="Manually configure partitions"
+        name="radio-8"
```
regexowl, line 38: And here it could be something like "fsc-radio-manual"

src/Components/CreateImageWizardV2/steps/ImageOutput/ReleaseLifecycle.tsx
```diff
@@ -103,7 +103,7 @@ export const chartMajorVersionCfg = {
   },
 };
 
-const MajorReleasesLifecyclesChart = () => {
+export const MajorReleasesLifecyclesChart = () => {
```
regexowl, line 106: Does this change belong to this PR?

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -50,9 +52,11 @@ import {
   selectGcpShareMethod,
   selectPackages,
   selectRegistrationType,
+  selectFileSystemConfiguration,
+  selectFileSystemPartitionMode,
 } from '../../../../store/wizardSlice';
 import { toMonthAndYear } from '../../../../Utilities/time';
-import { MajorReleasesLifecyclesChart } from '../../../CreateImageWizard/formComponents/ReleaseLifecycle';
+import { MajorReleasesLifecyclesChart } from '../ImageOutput/ReleaseLifecycle';
```
regexowl, line 59: Now I see why there is the change to the ReleaseLifecycle. This is great catch! Could you please open a separate PR for this one so the changes are not hidden in the FSC one?

src/Components/CreateImageWizardV2/steps/FileSystem/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { Text, Form, Title } from '@patternfly/react-core';
+
+import FileSystemAutomaticPartition from './FileSystemAutomaticPartitionInformation';
+import FileSystemConfiguration from './FileSystemConfiguration';
+import FileSystemPartition from './FileSystemPartition';
+
+import { useAppSelector } from '../../../../store/hooks';
+import { selectFileSystemPartitionMode } from '../../../../store/wizardSlice';
+export type FileSystemPartitionMode = 'automatic' | 'manual' | 'oscap';
```
regexowl, line 11: I'm still not very sure about adding 'oscap' as a partitioning mode, alas can't think of better solution right now :thinking: 

What we'll need to keep in mind is that the oscap would need to be parsed as "manual" in requestMapper.

src/Components/CreateImageWizardV2/steps/FileSystem/SizeUnit.js
```diff
@@ -0,0 +1,84 @@
+import React, { useEffect, useState } from 'react';
```
regexowl, line 1: Let's migrate this file to TypeScript.

src/Components/CreateImageWizardV2/steps/FileSystem/MountPoint.tsx
```diff
@@ -0,0 +1,109 @@
...
+          )}
+      </GridItem>
+    </Grid>
+  );
+};
+
+MountPoint.propTypes = {
+  mountpoint: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired,
+};
```
regexowl, line 107: We should replace propTypes with proper TypeScript typing. As this file is now 1:1 with V1 I don't mind if the typing gets split into separate task, but it's something we should keep in mind / create an issue for if we decide to split it.

src/store/wizardSlice.ts
```diff
@@ -67,7 +71,10 @@ type wizardState = {
       enabled: string[] | undefined;
     };
   };
-
+  fileSystem: {
+    mode: FileSystemPartitionMode;
+    fileSystemConfiguration: FileSystemConfigInfo[];
```
regexowl, line 76: This will be very nit picky, but I'm thinking about more descriptive names. What do you think about 
```typescript
fsc: {
  partitioningMode: 'automatic',
  fileSystemConfiguration: [
    ...
  ],
}
```

for the step name from what I can see we are mirroring the headers on the step mostly and mode -> partitioningMode is a small change, but makes things clearer even without any context.

regexowl (commented):
The PR is looking good, still reviewing, but here are some initial comments.

regexowl:
Is it possible to clean up the commit history a bit? Generally it could follow the usual structure:
1. commit copying over needed files
2. commit including changes to the fsc code
3. commit for related oscap changes
4. commit for review step
5. commit for test updates

In case it would be difficult to split 2. and 3. I don't mind all the code changes connected directly to the fsc step in one commit.

regexowl:
When adding a new partition the validation glitches on the screen for a second even though there is a root partition already and none of the mount points are duplicates:

validation_glitch

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -0,0 +1,536 @@
...
+} from '@patternfly/react-core';
+import {
+  HelpIcon,
+  MinusCircleIcon,
+  PlusCircleIcon,
+} from '@patternfly/react-icons';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import styles from '@patternfly/react-styles/css/components/Table/table';
+import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
+// @ts-ignore
```
regexowl, line 20: Ok, googled a bit and this can be removed. All that needs to be done for the error to disappear is to install `@types/uuid` as a dev dependency. The dependency needs to have the same version as uuid in the package.json file.

src/store/wizardSlice.ts
mgold1234: good point, I think the good solution here is to initialize the configuration to empty array, and if we choose manual. we have the first row initizalization, WDYT @regexowl ?

regexowl: Sounds good to me :+1:

src/Components/CreateImageWizardV2/steps/ImageOutput/ReleaseLifecycle.tsx
mgold1234: no, do you think I need to open new pr for that?

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
mgold1234: sure

src/store/wizardSlice.ts
mgold1234: amm @lucasgarfield advice me to change it to mode, @lucasgarfield WDYT?

lucasgarfield:
@mgold1234 Can you please clean up the commits a bit? I'd suggest you start by using fixup or squash to get them into a logical progression and fix the typo (Wizrad -> Wizard).

PR #1666, opened 2024-02-20
Blueprints: add blueprint out-of-sync alert

amirfefer:
Screenshot 2024-02-20 at 19 37 45

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -54,9 +54,11 @@ import {
   STATUS_POLLING_INTERVAL,
 } from '../../constants';
 import {
+  BlueprintItem,
   ComposesResponseItem,
   ComposeStatus,
   useGetBlueprintComposesQuery,
+  useGetBlueprintsQuery,
   useGetComposesQuery,
   useGetComposeStatusQuery,
 } from '../../store/imageBuilderApi';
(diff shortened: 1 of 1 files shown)


src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -73,6 +74,17 @@ type ImageTableProps = {
 const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
   const [page, setPage] = useState(1);
   const [perPage, setPerPage] = useState(10);
+  const { selectedBlueprintVersion } = useGetBlueprintsQuery(
+    {},
+    {
+      selectFromResult: ({ data }) => ({
+        selectedBlueprintVersion: data?.data?.find(
+          (blueprint: { id: string | undefined }) =>
```
lucasgarfield, line 82: Let's use `BlueprintItem` (exported from `imageBuilderApi.ts`) as the type here instead of `{ id: string | undefined }`.

lucasgarfield (dismissed):
This is really lovely! Just a minor nitpick about the typing and we can get this in.

src/Components/ImagesTable/ImagesTable.tsx
lucasgarfield: Force pushed these changes to your branch myself so we can hopefully just get this in, hope that's okay with you. :sweat_smile:

lucasgarfield (approved):
I really like this, good thinking.

PR #1674, opened 2024-02-21
V2Wizrd: add azure details to review step

mgold1234:
this commit add the azure details to review step


amirfefer (commented):
Thanks @mgold1234!
Could you please add the azure target to the [requestMapper](https://github.com/osbuild/image-builder-frontend/blob/main/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx#L37) so these changes will be applied E2E?

PR #1678, opened 2024-02-22
test: Up timeout for 'image name invalid for more than 63 chars'

regexowl:
This increases the timeout of 'image name invalid for more than 63 chars' to 20000.


regexowl:
This test times out for me on most of the runs marking dev checks as a failure. Can we increase the timeout a bit?

mgold1234 (approved):
/lgtm

PR #1681, opened 2024-02-26
Blueprints: Add guest-image and vsphere-ova to requestMapper

regexowl:
This adds guest-image and vsphere-ova as possible target environments to the requestMapper.


regexowl:
@lucasgarfield @amirfefer When monkey testing the packages addition to the requestMapper I've noticed that guest-image and vsphere-ova are missing from the `uploadTypeByTargetEnv`.

In case it's not intentional this PR should fix it. If it's intentional feel free to close this :sweat_smile:

PR #1693, opened 2024-02-28
Blueprints: filter composes by blueprint version

ezr-ondrej:
Refs: HMS-3412


lucasgarfield:
Tests are failing, I haven't had time to deep dive into them but just looking at your code I think you might need to hide the new JSX behind a feature flag like so: `{ experimentalFlag && ... }`.

ezr-ondrej:
> Tests are failing, I haven't had time to deep dive into them but just looking at your code I think you might need to hide the new JSX behind a feature flag like so: `{ experimentalFlag && ... }`.

I've added a new compose, which was first and failed other tests :)
Adding it last should fix things

src/Components/Blueprints/BlueprintVersionFilter.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const onSelect = (
+    _event: React.MouseEvent<Element, MouseEvent> | undefined,
+    value: string | undefined
+  ) => {
+    if (setBlueprintVersionFilter) setBlueprintVersionFilter(value);
+    setIsOpen(false);
+  };
+
+  return (
+    <div data-testid="blueprint-version-filter">
```
lucasgarfield, line 55: No need for a `<div>` here (we never use those), you can just put the `data-testid` directly into the `<Dropdown>`. You might need to surround the component with an empty set of `<></>` tags.

ezr-ondrej: yup `Unable to find an element by: [data-testid="blueprint-version-filter"]` when I move the testid

lucasgarfield (changes requested):
Sorry to do this @ezr-ondrej but would you mind rebasing this on top of https://github.com/osbuild/image-builder-frontend/pull/1712? 

This will require a fair amount of changes but I think it will be worth it to manage the filter state (especially once we start adding multiple filters) in Redux.

ezr-ondrej:
> Sorry to do this @ezr-ondrej but would you mind rebasing this on top of #1712?
> 
> This will require a fair amount of changes but I think it will be worth it to manage the filter state (especially once we start adding multiple filters) in Redux.

No worries, agreed, lets draft this and lets get #1712 in first :+1:

src/Components/Blueprints/BlueprintVersionFilter.tsx
lucasgarfield: Urgh, that's frustrating. Do you actually need a data-testid? I usually use the Testing Playground browser plugin (https://chromewebstore.google.com/detail/testing-playground/hejbmebodbijjdhflfknehhcgaklhano?pli=1) and try to avoid data-testid as much as possible... (the Testing Library folks consider using it to be a last resort: https://testing-library.com/docs/queries/bytestid/)

ezr-ondrej: Looks like a cool plugin, let me try! :)

src/Components/Blueprints/BlueprintVersionFilter.tsx
```diff
@@ -0,0 +1,85 @@
...
+  setBlueprintVersionFilter?: React.Dispatch<
+    React.SetStateAction<string | undefined>
+  >;
+}
+
+export const ALL_VERSIONS_CONST = 'All Versions';
+export const NEWEST_VERSION_CONST = 'Newest';
+const LATEST_VERSION = -1;
+
+export const getBlueprintVersionFilterAPIValue = (
```
lucasgarfield, line 22: This seems a bit over-engineered to me. I think that using Redux (and taking more advantage of Typescript types) will help simplify this a bit. 

Once we have a selector, we’ll be able to do something like
```Typescript
const blueprintVersionFilter = selectBlueprintVersionFilter(...)
```

I would have the type in the store be something like
```Typescript
type BlueprintVersionFilter = ‘all’ | ‘newest’
```

Then for this function you could just use a switch case, something like
```Typescript
const getBlueprintVersionFilter = (blueprintVersionFilter: BlueprintVersionFilter) => {
  switch (blueprintVersionFilter) {
    case 'all':
      return undefined;
    case 'newest':
      return '-1';
    default:
      return undefined;
  }
};
```

ezr-ondrej: Yeah I've completely rewritten it for it all be stored in the slice.

ezr-ondrej:
Ready for re-review IMO :)

ezr-ondrej:
One last change I've got in, improved the icon positioning (by using the MenuToggle prop instead of hacking the icon in manually):
image

amirfefer (commented):
Thanks @ezr-ondrej! 👍  Looks clean and clear :)
I just left a small nitpick

And a general question, if a blueprint is out of sync (no corresponding images for its latest version), wouldn't it be clearer to show the table out of sync alert and also with an empty state? 
In the empty state it mentioned to clear filters:
```
The selected blueprint does not contain any images, build an image
          from this version or adjust the filters.
```

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -205,6 +218,11 @@ const ImagesTable = () => {
                 </ToolbarItem>
               </>
             )}
+            {selectedBlueprintId && (
+              <ToolbarItem>
+                <BlueprintVersionFilter onFilterChange={() => setPage(1)} />
+              </ToolbarItem>
+            )}
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah, it's getting rather big, let me open a PR for it :+1:

ezr-ondrej:
> And a general question, if a blueprint is out of sync (no corresponding images for its latest version), wouldn't it be clearer to show the table out of sync alert and also with an empty state?
> In the empty state it mentioned to clear filters:

Hmm the code looks it should, but for some reason the backend apparently returns some images and we only filter them in the UI, I've now noticed the pagination shows three images, so there is something off.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -167,7 +173,14 @@ const ImagesTable = () => {
...
 
-  const composes = data?.data;
+  let filteredComposes = data?.data;
+  if (selectedBlueprintId && blueprintVersionFilter === 'latest') {
+    filteredComposes = filteredComposes?.filter((compose) => {
+      return compose.blueprint_version === selectedBlueprintVersion;
+    });
+  }
+
+  const composes = filteredComposes || data?.data;
```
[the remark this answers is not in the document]
ezr-ondrej: I've gotten rid of the filtered composes entirely and just used `let composes`;

src/test/Components/Blueprints/Blueprints.test.js
```diff
@@ -32,6 +32,20 @@ jest.mock('@unleash/proxy-client-react', () => ({
...
 
+const selectBlueprintById = async (user, bpId) => {
+  const nameMatcher = (_, element) =>
+    element.getAttribute('name') === 'blueprints';
+
+  const radioButtons = await screen.findAllByRole('radio', {
+    name: nameMatcher,
+  });
+  const elementById = radioButtons.find(
+    (button) => button.getAttribute('id') === bpId
```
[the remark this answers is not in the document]
ezr-ondrej: It's now a second commit - cleaning up all the other uses, we can improve the function itself later on :)

lucasgarfield:
Is it really correct for the filter only to appear after selecting an image? In the mocks (that I _think_ are the most up to date, although it is hard to be sure :sweat_smile: ) they show the filters appearing even if a blueprint isn't selected:
image
Also, the filters appear to the left of the Build images button in the mocks. 

I see why you put it on the right since it only appears when a blueprint is selected, that way the Build images button doesn't slide around, but if the filter is always visible I think we can also put it on the left to match the mocks.

ezr-ondrej:
> Is it really correct for the filter only to appear after selecting an image?

Well, at least initially this is the only way IMO.
Backend is not ready to apply this filter on all composes as that would require a change (even if small one) to the current endpoint and with that leak blueprints from the `/experimental`. We're holding off the version column information on all composes (https://github.com/osbuild/image-builder/pull/1046) for the same reason. 

I'd prefer to keep it out of scope, but if you think we should make it in scope, IMO it should wait until we are ready for preview at least.

src/Components/ImagesTable/ImagesTable.tsx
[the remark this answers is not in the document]
ezr-ondrej: https://github.com/osbuild/image-builder-frontend/pull/1769

lucasgarfield (approved):
Wow, using Redux made this so much cleaner! Thank you!

PR #1698, opened 2024-02-29
test: Re-align CreateImageWizard.test tests

regexowl:
The tests in the suite were improperly nested, this fixes the problem.

Structure before (steps starting by GCP are nested under AWS)
Screenshot from 2024-02-29 10-53-02

After the fix (all steps on top level)
Screenshot from 2024-02-29 10-50-46


lucasgarfield (approved):
Nice catch!

PR #1699, opened 2024-02-29
V2Wizard/Repositories: Add payload_repositories to request and test it

lucasgarfield:
Previously the V2 Wizard's request mapper was only adding the custom_repositories field. We also need to add a nearly duplicate payload_repositories field due to how the image-builder API works... which is admittedly not intuitive.

Tests are also added to ensure that requests are generated correctly when using the custom repositories feature.


lucasgarfield:
@croissanne I still get confused about how the required fields work when adding custom repositories, can you double check the tests I've written here and make sure that the expected request is correct?

regexowl:
The changes are looking great to me! I'll leave the final word to Sanne

regexowl (approved):
Looks great, thanks!

PR #1700, opened 2024-02-29
wizard: disable the unbuildable oscap profiles (HMS-2077)

lavocatt:
Some Oscap profiles can't get built. That is because they require to disable a set of services that aren't installed on the system at this moment.

A workaround is being worked on the backend side. But meanwhile, we decide to disable this set of profiles to insure the user doesn't stumble upon bound-to-fail configurations.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/Oscap.tsx
+++ b/src/Components/CreateImageWizard/formComponents/Oscap.tsx
@@ -22,6 +22,7 @@ import { HelpIcon } from '@patternfly/react-icons';
 
 import OscapProfileInformation from './OscapProfileInformation';
 
+import { RHEL_9 } from '../../../constants';
 import {
   DistributionProfileItem,
   useGetOscapCustomizationsQuery,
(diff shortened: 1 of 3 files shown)


lavocatt:
msw data get masked by the filter, I need to tweak the values

regexowl:
This looks great! Two questions - will this be also added in V2Wizard? And would it be possible to add a fixture with a blacklisted profile and check in the tests it does not get rendered?

lavocatt:
After checking out the `RHel8` builds, it turns out that they are all working, only the profiles for Rhel9 can bug. I've added a test for that.

lucasgarfield:
I agree with Thomas, this is very short term *and* this is only going into preview so manual testing is adequate. 

But it is surprising that this only seems to affect RHEL9... we should probably loop backend into this. @kingsleyzissou does that make sense to you?

One last thought... should we also check CentOS? Or even disable these profiles/the entire step for CentOS?

kingsleyzissou:
Question. Why don't just ignored the disabled services and keep the profiles? 

That way customers can still build images without the ability to disable services? It's super straightforward to do too once the image is booted

kingsleyzissou:
Also, I think I tried RHEL 8 & CIS last week and it had the same issue. But I could be wrong on that

kingsleyzissou:
https://github.com/osbuild/osbuild-composer/pull/3972 got merged, so I can work on getting this into stage for the service.

lucasgarfield (approved):

lucasgarfield:
@kingsleyzissou was able to get https://github.com/osbuild/image-builder-frontend/pull/1731 ready much faster than anticipated, so I'm closing this... thanks for the work @lavocatt, my apologies we didn't get to use it.

PR #1701, opened 2024-02-29
V2Wizard: Remove bulk select and refresh button from the Packages step

regexowl:
This removes bulk select and a refresh button from the Packages step and updates tests accordingly.


lucasgarfield (approved):
So fast! Thank you.

PR #1703, opened 2024-02-29
V2Wizard/Packages: Add tests for requests

lucasgarfield:
Adds tests that verify using the custom packages step to add a custom package results in a correctly generated API request.


regexowl (approved):
LGTM, thank you!

PR #1704, opened 2024-02-29
[WIP] test: Increase test timeout

regexowl:
This changes default test timeout from 10 to 20 seconds.


regexowl:
I don't think the global increase of the time out will be needed at the moment. Closing.

PR #1706, opened 2024-02-29
V2Wizard/test: Fix bug on registration step and enable tests (HMS-2781)

regexowl:
This fixes a bug on the registration step which made the text "Connect to Red Hat Insights" not to get rendered when registering with rhc.

All registration tests were also enabled in CreateImageWizard.test.tsx


lucasgarfield (approved):
Nice!

Issue #1691, opened 2024-02-28
Azure resource group


lucasgarfield:
The resource group dropdown should be cleared when the source is changed, because the resource groups are tied to a specific source.

mgold1234:
fix at https://github.com/osbuild/image-builder-frontend/pull/1707

PR #1707, opened 2024-02-29
V2Wizrd: clean resource group when user change the source name

mgold1234:
this commit fix https://github.com/osbuild/image-builder-frontend/issues/1691 when user change the source name, resource gruop should clean up


regexowl (approved):
Works great, thank you!

PR #1708, opened 2024-02-29
ghactions: Add PR best practice check (COMPOSER-2140)

ochosi:
Add an additional PR check that verifies that each pull request
1. Has a non-empty description
2. Has a title that complies with the schema: `component: This is the subject (JIRA-001)`
3. If both are ok, it adds the 'best practice' label to the PR (2 isn't blocking).

### DIFF
--- a/.github/workflows/pr_best_practices.yml
+++ b/.github/workflows/pr_best_practices.yml
@@ -1,0 +1,14 @@
+name: "Verify PR best practices"
+
+on:
+  pull_request_target:
+    branches: [main]
+
+jobs:
+  pr-best-practices:
+    runs-on: ubuntu-latest
+    steps:
+      - name: PR best practice check
+        uses: osbuild/pr-best-practices@main
+        with:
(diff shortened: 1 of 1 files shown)


ochosi:
TL;DR: Let's please merge this as is.

This took far too long to figure out...

1. GitHub Actions running in forks cannot use secrets that do more than `read`
2. Obviously, we're running in forks in general and need `write` to be able to add a label to the pull request.
3. We can use the `pull_request_target` trigger, which works around this by running code based on `main` (which is fine in our case).
4. The GitHub Action needs to be merged to `main` before it can be tested/executed.
5. This means that this PR needs to be merged, even though the GitHub Action didn't successfully run here.
6. I've tested it on another repository with more loose permissions so you can see that it works (the check was ok and the label was added): https://github.com/osbuild/pr-review-queue/pull/35

lucasgarfield (approved):
Extremely nice!

PR #1712, opened 2024-02-29
Blueprints: add blueprints slice for client global state 

amirfefer:
Made on top of  #1654, first two commits unrelated. 

This PR adds blueprints slice for managing blueprints client state globally. Multiple components use these data as props and args, and for keeping it consistent when a change occurs,  redux slice gives a clean solution.

### DIFF
--- a/src/Components/Blueprints/BlueprintsSideBar.tsx
+++ b/src/Components/Blueprints/BlueprintsSideBar.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react';
+import React, { useCallback } from 'react';
 
 import {
   Bullseye,
@@ -21,20 +21,18 @@ import debounce from 'lodash/debounce';
 import BlueprintCard from './BlueprintCard';
 
 import {
+  selectBlueprintSearchInput,
+  selectSelectedBlueprintId,
+  setBlueprintId,
+  setBlueprintSearchInput,
+} from '../../store/BlueprintSlice';
+import { useAppDispatch, useAppSelector } from '../../store/hooks';
+import {
   useGetBlueprintsQuery,
   BlueprintItem,
 } from '../../store/imageBuilderApi';
 
-type blueprintProps = {
-  selectedBlueprint: string | undefined;
-  setSelectedBlueprint: React.Dispatch<
-    React.SetStateAction<string | undefined>
-  >;
-};
-
 type blueprintSearchProps = {
-  filter: string | undefined;
-  setFilter: React.Dispatch<React.SetStateAction<string>>;
   blueprintsTotal: number;
 };
 
@@ -46,27 +44,13 @@ type emptyBlueprintStateProps = {
   bodyText: string;
 };
 
-const BlueprintsSidebar = ({
-  selectedBlueprint,
-  setSelectedBlueprint,
-}: blueprintProps) => {
-  const [blueprintsSearchQuery, setBlueprintsSearchQuery] = useState<
-    string | undefined
-  >();
-  const debouncedSearch = useCallback(
-    debounce((filter) => {
-      setBlueprintsSearchQuery(filter.length > 0 ? filter : undefined);
-    }, 300),
-    [setBlueprintsSearchQuery]
-  );
-  React.useEffect(() => {
-    return () => {
-      debouncedSearch.cancel();
-    };
-  }, [debouncedSearch]);
+const BlueprintsSidebar = () => {
+  const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
+  const blueprintSearchInput = useAppSelector(selectBlueprintSearchInput);
   const { data: blueprintsData, isLoading } = useGetBlueprintsQuery({
-    search: blueprintsSearchQuery,
+    search: blueprintSearchInput,
   });
+  const dispatch = useAppDispatch();
   const blueprints = blueprintsData?.data;
 
   const blueprintsTotal = blueprintsData?.meta?.count || 0;
@@ -79,7 +63,7 @@ const BlueprintsSidebar = ({
     );
   }
 
-  if (blueprintsTotal === 0 && blueprintsSearchQuery === undefined) {
+  if (blueprintsTotal === 0 && blueprintSearchInput === undefined) {
     return (
       <EmptyBlueprintState
         icon={PlusCircleIcon}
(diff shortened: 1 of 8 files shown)


lucasgarfield:
This had not occurred to me but I think it will probably end up being a good idea, I like it.

Haven't done an in-depth review yet but two things immediately:
1. Can you please move `BlueprintsSlice.tsx` into `/store`, use `pascalCase` for the title, and use a `.ts.` extension? Just to be consistent with, for instance, `wizardSlice.ts`.
2. Would it be possible to just rebase this onto `main`? Then I think we could get it in pretty quickly. Might take a while to get first two commits it is rebased onto in this PR reviewed and merged.

src/store/BlueprintSlice.ts
```diff
@@ -0,0 +1,53 @@
...
+type blueprintsState = {
+  selectedBlueprint: string | undefined;
+  searchInput?: string;
+  offset?: number;
+  limit?: number;
+};
+
+const initialState: blueprintsState = {
+  selectedBlueprint: undefined,
+  searchInput: undefined,
```
lucasgarfield, line 14: Let's initialize this as `''`.

src/store/BlueprintSlice.ts
```diff
@@ -0,0 +1,53 @@
+import { PayloadAction, createSlice } from '@reduxjs/toolkit';
+
+import { RootState } from '.';
+
+type blueprintsState = {
+  selectedBlueprint: string | undefined;
+  searchInput?: string;
+  offset?: number;
```
lucasgarfield, line 8: Would we ever want the offset and limit to be undefined? I suppose if we wanted the query to use whatever the default values are... but given that we'll be implementing pagination I think we will need fine control. Can we make these 
```
offset: number;
limit: number;
```
instead?

src/store/BlueprintSlice.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+const initialState: blueprintsState = {
+  selectedBlueprint: undefined,
+  searchInput: undefined,
+  offset: 0,
+  limit: 10,
+};
+
+export const selectSelectedBlueprint = (state: RootState) =>
```
lucasgarfield, line 19: Please change to `selectSelectedBlueprintId`.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -59,15 +64,14 @@ import {
...
-type ImageTableProps = {
-  selectedBlueprint?: string | undefined;
-};
-
-const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
+const ImagesTable = () => {
   const [page, setPage] = useState(1);
   const [perPage, setPerPage] = useState(10);
+  const selectedBlueprint = useAppSelector(selectSelectedBlueprint);
+  const blueprintsSearchQuery = useAppSelector(selectBlueprintSearchInput);
```
lucasgarfield, line 71: Would you mind changing this to `blueprintSearchInput`? I find it easier to understand the code when the name of the variable in the React component is the same as it is in the store.

src/Components/Blueprints/DeleteBlueprintModal.tsx
```diff
@@ -7,25 +7,27 @@ import {
...
   DeleteBlueprintModalProps
-> = ({
-  onDelete,
-  selectedBlueprint,
-  isOpen,
-  onClose,
-}: DeleteBlueprintModalProps) => {
+> = ({ onDelete, isOpen, onClose }: DeleteBlueprintModalProps) => {
+  const selectedBlueprint = useAppSelector(selectSelectedBlueprint);
+  const blueprintsSearchQuery = useAppSelector(selectBlueprintSearchInput);
```
lucasgarfield, line 27: Would you mind changing this to `blueprintSearchInput`? I find it easier to understand the code when the name of the variable in the React component is the same as it is in the store.

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -141,30 +124,37 @@ const BlueprintsSidebar = ({
...
   );
 };
 
-const BlueprintSearch = ({
-  filter,
-  setFilter,
-  blueprintsTotal,
-}: blueprintSearchProps) => {
+const BlueprintSearch = ({ blueprintsTotal }: blueprintSearchProps) => {
+  const blueprintsSearchQuery = useAppSelector((state) =>
```
lucasgarfield, line 136: Would you mind changing this to `blueprintSearchInput`? I find it easier to understand the code when the name of the variable in the React component is the same as it is in the store.

src/Components/Blueprints/BlueprintsSideBar.tsx
```diff
@@ -49,27 +47,13 @@ type emptyBlueprintStateProps = {
...
-    [setBlueprintsSearchQuery]
-  );
-  React.useEffect(() => {
-    return () => {
-      debouncedSearch.cancel();
-    };
-  }, [debouncedSearch]);
+const BlueprintsSidebar = () => {
+  const selectedBlueprint = useAppSelector(selectSelectedBlueprint);
+  const blueprintsSearchQuery = useAppSelector(selectBlueprintSearchInput);
```
lucasgarfield, line 52: Would you mind changing this to blueprintSearchInput? I find it easier to understand the code when the name of the variable in the React component is the same as it is in the store.

lucasgarfield:
Wow, this is really nice! The components are going to be so much easier to work with now. I know the variable name comments are a bit nitpicky, but I think that the comments about the types in the store are actually something really important to get correct from the beginning.

ezr-ondrej:
@amirfefer can you rebase this pls? I guess this would be nice to get in prioritized, to base other work on this.

src/store/BlueprintSlice.ts
amirfefer: When the backend get empty string (`''`)  in the search  param,  it will do an empty searching...

ezr-ondrej: Yeah, lets not xD or fix the backend first :)

src/store/BlueprintSlice.ts
```diff
@@ -0,0 +1,53 @@
+import { PayloadAction, createSlice } from '@reduxjs/toolkit';
+
+import { RootState } from '.';
+
+type blueprintsState = {
+  selectedBlueprint: string | undefined;
+  searchInput?: string;
```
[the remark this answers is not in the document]
amirfefer: There's a reason why undefined is needed here for empty search, as you can see in the endpoint implementation  - https://github.com/osbuild/image-builder/blob/main/internal/v1/handler_blueprints.go#L192

src/store/BlueprintSlice.ts
amirfefer: The default values for offset and limit in the endpoint are `0` and `100` respectively, we can have a default values here as well, but we need to keep it consistent.

src/store/BlueprintSlice.ts
```diff
@@ -0,0 +1,53 @@
+import { PayloadAction, createSlice } from '@reduxjs/toolkit';
+
+import { RootState } from '.';
+
+type blueprintsState = {
+  selectedBlueprint: string | undefined;
```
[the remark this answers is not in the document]
amirfefer: I  kept the previous name, to have less changes in code, tho `selectedBlueprintId` is better

src/store/BlueprintSlice.ts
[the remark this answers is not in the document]
lucasgarfield: Okay, as long as there's a good reason!

src/store/BlueprintSlice.ts
lucasgarfield: Fair enough :smile:

src/store/BlueprintSlice.ts
lucasgarfield: Okay that's fine.

ezr-ondrej:
/retest

npm acting out

lucasgarfield (dismissed):
Thanks Amir! I think this is going to be the last "big" PR, we're so close!

lucasgarfield (approved):

PR #1718, opened 2024-03-01
test: Remove "Click through all" (HMS-2781)

regexowl:
This removes the "Click through all" test as it will be replaces by testing the requests for separate customisations.


lucasgarfield (approved):
One weird trick for getting your tests passing!

PR #1719, opened 2024-03-01
V1 Wizard: Add search parameter for release selection (HMS-3684)

lucasgarfield:
Adding an optional search parameter to the wizard like so:

/insights/image-builder/imagewizard?release=rhel8

results in the wizard being opened and RHEL 8 being pre-selected as the release.

The Insights assistant chat bot will make use of this feature.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/ImageOutputReleaseSelect.js
+++ b/src/Components/CreateImageWizard/formComponents/ImageOutputReleaseSelect.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { FormSpy } from '@data-driven-forms/react-form-renderer';
 import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';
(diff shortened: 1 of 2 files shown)


regexowl:
The re-renders are not optimal, but if it unblocks PA, then I'd say we can merge and look into those later.

There are currently tests failing. From the logs it seems that the failure is happening in Keyboard Accessibility, but fails the "set release using query parameter", there could be an early exit or something similar going on.

regexowl:
Found the problem - the `testTile` function in "target environment tiles are keyboard selectable" test isn't properly awaited, leaving no time for attributes to change. So changing it to this will do the trick:
```typescript
    const testTile = async (tile) => {
      tile.focus();
      await user.keyboard('{space}');
      await waitFor(() => expect(tile).toHaveClass('pf-m-selected'));
      await user.keyboard('{space}');
      await waitFor(() => expect(tile).not.toHaveClass('pf-m-selected'));
    };
```
I'd add a commit, but unsure how to do that properly. :sweat_smile:

PR #1720, opened 2024-03-01
V2Wizard: Update Review step

regexowl:
This updates the ReleaseLifecycle dates used on the Review step and adds a skeleton for FSC step.


src/Components/CreateImageWizardV2/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -103,8 +107,80 @@ export const ImageOutputList = () => {
   );
 };
 export const FSCList = () => {
+  //const isManual =
+  //  getState()?.values?.['file-system-config-radio'] === 'manual';
+  //const partitions = getState()?.values?.['file-system-configuration'];
```
lucasgarfield, line 112: Do we want to just delete this commented code? Since it's calling a DDF function, I can't think of a good reason to keep it.

lucasgarfield: I suppose this is part of the skeleton for FSC... To be honest, I would prefer not adding a skeleton in the form of comments. If we could figure out a way to do this without commenting the code, that would be better.

Can we split out the ReleaseLifecycle dates fix and the FSC list stuff into separate commits? Maybe for this PR we can just have the ReleaseLifeCycle date fix and then we can worry about the FSC review step later, once we have the actual step merged.

regexowl: Fair enough, I've updated the PR to fix the ReleaseLifecycle dates only :+1:

lucasgarfield (approved):
Thanks!

Issue #1721, opened 2024-03-01
V2Wizard: Add currentStep logic to enable conditional rendering of shared components


regexowl:
We should implement currentStep to the Wizard https://www.patternfly.org/components/wizard#get-current-step as we are reusing some of the components on the Review step and this will allow us to have one source of truth.

Currently the things that should be rendered conditionally would be FormGroup in ActivationKeyInformation and alert in OscapProfileInformation.

regexowl:
The two PRs above sorted out current problems so closing this issue. We can revisit the implementation of currentStep if we find another use for it in the future.

PR #1723, opened 2024-03-01
V2Wizard: Move OSCAP alert outside of OscapProfileInformation (HMS-2781)

regexowl:
This moves the "Additional customizations" alert outside of OscapProfileInformation which ensures it is not shown in the Oscap expandable on the Review step.

User can see the added customisations in the individual expandables, making the alert non-actionable extra information.


regexowl:
Related to https://github.com/osbuild/image-builder-frontend/issues/1721

lucasgarfield (approved):
Makes sense!

PR #1729, opened 2024-03-04
V2Wizard: Move FormGroup outside of ActivationKeyInformation (HMS-2781)

regexowl:
This moves the FormGroup outside of the ActivationKeyInformation component and puts it higher into index of the Registration step. This way the group label does not get rendered on the Review step.

Before:
image

After:
image


regexowl:
Related to https://github.com/osbuild/image-builder-frontend/issues/1721

PR #1730, opened 2024-03-04
V1/V2Wizard: Capitalise ID on Azure step

regexowl:
This capitalises id -> ID in the test of an alert on Azure step to keep consistent with the capitalisation throughout the step.


lucasgarfield (approved):
Nice catch.

PR #1733, opened 2024-03-04
feat(HMS-3686): Update position of blueprint buttons

avitova:
This is a followup of #1631 and an update according to our new mockups. :)

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -58,12 +58,21 @@ import {
   computeHoursToExpiration,
   timestampToDisplayString,
 } from '../../Utilities/time';
+import { BlueprintActionsMenu } from '../Blueprints/BlueprintActionsMenu';
+import { BuildImagesButton } from '../Blueprints/BuildImagesButton';
+import { DeleteBlueprintModal } from '../Blueprints/DeleteBlueprintModal';
 
 type ImageTableProps = {
   selectedBlueprint?: string | undefined;
+  setSelectedBlueprint: React.Dispatch<
+    React.SetStateAction<string | undefined>
+  >;
 };
 
-const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
+const ImagesTable = ({
+  selectedBlueprint,
+  setSelectedBlueprint,
+}: ImageTableProps) => {
   const [page, setPage] = useState(1);
   const [perPage, setPerPage] = useState(10);
   const { selectedBlueprintVersion } = useGetBlueprintsQuery(
@@ -85,6 +94,7 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
     setPerPage(perPage);
   };
 
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
   const {
     data: blueprintsComposes,
     isSuccess: isBlueprintsSuccess,
@@ -163,6 +173,12 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
   return (
     <>
       <>
+        <DeleteBlueprintModal
+          selectedBlueprint={selectedBlueprint}
+          setSelectedBlueprint={setSelectedBlueprint}
+          setShowDeleteModal={setShowDeleteModal}
+          isOpen={showDeleteModal}
+        />
         <Toolbar>
           <ToolbarContent>
             <ToolbarItem>
@@ -174,6 +190,19 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
                 Create image
               </Link>
             </ToolbarItem>
+            {experimentalFlag && (
+              <>
+                <ToolbarItem>
+                  <BuildImagesButton selectedBlueprint={selectedBlueprint} />
+                </ToolbarItem>
+                <ToolbarItem>
+                  <BlueprintActionsMenu
+                    selectedBlueprint={selectedBlueprint}
+                    setShowDeleteModal={setShowDeleteModal}
+                  />
+                </ToolbarItem>
+              </>
+            )}
             <ToolbarItem variant="pagination" align={{ default: 'alignRight' }}>
               <Pagination
                 itemCount={itemCount}
--- a/src/Components/Blueprints/BlueprintActionsMenu.tsx
+++ b/src/Components/Blueprints/BlueprintActionsMenu.tsx
@@ -1,0 +1,57 @@
+import React, { useState } from 'react';
+
+import {
+  Dropdown,
+  DropdownItem,
+  DropdownList,
+  MenuToggle,
+  MenuToggleElement,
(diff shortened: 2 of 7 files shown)


src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -171,16 +207,53 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
       {itemCount === 0 && <EmptyImagesTable />}
       {itemCount > 0 && (
         <>
+          {/*@ts-ignore*/}
```
lucasgarfield, line 210: Leftover I think, let's remove.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -93,6 +105,30 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
     setPage(1);
     setPerPage(perPage);
   };
+  const [buildBlueprint, { isLoading: imageBuildLoading }] =
+    useComposeBlueprintMutation();
+
+  const onBuildHandler = async () => {
+    selectedBlueprint && (await buildBlueprint({ id: selectedBlueprint }));
+  };
+  const [isOpen, setIsOpen] = useState<boolean>(false);
```
lucasgarfield, line 114: We don't need to specify `<boolean>` here, Typescript can infer it since the hook was initialized with `false`.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -93,6 +105,30 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
...
+  const onSelect = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const [showDeleteModal, setShowDeleteModal] = React.useState(false);
```
lucasgarfield, line 119: `useState` has been imported, so you don't need to use `React.`.

lucasgarfield (changes requested):
This is looking great! One oversight, the changes leaked out of experimental. Here's what I see running `npm run stage-beta`:

image

You'll just need to put all of this behind the experimental flag. You can use conditional rendering like so: `{experimentalFlag && ... }`... you should be able to find plenty of examples of this in the code already.

avitova:
Thank you for your review @lucasgarfield 
I have rebased, addressed the comments, fixed a small bug (when deleting a blueprint, we should set the selectedBlueprint to undefined), and put back the Create image button since that is needed out of experimental. :)

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -84,6 +100,31 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
     setPage(1);
     setPerPage(perPage);
   };
+  const [buildBlueprint, { isLoading: imageBuildLoading }] =
```
lucasgarfield, line 103: We can move this into a new `<BlueprintActionsMenu>` component.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -84,6 +100,31 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
...
+    useComposeBlueprintMutation();
+
+  const onBuildHandler = async () => {
```
lucasgarfield, line 106: We can move this into a new `<BuildImagesButton>` component.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -84,6 +100,31 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
...
+    selectedBlueprint && (await buildBlueprint({ id: selectedBlueprint }));
+  };
+  const [isOpen, setIsOpen] = useState(false);
+  const onSelect = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
+  const [deleteBlueprint] = useDeleteBlueprintMutation({
```
lucasgarfield, line 115: We can move this to `<DeleteBlueprintModal>` I think.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -84,6 +100,31 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
...
+    fixedCacheKey: 'delete-blueprint',
+  });
+  const handleDelete = async () => {
```
lucasgarfield, line 118: We can move this to `<DeleteBlueprintModal>` I think.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -84,6 +100,31 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
...
+    if (selectedBlueprint) {
+      setShowDeleteModal(false);
+      await deleteBlueprint({ id: selectedBlueprint });
+      setSelectedBlueprint(undefined);
+    }
+  };
+  const onDeleteClose = () => {
```
lucasgarfield, line 125: We can move this to `<DeleteBlueprintModal>` I think.

lucasgarfield (changes requested):
Thanks for the updates! I have some opinionated nitpicks next. I’m trying not to block on things that are too nitpicky but I think these won’t be too hard and should be worth it.

I would like to extract some of the JSX into new React components, and move some code (handlers and RTKQ hooks) deeper where it is closer to where it’s needed. This will help keep the scope of `<ImagesTable>` smaller, and keeping the handlers closer to where they called makes the code easier to read and understand. In general I try to avoid passing down props as much as possible, and since RTKQ hooks are cached I prefer to just call them in the component they’re needed in. So with that in mind, I hope you think these requests are reasonable. :sweat_smile:

avitova:
Thank you, Lucas, for the amount of suggestions. I think it all makes sense, and the code looks cleaner. All of the mentioned components are in a sense logical units, so I like the idea of not keeping this all in the images table. :) I decomposed based on your suggestion.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -186,6 +215,7 @@ const ImagesTable = ({ selectedBlueprint }: ImageTableProps) => {
                 isCompact
               />
             </ToolbarItem>
+            )
```
lucasgarfield, line 218: This `)` is getting rendered to the right of the pagination :wink: 

image

lucasgarfield:
Thanks for refactoring, @avitova! This looks really nice! 

Please just fix the tiny bug with the `)` I mentioned and I'll approve and merge.

We will probably also need to do a follow-up and convert these components to use the Redux store when https://github.com/osbuild/image-builder-frontend/pull/1712 gets merged.

lucasgarfield (approved):
Thank you!

PR #1739, opened 2024-03-05
V2Wizard: Update repository icons and names on Packages step (HMS-2864)

regexowl:
This updates the naming and the icons in the "Packages repository" column to be consistent with current mocks.

Before:
image

After:
image

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
@@ -7,7 +7,6 @@ import {
   EmptyStateHeader,
   EmptyStateIcon,
   EmptyStateVariant,
-  Icon,
   Pagination,
   SearchInput,
   ToggleGroup,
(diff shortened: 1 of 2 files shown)


regexowl:
I've noticed my "placeholder red person" made it through the merge :sweat_smile:

I looked for a red fedora only logo in frontend-assets, but the stacked one is closest to what we want. Is there a logo we could use somewhere else?

lucasgarfield:
Did you check here? https://www.redhat.com/en/about/brand/standards/the-hat

regexowl:
> Did you check here? https://www.redhat.com/en/about/brand/standards/the-hat

I did some time ago (from a quick check I think nothing changed) and we could download the logo from there and add it to our own assets, but I was wondering if there is some repo or something where we could just link to to not have to store the image in FE repo.

lucasgarfield:
I don't think so... we can serve it. I don't think we serve any static assets now so you will need to look into patterns for that. I think we might be able to just put the file into `src/images/redhat.png` (for instance), import it and use it as needed... I think that simply importing it will let Webpack know it should be included in the bundle. I wonder if we can reduce the size a bit to keep bundle size down? It's only 2.2 kb but still its bigger than we need...

regexowl:
The logo was added to the frontend-assets repo so we don't need to serve it, just link to it.

lucasgarfield (approved):
Nice job!

PR #1740, opened 2024-03-05
App: experimental features in beta only

lavocatt:
When the unleash switch `image-builder.new-wizard.enabled` will get activated on stage the blueprint and new wizard will be visible on the platform.
To make them only visible in preview we need to change the condition to check if Beta is selected by the user.

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -22,7 +22,6 @@ import {
   Thead,
   Tr,
 } from '@patternfly/react-table';
-import { useFlag } from '@unleash/proxy-client-react';
 import { Link, NavigateFunction, useNavigate } from 'react-router-dom';
 
 import './ImagesTable.scss';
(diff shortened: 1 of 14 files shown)


src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -468,8 +469,10 @@ const Row = ({
 }: RowPropTypes) => {
   const [isExpanded, setIsExpanded] = useState(false);
   const handleToggle = () => setIsExpanded(!isExpanded);
+  const { isBeta } = useGetEnvironment();
   const experimentalFlag =
-    useFlag('image-builder.new-wizard.enabled') || process.env.EXPERIMENTAL;
+    (useFlag('image-builder.new-wizard.enabled') || process.env.EXPERIMENTAL) &&
+    isBeta();
```
ezr-ondrej, line 475: I'd prefer `isPreview()`, but I guess we don't have that in our wrappers, right?

ezr-ondrej (commented):
We'll need to update our tests to say we are in preview, currently we are using just the flag, but apart of that an one nit, nice work! :orange_heart:

lucasgarfield:
Tests should be passing now. :smile:

lucasgarfield:
Just noticed we missed one of the `const experimentalFlag` variables and that somehow we've also introduced a Typing bug so the PR check is failing.

We really should probably consolidate all of these into a single variable.

ezr-ondrej:
> We really should probably consolidate all of these into a single variable.

Depends where we are on the decision for the wizard, but might make sense to have that var renamed from experimental to blueprint (or similar) and store it in the [Blueprint slice](https://github.com/osbuild/image-builder-frontend/pull/1712) as `enabled`, WDYT?

lucasgarfield:
Ugh tests still failing, looking into it...

lucasgarfield (approved):
Passing all tests, going to merge. We will still need to switch the flag value in Unleash to see the new features in stage-preview.

PR #1742, opened 2024-03-05
change the import path for MajorReleasesLifecyclesChart

mgold1234:
this commit change the path for MajorReleasesLifecyclesChart in the review. this fix the correct path of import


regexowl (approved):
Great catch, thank you!

PR #1743, opened 2024-03-05
V2Wizard: Remove aspectRatio

regexowl:
This removes aspectRatio from the ReleaseLifecycle chart. The set aspectRatio was causing incorrect rendering of the annotation line, making time travel during image building inevitable.


regexowl:
This is V2 version of this V1 fix: https://github.com/osbuild/image-builder-frontend/pull/1643

PR #1747, opened 2024-03-06
test: Increase timeout for CreateImageWizard test

regexowl:
The "Step Details image name invalid for more than 63 chars" is often timing out. The timeout was already increased in V2, this makes the same change for V1.


lucasgarfield (approved):
I wonder why this one is so slow?

PR #1752, opened 2024-03-06
Blueprints table: Add title (HMS-3385)

lucasgarfield:
Adds title to blueprints table. Later we will need to extract this into a new component that changes the value depending on whether a blueprint is selected or not (if blueprint is selected, it will need to show blueprint name).

image


regexowl (approved):
LGTM

lavocatt (approved):

PR #1757, opened 2024-03-06
Blueprint: Empty state action is link now

ezr-ondrej:
Convert Empty state action to a link and add test for it.

Screenshot from 2024-03-06 13-59-16


src/test/Components/Blueprints/Blueprints.test.js
```diff
@@ -55,8 +55,14 @@ describe('Blueprints', () => {
       )
     );
 
+    // const { router } = await renderWithReduxRouter('', {});
```
amirfefer, line 58: leftover?

ezr-ondrej: oh yeah, I meant to fix the test to actually test the button works - found the fix :)

PR #1758, opened 2024-03-06
Blueprints Table: Add "new" alert

lucasgarfield:
Adds an alert to explain the basic concept of blueprints to users. Actual copy (and presence of emojis) is pending feedback from UX/copy, this is just a placeholder for now.
image


lucasgarfield:
The emojis have been shot down. :cry: However, we are allowed to keep the "!", which connotates excitement!

lucasgarfield:
Now with 100% less fun (but I did fix the padding):
image

regexowl (approved):
:sad thumbs up:

PR #1764, opened 2024-03-07
V2Wizard: Gate table title behind experimental

regexowl:
This adds gating the ImagesTable title behind experimental so it doesn't render in the "classic" table.


lucasgarfield (approved):
Oops, my bad. Nice catch, thanks.

PR #1769, opened 2024-03-07
ImagesTable: extract toolbar to separate component

ezr-ondrej:
This extracts the toolbar, that was getting bit too big to a separate component. It also adjusts the title to correspond to the filter of the images. If there is a Blueprint selected, it changes the title to '<BlueprintName> images'

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -11,7 +11,6 @@ import {
   Spinner,
   Bullseye,
   Badge,
-  Title,
 } from '@patternfly/react-core';
 import {
   ActionsColumn,
@@ -23,7 +22,7 @@ import {
   Thead,
   Tr,
 } from '@patternfly/react-table';
-import { Link, NavigateFunction, useNavigate } from 'react-router-dom';
+import { NavigateFunction, useNavigate } from 'react-router-dom';
 
 import './ImagesTable.scss';
 import ImagesEmptyState from './EmptyState';
@@ -34,6 +33,7 @@ import {
   GcpDetails,
   OciDetails,
 } from './ImageDetails';
+import ImagesTableToolbar from './ImagesTableToolbar';
 import { AwsS3Instance, CloudInstance, OciInstance } from './Instance';
 import Release from './Release';
 import { ExpiringStatus, CloudStatus } from './Status';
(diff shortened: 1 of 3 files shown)


src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -173,54 +169,13 @@ const ImagesTable = () => {
   return (
     <>
       <>
-        <DeleteBlueprintModal
-          setShowDeleteModal={setShowDeleteModal}
-          isOpen={showDeleteModal}
+        <ImagesTableToolbar
+          itemCount={itemCount}
```
lucasgarfield, line 173: Let's use our Redux `useAppSelector` hook to get all of this data in the component instead of passing props.

src/Components/ImagesTable/ImagesTableToolbar.tsx
```diff
@@ -0,0 +1,114 @@
...
+}
+
+const ImagesTableToolbar: React.FC<imagesTableToolbarProps> = ({
+  itemCount,
+  perPage,
+  page,
+  onSetPage,
+  onPerPageSelect,
+}: imagesTableToolbarProps) => {
+  const experimentalFlag = useExperimentalFlag();
```
lucasgarfield, line 42: I think we might avoid bugs more if this header component is _only_ used for the Blueprints table. Then we can just check the experimentalFlag one time, in the <Table> component, and render either the old header or this new header component conditionally.

src/Components/ImagesTable/ImagesTable.tsx
ezr-ondrej: Ugh? I don't know what you mean, I could probably calculate the itemCount here if I get the composes, but that's about it, everything else is local state in parent component :(

What we **could** do is to have pagination as children and do

```jsx
<ImagesTableToolbar>
  <Pagination ...>
</ImagesTableToolbar>
```

lucasgarfield: Oh, I got confused, this is the pagination related to the images table. I was thinking about Amir's PR that added these properties to the _blueprints_ yesterday. My bad. 

Maybe we should do something similar and add an `imagesSlice` so that the way we handle pagination is consistent.

Anyway, my bad, disregard. :sweat_smile:

src/Components/ImagesTable/ImagesTableToolbar.tsx
ezr-ondrej: It's done now ;)

amirfefer (approved):
Thanks @ezr-ondrej! seems to work as expected

lucasgarfield (approved):
Yep, works great! Thank you!

PR #1771, opened 2024-03-07
Blueprints: keep blueprint card selected after editing (HMS-3385)

### LINKED ISSUE #1770: After Editing a blueprint and save it, the blueprint's card is not marked as selected.

amirfefer:
Fixes #1770.

If you select a blueprint, edit it, and return to the table, the blueprint should still be selected and its images should appear in the table. 

This fixes a bug where the blueprint card was not selected upon returning to the table after editing. Note that the images table was showing only the images associated with that blueprint as intended.


regexowl:
@amirfefer Looks good to me! Could you please add a short description to the PR for the pr-best-practices check? :)

lucasgarfield (approved):
Works perfectly!

PR #1776, opened 2024-03-10
V2Wizard: fix partition at review step (HMS-3754)

mgold1234:
this commit fix the partition table at review step when user choose manual partition

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
@@ -28,6 +28,7 @@ import {
   selectImageTypes,
   removePartition,
   selectPartitions,
+  changePartitionUnit,
 } from '../../../../store/wizardSlice';
 import UsrSubDirectoriesDisabled from '../../UsrSubDirectoriesDisabled';
 import { ValidatedTextInput } from '../../ValidatedTextInput';
@@ -36,6 +37,7 @@ export type Partition = {
   id: string;
   mountpoint: string;
   min_size: string;
+  unit: Units;
 };
 
 const FileSystemConfiguration = () => {
@@ -50,7 +52,8 @@ const FileSystemConfiguration = () => {
       addPartition({
         id,
         mountpoint: '/home',
-        min_size: '1',
+        min_size: UNIT_GIB.toString(),
+        unit: 'GiB',
       })
     );
   };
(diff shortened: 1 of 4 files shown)


src/Components/CreateImageWizardV2/steps/Review/ReviewStepTables.tsx
```diff
@@ -64,7 +66,15 @@ export const FSReviewTable = () => {
               <Th>Minimum size</Th>
             </Tr>
           </Thead>
-          <Tbody data-testid="file-system-configuration-tbody-review"></Tbody>
+          <Tbody data-testid="file-system-configuration-tbody-review">
+            {partitions.map((partition, partitionIndex) => (
+              <Tr key={partitionIndex}>
+                <Td className="pf-m-width-30">{partition.mountpoint}</Td>
+                <Td className="pf-m-width-30">xfs</Td>
+                <Td className="pf-m-width-30">{partition.min_size} </Td>
```
regexowl, line 74: What we want to display in this column is a specific size of the partition and the unit. Here's how the table looks like in V1:

image

regexowl (commented):
Looks good :) I've added one comment.

regexowl:
Also, could you please update the name of the PR to "V2Wizard: fix partition at review step (HMS-3754)"?

The `<component>:<name of PR> (<JIRA ticket>)` is now a standard format we want to use. There's also new bot that checks for best practices which will automatically add a label to your PR when the name is in the specific format :sparkles:

src/Components/CreateImageWizardV2/steps/Review/ReviewStepTables.tsx
regexowl: There are probably two ways to handle this - either we store both the value and the unit and then use those or we could save the size in bytes and do a small computation before displaying the minimum size including the unit. How did we work with the values in V1?

lucasgarfield: Oops, I didn't realize we would also need to remember the display units for the review step.

In that case, let's add a 'displayUnits' field to the partition object we store in the wizardSlice's array of partitions. We will always store the min_size in bytes (I'm hoping that leads to less bugs when creating requests or editing the blueprint).

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -328,35 +330,39 @@ const MinimumSize = ({ partition, units }: MinimumSizePropTypes) => {
...
+        );
       }}
     />
   );
 };
 
 type SizeUnitPropTypes = {
-  units: Units;
-  setUnits: React.Dispatch<React.SetStateAction<Units>>;
+  //setUnits: React.Dispatch<React.SetStateAction<Units>>;
```
regexowl, line 342: Can we remove this comment since we don't need the code inside?

regexowl (commented):
Looks good! Just one small clean up request and then I'd say we're good to merge :)

PR #1778, opened 2024-03-10
V2Wizard: fix remove partition funtionallity (HMS-3753)

mgold1234:
ths commit fix the remove partition functionalli file system step

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
@@ -26,6 +26,7 @@ import {
   changePartitionMinSize,
   changePartitionMountpoint,
   selectImageTypes,
+  removePartition,
   selectPartitions,
 } from '../../../../store/wizardSlice';
 import UsrSubDirectoriesDisabled from '../../UsrSubDirectoriesDisabled';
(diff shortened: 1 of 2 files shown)


lavocatt (approved):
This works fine!

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -163,6 +164,12 @@ const getSuffix = (mountpoint: string) => {
 
 const Row = ({ partition }: RowPropTypes) => {
   const [units, setUnits] = useState<Units>('MiB');
+  const dispatch = useAppDispatch();
+
+  // @ts-ignore
+  const handleRemovePartition = (id: uuidv4) => {
```
regexowl, line 170: Let's solve the problem so we don't have to silence it here. When you remove the `@ts-ignore` the message will say that a value is used as a type. `id` is initiated as a `uuidv4`, but the type of uuid is a string. So this should fix the problem :)
```suggestion
  const handleRemovePartition = (id: string) => {
```

regexowl (commented):
Works great! Just one comment

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
lavocatt: @regexowl what about preventing the usage of ts-ignore?

lavocatt: I mean with some rules for the linter.

regexowl: Thanks for the inspiration! :smile: https://github.com/osbuild/image-builder-frontend/pull/1799

regexowl (approved):

PR #1781, opened 2024-03-11
deps: Remove @types/uuid

regexowl:
This removes @types/uuid from devDependencies. The dependency was initially added as a part of https://github.com/osbuild/image-builder-frontend/pull/1658, but the import which caused the addition was later removed from the code.


regexowl:
Just noticed that other FSC PRs might use the dependency again. I'll keep this one open as a draft for now as a reminder to remove or bump the dependency later.

PR #1782, opened 2024-03-11
V2Wizard: Repository recommendations (HMS-3249)

regexowl:
This PR is a part of Proactive Assistance - Repository recommendations.

New toggle group was added which allows user to see results from non-added EPEL repository in a case that searched term didn't return any results from added custom and distribution repos.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
@@ -1,27 +1,37 @@
 import React, { useEffect, useState } from 'react';
 
 import {
+  Alert,
   Bullseye,
+  Button,
   EmptyState,
   EmptyStateBody,
+  EmptyStateFooter,
   EmptyStateHeader,
   EmptyStateIcon,
   EmptyStateVariant,
+  Icon,
   Pagination,
   PaginationVariant,
+  Popover,
   SearchInput,
+  Text,
+  TextContent,
   ToggleGroup,
   ToggleGroupItem,
   Toolbar,
   ToolbarContent,
   ToolbarItem,
 } from '@patternfly/react-core';
-import { SearchIcon } from '@patternfly/react-icons';
+import { HelpIcon, OptimizeIcon, SearchIcon } from '@patternfly/react-icons';
 import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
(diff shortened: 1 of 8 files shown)


regexowl:
Alright soo... this is a baby step for the repo recommendations. To try it just ignore any custom repos on repository step, go directly to packages and search for "gnuplot".

I went around some problems, but there is a thing. The thing is that workarounds work quite nicely until the very last step of dispatching the repo change. 

I've tried to get recommendations for all the non-added repositories, but once a package gets back in the flat response, there is no easy way to figure out which repo it belongs to so the repo can't be added to the store.

The only way to kind of infer the connection seems be making an API call for each of the repos, getting the complete lists of their rpms and then searching for the packages by package names within the lists...

So for now there's only EPEL hard coded in. My question is - should I try to hack in the support for "all repo search" on the front end side? If so, do we want some limit on the number of searched repos? Thinking this might not scale the best.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -33,93 +43,53 @@ import {
...
+
   const customRepositories = useAppSelector((state) =>
     selectCustomRepositories(state)
   );
+
+  const { data: epelRepo, isSuccess: isSuccessEpelRepo } =
+    useGetRepositoryQuery({
+      uuid:
+        distribution === 'rhel-8'
+          ? '3012f536-eb7c-4cfc-838d-a0cfd553224c'
```
lucasgarfield, line 71: This feels like a "magic uuid"... how do we know the appropriate UUIDs?

regexowl: Those are queried from the content sources API and also seem to be the reason why the search doesn't work on popular repos that are not added - non-added repos don't have any uuid nor existing_name (although the suggested_name seems to be working for search).

image

I wonder if `searchRpm` goes through the added repos only then. I'm not getting any results with popular non-added repo, but when it's added the results are there even though the url of the repo remains the same.

regexowl: Alright, so I was able to get rid of the uuids which indeed felt a bit like magic values. The repo is now fetched based on url which is much easier to confirm in content sources.

lucasgarfield (approved):
Love it!

PR #1786, opened 2024-03-12
Wizard: Add FSC alert for bare metal images

regexowl:
FSC customisations are not being applied to bare metal - installer images, this adds an alert to make users aware of the fact.

fsc-alert


ochosi (approved):
Wonderful - thank you!

PR #1787, opened 2024-03-12
V2Wizard: Add FSC alert for bare metal images

regexowl:
This mirrors the functionality of https://github.com/osbuild/image-builder-frontend/pull/1786 in V2Wizard.

An alert was added to the FSC step to make users aware that customisations are not being applied to bare metal - installer images.

v2-fsc-alert


ochosi (approved):
Very straightforward - thanks for fixing this <3

PR #1793, opened 2024-03-13
deps: Pin @types/uuid

regexowl:
This pins the @types/uuid version as pinned version of dependencies work best with dependabot.


mgold1234 (approved):
/lgtm

PR #1799, opened 2024-03-14
ESLint: Turn ban-ts-comment rule on

regexowl:
This turns ban-ts-comment rule on, adds some configuration to it and fixes the errors caused by changing linting.

The current rule allows ts-expect-error and ts-ignore, but only when accompanied by a description with a minimum length of 5 characters.


lavocatt (approved):
That's amazing! :rocket:

PR #1798, opened 2024-03-14
V1 Wizard: Add search parameter for arch selection (HMS-3684)

ochosi:
Add an optional search parameter to the wizard like so:

```
/insights/image-builder/imagewizard?arch=aarch64
```

This results in the wizard being opened and `aarch64` being pre-selected as the architecture (as opposed to the default `x86_64`). The Insights assistant chat bot and our websites (access.redhat.com and developers.redhat.com) will make use of this feature.

This has been shamelessly cargo-culted from 349377267251a395e2b96a8f1afb97af85a8f401

PR #1805, opened 2024-03-15
V2 Wizard: Add search parameter for arch selection (HMS-3684)

ochosi:
Add an optional search parameter to the wizard like so:

/insights/image-builder/imagewizard?arch=aarch64

This results in the wizard being opened and 'aarch64' being pre-selected as the architecture. The Insights assistant chat bot and our websites (access.redhat.com and developers.redhat.com) will make use of this feature.

This is a follow-up of #1798, porting the same functionality to the v2 wizard.


regexowl (approved):

PR #1806, opened 2024-03-15
V2Wizard: Put Save button into div

regexowl:
This encloses Save button inside a div to make testing with widgetastic.patternfly library's pattern possible.


mgold1234 (approved):
/lgtm

jrusz (approved):
Works great now, thank you!

PR #1807, opened 2024-03-15
Blueprints: Make Blueprints! alert dismissible

### LINKED ISSUE #1804: Blueprints: Make banner Blueprints! dismissible
The alert showing that blueprints are now available should be dismissible.

regexowl:
Fixes #1804

This adds a close button to the Blueprints! alert, making it dismissible.

The dismissal persists only throughout a non-refreshed session.

### DIFF
--- a/src/Components/LandingPage/NewAlert.tsx
+++ b/src/Components/LandingPage/NewAlert.tsx
@@ -1,16 +1,33 @@
-import React from 'react';
+import React, { useState } from 'react';
 
-import { Alert, Text } from '@patternfly/react-core';
+import { Alert, AlertActionCloseButton, Text } from '@patternfly/react-core';
 
 export const NewAlert = () => {
-  return (
-    <Alert title="New in Images: Blueprints!">
-      <Text>
-        Blueprints make it easier for you to manage your images. Images expire
-        after two weeks, but blueprints last forever. Create a blueprint for
-        your “golden image”, modify it over time as your needs change, and use
-        it to build and deploy images on demand.
-      </Text>
-    </Alert>
+  const isAlertDismissed = window.localStorage.getItem(
+    'imageBuilder.alertDismissed'
   );
+  const [displayAlert, setDisplayAlert] = useState(!isAlertDismissed);
+
+  const dismissAlert = () => {
+    setDisplayAlert(false);
+    window.localStorage.setItem('imageBuilder.alertDismissed', 'true');
+  };
+
+  if (displayAlert) {
+    return (
+      <Alert
+        title="New in Images: Blueprints!"
+        actionClose={<AlertActionCloseButton onClose={dismissAlert} />}
+      >
+        <Text>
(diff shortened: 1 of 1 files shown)


ochosi:
LGTM!
I presume for anything other than per-session we'd have to resort to cookies..?

regexowl:
> LGTM! I presume for anything other than per-session we'd have to resort to cookies..?

I presume so as well :sweat_smile: Persistence throughout sessions would probably need cookies or be linked to an account id.

lucasgarfield:
Should we try this using the web storage api so we have session persistence?
https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API

I think we can do it very easily. We just need to use `localStorage.setItem('imageBuilder.alertDismissed', 'true');` whenever the alert is dismissed and conditionally render the alert based on something like this: `const alertDismissed = localStorage.getItem('alertDismissed') === 'true';`.

regexowl:
> Should we try this using the web storage api so we have session persistence? https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API
> 
> I think we can do it very easily. We just need to use `localStorage.setItem('imageBuilder.alertDismissed', 'true');` whenever the alert is dismissed and conditionally render the alert based on something like this: `const alertDismissed = localStorage.getItem('imageBuilder.alertDismissed') === 'true';`.

Great point, localStorage seems to do the trick! I've updated the PR

lucasgarfield:
I checked with the Hybrid Cloud Console folks and it is fine if we use local storage, we just need to make sure we use unique keys for things (I proposed we just prefix everything with `imageBuilder.`).

src/Components/LandingPage/NewAlert.tsx
```diff
@@ -1,16 +1,43 @@
...
-        Blueprints make it easier for you to manage your images. Images expire
-        after two weeks, but blueprints last forever. Create a blueprint for
-        your “golden image”, modify it over time as your needs change, and use
-        it to build and deploy images on demand.
-      </Text>
-    </Alert>
-  );
+  const [displayAlert, setDisplayAlert] = useState(true);
+
+  useEffect(() => {
```
lucasgarfield, line 8: Could we get rid of this useEffect if we just declared `alertDismissed` before the useState hook for displayAlert and initialized it like `const [displayAlert, setDisplayAlert] = useState(alertDismissed || true);`?

lucasgarfield: This seems to be working on my machine, I was able to get rid of both useEffect hooks:
image

regexowl: You're right, this works perfectly fine. I've pushed the changes.

lucasgarfield: Wow, I'm glad we have this figured out! Now we can do all sorts of fun stuff with persistence across sessions!

lucasgarfield (approved):
Thank you!

PR #1808, opened 2024-03-17
V2Wizrad: fix validation at file system step (HMS-3733)

mgold1234:
this commit fix validation when usr choose duplicate mount point andwhen there is no root partition


src/store/wizardSlice.ts
```diff
@@ -127,6 +128,7 @@ const initialState: wizardState = {
   fileSystem: {
     mode: 'automatic',
     partitions: [],
+    hasErrorOnSubmit: false,
```
lucasgarfield, line 131: I'm not really comfortable with the idea of keeping error state in the store, it feels like that is something we should be figuring out using a validator or helper function where needed.

So in this case we need to know both in the custom footer (to decide whether or not to disable the next button) and in the component (to decide whether to render the error messages). I think that should be possible without adding a new field to the store's wizardSlice.

Let's remove this and see if we can get it working without it.

lucasgarfield (changes requested):

src/store/wizardSlice.ts
mgold1234: We agree to change the name to isNextButtonPristine and initialize to true, and when the user presses on the 'Next' button this variable will change to false, and will not return to true.
if there are duplicate rows after the user presses on the next button at the first time, the error message will show immediately.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -36,3 +38,38 @@ export const isBlueprintNameValid = (blueprintName: string) =>
...
 };
+export const isFileSystemConfigValid = (partition: Partition[]) => {
+  const duplicate = getDuplicateMountPoints(partition);
+  return !(
+    (duplicate?.duplicates?.length !== 0 &&
+      duplicate?.duplicates?.length !== undefined) ||
+    duplicate?.root === false
+  );
+};
+export const getDuplicateMountPoints = (partition: Partition[]) => {
```
lucasgarfield, line 49: I think we can refactor this function to make it a lot simpler. Right now it can return either undefined, or an object with a duplicates field that can be undefined, as well as a root field that doesn't seem related to the duplicate mountpoints... it is very confusing.

I think that it should return either (1) an array of partition ids that have duplicate mount points or (2) an array of duplicate mountpoints (just strings). In case there are no duplicate mount points, it would just return an empty array.

lucasgarfield: I don't think we need it anymore do we? The root '/' mountpoint is provided by default, and you aren't allowed to add another one - you can only add mountpoints from the list:
image

mgold1234: there is an error in case that there is no root 
 getDuplicateMountPoints(partitions)?.root === false && (
          <div style={{ padding: '15px 0' }}>
            <Alert
              isInline
              variant="danger"
              title="No root partition configured."
            />
we can remove it?

lucasgarfield: I think so... I don't see how you could ever get to the state where you have two '/' partitions since '/' isn't an option the user can select.

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -111,6 +111,7 @@ export const mapRequestToState = (request: BlueprintResponse): wizardState => {
     fileSystem: {
       mode: 'automatic',
       partitions: [],
+      isNextButtonPristine: false,
```
regexowl, line 114: Should this be set to `true`? Since we are mapping the request to the wizard state the form has technically not been interacted with as of yet and the Next button should be pristine.

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -63,6 +119,17 @@ const FileSystemConfiguration = () => {
       {partitions?.find((partition) =>
         partition?.mountpoint?.includes('/usr')
       ) && <UsrSubDirectoriesDisabled />}
+      {!isNextButtonPristine &&
+        getDuplicateMountPoints(partitions)?.length !== 0 &&
+        getDuplicateMountPoints(partitions)?.length !== undefined && (
+          <div style={{ padding: '15px 0' }}>
```
regexowl, line 125: The spacing looks a bit weird here. I think we can go with default one and remove the "style"

```suggestion
          <div>
```

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -148,6 +150,12 @@ const CreateImageWizard = ({ startStepIndex = 1 }: CreateImageWizardProps) => {
     selectBlueprintDescription(state)
   );
   const activationKey = useAppSelector((state) => selectActivationKey(state));
+  const partitions = useAppSelector((state) => selectPartitions(state));
+  const [isNextDisable, setNextDisable] = useState(false);
```
regexowl, line 154: Very nit picky, but can we update the names of the variables to `isNextDisabled` and `setNextDisabled` ?

```suggestion
  const [isNextDisabled, setNextDisabled] = useState(false);
```

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -148,6 +150,12 @@ const CreateImageWizard = ({ startStepIndex = 1 }: CreateImageWizardProps) => {
...
+
+  useEffect(() => {
+    setNextDisable(false);
```
regexowl, line 157:
```suggestion
    setNextDisabled(false);
```

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -260,7 +268,12 @@ const CreateImageWizard = ({ startStepIndex = 1 }: CreateImageWizardProps) => {
           <WizardStep
             name="File system configuration"
             id="step-file-system"
-            footer={<CustomWizardFooter disableNext={false} />}
+            footer={
+              <FileSystemStepFooter
+                setNextDisable={setNextDisable}
+                isNextDisable={isNextDisable}
```
regexowl, line 274:
```suggestion
                setNextDisabled={setNextDisabled}
                isNextDisabled={isNextDisabled}
```

src/store/wizardSlice.ts
```diff
@@ -376,6 +381,9 @@ export const wizardSlice = createSlice({
     ) => {
       state.fileSystem.partitions = action.payload;
     },
+    changeIsNextButtonPristine: (state) => {
+      state.fileSystem.isNextButtonPristine = false;
+    },
```
regexowl, line 386: Could there be a use case for needing to set `isNextButtonPristine` to true again? I'd say let's either set the value based on action.payload or we can rename the function so the name fits the functionality a bit better. Something like `setIsNextButtonPristineToFalse`, what do you think?

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -28,22 +30,76 @@ import {
...
   min_size: string;
 };
+type FileSystemConfigurationStepFooterPropTypes = {
+  setNextDisable(isNextDisable: boolean): void;
+  isNextDisable: boolean;
+};
+
+export const FileSystemStepFooter: React.FunctionComponent<
+  FileSystemConfigurationStepFooterPropTypes
+> = ({ setNextDisable, isNextDisable }) => {
```
regexowl, line 56: The prop typing should be applied to the props.

```suggestion
export const FileSystemStepFooter = ({
  setNextDisable,
  isNextDisable,
}: FileSystemConfigurationStepFooterPropTypes) => {
```

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -28,22 +30,76 @@ import {
...
+  } else isValid = false;
+
+  const onValidate = () => {
+    dispatch(changeIsNextButtonPristine());
+    if (!isValid) {
+      setNextDisable(true);
+    } else {
+      goToNextStep();
+    }
+  };
```
regexowl, line 78: The pristinity (is that a word? :sweat_smile:) of the button might need to be reset at some point, which is connected to my other comment in wizardSlice file.

Try going to FSC step, creating a duplicate mount point and clicking Next. Everything renders correctly and `isNextButtonPristine` is set to false. So far so good.

Resolve the duplicity, the Next button should be enabled again.

But now if you again add a duplicate mount point, the error is rendered immediately and the Next button stays enabled until it is clicked.

image

Can we reset the pristinity of the Next button after the duplicity was resolved somehow? If that's too complicated, could we at least disable the Next button when the alert is rendered?

regexowl (commented):
Added some comments :nerd_face:

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -148,6 +150,12 @@ const CreateImageWizard = ({ startStepIndex = 1 }: CreateImageWizardProps) => {
     selectBlueprintDescription(state)
   );
   const activationKey = useAppSelector((state) => selectActivationKey(state));
+  const partitions = useAppSelector((state) => selectPartitions(state));
+  const [isNextDisabled, setNextDisabled] = useState(false);
+
+  useEffect(() => {
+    setNextDisabled(false);
+  }, [partitions, dispatch]);
```
lucasgarfield, line 158: I think that we can drop this lower into the child `<FileSystemStepFooter>` component.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -36,3 +38,24 @@ export const isBlueprintNameValid = (blueprintName: string) =>
 export const isBlueprintDescriptionValid = (blueprintDescription: string) => {
   return blueprintDescription.length <= 250;
 };
+export const isFileSystemConfigValid = (partition: Partition[]) => {
```
lucasgarfield, line 41: Let's refactor this a bit so it is consistent, and change `partition` to `partitions`.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -36,3 +38,24 @@ export const isBlueprintNameValid = (blueprintName: string) =>
...
+  const duplicates = getDuplicateMountPoints(partition);
+  return !(duplicates?.length !== 0);
```
lucasgarfield, line 43: This is a bit confusing, what if we just did `duplicates.length === 0 ? true : false`?

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -36,3 +38,24 @@ export const isBlueprintNameValid = (blueprintName: string) =>
...
+};
+export const getDuplicateMountPoints = (partitions: Partition[]) => {
+  if (!partitions) {
+    return [];
+  }
+  const mountPointFreqs = {} as { [key: string]: number | boolean };
```
lucasgarfield, line 49: I think this is a bit confusing, especially how the value can be either a number or boolean.

What if we used a `Set()` to keep track of the duplicates? I'm not sure if we should return the set itself or convert it to an array... we don't use sets that often in our code base so returning an array might make it a bit easier to work with, because we will be able to call array methods on it...

lucasgarfield: I think if we take this approach we can probably also get rid of that early return...

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -28,22 +30,77 @@ import {
...
+  setNextDisabled(isNextDisabled: boolean): void;
+  isNextDisabled: boolean;
+};
+
+export const FileSystemStepFooter = ({
+  setNextDisabled,
+  isNextDisabled,
+}: FileSystemConfigurationStepFooterPropTypes) => {
+  const { goToNextStep, goToPrevStep, close } = useWizardContext();
+  let isValid = false;
```
lucasgarfield, line 59: I think we can use a `useState()` hook here instead of `let`...

lucasgarfield (changes requested):

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
lucasgarfield: I agree, let's aim for 1:1 behavior with what we have in prod for now.

lucasgarfield (approved):
Working nicely!

PR #1811, opened 2024-03-18
fixes(HMS-3778): Show image name only after it is loaded

avitova:
Sometimes, when the response from the API was a bit slow, and the name was not yet loaded, while the id was selected, this would end up by showing "undefined images".


ezr-ondrej (approved):
Might be better to have a loader, but this is better then what we have so :+1: 
Thank you @avitova, lets play the CI dance now xD

PR #1816, opened 2024-03-19
V2Wizard: Add search parameter for target selection (HMS-3684)

regexowl:
Add an optional search parameter to the wizard like so:

`/insights/image-builder/imagewizard?target=iso`
or
`/insights/image-builder/imagewizard?target=qcow`

This results in wizard being opened and iso or qcow target being pre-selected. The Insights assistant chat bot will make use of this feature.

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -53,6 +53,7 @@ import {
   selectGcpShareMethod,
   selectImageTypes,
   selectRegistrationType,
+  addImageType,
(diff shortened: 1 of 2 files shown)


ochosi:
My comment about the search parameter value name also applies to this PR. So whatever you go with in the end, it'd be great to make it a consistent choice :)
https://github.com/osbuild/image-builder-frontend/pull/1817#issuecomment-2009697302

lucasgarfield (approved):
I also took the liberty of changing `qcow` to `qcow2` here just so we can get the review backlog down faster... Great work @regexowl, thank you!

PR #1817, opened 2024-03-19
V1Wizard: Add search parameter for target selection (HMS-3684)

regexowl:
Add an optional search parameter to the V1Wizard like so:

`/insights/image-builder/imagewizard?target=iso`
or
`/insights/image-builder/imagewizard?target=qcow`

This results in wizard being opened and iso or qcow target being pre-selected. The Insights assistant chat bot will make use of this feature.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
+++ b/src/Components/CreateImageWizard/formComponents/TargetEnvironment.js
@@ -18,6 +18,7 @@ import {
 import { HelpIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
 import { useField } from 'react-final-form';
+import { useSearchParams } from 'react-router-dom';
 
 import { useGetArchitecturesQuery } from '../../../store/imageBuilderApi';
 import { provisioningApi } from '../../../store/provisioningApi';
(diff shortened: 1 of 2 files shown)


regexowl:
Alright, the newly added tests were failing, but when looking into it the failure came from the "Keyboard accessibility target environment tiles are keyboard selectable" test. Which means that the async is not asyncing correctly somewhere. Tried to debug that part, but it seems to be a bit more tangled.

For now the url params tests have been moved before keyboard accessibility which magically fixed it.

(the `testTile` function should be awaited in the test mentioned above, which explains which the failure fell through to the url params tests, but manifested only for target params as the tests needed to interact with the part of code common for both target param and target tiles)

ochosi:
I tested this and it works as advertised - wonderful job, thank you @regexowl ! <3

I have one nitpick and one question.

**Nitpick:** I'd go with `qcow2` for the value of the search parameter since that's the complete extension. When I tested this, it originally failed because that's what I supplied (until I noticed I had to remove the `2`).

**Question:** Do we want to allow multiple targets? This currently does not work yet, but could be a thing to consider for the future (there's no specific use case yet though).

lucasgarfield (approved):
Wow, great work @regexowl! I took the liberty of swapping `qcow` to `qcow2` for you, hope you don't mind. :wink:

@ochosi, I think supporting multiple targets in the future is a good idea, let's plan for that. For now in the V1 wizard I think this is good enough for an MVP so I'm going to merge it.

PR #1818, opened 2024-03-19
V2Wizard: Initialize packages with OpenSCAP

regexowl:
This initializes packages upon a selection of an OpenSCAP profile.

Current logic also fixes a bug previously present in V1 - when changing an OpenSCAP profile with other previously chosen packages, the packages outside of the profile persist and don't get removed.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
@@ -34,6 +34,10 @@ import {
   selectEnabledServices,
   changeDisabledServices,
   changeEnabledServices,
+  clearOscapPackages,
+  addPackage,
+  selectPackages,
+  removePackage,
 } from '../../../../store/wizardSlice';
 
 const ProfileSelector = () => {
@@ -42,6 +46,7 @@ const ProfileSelector = () => {
   let disabledServices = useAppSelector(selectDisabledServices);
   let enabledServices = useAppSelector(selectEnabledServices);
   const release = useAppSelector(selectDistribution);
+  const packages = useAppSelector(selectPackages);
   const dispatch = useAppDispatch();
   const [profileName, setProfileName] = useState<string | undefined>('None');
   const [isOpen, setIsOpen] = useState(false);
@@ -96,6 +101,25 @@ const ProfileSelector = () => {
     }
   }, [data]);
 
+  useEffect(() => {
+    dispatch(clearOscapPackages());
+    for (const pkg in data?.packages) {
+      if (
+        packages.map((pkg) => pkg.name).includes(data?.packages[Number(pkg)])
+      ) {
+        dispatch(removePackage(data?.packages[Number(pkg)]));
+      }
+      dispatch(
+        addPackage({
+          name: data?.packages[Number(pkg)],
+          summary: 'Required by chosen OpenSCAP profile',
+          repository: 'distro',
+          isRequiredByOpenScap: true,
+        })
+      );
+    }
+  }, [data?.packages, dispatch]);
+
   const handleToggle = () => {
     if (!isOpen) {
       refetch();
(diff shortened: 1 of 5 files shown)


src/store/wizardSlice.ts
```diff
@@ -438,6 +438,11 @@ export const wizardSlice = createSlice({
         1
       );
     },
+    clearOscapPackages: (state) => {
+      state.packages = state.packages.filter(
+        (pkg) => pkg.repository !== 'openscap'
```
lucasgarfield, line 443: I’m a little wary of using the `repository` field to keep track of whether a package is included due to OpenSCAP or not… it just feels a confusing to me because it isn’t _technically_ true – the actual repository the package would get installed from would still be the `distro` repo, right?.

What if we add a field to `IBPackagewithRepositoryInfo` that is specifically related to whether the package is being added due to an OpenSCAP profile? Something like `isRequiredByOpenScapProfile`? And I suppose it would be a boolean. 

That's just one idea, there might be a better way to do it. But my main point is that I think we should keep track of this in a new field in the object.

Also a bit of refactoring I’d like to request that's not really related to this PR but would be nice – right now `repository` is a string. Can we create a new type for this? Just something like `type PackageRepository = 'distro' | ...`.

lucasgarfield (changes requested):

src/store/wizardSlice.ts
regexowl: Good point, I've decoupled the oscap info from the repository. The added field it is!

The `PackageRepository` type is also added with the possible values being 'distro', 'custom' and an empty uninitialized string.

lucasgarfield: Awesome, thank you!

lucasgarfield (changes requested):
I found an edge case. If you add a package (for instance `firewalld`) and then add an OpenSCAP profile that includes that package, the package will appear in the list twice:

image

I suppose the "correct" behavior would be that the "OpenSCAP version" of the package would over-write the pre-existing one?

regexowl:
> I found an edge case. If you add a package (for instance `firewalld`) and then add an OpenSCAP profile that includes that package (CIS Red Hat Enterprise Linux 9 Benchmark for Level 1 - Workstation), the package will appear in the list twice:
> 
> I suppose the "correct" behavior would be that the "OpenSCAP version" of the package would over-write the pre-existing one?

Great catch! I've updated the code so there shouldn't be duplicates now.

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
```diff
@@ -98,6 +103,25 @@ const ProfileSelector = () => {
     }
   }, [data]);
 
+  useEffect(() => {
```
lucasgarfield, line 106: This is this components 3rd useEffect... it is getting a bit hard to keep track of all the side effects...

I'm also a bit worried about a possible async bug. I'm not sure it is a valid concern, but could it ever happen that this useEffect doesn't get triggered because data?.packages doesn't change in time? For instance, if the Next button somehow gets clicked very quickly after changing the profile and the request is still in flight?

Could we run this code inside the `handleSelect()` function instead? We might need to create a small helper function to keep the code clean. Maybe we could make it async and somehow ensure that data.packages is correct, perhaps by waiting if isFetching is true or something?

lucasgarfield (approved):

lucasgarfield (changes requested):
Accidentally approved, should have requested changes :facepalm:

src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
regexowl: Fair point. I've previously tried to handle dispatching within the `handleSelect` function. 

What I got stuck on is the asynchronicity of store update - when handling things within `handleSelect` (or a helper function called from it) the store stays one step behind the desired state - when you select a profile, no packages get added. When you change the profile, only packages for the first profile get added.

The only solution I've found so far was the `useEffect`. :thinking:

lucasgarfield (approved):
I don't love the useEffect but it works and we're going to try to do some pair programming on the OpenSCAP step to get it finished tomorrow, so I'm just going to get this in now so that we have the correct behavior in the packages step for that work.

PR #1827, opened 2024-03-20
Experimental: fix the ability to force experimental in developement

ezr-ondrej:
The useExperimental broken the ability to enforce experimental on developement environment, it is a good idea to be able to do so at least for now.
This fixes it.


lucasgarfield (approved):
Thank you!

PR #1830, opened 2024-03-21
fix one warning at useEffect on Oscap file

mgold1234:
this commit fix a warning-
React Hook useEffect has a missing dependency: 'change'. Either include it or remove the dependency array  react-hooks/exhaustive-deps


lucasgarfield (approved):
Thank you!

PR #1832, opened 2024-03-21
Blueprint: Validate Blueprint name is not empty (HMS-3801)

ezr-ondrej:
We are now validating Blueprint name on the backend to be non-empty. Adds same validation for frontend

Refs HMS-3801

### DIFF
--- a/src/Components/CreateImageWizardV2/validators.ts
+++ b/src/Components/CreateImageWizardV2/validators.ts
@@ -31,7 +31,9 @@ export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
 };
 
 export const isBlueprintNameValid = (blueprintName: string) =>
-  blueprintName.length > 0 && blueprintName.length <= 100;
+  blueprintName.length >= 2 &&
+  blueprintName.length <= 100 &&
+  /\w+/.test(blueprintName);
 
 export const isBlueprintDescriptionValid = (blueprintDescription: string) => {
   return blueprintDescription.length <= 250;
--- a/src/Components/CreateImageWizardV2/steps/Details/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Details/index.tsx
@@ -64,7 +64,9 @@ const DetailsStep = () => {
         />
         <FormHelperText>
           <HelperText>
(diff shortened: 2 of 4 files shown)


regexowl:
The code looks good to me, one very nit picky question - do we want to set the minimal length to three? As that's what the current regex allows even thought the helper text insists 1-100 characters are possible:

image

Other option would be to adjust the regex for example with a non-capturing optional group so we really do allow 1 or 2 char names.

regexowl:
Maybe something like this? `^[a-zA-Z0-9](?:[a-zA-Z0-9-_ ]*[a-zA-Z0-9])?$`

image

Could be also probably simplified to `^[a-z0-9](?:[a-z0-9-_ ]*[a-z0-9])?$` since the test is set as case insensitive.

src/Components/CreateImageWizardV2/validators.ts
```diff
@@ -31,7 +31,9 @@ export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
 };
 
 export const isBlueprintNameValid = (blueprintName: string) =>
-  blueprintName.length > 0 && blueprintName.length <= 100;
+  blueprintName.length >= 2 &&
+  blueprintName.length <= 100 &&
+  /\w+/.test(blueprintName);
```
ezr-ondrej, line 36: @regexowl I've changed this to only require at least two characters, I've realized we do not care about cool names, as long as we can transpose those to machine friendly image name.
I've filed https://github.com/osbuild/image-builder/pull/1093 to make the transposition happen :)

regexowl: I see, nice! Should this be blocked until we merge https://github.com/osbuild/image-builder/pull/1093 then? To avoid allowing cool names that will break the image during build?

ezr-ondrej: well we allow them today, so this would be improvement to today's situation, but sure, I'm all for fixing it properly on backend first ;)

regexowl: Ah yeah, you're completely right. I got the validation mixed up, sorry about that :sweat_smile: I think we can merge then! :)

regexowl (approved):
Thank you!

PR #1833, opened 2024-03-21
V2 Wizard: Add blueprints exporting (HMS-3687)

avitova:
As a part of exporting, we might want to show a toast notification in the future. But we can not really get a status response from the "browser" if the download was really performed, so I would maybe leave this to a followup to come up with a solution for that.


src/Components/Blueprints/BlueprintActionsMenu.tsx
```diff
@@ -59,11 +67,39 @@ export const BlueprintActionsMenu: React.FunctionComponent<
...
+    suggestedName: blueprintName.replace(/\s/g, '_').toLowerCase() + '.json',
+    types: [
+      {
+        description: 'Text file',
+        accept: { 'text/plain': ['.json'] },
+      },
+    ],
+  };
+
+  const fileHandle = await (window as any).showSaveFilePicker(opts);
```
avitova, line 101: The linter fails because of this, do we have a solution other than ignoring the linter?

src/Components/Blueprints/BlueprintActionsMenu.tsx
```diff
@@ -28,7 +33,10 @@ export const BlueprintActionsMenu: React.FunctionComponent<
   };
   const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
   const navigate = useNavigate();
-
+  const { data: blueprintDetails } = useGetBlueprintQuery({
+    id: selectedBlueprintId || '',
+  });
```
amirfefer, line 38: This will fetch the details even when the user doesn't need it,
How about using `useLazyQuery`? 
```js
  const [trigger, { data, error, isLoading }] = useLazyQuery(api.endpoints.getBlueprint);
```

The `trigger` can be invoked in the onClickHandler, after the user clicks for downloading the blueprint file

avitova:
I forgot we do docs-first open api generating. xD This should work now. :) Thanks for the comment, Amir, let me know what you think now. :+1: I do not think that the PR failure is relevant. :)

lucasgarfield:
Can you add a commit description please?

src/Components/Blueprints/BlueprintActionsMenu.tsx
```diff
@@ -60,11 +71,36 @@ export const BlueprintActionsMenu: React.FunctionComponent<
         >
           Edit details
         </DropdownItem>
-        <DropdownItem>Download blueprint (.json)</DropdownItem>
+        <DropdownItem
+          onClick={() => {
+            trigger({ id: selectedBlueprintId || '' });
+          }}
+        >
+          Download blueprint (.json) <BetaLabel />
```
lucasgarfield, line 79: Since we've got Unleash in place, do we really need the `<BetaLabel />`? I think we can just ship this straight to prod-stable when we're ready.

lucasgarfield (changes requested):

src/Components/Blueprints/BlueprintActionsMenu.tsx
ezr-ondrej: I think that's the thing ... we should ship it to prod-stable with beta label :P
It should mark the exported schema as "we can't promise it'll be importable next month".

src/Components/Blueprints/BlueprintActionsMenu.tsx
```diff
@@ -60,11 +71,36 @@ export const BlueprintActionsMenu: React.FunctionComponent<
...
+        </DropdownItem>
         <DropdownItem onClick={() => setShowDeleteModal(true)}>
           Delete blueprint
         </DropdownItem>
       </DropdownList>
     </Dropdown>
   );
 };
+
+async function handleExportBlueprint(
```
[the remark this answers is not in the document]
lucasgarfield: Alright, as per discussion on Slack, this situation is different enough that the way you've done it here is better. Namely, we don't have to trigger a lazyQuery in an onClick handler when using `<a>` since the data is already available.

src/Components/Blueprints/BlueprintActionsMenu.tsx
```diff
@@ -60,11 +71,36 @@ export const BlueprintActionsMenu: React.FunctionComponent<
...
+  blueprintName: string | undefined,
+  blueprint: BlueprintResponse | undefined
```
lucasgarfield, line 91: Can we refactor this so that blueprintName is only `string` and blueprint is only `BlueprintResponse`? Let's deal with the undefined case at the call site.

src/Components/Blueprints/BlueprintActionsMenu.tsx
```diff
@@ -28,7 +33,13 @@ export const BlueprintActionsMenu: React.FunctionComponent<
   };
   const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
   const navigate = useNavigate();
-
+  const [trigger, { data: blueprintDetails }] = useLazyGetBlueprintQuery();
+  const blueprintName = blueprintDetails?.name;
+  useEffect(() => {
```
lucasgarfield, line 38: I haven't had time to look at this closely, but if we could do this without a useEffect somehow that would be really nice.

lucasgarfield (approved):
Works great and nice job getting rid of the useEffect, this is a lot cleaner!

avitova:
Thanks!:)

PR #1846, opened 2024-03-26
store: Turn off immutableCheck and serializableCheck 

regexowl:
This turns off immutableCheck and serializableCheck which removes warnings `SerializableStateInvariantMiddleware took XXms, which is more than the warning threshold of 32ms.` from the test output.

Another option would be to disable the middleware in development mode as it can causes a slowdown.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -86,7 +86,10 @@ startAppListening({
 // Listener middleware must be prepended according to RTK docs:
 // https://redux-toolkit.js.org/api/createListenerMiddleware#basic-usage
 export const middleware = (getDefaultMiddleware: Function) =>
-  getDefaultMiddleware()
+  getDefaultMiddleware({
+    immutableCheck: false,
+    serializableCheck: false,
+  })
     .prepend(listenerMiddleware.middleware)
     .concat(
       promiseMiddleware,


regexowl:
Wait, no, there are some checks running for 120ms... I still think we can probably just turn those of as this doesn't have any effect outside of development mode. I'll update the PR and then we can discuss if we want to merge it.

lucasgarfield:
Hm... I'm not 100% sure on this. Even though RTK uses Immer, I would assume the immutableCheck ensures that what we're doing is legal according to Immer... and similarly, the serializableCheck seems important too. It's easy to forget the "rules" of RTK state management and removing these makes me a bit uncomfortable.

I guess the real question I'd like to answer is, why are we exceeding the warning threshold? Is that a red flag that we've done something inefficiently?

lucasgarfield:
Here's a good post by the RTK author about why this check is useful, for instance: https://stackoverflow.com/a/72095474

lucasgarfield:
I'm trying to clean up PR back log, I'm going to close this if that's okay with you @regexowl (re-open it if not :sweat_smile:) because as I mentioned in the comments, I think these are worthwhile checks.

PR #1849, opened 2024-03-26
feat(HMS-3688): Add blueprints import dialog

avitova:


src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+          variant="info"
+          title="For expected file format, refer to documentation"
+          ouiaId="InfoAlert"
+        >
+          <AlertActionLink component="a" href="#">
+            View documentation
+          </AlertActionLink>
+        </Alert>
+      )}
+      <FileUpload
```
amirfefer, line 58: You can limit the file  type and size with this prop
```js
         dropzoneProps={{
            accept: { 'text/csv': ['.csv'] },
            maxSize: 1024,
            onDropRejected: handleFileRejected
          }}
```

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+      {false && (
+        <Alert
+          variant="info"
+          title="For expected file format, refer to documentation"
+          ouiaId="InfoAlert"
+        >
+          <AlertActionLink component="a" href="#">
+            View documentation
+          </AlertActionLink>
+        </Alert>
```
amirfefer, line 56: `false` makes it never being render,  is this  for validation?
We can use `validated` prop and wrap the <FileUpload> with <FormGroup>

amirfefer (commented):
Thank you @avitova ! 👍

src/Components/Blueprints/ImportBlueprintModal.tsx
avitova: We do not have the docs yet, there should be a link, but since we have no place to refer to, I wanted we might want to hide this for now.

ezr-ondrej: dead code looks weird and might accidently stay there forever, IMO its better just add the code once we have the docs and have a reminder in Jira instead of in code :)

avitova:
I addressed all the concerns, and I do not think the PR failure is relevant. :)

regexowl:
@avitova Can you please resolve the conflict?

avitova:
@regexowl Conflicts resolved :)

regexowl (approved):
LGTM, thank you! :sparkles:

PR #1853, opened 2024-03-27
ESLint: Add recommended redux ruleset

regexowl:
Trying to figure out the cause of warnings in https://github.com/osbuild/image-builder-frontend/pull/1846

Didn't come up with the solution yet, but I've noticed we were missing redux ruleset in our ESLint config, this adds the recommended rules.


lucasgarfield (dismissed):
Cool! Looks like we were doing a pretty decent job, seems like only the selectors were problematic... thanks for fixing those!

regexowl:
Fixed some new linting errors from other merges, should be all good now! :wrench:

lucasgarfield (approved):

Issue #1824, opened 2024-03-20
Bug- recreate image


kelseamann:
Was giving a demo of this yesterday, noticed that when I select "Recreate image" from the kebab in the images menu, the Review step in the wizard flashes for a second before the image is recreated. confusing for the user- take this out so they know there is only one way to access the review screen, which is through "Edit blueprint".

lucasgarfield:
We've removed "Recreate image" functionality, even for blueprint-less images (users should migrate to using blueprints and the 'build images' function).

PR #1855, opened 2024-03-27
fix(HMS-3851): remove 'recreate image' from kebab dropdown

amirfefer:
This PR solves issue #1824 and removes the `recreate image` option.

Screenshot 2024-03-27 at 19 15 34

As you can see, `Download compose request` is the only left item, but I'm not sure what is the point to keep it? Blueprints import/export should replace it, no?

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -426,6 +426,7 @@ type AwsRowPropTypes = {
 
 const AwsRow = ({ compose, composeStatus, rowIndex }: AwsRowPropTypes) => {
   const navigate = useNavigate();
+  const experimentalFlag = useExperimentalFlag();
 
   const target = <AwsTarget compose={compose} />;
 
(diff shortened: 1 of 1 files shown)


regexowl:
There is also "Share to new region" action for aws images, but I see what you mean. I guess for now we should keep the option to download compose request until export/import are implemented (unless they already are and I've missed it :sweat_smile:)

Also the tests for ImagesTable will need a slight adjustment as `Images Table › check recreate action` is failing with the Recreate action no longer being that. 

That being said - how should we handle compatibility between V1Wizard + image based table and V2Wizard + blueprints. We have only one ImagesTable which means recreating would no longer be available to original Image Builder users, is that right?

amirfefer:
> That being said - how should we handle compatibility between V1Wizard + image based table and V2Wizard + blueprints. We have only one ImagesTable which means recreating would no longer be available to original Image Builder users, is that right?

That's right @regexowl , I wrapped it with `experimentalFlag`, so the `recreate image` is removed only when V2 wizard is turn on.

regexowl:
Nice one, thank you! I suppose we can clean the actions a bit more when we fully migrate to Blueprints/V2Wizard, but for now this works nicely :)

There is still an option to recreate an image present in the Instance column of the ImagesTable. I've created an issue for that: https://github.com/osbuild/image-builder-frontend/issues/1884

PR #1862, opened 2024-03-28
Minor improvements to Blueprints UI (HMS-3848)

avitova:
The position of each header for the images table differs based on the content. We might want to try and have it fixed (percentage?), but I think it might make sense to leave the formatting based on the content. Therefore, only for empty table, I tried to make it a bit more consistent. Let me know what you think about hardcoding this.


ezr-ondrej (approved):
Looks great to me :orange_heart:

regexowl (approved):
Looks great, thank you!

PR #1863, opened 2024-03-28
V2Wizard: Update "Too many results" copy

regexowl:
This updates copy shown in a table when too many results are fetched. Also a period was added to the end of the empty state bodies.

Without an exact match
image

With an exact match
image


lucasgarfield (approved):
Thank you!

PR #1869, opened 2024-04-02
V2Wizard: Add full bottom pagination to Repositories and Packages

regexowl:
This adds full pagination to the bottom of Repositories and Packages steps which will allow the user to get to the very end of the tables in a case of many results.

image


lucasgarfield (approved):
Thank you!

PR #1848, opened 2024-03-26
templates/testing: Update sel testing image frontend (IQE-2748)

akhil-jha:
We have switched from using a custom selenium image to the upstream variant so updating the testing template accordingly.

PR #1870, opened 2024-04-02
templates/testing: IQE-2748 Revert Update sel testing image frontend

akhil-jha:
New sel image is failing the test pipeline because it does not have a controlled shutdown command, making the pod stuck and resulting into timeout.


regexowl:
Comparing with the [original PR](https://github.com/osbuild/image-builder-frontend/pull/1848/files) - should the VNC_GEOMETRY value be reverted as well?

akhil-jha:
> Comparing with the [original PR](https://github.com/osbuild/image-builder-frontend/pull/1848/files) - should the VNC_GEOMETRY value be reverted as well?

It's the same value. Just no new line diff.

regexowl:
Ah yeah, my bad. Can't read :sweat_smile:

regexowl (approved):

PR #1871, opened 2024-04-02
V1Wizard: Add full bottom pagination to Repositories

regexowl:
This adds a full pagination to the bottom of the Repositories step. This allows the user to get to the very end of the repositories list easily.


lucasgarfield (approved):
good idea & lgtm

PR #1872, opened 2024-04-02
V2Wizard: Center messages in packages table (HMS-2781)

lucasgarfield:
The copy in the packages table (too many results to show, etc...) was not centered because the table actually has 5 columns, not 4. This fixes the problem.

image


mgold1234 (approved):
/lgtm

PR #1879, opened 2024-04-03
WizardV2: add 'register-later' to state mapper

### LINKED ISSUE #1865: V2: Editing blueprint shows incorrect Registration status
When I pick "Register later" during blueprint creation, create it and then try to Edit it, it will show that I want to automatically Register instead of "Register later".

amirfefer:
Fix #1865 

This PR maps the `register-later` subscription type during editing.


regexowl (approved):
Works great, thanks!

PR #1886, opened 2024-04-04
Blueprints: Update button in header

regexowl:
This updates the Create blueprint and Import buttons based on recent mocks.

mocks: 
image


lucasgarfield (approved):
Looks really sharp with the icons!

PR #1887, opened 2024-04-04
Blueprints: No Build Image button without a blueprint

regexowl:
This hides the Build Image button and blueprint actions kabob when no blueprint is selected as per recent mocks.

mocks:
image


lucasgarfield (approved):
Thanks!

PR #1889, opened 2024-04-04
WizardV2: map oscap to wizard request

amirfefer:
This PR adds the Openscap step to the wizard request mapper, enabling creation and editing.
I simplify the Openscap component, by reducing the duplicated data in RTK query and wizard slice.
I added a fixed `max-height` for better scrolling, and added description for each policy.

https://github.com/osbuild/image-builder-frontend/assets/11807069/c36b7d1e-56d8-4bb2-be3f-d19fc3b4aab8

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Oscap/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Oscap/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import { Button, Form, Text, Title } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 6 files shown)


src/Components/CreateImageWizardV2/steps/Oscap/index.tsx
```diff
@@ -10,9 +10,14 @@ import { useAppSelector } from '../../../../store/hooks';
 import { selectDistribution } from '../../../../store/wizardSlice';
 
 const OscapStep = () => {
-  const prefetchOscapProfile = imageBuilderApi.usePrefetch('getOscapProfiles');
+  const prefetchOscapProfile = imageBuilderApi.usePrefetch(
+    'getOscapProfiles',
+    {}
+  );
   const release = useAppSelector(selectDistribution);
-  prefetchOscapProfile({ distribution: release });
```
amirfefer, line 15: prefetching in the top level of the component is a bad practice, it might triggered every re-render.
I guess a better prefetch would need here, not for the profiles list, but for the profiles details.

src/Components/CreateImageWizardV2/steps/Review/Footer/Footer.tsx
```diff
@@ -33,6 +35,10 @@ const ReviewWizardFooter = () => {
       reset: resetCreate,
     },
   ] = useCreateBlueprintMutation({ fixedCacheKey: 'createBlueprintKey' });
+
+  // initialize the server store with the data from RTK query
+  const oscapData = useOscapData(useAppSelector(selectProfile));
+  const serverStore = { ...oscapData };
```
amirfefer, line 41: This is my suggestion for reducing data duplication, instead of storing and mange the same data twice in redux store and RTK query, we can reuse the data from RTK and create a `ServerStore`

src/Components/CreateImageWizardV2/steps/Oscap/index.tsx
lucasgarfield: I support removing pre-fetching for now and adding back later in a very specific PR.

lucasgarfield (approved):
Thanks, @amirfefer!

PR #1895, opened 2024-04-05
Promote OpenSCAP to prod stable (HMS-2077)

kingsleyzissou:
- disable `stig` and `stig_gui` profiles since they don't boot
- ~add `firewalld` to packages needed for `PCI-DSS` profile.~
- promote OpenSCAP to prod stable

I removed the second step since https://github.com/osbuild/image-builder/pull/1110 landed

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.js
+++ b/src/Components/CreateImageWizard/CreateImageWizard.js
@@ -614,8 +614,7 @@ const CreateImageWizard = () => {
 
   // Only allow oscap to be used in Beta even if the flag says the feature is
   // activated.
-  const oscapFeatureFlag =
-    useFlag('image-builder.wizard.oscap.enabled') && isBeta();
+  const oscapFeatureFlag = useFlag('image-builder.wizard.oscap.enabled');
   let initialState = requestToState(composeRequest, isProd(), oscapFeatureFlag);
   const stepHistory = formStepHistory(
     composeRequest,
--- a/src/Components/CreateImageWizard/formComponents/Oscap.tsx
+++ b/src/Components/CreateImageWizard/formComponents/Oscap.tsx
@@ -197,16 +197,27 @@ const ProfileSelector = ({ input }: ProfileSelectorProps) => {
                 key="oscap-none-option"
               />,
             ].concat(
-              profiles.map((profile_id, index) => {
-                return (
-                  <OScapSelectOption
-                    key={index}
-                    profile_id={profile_id}
-                    setProfileName={setProfileName}
-                    input={value}
-                  />
-                );
-              })
+              profiles
+                // stig and stig_gui don't boot at the moment,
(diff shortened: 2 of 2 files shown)


src/Components/CreateImageWizard/formComponents/Oscap.tsx
```diff
@@ -91,6 +91,15 @@ const ProfileSelector = ({ input }: ProfileSelectorProps) => {
       change('kernel', data.kernel);
     }
     if (data?.services?.enabled) {
+      if (data.services.enabled.includes('firewalld')) {
```
kingsleyzissou, line 94: This is a tad bit ugly (but also temporary). Any suggestions on how to handle this better, or is this okay for now?

kingsleyzissou: It can get removed once https://github.com/osbuild/image-builder/pull/1110 lands

src/Components/CreateImageWizard/CreateImageWizard.js
```diff
@@ -614,8 +614,7 @@ const CreateImageWizard = () => {
 
   // Only allow oscap to be used in Beta even if the flag says the feature is
   // activated.
-  const oscapFeatureFlag =
```
kingsleyzissou, line 617: Is there anything else I might have missed for us to move this to `prod-stable`?

src/Components/CreateImageWizard/formComponents/Oscap.tsx
regexowl: https://github.com/osbuild/image-builder/pull/1110 seems close to landing. If it gets merged today/tomorrow then I'd say let's wait for it and remove the temporary part right away so we don't forget to remove it :upside_down_face: What do you think?

regexowl:
There have been some problems with ephemeral so I guess the pr_check failure might not be entirely relevant.

But still - @jrusz are there some tests checking whether OSCAP step is visible/hidden in prod that would need to get updated when it's moved to stable?

regexowl:
Yup, this change would make OpenSCAP available in stage-stable as well as the visibility of the step will rely on on the oscap unleash flag only.

src/Components/CreateImageWizard/formComponents/Oscap.tsx
jrusz: So https://github.com/osbuild/image-builder/pull/1110 got merged, when is this planned to go in?

kingsleyzissou: I'll push the changes for this today

kingsleyzissou:
> That won't work now, image builds on stage are failing, we need updates, Sanne is working on it.

My bad, thanks for letting me know!

regexowl (approved):
LGTM, thanks! :rocket:

PR #1902, opened 2024-04-08
WizardV2: Filesystem draggable mount point

amirfefer:
This PR adds support for re-ordering mount points in file system step by adding `draggable` support for the table.

https://github.com/osbuild/image-builder-frontend/assets/11807069/7ecda742-6288-42db-ab7e-489db4f1e3b2

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
@@ -3,7 +3,6 @@ import React, { useEffect, useState } from 'react';
 import {
   Alert,
   Button,
-  Popover,
   Text,
   TextContent,
   TextInput,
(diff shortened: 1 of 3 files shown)


ezr-ondrej (approved):
Thanks @amirfefer :+1: Works well and code looks similar to what we had in the v1 wizard :orange_heart: 

Do you think it makes sense to add any tests here?

ezr-ondrej:
@regexowl @mgold1234 WDYT about this one? It might be good to get this one in first and only then #1905 :)

mgold1234:
yes it may be good to merge this first, I am reviewing it now @ezr-ondrej

amirfefer:
Testing drag and drop with jsdom is not trivial, jsdom has it own benefits, but  it still an abstraction of a real browser.
React testing library official `drag` event doesn't work with jsdom - https://testing-library.com/docs/example-drag/

regexowl (approved):
Looks good and works great, thanks a lot @amirfefer!

Even though it would be nice to have a unit test for this in the future, testing draggability in jsdom is not easy (which I guess was the reason why the functionality didn't have any tests in V1 either :sweat_smile:). So let's not block on that

ezr-ondrej:
> Testing drag and drop with jsdom is not trivial

Thanks for looking into it :orange_heart:

Issue #1904, opened 2024-04-08
Partitioning units are missbehaving


ezr-ondrej:
[Screencast from 2024-04-08 21-18-48.webm](https://github.com/osbuild/image-builder-frontend/assets/2884324/b19d1269-d917-409d-be5f-b87cf789f54f)

regexowl:
Fixed by https://github.com/osbuild/image-builder-frontend/pull/1905

PR #1905, opened 2024-04-08
WizzardV2: Filesystem step partitioning suffix and size (HMS-2863)

### LINKED ISSUE #1903: Second part of partition is not filled in a second column
[Screencast from 2024-04-08 21-18-26.webm](https://github.com/osbuild/image-builder-frontend/assets/2884324/93663cae-4cfa-40ba-85a2-8dc2902553a6)

ezr-ondrej:
Fixes #1903, #1904


mgold1234:
@ezr-ondrej I compared the behavior to production,
and in production the user can add '/' to the beginning of the suffix, and with your changes I see that the user can't, don't we want to keep the behavior the same?


prodcution behavior:
Screenshot 2024-04-09 at 11 02 21

local env:
Screenshot 2024-04-09 at 11 09 16

ezr-ondrej:
Ah, sure :) I did not realize, but it make sense, I'd guess the `/` is required tho, so I'll try to make it so it appears automatically :)

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -380,11 +380,15 @@ const MinimumSize = ({ partition, units }: MinimumSizePropTypes) => {
   const conversionFactor = getConversionFactor(units);
 
   const convertToDisplayUnits = (minSize: string) => {
-    return (parseInt(minSize) / conversionFactor).toString();
+    return minSize.length > 0
+      ? (parseInt(minSize) / conversionFactor).toString()
+      : '';
```
regexowl, line 385: Should there be a warning/error in a case of empty minimum size? Or we can simply default to zero as the empty value. What do you think?

regexowl: Ah, default to zero seems to be consistent with V1 behaviour so I'd go in that direction
v1-min-size

ezr-ondrej: not allowing empty value during editing makes for a weird behavior, when you want to delete current value and type in a new value.
We could add default to 0 on Blur, but that's still invalid value, so it should then fail validation, right? :thinking:

ezr-ondrej: Right, so I've fixed it by adding the validator :)

ezr-ondrej: Ah I see, when you add 0, it allows you to type after 0... that makes sense actually :) let me do that :+1:

ezr-ondrej: oki, done, no empty strings allowed! ;)

ezr-ondrej:
Rebased, so this is ready for re-review :)

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -369,19 +379,21 @@ const MinimumSize = ({ partition, units }: MinimumSizePropTypes) => {
     <ValidatedTextInput
       ariaLabel="minimum partition size"
       helperText=""
```
regexowl, line 381: Let's add some error message
image


```suggestion
      helperText="Must be larger than 0"
```

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -369,19 +379,21 @@ const MinimumSize = ({ partition, units }: MinimumSizePropTypes) => {
...
-        dispatch(
-          changePartitionMinSize({
-            id: partition.id,
-            min_size: convertToBytes(minSize),
-          })
-        );
-        dispatch(
-          changePartitionUnit({ id: partition.id, unit: partition.unit })
-        );
+        if (minSize === '' || /^\d+$/.test(minSize)) {
```
regexowl, line 386: With minSize defaulting to 0, the first part of condition can be probably removed?

```suggestion
        if (/^\d+$/.test(minSize)) {
```

regexowl (commented):
Looks great! Added two teeny tiny nit picks :)

src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
ezr-ondrej: This is required here, that's what actually transform empty field to 0 :)
But I've removed it from validator :)

regexowl: Ah yeah, I see :sweat_smile:

regexowl (approved):
Works like magic, thanks!!

PR #1909, opened 2024-04-09
V2Wizard: Fix plural sources

jrusz:
We're actually selecting a singular source to share the image with. It's the same in V1 wizard.


regexowl (approved):
Nice catch! Thank you :sparkles:

PR #1918, opened 2024-04-10
V2 Wizard: Add feature flag for import/export (HMS-3906)

avitova:
So I was thinking how to make it possible to develop the import/export, and added also an experimental flag together to the import/export that should be disabled. Does this make sense to you?


lucasgarfield (approved):
Looks good @avitova!

PR #1919, opened 2024-04-10
V2Wizard: Replace "Adding recommended repo" alert with modal

regexowl:
This replaces the previously used Alert displayed when adding a package from repo recommendations with a modal as per mocks.

mocks:
image


lucasgarfield:
I'm getting a little worried the Wizard's state is becoming too confusing here - we've got a `Payload Repositories` field, a `Custom Repositories` field, and now a `Recommended Repositories` field. 

I think that we should collapse these into a single field, `SelectedRepositories`, and then we can generate the payload and custom repositories fields when we generate the request.

Is there any particular reason we would want to keep track of the `Recommended Repositories` separately?

lucasgarfield (approved):
I created an issue to track the Wizard state refactor and bug I mentioned. I think it is better to refactor the Wizard state first, then make sure the behavior is correct on the Repositories step (specifically, that repositories added via the modal are selected in the table on the Repositories step).

PR #1925, opened 2024-04-11
V2Wizard: Allow enabling repos and update state logic (HMS-3249)

regexowl:
This commit contains two changes:
1. When a package from popular repository gets added, the EPEL repository gets enabled in content services in case it wasn't before.
2. The logic of rendering states for the different toggle combinations on Packages step got updated.


regexowl:
Okay, the complexity of the Packages step is starting to be a bit too much. Not saying nightmarish levels, but close. 

I think this should do the trick, but I'm also afraid this one will require a thorough review.

lucasgarfield (approved):
This works really well! I agree about the complexity growing but I think we can make it manageable with a bit of refactoring.

PR #1928, opened 2024-04-12
Blueprints: Add import Wizard (HMS-3690)

avitova:
Sooo...I think we are ready for a discussion on what we want to export AND import, since this is ready. :)

### DIFF
--- a/src/Components/Blueprints/ImportBlueprintModal.tsx
+++ b/src/Components/Blueprints/ImportBlueprintModal.tsx
@@ -13,6 +13,12 @@ import {
   Modal,
   ModalVariant,
 } from '@patternfly/react-core';
+import { useNavigate } from 'react-router-dom';
+
+import { BlueprintResponse } from '../../store/imageBuilderApi';
+import { wizardState } from '../../store/wizardSlice';
+import { resolveRelPath } from '../../Utilities/path';
+import { mapRequestToState } from '../CreateImageWizardV2/utilities/requestMapper';
 
 interface ImportBlueprintModalProps {
   setShowImportModal: React.Dispatch<React.SetStateAction<boolean>>;
@@ -26,6 +32,9 @@ export const ImportBlueprintModal: React.FunctionComponent<
     setShowImportModal(false);
   };
   const [jsonContent, setJsonContent] = React.useState('');
+  const [importedBlueprint, setImportedBlueprint] =
+    React.useState<wizardState>();
+  const [isInvalidFormat, setIsInvalidFormat] = React.useState(false);
   const [filename, setFilename] = React.useState('');
   const [isLoading, setIsLoading] = React.useState(false);
   const [isRejected, setIsRejected] = React.useState(false);
@@ -36,11 +45,13 @@ export const ImportBlueprintModal: React.FunctionComponent<
   ) => {
     setFilename(file.name);
     setIsRejected(false);
+    setIsInvalidFormat(false);
   };
   const handleClear = () => {
     setFilename('');
     setJsonContent('');
     setIsRejected(false);
+    setIsInvalidFormat(false);
   };
   const handleTextChange = (
     _: React.ChangeEvent<HTMLTextAreaElement>,
@@ -49,10 +60,19 @@ export const ImportBlueprintModal: React.FunctionComponent<
     setJsonContent(value);
   };
   const handleDataChange = (_: DropEvent, value: string) => {
-    setJsonContent(value);
+    try {
+      const importedBlueprint: BlueprintResponse = JSON.parse(value);
+      const importBlueprintState = mapRequestToState(importedBlueprint);
+      setImportedBlueprint(importBlueprintState);
+      setJsonContent(value);
+    } catch (error) {
+      setIsInvalidFormat(true);
+    }
   };
   const handleFileRejected = () => {
     setIsRejected(true);
+    setJsonContent('');
+    setFilename('');
   };
   const handleFileReadStarted = () => {
     setIsLoading(true);
@@ -60,6 +80,7 @@ export const ImportBlueprintModal: React.FunctionComponent<
   const handleFileReadFinished = () => {
     setIsLoading(false);
   };
+  const navigate = useNavigate();
 
   return (
     <Modal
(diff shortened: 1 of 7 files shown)


src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -96,15 +102,25 @@ export const ImportBlueprintModal: React.FunctionComponent<
             <HelperText>
               <HelperTextItem variant={isRejected ? 'error' : 'default'}>
                 {isRejected
-                  ? 'Must be a JSON file no larger than 1 KB'
+                  ? 'Must be a JSON file no larger than 6 KB'
```
avitova, line 105: In case we want to export even the gpg key and overall everything, 1 KB does not seem to be enough.

lucasgarfield:
Should this be working, @avitova? When I tried importing a blueprint, I just got kicked back to the table.

avitova:
@lucasgarfield Yeah, you are right, let me correct this:)

amirfefer (commented):
Thanks @avitova, this looks and works great! 👍 
I have a suggestion regarding the json parsing

src/Components/CreateImageWizardV2/ImportImageWizard.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const dispatch = useAppDispatch();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const locationState = location.state as { blueprint?: string };
+  const blueprint = locationState?.blueprint;
+  useEffect(() => {
+    if (blueprint) {
+      try {
+        const importedBlueprint: BlueprintResponse = JSON.parse(blueprint);
+        const importBlueprintState = mapRequestToState(importedBlueprint);
```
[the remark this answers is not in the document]
avitova: I agree, thank you for this idea. :)

avitova:
I addressed your comments @amirfefer :) Ready for re-review.

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -87,23 +109,35 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+                  : isInvalidFormat
+                  ? 'Not compatible with the blueprints format.'
                   : 'Upload a JSON file'}
               </HelperTextItem>
             </HelperText>
           </FormHelperText>
         </FormGroup>
         <ActionGroup>
-          <Button type="button">Review and finish</Button>
+          <Button
```
amirfefer, line 130: can you please add `ouid` prop for testing?

amirfefer (commented):
Thanks @avitova, tested a full export-import scenario, and it works well!  👍 
I left a few minor inline comments, mainly for wrong casting.

This feature is guarded by a feature flag, so I'm ok to add unit  tests as follow-up. react testing library has `upload` API for mimicking upload input, like [here](https://github.com/osbuild/image-builder-frontend/pull/1960/files#diff-f3153ea47023210c886d905eabef84599e8b22b39c5abb7fdaad04af76dca81cR98-R102)

src/Components/CreateImageWizardV2/ImportImageWizard.tsx
```diff
@@ -0,0 +1,34 @@
...
+      navigate(resolveRelPath(''));
+      dispatch(
+        addNotification({
+          variant: 'warning',
+          title: 'No blueprint was imported',
+        })
+      );
+    }
+  }, [blueprint, dispatch]);
+  return <CreateImageWizard startStepIndex={1} />;
```
amirfefer, line 31: This is the reason why this fails to build, `startStepIndex` prop is no longer exists, and doesn't needed here, the default is 1 anyway.

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -49,17 +60,28 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
   };
   const handleDataChange = (_: DropEvent, value: string) => {
-    setJsonContent(value);
+    try {
+      const importedBlueprint: BlueprintResponse = JSON.parse(value);
+      const importBlueprintState = mapRequestToState(importedBlueprint);
+      setImportedBlueprint(importBlueprintState);
+      setJsonContent(value);
+    } catch (error) {
+      console.log(error);
```
regexowl, line 69: I don't think we should be logging errors to the console as those have very little value to user.

Since this would be an error with the blueprint format, would a toast notification like "Failed to parse the blueprint" or something similar work?

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -87,23 +109,36 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+            type="button"
+            isDisabled={isRejected || isInvalidFormat || !jsonContent}
+            onClick={() =>
+              navigate(resolveRelPath(`imagewizard/import`), {
+                state: { blueprint: importedBlueprint },
+              })
+            }
+            ouiaId="import-blueprint-finish"
+          >
+            Review and finish
```
regexowl, line 140: More of a question - do we want to drop the user on the first step or on the review step after importing a blueprint?

Dropping them directly on the Review with all the information prefilled makes more sense to me as it saves the user the work of clicking through the Wizard. Especially since the left side navigation is disabled so there's no quick way to get to the review.

(also I thought the importing doesn't work for a moment, because I didn't notice that there are already some preselected values in the Wizard :sweat_smile:)

What do you think? @avitova @lucasgarfield

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -87,23 +109,36 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+          </Button>
```
regexowl, line 141: When I tried to import a compose request, it didn't work which makes sense. The request is missing values that blueprint needs to have.

But the failure currently crashes the service and leaves the user stuck on the `import` path which means that even refreshing the page will not help.

image

regexowl (commented):
Thank you for working on this! Added some comments

src/Components/Blueprints/ImportBlueprintModal.tsx
avitova: Oh yeah, you are right. This was a leftover from testing. :eyes:

src/Components/Blueprints/ImportBlueprintModal.tsx
avitova: We discussed this with Kelsea on one of the UI meetings. It makes sense to drop them on the first page. In the future, we will possibly omit some of the data (and also some of the mandatory data) that we allow to export/import, so users would have to go through the wizard anyways:)

avitova:
So I added a test and rebased. The tests mostly cover which files should be accepted and which should not. It is quite plain, but I think that we can go with this for a while.

regexowl:
@avitova Looks great! Can you please resolve the conflict and rebase?

src/Components/Blueprints/ImportBlueprintModal.tsx
regexowl: Got it, makes sense!

avitova:
@regexowl rebased:)

regexowl (approved):
Excellent, thank you! Looks and works great :rocket:

avitova:
:partying_face: Thank you!

PR #1935, opened 2024-04-15
V2 Wizard: Add custom file system to request customizations (HMS-3690)

lucasgarfield:
The blueprint request now contains a field for the filesystem in its customizations.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
@@ -322,7 +322,7 @@ const MountpointSuffix = ({ partition }: MountpointSuffixPropTypes) => {
           })
         );
       }}
-      aria-label="text input example"
+      aria-label="mountpoint suffix"
       ouiaId="mount-suffix"
     />
   );
--- a/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
+++ b/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
@@ -8,6 +8,7 @@ import {
   CreateBlueprintRequest,
   Customizations,
   DistributionProfileItem,
+  Filesystem,
   GcpUploadRequestOptions,
   ImageRequest,
   ImageTypes,
@@ -43,6 +44,8 @@ import {
   selectRegistrationType,
   selectServerUrl,
   wizardState,
+  selectFileSystemPartitionMode,
+  selectPartitions,
 } from '../../../store/wizardSlice';
 import {
   convertSchemaToIBCustomRepo,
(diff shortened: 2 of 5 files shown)


src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -322,7 +322,7 @@ const MountpointSuffix = ({ partition }: MountpointSuffixPropTypes) => {
           })
         );
       }}
-      aria-label="text input example"
+      aria-label="mountpoint suffix"
```
mgold1234, line 325: can you please change it to "Mount point suffix text input" as it was in V1wizard?

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -8,6 +8,7 @@ import {
   CreateBlueprintRequest,
   Customizations,
   DistributionProfileItem,
+  Filesystem,
```
mgold1234, line 11:
```suggestion
  FileSystem,
```

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -317,6 +320,18 @@ const getOpenscapProfile = (state: RootState): OpenScap | undefined => {
   return undefined;
 };
 
+const getFileSystem = (state: RootState): Filesystem[] | undefined => {
+  const mode = selectFileSystemPartitionMode(state);
+  if (mode === 'manual') {
+    const partitions = selectPartitions(state);
+    const filesystem = partitions.map((partition) => {
```
mgold1234, line 327:
```suggestion
    const fileSystem = partitions.map((partition) => {
```

mgold1234:
there is a problem with getFileSystem function at 'requestMapper.tsx when I create an image I got an error - 
```Blueprint could not be created
Status code 400: cannot parse request body: json: cannot unmarshal string into Go struct field Filesystem.customizations.filesystem.min_size of type uint64```

lucasgarfield:
> there is a problem with getFileSystem function at 'requestMapper.tsx when I create an image I got an error -
> 
> ```
> Status code 400: cannot parse request body: json: cannot unmarshal string into Go struct field Filesystem.customizations.filesystem.min_size of type uint64```
> ```

Oh nice catch. The image builder API specifies the type of the min_size as `any` so I thought it would be okay to use a `string`. I'll change to `number`.

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
lucasgarfield: I agree we should use `FileSystem` camel case but this is an import from the `image-builder` API so I'll suggest we keep it and don't rename it.

mgold1234 (approved):
/lgtm

PR #1941, opened 2024-04-16
ImagesTable: Update column headers

regexowl:
This updates the name of column headers as per recent mocks.

mocks:
image


lucasgarfield (approved):
Thank you!

PR #1942, opened 2024-04-16
LandingPage: Remove "New" label from Immutable tab

regexowl:
This removes "New" label from the Immutable tab.


regexowl:
Should we remove the "New" label from Immutable tab on the Landing page? I can't see it in any of the recent mocks. How long is a new feature new?

lucasgarfield (approved):
Thank you!

PR #1944, opened 2024-04-16
WizardV2: validate uniqueness of name (HMS-3801)

ezr-ondrej:
Fetches Bluepritnts with the same name as inputed and when there is already one in the database, fails the validation.

This is implemented by adding async validation to the StateValidatedInput, which is bit complex as we wanted to add debounce for that validation as well, so it doesn't fire too frequently.

### DIFF
--- a/src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
+++ b/src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import {
   HelperText,
@@ -31,6 +31,10 @@ interface StateValidatedTextInputPropTypes extends TextInputProps {
   ariaLabel: string | undefined;
   helperText: string | undefined;
   validator: (value: string | undefined) => boolean;
+  asyncValidator?: (
+    value: string,
+    callback: (isValid: boolean) => void
+  ) => void;
   value: string;
   placeholder?: string;
 }
@@ -43,6 +47,7 @@ export const StateValidatedInput = ({
   ariaLabel,
   helperText,
   validator,
+  asyncValidator,
   value,
   placeholder,
   onChange,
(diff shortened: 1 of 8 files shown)


src/Components/CreateImageWizardV2/steps/Details/index.tsx
```diff
@@ -27,13 +28,28 @@ const DetailsStep = () => {
...
     _event: React.FormEvent<HTMLInputElement>,
     name: string
   ) => {
     dispatch(changeBlueprintName(name));
   };
 
+  const { data: sameNameBlueprint, isFetching } = useGetBlueprintsQuery(
+    { name: blueprintName },
+    { skip: blueprintName.length < 2 }
+  );
```
ezr-ondrej, line 42: We're firing for every name change, which is bit unfortunate, but I didn't manage to make debounce work for this.

amirfefer: have you tried something like:

```js
  const [debouncedBlueprintName, setDebouncedBlueprintName] = useState(blueprintName);

  useEffect(() => {
    const handler = setTimeout(() => {
      setDebouncedBlueprintName(blueprintName);
    }, 500); 
    return () => {
      clearTimeout(handler);
    };
  }, [blueprintName]);

  const { data: sameNameBlueprint, isFetching } = useGetBlueprintsQuery(
    { name: debouncedBlueprintName },
    { skip: debouncedBlueprintName.length < 2 }
  );
```

ezr-ondrej: I haven't, I guess this would work, but it would be setting the name to the local state on top of the global one.
Tho I guess it's simpliest solution we have now.

ezr-ondrej:
Actually #1949 might make it bit easier :)

src/Components/CreateImageWizardV2/steps/Details/index.tsx
ezr-ondrej: I've fixed this in new version, but unfortunatelly it requires the caller to provide a memoized function, so it's prone to mistakes.

ezr-ondrej:
I've rebased this to use the new state validation :)

src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
```diff
@@ -31,6 +31,10 @@ interface StateValidatedTextInputPropTypes extends TextInputProps {
   ariaLabel: string | undefined;
   helperText: string | undefined;
   validator: (value: string | undefined) => boolean;
+  asyncValidator?: (
+    value: string,
+    callback: (isValid: boolean) => void
+  ) => void;
```
ezr-ondrej, line 37: We now have second validator prop, that allows for async validation, when provided, it waits for the async validation to finish to mark validation success.

lucasgarfield:
Is this ready for review, @ezr-ondrej ?

ezr-ondrej:
> Is this ready for review, @ezr-ondrej ?

Definitelly ready for review, but if we prefer to wait until we figure out the validation properly, we can do that as well :)

ezr-ondrej:
I'm keeping it open as a reminder, but once https://github.com/osbuild/image-builder-frontend/pull/2133 is in, I'll rewrite it :)

ezr-ondrej:
Superseeded by #2178

PR #1945, opened 2024-04-16
V2Wizard: Select a repository when added from recommendations

regexowl:
This selects a repository on the Custom repositories step when it's added on the Packages step from the recommendations.

The check box is also disabled as removing the repository would have to trigger removal of the added packages as well.

And expandable with explanation about the disabled check box was added.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
@@ -49,6 +49,7 @@ import {
   selectArchitecture,
   selectCustomRepositories,
   selectDistribution,
+  selectRecommendedRepositories,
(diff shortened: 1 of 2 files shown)


regexowl:
I know this is quite ugly, but hear me out - this should allow fixing the bug with recommended repos not being selected without complete refactor of the repositories step and state. 

The refactoring is high on my to do list, alas it will be a bit more complex so my current plan is to take a look at it after the PA features are delivered.

lucasgarfield:
I think just extracting all that conditional logic into clearly named functions would go a long ways towards making this code cleaner - what do you think?

regexowl:
Yeah, great point. I think that helped a lot with the readability :relieved: also added a comment to each of the function.

lucasgarfield (approved):
I think it reads much nicer now, thanks!

PR #1946, opened 2024-04-16
Blueprints: Remove icon from "Create blueprint" button

regexowl:
This removes the plus icon from the "Create blueprint" button as per recent review.


amirfefer (approved):
Thanks @regexowl 👍

PR #1948, opened 2024-04-16
Blueprints: Update Import modal padding

regexowl:
The `<ActionGroup>` should be a child of `<Form>`, moving it improves the padding on the Import modal.


amirfefer (approved):
Thanks @regexowl  👍

PR #1949, opened 2024-04-16
WizardV2: Validate steps through redux state (HMS-3801)

ezr-ondrej:
Store validation status in redux state.
This is bit complex on the redux side, but pretty simple on the components. It allows for reuse of the validation state instead of revalidating wherever needed.

ATM it's just a PoC and it's applied just to Details step, next step would be to apply everywhere, but details are best to showcase the advantage, notice disabling the create button on Review step:

[Screencast from 2024-04-17 20-36-48.webm](https://github.com/osbuild/image-builder-frontend/assets/2884324/15fde597-05b2-4c40-8b50-a1d02002e639)

### DIFF
--- a/src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
+++ b/src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
@@ -7,15 +7,104 @@ import {
   TextInputProps,
 } from '@patternfly/react-core';
 
+import { useAppDispatch, useAppSelector } from '../../store/hooks';
+import {
+  setStepInputValidation,
+  selectInputValidation,
+} from '../../store/wizardSlice';
+
 interface ValidatedTextInputPropTypes extends TextInputProps {
   dataTestId?: string | undefined;
   ouiaId?: string;
   ariaLabel: string | undefined;
   helperText: string | undefined;
-  validator: (value: string | undefined) => Boolean;
+  validator: (value: string | undefined) => boolean;
   value: string;
   placeholder?: string;
 }
+
+interface StateValidatedTextInputPropTypes extends TextInputProps {
+  dataTestId?: string | undefined;
+  ouiaId?: string;
+  stepId: string;
+  inputId: string;
+  ariaLabel: string | undefined;
+  helperText: string | undefined;
+  validator: (value: string | undefined) => boolean;
+  value: string;
+  placeholder?: string;
+}
+
+export const StateValidatedInput = ({
+  dataTestId,
+  ouiaId,
+  stepId,
+  inputId,
+  ariaLabel,
+  helperText,
+  validator,
+  value,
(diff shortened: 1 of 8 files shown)


src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
```diff
@@ -7,16 +7,105 @@ import {
...
+  validator: (value: string | undefined) => boolean;
   value: string;
   placeholder?: string;
 }
 
+interface StateValidatedTextInputPropTypes extends TextInputProps {
+  dataTestId?: string | undefined;
+  ouiaId?: string;
+  stepId: string;
+  inputId: string;
```
ezr-ondrej, line 30: New component, but it requires `stepId` and `inputId` so I've decided just to copy the component to ease up the transition, but the end goal would be to get all fields validated through this component

regexowl:
@ezr-ondrej Thank you! I wasn't sure about persistent validation at first, but I must admit I quite like it. Especially with how we allow the user to jump across the steps now, showing the validation of steps in nav looks really nice:

image

I have one nit pick - when the "Create/Edit" button gets disabled it's unreadable, can we make the blue part "disabled" grey? The dark grey text should get more contrast then.

image

But one bigger problem I see is this:

image

if you fill in the name, but don't do anything else the Next button remains disabled which is quite confusing. 

The user needs to tab, enter or click somewhere to trigger the validation and enable the button. That's different from current behaviour when the input is validated as it's being typed. 

Can we do the validation during typing here as well?

ezr-ondrej:
> Can we do the validation during typing here as well?

The Pristine prevents from premature validation, so it's intended to work this way (only validate once User leaves the field).
Previously we validated for the next button separately, but we now rely on the field validation.

I guess we'd need to either forfeith the pristine validation and always validate once user start typing in the field, or just allow the click of the next button before the fields are validated (when validation is default).

ezr-ondrej:
> button gets disabled it's unreadable, can we make it gray?

Yeah, let me fix it, I noticed that, but didn't find an easy way, so skipped it until we agree on this.

regexowl:
Yeah, premature validation is not great either... from the options you've mentioned I think I would go with enabling the Next button by default and disabling it after validation fails. We have a similar pattern on the FSC step already so it wouldn't be completely new behaviour. What do you think?

ezr-ondrej:
Well the default state is valid in FSC.
I'm gonna test another idea of validating always, but not surfacing the validation to the user prior blur. So pristine would just control the UI validation, not the state.

ezr-ondrej:
I think I've addressed all findings :)
I've implemented what I commented in the last comment ;)

ezr-ondrej:
Thank you! I'll have a look if I can migrate all the other steps too :)

PR #1950, opened 2024-04-16
WizardV2: use the alert color darker

ezr-ondrej:
image


amirfefer (approved):
Thanks @ezr-ondrej :) 👍

PR #1953, opened 2024-04-17
Blueprints: selected Blueprint Empty state action

ezr-ondrej:
Add action in the Images Table empty state when Blueprint is selected. Change the icon to plus as it better underlines the action needed to remedy.

image


regexowl (approved):
Works great, thanks! :rocket:

PR #1954, opened 2024-04-17
V2Wizard: Packages recommendation (HMS-3890)

regexowl:
This adds an expandable with package recommendations to the Packages step.

The recommendations are populated when two or more packages are selected.


lucasgarfield:
Let's add an Unleash flag for this and expose it in stage-preview for now while the final back-end bits are being figured out.

regexowl:
Ah, right, I guess the tests would be affected by gated functionality, is that correct? Should I just comment them out for now?

lucasgarfield (approved):
This is fantastic, thank you!

PR #1955, opened 2024-04-17
ImagesTable: old images alert position

ezr-ondrej:
Probably a good idea to wait for Kelsea to confirm this is what we want :)

image


kelseamann:
Hi there- I looked back in my old mocks, please see this Sketch file for the correct positioning- https://www.sketch.com/s/669c80df-098c-4296-b689-b4e0702d8261/a/MmnrMom

kelseamann:
Also for the copy, let's change it to "The selected blueprint is at version X, images are at version Y. Build images to synchronize them."

src/Components/ImagesTable/ImagesTableToolbar.tsx
```diff
@@ -105,6 +131,14 @@ const ImagesTableToolbar: React.FC<imagesTableToolbarProps> = ({
               : 'All images'}
           </Title>
         </ToolbarContent>
+        {itemCount > 0 && experimentalFlag && isBlueprintOutSync && (
+          <Alert
+            style={{ margin: '0 var(--pf-v5-c-toolbar__content--PaddingRight) 0 var(--pf-v5-c-toolbar__content--PaddingLeft)' }}
```
ezr-ondrej, line 136: Be wary, I'm sneaking in inline style, let me know if I shouldn't :stuck_out_tongue:

ezr-ondrej:
Thank you @kelseamann! I've updated the screenshot :)

src/Components/ImagesTable/ImagesTableToolbar.tsx
```diff
@@ -105,6 +130,17 @@ const ImagesTableToolbar: React.FC<imagesTableToolbarProps> = ({
...
           </Title>
         </ToolbarContent>
+        {itemCount > 0 && experimentalFlag && isBlueprintOutSync && (
+          <Alert
+            style={{
+              margin:
+                '0 var(--pf-v5-c-toolbar__content--PaddingRight) 0 var(--pf-v5-c-toolbar__content--PaddingLeft)',
+            }}
+            isInline
+            title={`The selected blueprint is at version ${selectedBlueprintVersion}, images are at version ${latestImageVersion}. Build images to synchronize with the latest version.`}
```
regexowl, line 140: As we mention only one version of the images in the alert, would it make sense to say "the newest" or "latest" images are at a certain version?

ezr-ondrej: I'd think so :thinking:

src/Components/ImagesTable/ImagesTableToolbar.tsx
regexowl: Thanks for the warning :smile: there were some discussions about inline styling previously, but I don't really remember if we were set on not using them at all to be honest :sweat_smile: 

Could this be moved to a style file so it's consistent with the rest of the code base for now? To be honest I'm not super dead set on this, we'll probably need to do a bigger style review in the future anyway.

lucasgarfield: Are we sure we need style at all? Custom CSS is a red flag for me, my goal is to have zero CSS in the repo.

If the rendered React Patternfly components look different than what UX has made in Figma, it seems more like there is a problem somewhere in Patternfly - either in the component or in the Figma extension. 

I guess the question is, how does it look without the CSS? Bad, or just different? 

If bad - I will begrudgingly accept. If  just different - what do you say to the idea that we just accept it is slightly different?

ezr-ondrej: It's just about the spacing around the alert, so it's "Just different" :)

lucasgarfield: I tried using `<ToolbarContent>` and the alert didn't fill the table to the full width... so let's go with the custom CSS that at least works!

lucasgarfield (approved):
Thank you!

PR #1960, opened 2024-04-18
feat(HMS-3582): add first boot step to wizard

amirfefer:
This PR adds the ability to add a first boot script (currently just bash/shell) in the blueprint wizard.

Screenshot 2024-04-18 at 18 33 12

- [x] Add first boot step into v2 wizard
- [x] Add an alert (warning?) for sensitive data
- [x] Map the script into the blueprint request and wizard state
- [x]  Add full test coverage for testing the request
- [x] Align other tests
### Demo 
_shows a full roundtrip to backend - editing an existing blueprint._

https://github.com/osbuild/image-builder-frontend/assets/11807069/5b5374df-9c0b-447f-b8bb-7399f6b6e52d


amirfefer:
The alert content:
Screenshot 2024-04-18 at 18 47 32

ezr-ondrej:
> Do we want this in production immediately, or should we put behind Unleash flags so we can keep in stage? @croissanne?

I believe we wanted to have it behind feature flag, right?

src/test/Components/CreateImageWizardV2/steps/FirstBoot/Firstboot.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+];
+
+describe('First Boot step', () => {
+  test('should render First Boot step', async () => {
+    await render();
+    await goToFirstBootStep();
+    expect(screen.getByText('First boot configuration')).toBeInTheDocument();
+  });
+  describe('validate first boot request ', () => {
+    test('should validate first boot request', async () => {
```
lucasgarfield, line 85: @amirfefer would you mind refactoring this a bit by extracting the steps into small functions with clear names?

For instance this might be something like:
```typescript
await render();
await goToFirstBootStep();
await addShellScript();
await goToReviewStep();
...
```

src/Components/CreateImageWizardV2/EditImageWizard.tsx
```diff
@@ -36,7 +36,7 @@ const EditImageWizard = ({ blueprintId }: EditImageWizardProps) => {
       navigate(resolveRelPath(''));
     }
   }, [error, navigate]);
-  return <CreateImageWizard startStepIndex={13} />;
+  return <CreateImageWizard startStepIndex={14} />;
```
lucasgarfield, line 39: Will probably need to bump this to 15 if we get the Snapshotting PR in first :thinking:

src/Components/CreateImageWizardV2/steps/FirstBoot/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </Alert>
+      <CodeEditor
+        isUploadEnabled
+        isDownloadEnabled
+        isCopyEnabled
+        isLanguageLabelVisible
+        language={Language.shell}
+        onCodeChange={(code) => dispatch(setFirstBootScript(code))}
+        code={selectedScript}
+        height="400px"
```
lucasgarfield, line 46: Do we not want this to just fill the vertical space?

src/store/wizardSlice.ts
```diff
@@ -272,6 +276,10 @@ export const selectInputValidation =
     return isValid ? 'success' : 'error';
   };
 
+export const selectFirstBootScript = (state: RootState) => {
+  return state.wizard.firstBoot?.script;
```
lucasgarfield, line 280: No need for `?` here I think if we use `''` as the empty/none state.

src/test/Components/CreateImageWizardV2/steps/FirstBoot/Firstboot.test.tsx
lucasgarfield, line 1: Can we name this FirstBoot.test.tsx to be consistent?

src/test/Components/CreateImageWizardV2/steps/FirstBoot/Firstboot.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+const goToReviewStep = async () => {
+  await clickNext(); // Details
+  await enterBlueprintName();
+  await clickNext(); // Review
+};
+
+const SCRIPT = `#!/bin/bash
+systemctl enable cockpit.socket`;
+
+const getFirstbootData = (): ImageBuilderFile[] => [
```
lucasgarfield, line 62: Would it be simpler for now to have this just be an object and not a function that returns an object?

src/store/wizardSlice.ts
amirfefer: if you edit an old blueprint, `state.wizard.firstBoot` is undefined.

ezr-ondrej (changes requested):
I've tested this and it creates the required files, but doesn't run the firstboot, because we did not enabled it, we are missing following in the request:
```
    "services": {
      "enabled": [
        "custom-first-boot"
      ]
    },
```

src/Components/CreateImageWizardV2/steps/FirstBoot/index.tsx
amirfefer: leaving it empty or with `100%` doesn't work well, so instead I used the responsive `vh` value.

amirfefer:
@lucasgarfield, @ezr-ondrej  I addressed your requests, plus:

- [ ] Rebased with snapshot
- [ ] Adding `custom-first-boot` enabled service
- [ ] Added a feature flag
- [ ] Aligned All tests

ezr-ondrej (commented):
It does work! :tada:
Just copy [lzaps gist](https://gist.github.com/lzap/d074df43ce2be3acc01ffbf8830aa2d5#file-custom-first-boot) and you'll get cockpit on your machine! 

Nitpicks: 
* we are missing info about the script being added in the Review step.
* the flag allows for enable/disable only for both environments, which might get us in tricky position, but it can be solved in a follow-up

amirfefer:
Thanks @ezr-ondrej.
Added `isBeta` and also an indicator in review step:
Screenshot 2024-05-02 at 16 07 50

src/Components/CreateImageWizardV2/steps/Review/ReviewStep.tsx
```diff
@@ -175,6 +179,17 @@ const Review = ({ snapshottingEnabled }: { snapshottingEnabled: boolean }) => {
         {/* Intentional prop drilling for simplicity - To be removed */}
         <ContentList snapshottingEnabled={snapshottingEnabled} />
       </ExpandableSection>
+      <ExpandableSection
```
ezr-ondrej, line 182: Should this be behind the feature flag?

ezr-ondrej: Ah, disregard :)

src/Components/CreateImageWizardV2/CreateImageWizard.tsx
```diff
@@ -126,6 +128,8 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
 
   // =========================TO REMOVE=======================
 
+  const firstbootFlag = useFlag('image-builder.firstboot.enabled');
+  const isFirstBootEnabled = isBeta() && firstbootFlag;
```
ezr-ondrej, line 132: Given we should use it to hide the review as well, should this live in Utilities?
Something like https://github.com/osbuild/image-builder-frontend/blob/main/src/Utilities/useExperimentalFlag.ts

lucasgarfield: What if we had a single hook, `useFirstBootEnabled`, that would return true or false depending on whether we were in stage or prod _and_ whether the flags `image-builder.firstboot.stable.enabled` or image-`builder.firstboot.beta.enabled` were on/off?

lucasgarfield (approved):
Works nicely! I think we need to make a new pattern for how to handle blueprints that use preview features as they will still appear in the stable version, but that's for a follow-up PR.

ezr-ondrej:
> Works nicely! I think we need to make a new pattern for how to handle blueprints that use preview features as they will still appear in the stable version, but that's for a follow-up PR.

Bit fast on the button IMO :stuck_out_tongue: @amirfefer can you hide it from review in a follow-up pls? :)

PR #1964, opened 2024-04-19
V2Wizard: Add `wizardMode` and repo removal modal

regexowl:
This adds a new field indicating create/edit mode of the wizard to the state.

An alert an a modal were also added to warn users when removing custom repositories that this action can break their packages.


lucasgarfield (approved):
Clever!

Issue #1788, opened 2024-03-12
Wizard/s: Make Manual partitioning radio disabled for ISO only images


regexowl:
FSC is not applied to bare metal - installer images. When an iso only image gets built, the FSC step displays an alert to make the user aware of the fact and the manual radio button could get disabled.

Careful about multiple target images. In a case when there are more target types the radio should be enabled even though there is iso among them.

![image](https://github.com/osbuild/image-builder-frontend/assets/49452678/f83a355a-4547-4ad9-96b0-b8272429fe5e)

PR #1965, opened 2024-04-19
V2 Wizard: Fix OpenSCAP step (HMS-2781)

lucasgarfield:
OpenSCAP profile info is now correctly loaded into the state and verified to be in the final request with tests.

I modified the fixtures for the OpenSCAP profiles. My changes ensure we have a Venn-diagram like overlap when changing from one profile to another where one package is the same, one package is removed, and one package is added. The same is true for the services and partitions. For kernel args it is not so important as that is just a string (as opposed to an array), so it is enough to be different.

I was able to eliminate a useEffect by replacing it with a lazy query trigger function. Setting the second arg `preferCacheValue` to `true` means that a request is only made if cached data is not available.

I modified the Redux reducers a bit to add some additional safety. 

`changeFileSystemPartitionMode` is now responsible for initializing the partitions field in the state by adding the root partition – previously this was done in the components themselves by dispatching `addPartition`. This reducer can always be safely dispatched – if the mode is ‘manual’, dispatching it with a payload of ‘manual’ will not result in any changes to the state. 

`addPackage` is also safer now. When a package is added, the list of packages is checked. If there is a package with an identical name, the new package overwrites the previous package. This is useful because the description may be different for the same package – for instance, when adding an OpenSCAP package, we use a custom description (‘this package required by OpenSCAP’).

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemPartition.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemPartition.tsx
@@ -1,20 +1,23 @@
 import React from 'react';
 
 import { FormGroup, Label, Radio } from '@patternfly/react-core';
-import { v4 as uuidv4 } from 'uuid';
 
-import { UNIT_GIB } from '../../../../constants';
 import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
 import {
-  changeFileSystemConfiguration,
   changeFileSystemPartitionMode,
   selectFileSystemPartitionMode,
+  selectProfile,
 } from '../../../../store/wizardSlice';
 
 const FileSystemPartition = () => {
-  const id = uuidv4();
   const dispatch = useAppDispatch();
   const fileSystemPartitionMode = useAppSelector(selectFileSystemPartitionMode);
+  const hasOscapProfile = useAppSelector(selectProfile);
+
+  if (hasOscapProfile) {
+    return undefined;
+  }
+
   return (
     <FormGroup>
       <Radio
@@ -33,7 +36,6 @@ const FileSystemPartition = () => {
         isChecked={fileSystemPartitionMode === 'automatic'}
         onChange={() => {
           dispatch(changeFileSystemPartitionMode('automatic'));
-          dispatch(changeFileSystemConfiguration([]));
         }}
       />
       <Radio
@@ -45,16 +47,6 @@ const FileSystemPartition = () => {
         isChecked={fileSystemPartitionMode === 'manual'}
         onChange={() => {
           dispatch(changeFileSystemPartitionMode('manual'));
-          dispatch(
-            changeFileSystemConfiguration([
-              {
-                id: id,
-                mountpoint: '/',
-                min_size: (10 * UNIT_GIB).toString(),
-                unit: 'GiB',
-              },
-            ])
-          );
         }}
       />
     </FormGroup>
--- a/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 
 import {
   Alert,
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -21,3 +21,4 @@ coverage
 
 *~
 bots
+.aider*
```
mgold1234, line 24: is it relate to that pr?

mgold1234:
It looks good overall, but there are some inconsistencies in the environment:
1) When a user adds an oscap profile, it takes a moment to see the impact of that change - switching from 'None' to another profile. Additionally, when quickly navigating to the File System configuration step, the 'Manually configure partitions' switch toggles after a second.
2) In the Open File System configuration step, if you check the 'Manually configure partitions' radio button and press 'Add partition' twice, the second attempt is blocked. Only after the user changes the first added partition can the 'Add partition' button be pressed again.

Uploading Screen Recording 2024-04-21 at 16.35.12.mov…

mgold1234:
https://github.com/osbuild/image-builder-frontend/assets/73419853/cb0e357e-1e52-4537-98fd-dd12d909847f

regexowl:
> In the Open File System configuration step, if you check the 'Manually configure partitions' radio button and press 'Add partition' twice, the second attempt is blocked. Only after the user changes the first added partition can the 'Add partition' button be pressed again.

Great catch! The "Add partition" button doesn't seem to do anything in cases when a `/home` partition is added. Adding duplicates of other mount points seems to be possible still though.

regexowl:
Should the automatic partitioning option get disabled when there's a file system pre-filled from an OSCAP profile? Clicking on automatic and back to manual deletes all the changes.

drop-fsc

lucasgarfield:
Thanks for QA @regexowl and @mgold1234, I'll get those issues fixed up asap.

> edit: Or we could just not render the radio buttons which would be consistent with the V1 behaviour.

I'll do this!

lucasgarfield:
I have fixed the slowness issue by setting `preferCacheValue: true` in the `trigger()` function that is called in the select handler for profiles. Previously we were refetching the profile data upon selection, now we use cached data if it is available.

I've also simplified how we handle the file system step partitions when a profile is selected/changed/cleared so it is equivalent to the V1 wizard: changing a profile clears all existing partitions, selecting none changes the mode back to automatic. I wanted it to be a bit more clever but there are just too many edge cases to think through, so let's just mimic V1 for now.

I need to do a little more clean up and this will be ready for review again. Thanks!

kingsleyzissou (commented):
This looks great! I have two quick comments:
- we probably want to disable `stig` and `stig_gui` like we did [here](https://github.com/osbuild/image-builder-frontend/pull/1895/commits/ee449b602768397a79cbb0b8390ababe4ca63a40) for the V1 wizard 
- I noticed that if we select ISO as the target and we select a profile, we still populate the partitions. This is maybe another edge case that we could look at in a follow up? (I've checked and seen that the V1 wizard does the same).

regexowl:
> I noticed that if we select ISO as the target and we select a profile, we still populate the partitions. This is maybe another edge case that we could look at in a follow up? (I've checked and seen that the V1 wizard does the same).

Great point, we have an open issue for disabling FSC when there's a bare metal target only: https://github.com/osbuild/image-builder-frontend/issues/1788

For now there's an alert, that any FSC customisation won't get applied. It's hidden under manual config though, maybe we could bump it a bit higher.

kingsleyzissou:
> For now there's an alert, that any FSC customisation won't get applied. It's hidden under manual config though, maybe we could bump it a bit higher.

I think the placement of the alert was fine :)

lucasgarfield:
> This looks great! I have two quick comments:
> 
>     * we probably want to disable `stig` and `stig_gui` like we did [here](https://github.com/osbuild/image-builder-frontend/pull/1895/commits/ee449b602768397a79cbb0b8390ababe4ca63a40) for the V1 wizard
> 
>     * I noticed that if we select ISO as the target and we select a profile, we still populate the partitions. This is maybe another edge case that we could look at in a follow up? (I've checked and seen that the V1 wizard does the same).

Good points, thank you! I might do these in a follow-up PR if that's okay with everyone, this one is already getting a bit complex.

mgold1234:
unit tests fail

mgold1234:
@lucasgarfield you fixed the 2 problem I saw, its looking gtreat

.gitignore
lucasgarfield: Nope :)

regexowl (approved):
Looks great! :sparkles: :rocket:

PR #1970, opened 2024-04-22
V2Wizard: Update save button

regexowl:
This removes the second "Save" from a save button on Review step.

before:
image

after:
image


lucasgarfield (approved):
Thanks!

PR #1975, opened 2024-04-23
V2 Wizard: Add AWS tests (HMS-2781)

lucasgarfield:
Adds tests to verify that building an AWS image using either sources or an account id results in a correctly generated request.


regexowl (approved):
LGTM

PR #1978, opened 2024-04-23
V2Wizard: Remove "Recreate" option for expired images

regexowl:
This replaces the "Recreate" option for expired imaged with a disabled download link.

The change is gated behind experimental mode so V1 imagesTable composes are not affected.


lucasgarfield (approved):
Thank you!

PR #1981, opened 2024-04-23
V2 Wizard: Disable back on first step (HMS-2781)

lucasgarfield:
Disables the back button on the first step of the wizard. Disabling the back button (as opposed to hiding it) is how the wizard works in the Patternfly official examples.

image


regexowl (approved):
Nice catch!

PR #1983, opened 2024-04-23
V2Wizard: Show selected packages in included repos

regexowl:
Previously when a package from an other repo was selected, it was still shown under "Other repos" toggle.

As a recommended repository gets added together with the package, it should be shown under "Included repos".

empty-state


lucasgarfield (approved):
Thank you!

PR #1989, opened 2024-04-24
V2Wizard: Fix VMware popovers

regexowl:
The popovers for VMware sphere types were not buttons. This caused weird behaviour with "ban" cursor when disabled and not opening popover when enabled.

"ban" cursor:
image


lucasgarfield (approved):
Nice catch!

PR #1990, opened 2024-04-24
V2Wizard: Re-use activation popover for Review step

regexowl:
I was bothered by the organization's id in Activation key popover not being separated by an empty line. While adding it I've noticed we duplicated the component on Review step. Cleaned up the code a bit and got rid of a useEffect.


lucasgarfield (approved):
Get the heck out of here, useEffect! Nice work! :smile:

PR #1994, opened 2024-04-24
V2Wizard: Update copy on empty state

regexowl:
This updates copy on empty state when no packages are found with "Available" and "Included repos" toggled.

image

Copy on "Selected" and "Other repos" was also updated.

image


lucasgarfield (approved):
Thank you!

PR #1995, opened 2024-04-24
V2Wizard: Fix padding around registration options

regexowl:
This updates the padding around registration options.

before:
image

after:
image


lucasgarfield (approved):
I wish we could do this without CSS somehow but at least this works! Thank you!

PR #1996, opened 2024-04-24
V2Wizard: Fix padding and size of `<HelpIcon>` button

regexowl:
This updates padding around the popover button on OpenSCAP step and makes all `<HelpIcon>` icons the default size so it's consistent throughout the Wizard.

before:
image

after:
image


regexowl:
Also we don't use popover position to open it to left anywhere else, so removed it for consistency.

lucasgarfield (approved):
Looks good for now, maybe we can get rid of CSS in future somehow.

PR #1997, opened 2024-04-24
V2Wizard: Deduplicate minimum size popover and set width

regexowl:
This sets a maximum width to the minimum size popover. Also moved code to a reusable component.

before:
image

after:
image


lucasgarfield (approved):
Thank you, looks much better!

PR #1999, opened 2024-04-25
ImagesTable: Update empty state for Blueprints

regexowl:
This adds an empty state for blueprint table when there are no blueprints created.

Changes are gated behind experimental flag and preview so that the original ImagesTable isn't affected.

### DIFF
--- a/src/Components/ImagesTable/EmptyState.tsx
+++ b/src/Components/ImagesTable/EmptyState.tsx
@@ -16,10 +16,12 @@ import {
   ExternalLinkAltIcon,
   PlusCircleIcon,
   PlusIcon,
+  SearchIcon,
 } from '@patternfly/react-icons';
 import { Link } from 'react-router-dom';
 
 import { resolveRelPath } from '../../Utilities/path';
+import { useExperimentalFlag } from '../../Utilities/useExperimentalFlag';
 import { BuildImagesButton } from '../Blueprints/BuildImagesButton';
 
 type ImagesEmptyStateProps = {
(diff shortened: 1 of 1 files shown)


src/Components/ImagesTable/EmptyState.tsx
```diff
@@ -52,72 +55,89 @@ const EmptyBlueprintsImagesTable = () => (
 );
 
 const EmptyImagesTable = () => {
+  const { isBeta } = useGetEnvironment();
+  const experimentalFlag = useExperimentalFlag();
   return (
     <Bullseye>
       <EmptyState variant={EmptyStateVariant.lg} data-testid="empty-state">
-        <EmptyStateHeader
```
lucasgarfield, line 58: Wouldn't we still want to render this for the original images table? I.e. `{ !experimentalFlag && ... }`?

lucasgarfield (changes requested):
Oof, I forgot to let you know that I actually created a custom hook to handle the `isBeta() && experimentalFlag` case. Can you please use the `useExperimentalFlag` hook found in `useExperimentalFlag.ts`?

src/Components/ImagesTable/EmptyState.tsx
regexowl: Didn't realize `isBeta()` is already checked in `useExperimentalFlag` hook :sweat_smile: Fixed! I'll go through the code and check the other places I've recently updated.

What do you mean by "render this" in original imagesTable? The header? Currently it should be split so it's the same empty state for ImagesTable as it was, so "Create an RPM-DNF image" and for blueprints it will show the "No images" as per Kelsea's mocks.

lucasgarfield: :facepalm: I missed the ternary operator, I understand now. Ignore me. :sweat_smile:

src/Components/ImagesTable/EmptyState.tsx
```diff
@@ -52,72 +54,88 @@ const EmptyBlueprintsImagesTable = () => (
...
-        </EmptyStateFooter>
+        {experimentalFlag ? (
+          <>
+            <EmptyStateHeader
+              titleText="No images"
+              icon={<EmptyStateIcon icon={SearchIcon} />}
+              headingLevel="h4"
+            />
+            <EmptyStateBody>
+              <Text>Images are BLANK. Create blueprints to create images.</Text>
```
lucasgarfield, line 69: Is it really supposed to say BLANK in all caps? I didn't check the mocks. :sweat_smile:

lucasgarfield: It should be this state, isn't it?
image

lucasgarfield: I have checked the mocks and indeed they say BLANK. Let's double check with UX but for now... capslock is cruise control for cool :sunglasses:

PR #2000, opened 2024-04-25
V2Wizard: Add masked services to the "Disabled services" code block

### LINKED ISSUE #1985: OpenSCAP profile does not display disabled services
These are correctly added to the masked services field in the request but they do not show up in the UI.

regexowl:
Fixes #1985

This concatenates masked service to the disabled services so both get rendered in the "Disabled service" code block.

Test to check this was also added.


lucasgarfield (approved):
Thank you! This will result in masked services overwriting disabled services if there are duplicates (which there should not be, ideally). That's probably the safest way to do it and ensure that image builds don't fail if there are.

PR #2001, opened 2024-04-25
V2Wizard: Debounce package search

regexowl:
This debounces package search so it doesn't overwhelm content API with two requests for every character.


croissanne (dismissed):
Thanks!

regexowl:
Ugh, the timing doesn't play nice with the tests. I'll update those.

regexowl:
All good now :wrench:

lucasgarfield (approved):
Extremely cool and good!

PR #2005, opened 2024-04-25
experimental: Allow switching experimental on in stable

ezr-ondrej:
This allows us to flip the new wizard in stable :)
I've created the feature flag `image-builder.new-wizard.stable` in both stage and prod, so we can test it out in stage-stable first


lucasgarfield (approved):
Makes sense, can't wait to flip this bad boy in prod.

PR #2011, opened 2024-04-26
constants: Move URLs to constants

### LINKED ISSUE #1923: Move all URLs to constants.ts
We have quite a few links in Images these days, mostly pointing to either the documentation or to other Insights services. The URLs are, for the most part, embedded directly in the code. We should put these into `constants.ts` and export them so there is a single source of truth.

A good way to locate URLs apart from manually going through the UI might be to grep for the `href` attribute.

regexowl:
Fixes #1923

This moves all URLs from code to constants.


lucasgarfield (approved):
This is great, it should really reduce bug surface area. Thank you!

PR #2014, opened 2024-04-26
V2 Wizard: OpenSCAP edit tests (HMS-2781)

lucasgarfield:
Edit mode is now fully tested and working for OpenSCAP profiles. A
handler for the PUT request was added and the fixtures were updated to
support this.

`EditImageWizard.tsx`: Previously we dispatched `initializeWizard()`
while waiting for the blueprintDetails to load. This meant that the
state was incorrect (empty) while the blueprintDetails request was
in flight.

`requestMapper.tsx`: Correctly populate state in edit mode if the
blueprint contains a custom file system – previously custom mountpoints
were dropped and automatic mode was selected.

`spyOnRequest()`: Differentiate between request types (e.g. GET, PUT)
for the same endpoint.

### DIFF
--- a/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
+++ b/src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
@@ -120,6 +120,18 @@ export const mapRequestToState = (request: BlueprintResponse): wizardState => {
   const azureUploadOptions = azure?.upload_request
     .options as AzureUploadRequestOptions;
 
+  const fileSystem = request.customizations.filesystem
+    ? {
+        mode: 'manual',
+        partitions: request.customizations.filesystem,
+        isNextButtonTouched: true,
+      }
+    : {
+        mode: 'automatic',
+        partitions: [],
+        isNextButtonTouched: true,
+      };
+
   return {
     wizardMode,
     details: {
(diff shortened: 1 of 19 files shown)


regexowl:
The Oscap test part looks great!

Looking through the FSC part I've found a bit of a problem behaviour with the default 'GiB' unit. If the user chooses a smaller partition, goes to Edit mode and the FSC step, they will not be able to sipmly continue because the automatic conversion to 'GiB' will cause errors for them:
image

I've tried to add a commit that partially solved the problem in a way it was dealt with in V1, but it seems to have broken something else.

The question is - do we want to reach parity with V1 in how the minSizes/units are displayed? If not, I'd suggest setting the default unit to KiB so there are maybe a ridiculously high numbers in partition size inputs, but the validation doesn't block progressing through the wizard.

src/Components/CreateImageWizardV2/utilities/requestMapper.tsx
```diff
@@ -92,14 +93,35 @@ export const mapRequestFromState = (
   };
 };
 
+const parseSizeUnit = (bytesize: string) => {
+  let size;
+  let unit;
+
+  if (parseInt(bytesize) % UNIT_GIB === 0) {
+    size = parseInt(bytesize) / UNIT_GIB;
+    unit = 'GiB' as Units;
```
lucasgarfield, line 102: Can you change to `unit: Units = 'GiB'`? Using `as` is like a wrecking ball, it will allow you to do total nonsense. If you use what I proposed type checking will actually occur.

regexowl:
Also could we cherry pick the FSC changes to a separate PR so the OSCAP tests can get merged?

regexowl (approved):
Looks great, thanks! I'll open a follow up for FSC shortly.

PR #2016, opened 2024-04-26
V2 Wizard: Fix snapshot_date field

lucasgarfield:
The snapshot_date in the image_request must be a date. snapshot_date is an optional field and if no snapshot is used, the field should be undefined. Previously it was an empty string and this caused image builder to return a code 400 with message "Snapshot date is not in DateOnly (yyyy-mm-dd) format".


lucasgarfield:
Tested and image builds work in stage again for both images with and without snapshotting.

regexowl (approved):
Great catch! Works great, now. Thanks!

Issue #2017, opened 2024-04-26
 configuration type alwyas 'automatic' when user edit blueprint 


mgold1234:
1) user press on Edit blueprint - blueprint with file system configuration
2) open  'file system configuration'
actual:
 configuration type alwyas 'automatic' when user edit blueprint 
<img width="426" alt="Screenshot 2024-04-26 at 19 19 18" src="https://github.com/osbuild/image-builder-frontend/assets/73419853/0eacb07f-870f-4ede-ac05-1e41ebbffbe5">

regexowl:
Fixed in https://github.com/osbuild/image-builder-frontend/pull/2014/commits/02055746dacc6064e23dc85614f3387bd76ece0e

PR #2020, opened 2024-04-29
V2Wizard: Hide an option to manually configure FSC for ISO targets

### LINKED ISSUE #1788: Wizard/s: Make Manual partitioning radio disabled for ISO only images

regexowl:
Fixes #1788

This removes an option to manually configure FSC for ISO targets only.

A custom hook `useHasIsoTargetOnly` was added as it will be later needed in `requestMapper`. The current behaviour is so the option to manually configure partitions is not shown to the user for ISO targets only, but the possible FSC customisations from OSCAP profile get applied anyway.

The easiest way to solve this will be checking for `hasIsoTargetOnly` in `requestMapper` and populating the request with customisations / empty array based on the result.

The other way would be stopping the FSC from being populated during OSCAP profile selection. This created a bug when user was still able to click between steps in the nav menu and when they returned to the FSC step the partitions were not populated until the OSCAP profile got changed.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/index.tsx
@@ -8,10 +8,12 @@ import FileSystemPartition from './FileSystemPartition';
 
 import { useAppSelector } from '../../../../store/hooks';
 import { selectFileSystemPartitionMode } from '../../../../store/wizardSlice';
+import { useHasIsoTargetOnly } from '../../utilities/hasIsoTargetOnly';
 export type FileSystemPartitionMode = 'automatic' | 'manual';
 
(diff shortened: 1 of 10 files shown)


regexowl:
The next step will be adding the condition for populating FSC part of image request in `requestMapper`. But since there's already an issue for that https://github.com/osbuild/image-builder-frontend/issues/2017 and it looked like someone is already working on it I didn't add it to this PR.

lucasgarfield (approved):
lgtm thank you!

PR #2024, opened 2024-04-29
Wizard V2: Update packages debounce functionality and minimum api search value 

Andrewgdewar:
- Updates the functionality of the packages tab to handle a little more smooth. 
- Adds validation for minimum characters
- Adds placeholder text to push help the user understand they need to type.


regexowl:
Didn't go through the code changes yet, but a question based on the screenshots - should we allow lower minChar limit? There are packages that have only two chars in their name (not sure if there are any with just one). 

current search with an exact match for 'xa':
image

after proposed changes:
image

(also there is the overlap of states "no results" and "too short search" that should get sorted out, with the two toggles there are many combinations of possible states so it's currently a bit tedious to untangle :sweat_smile:)

src/Components/CreateImageWizardV2/CreateImageWizard.scss
```diff
@@ -82,3 +82,7 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
         margin-block-start: 0;
     }
 }
+
+.pf-v5-c-pagination.pf-m-bottom .pf-v5-c-menu-toggle {
+    display: flex
+}
```
regexowl, line 88: Just a heads up - the position of the arrow in bottom pagination is a bug directly in PF, they know about it and are working on releasing a quick patch fix for it.

Which means we'll be probably able to remove it :)

api/schema/imageBuilder.yaml
```diff
@@ -671,7 +671,6 @@ paths:
           content:
             application/json:
               schema:
-                type: array
```
regexowl, line 674: @mgold1234 Added the change directly in the image-builder repo https://github.com/osbuild/image-builder/pull/1165

The check for manual changes in the API should be hopefully passing after running `npm run api:generate`

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
```
regexowl, line 1: Found one bug when searching for repository recommendation. 

When searching for an EPEL package without adding the EPEL repo there are results listed in Available - Other repos. Adding any of those packages should add the EPEL repo as well which is explained in a modal that will pop up as soon as one of the packages is selected.

Previously the modal popped up only for the first package, but now it pops up twice in a row even though the repo being added should be the same EPEL repo.

two-modals

regexowl (commented):
This looks great, thanks a lot for looking into it! I've added some comments, two are mostly just a heads up :)

api/schema/imageBuilder.yaml
Andrewgdewar: I made the change prior to her PR being merged, I'll rerun the api-gen to confirm the changes though.

src/Components/CreateImageWizardV2/CreateImageWizard.scss
Andrewgdewar: Will add a note to check to remove it.

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
Andrewgdewar: hmmm will look at this.

src/Components/CreateImageWizardV2/CreateImageWizard.scss
regexowl: Perfect, thanks!

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
Andrewgdewar: Found the problem, the component wasn't getting rerendered, so wasn't receiving updated props in it's parent function. 

But I also noticed that we should likely only be adding that initial selection if the user "accepts" the modals proposition. 
So added that change.

api/schema/imageBuilder.yaml
Andrewgdewar: hmm I'm not seeing the API-gen update as per @mgold1234's changes on the above ticket.
Screenshot 2024-05-01 at 9 02 41 PM

It keeps attempting to append the above array. 
Maybe I need some permission that I don't understand? Perhaps for the bash script?

mgold1234: maybe you should update the main branch and rebase this one?

mgold1234: after this pr at the backend side getting merge, it will solve that problem in the frontend side https://github.com/osbuild/image-builder/pull/1168

regexowl: Thank you for looking into the API problem @mgold1234! :heart:

PR #2026, opened 2024-04-29
ImagesTable: show detailed time when hovering over the "Updated" column

schuellerf:
I just wanted to see the time as I build a lot locally :-D

would this be useful?

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -66,6 +66,7 @@ import { resolveRelPath } from '../../Utilities/path';
 import {
   computeHoursToExpiration,
   timestampToDisplayString,
+  timestampToDisplayStringDetailed,
 } from '../../Utilities/time';
 import { useExperimentalFlag } from '../../Utilities/useExperimentalFlag';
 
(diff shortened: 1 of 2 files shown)


regexowl:
@schuellerf I think this is really great idea! Thanks! :sparkles: Could you please fix the two linting errors, I'll be happy to merge then

schuellerf:
Absolutely!
Just didn't see it too late yesterday

regexowl (approved):
Thank you! :heart: :rocket:

schuellerf:
https://ci.int.devshift.net/job/osbuild-image-builder-frontend-pr-check/1517/console
```
10:04:09 FAIL src/test/Components/CreateImageWizardV2/CreateImageWizard.content.test.tsx (103.955 s)
10:04:09   Step Packages search results should be sorted with most relevant results first
10:04:09 
10:04:09     expect(received).toHaveLength(expected)
10:04:09 
10:04:09     Expected length: 6
10:04:09     Received length: 3
```
seems to be a real test problem or is this really due to some race condition?

regexowl:
That's one of the possible flakes :sweat: Both pr_check attempts failed while running frontend unit tests. Now it got over so if nothing comes up we're good to merge.

There's definitely a problem with the unit test flakiness, but it's not a problem of this PR :sweat_smile:

PR #2031, opened 2024-04-30
CreateImageWizard/Repositories: show hint why a repo is disabled

schuellerf:
Some repos in my local test were disabled to ✅, so I wanted to see why.
Maybe this is useful in general

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
@@ -261,17 +261,22 @@ const Repositories = () => {
   };
 
   const isRepoDisabled = (
-    repo: ApiRepositoryResponseRead
+    repo: ApiRepositoryResponseRead,
+    isSelected: boolean
   ): [boolean, string] => {
     if (isFetching) {
       return [true, 'Repository data is still fetching, please wait.'];
     }
 
-    if (recommendedRepos.length > 0 && repo.url?.includes('epel')) {
+    if (
+      recommendedRepos.length > 0 &&
+      repo.url?.includes('epel') &&
+      isSelected
+    ) {
       return [
         true,
-        'Selecting this repository would invalidate packages added from recommended repositories.\n' +
-          'Please use another combination of custom and recommended repositories.',
+        'This repository was added because of previously recommended packages added to the image.\n' +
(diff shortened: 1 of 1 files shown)


regexowl:
There is an expandable that partially covers this (the EPEL portion of this PR)
image

but I quite like putting the information in the title as I feel I'd be also trying to hover over the disabled checkbox to see what's up. The expandable doesn't seem to be visible enough when collapsed :sweat_smile: 

Maybe we could combine the information about specific packages with displaying the info in the hover title? So something like this?
image

What do you think @mmenestr

mmenestr:
@regexowl What about having an inline alert at the top? https://www.patternfly.org/components/alert#plain-inline-alert-variations
and maybe instead of listing out the packages in bullets, we can just separate by commas to make it less tall?

regexowl:
The expandable can be switched for an alert for sure. Warning variant? Also would it make sense to keep some kind of simplified information in the hover title as well? @schuellerf 's original proposal looks like this:

image

mmenestr:
Probably just info alert, not warning since they can't actually do anything on that page!

For the hover tooltip I would re \word it a bit since they can't physically deselect it and it's just an explanation of why -- 
Maybe something like: "This repository was added because of previously recommended packages added to the image. To remove the repository, its related packages must be removed first." or something like that?

regexowl:
Sounds good, thank you!

@schuellerf Can you please update the wording? Otherwise looks great to me :)
I'll take care of the expandable -> alert part.

schuellerf:
The tooltip is actually for the three cases where I _can't_ select the repo.
So I think "This repository was added because of …" is the inverse usecase? 🤔

Screenshot_20240502_105247

maybe
```
        'Repository can't be selected as it would invalidate packages added from recommended repositories.\n' +
          'Please use another combination of custom and recommended repositories.',
```
if I got the context right.

(I can't reproduce this case locally :-/ )

regexowl:
@schuellerf Heya, that's a case where the repo would be selected, but deselectable. 

You can reproduce it by:
- not adding EPEL on repo step
- going to the packages step
- searching for an EPEL only package (gnuplot, caja, caddy)
- going to "Other repos" and selecting any of the recommendations
- then going back to the Repository step.

Looks like this afterwards: 
image

regexowl (approved):
This looks great, thank you! :sparkles:

PR #2032, opened 2024-04-30
V2Wizard: Fix FSC units on Edit

### LINKED ISSUE #2015: 'Image size(minimum)' is missing at Review step when user define manual partition
scenraio:
1) create blueprint - blueprint should create with partitions in the file system configuration step
2) go to review step and check the file system configuration. 
actual:
'Image size(minimum)' is missing at Review step when user define manual partition
Screenshot 2024-04-26 at 15 34 31

regexowl:
Fixes #2015

This updates the logic of parsing bytesize into size/unit pair.

Currently 'GiB' is set as a default which causes problem with smaller partitions on Edit.

A validation for the min_size is in place and the rule is that the size cannot be smaller than 0 of a given unit. If there was a partition with smaller size in KiB it will break the rule when converted to GiB.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
+++ b/src/Components/CreateImageWizardV2/steps/FileSystem/FileSystemConfiguration.tsx
@@ -105,7 +105,7 @@ const FileSystemConfiguration = () => {
       addPartition({
         id,
         mountpoint: '/home',
-        min_size: UNIT_GIB.toString(),
+        min_size: '1',
         unit: 'GiB',
       })
     );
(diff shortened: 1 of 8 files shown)


regexowl:
Ok, finally untangled what I got tangled in. So how does this thing currently work?

The `min_size` value is kept in the store as bytes. No units added there. But in the frontend we were mixing two approaches - getting the size in bytes and adding arbitrary units (mostly GiB). This resulted in behaviour like this:

min_size_changes

and this on Edit:
min_size_on_edit

What changed - the `parseSizeUnit.ts` was introduced and it's called every time the byte size needs to be converted to size + unit pair. This should ensure that the size + unit get parsed from the request / oscap profile properly.

regexowl:
Also added second commit which fixes the missing minimum size on Review step.

mgold1234 (approved):
@regexowl its work great , /lgtm

regexowl:
The pr_check keeps failing so I'll keep this one open and check with @jrusz tomorrow if it's safe to merge :)

lucasgarfield:
PR check was green so I merged!

PR #2034, opened 2024-04-30
LandingPage: Update 'new' alert for blueprints to support quickstarts

lucasgarfield:
image


kelseamann:
I don't think we need the | between the two actions, according to our examples here- https://www.patternfly.org/components/alert/

lucasgarfield:
Oh nice, good catch @kelseamann. I found the `|` in old Patternfly documentation example (from version 3 haha). I guess things have changed since then!

ezr-ondrej:
The quickstart is in :)

ezr-ondrej (approved):
Thanks @lucasgarfield !
I've seen it working.
Inline style in this case is IMO completely fine as the alert will go away eventually.

regexowl (approved):

PR #2035, opened 2024-04-30
BlueprintTable: Invalidate tags on deleteBlueprint

lucasgarfield:
Invalidates Composes and ComposeBlueprints tags upon deleting a blueprint so that composes are re-fetched and the table is updated appropriately.

Previously, a deleted blueprints composes would still show in the table until the page was refreshed.


regexowl (approved):
Thank you!

PR #2041, opened 2024-05-01
V2Wizard: Expandable to alert for disabled EPEL repo

regexowl:
This converts previously used expendable into an info alert.

image


lucasgarfield (approved):
lgtm!

PR #2048, opened 2024-05-02
V2Wizard: Hide first boot from review step (HMS-3582)

amirfefer:
Follow-up from #1960

### DIFF
--- a/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizardV2/CreateImageWizard.tsx
@@ -8,7 +8,6 @@ import {
   WizardStepType,
   useWizardContext,
 } from '@patternfly/react-core';
-import { useFlag } from '@unleash/proxy-client-react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 
 import DetailsStep from './steps/Details';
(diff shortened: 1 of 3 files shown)


src/store/hooks.ts
```diff
@@ -39,3 +42,9 @@ export const useServerStore = () => {
   const oscap = useOscapData();
   return { ...oscap };
 };
+
+export const useBetaFlag = (flag: string): boolean => {
```
ezr-ondrej, line 46: not sure it belongs in store hooks, but it's fine IMO ;)

ezr-ondrej (dismissed):
I love `useBetaFlag` helper idea ;)

src/store/hooks.ts
```diff
@@ -39,3 +42,9 @@ export const useServerStore = () => {
...
+  const firstbootFlag = useFlag(flag);
```
regexowl, line 47: If I understand correctly, this hook will be re-usable outside of first boot, right? Could you give the const more universal name? Like `unleashFlag` or something like that?

regexowl (commented):
Nice! Love the `useBetaFlag` hook idea. Can we please move it outside of the store hooks and into a separate `useBetaFlag.ts` file in utilities folder? (doesn't matter if it's the one under components or V2Wizard, we'll have to clean those up when we remove V1)

PR #2049, opened 2024-05-02
test: Replace 'rhel-94' with a constant

regexowl:
This replaces 'rhel-94' with a RHEL_9 constant in the tests.


lucasgarfield (approved):
LGTM!

PR #2050, opened 2024-05-02
Wizard V2: Custom Repositories refactor

Andrewgdewar:
- Updated content api usage
- Refactored repositories
- A lot of small changes 

One design change: 
My pm suggested removing the confirmation modal on deletion during the edit screen, I have commented that out and will ask for feedback from all of you. We believe the warning below should be sufficient.
Screenshot 2024-05-06 at 10 46 51 PM

Also updated the empty state screen: 
Screenshot 2024-05-08 at 8 34 19 PM

Screenshot 2024-05-08 at 8 39 01 PM

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
@@ -1,37 +1,32 @@
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 
 import {
   Alert,
   Button,
-  EmptyState,
-  EmptyStateBody,
-  EmptyStateIcon,
-  EmptyStateVariant,
   Pagination,
   Panel,
   PanelMain,
   SearchInput,
-  Spinner,
   Toolbar,
   ToolbarContent,
   ToolbarItem,
-  EmptyStateHeader,
-  EmptyStateFooter,
   ToggleGroup,
   ToggleGroupItem,
   PaginationVariant,
+  Grid,
   Modal,
 } from '@patternfly/react-core';
-import {
-  Dropdown,
-  DropdownItem,
-  DropdownToggle,
-  DropdownToggleCheckbox,
-} from '@patternfly/react-core/deprecated';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
-import { RepositoryIcon } from '@patternfly/react-icons';
 import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
 
+import { BulkSelect } from './components/BulkSelect';
+import Empty from './components/Empty';
+import { Error } from './components/Error';
+import { Loading } from './components/Loading';
+import {
+  convertSchemaToIBCustomRepo,
+  convertSchemaToIBPayloadRepo,
+} from './components/Utilities';
 import RepositoriesStatus from './RepositoriesStatus';
 import RepositoryUnavailable from './RepositoryUnavailable';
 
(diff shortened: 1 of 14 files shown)


src/test/Components/CreateImageWizardV2/CreateImageWizard.content.test.tsx
```diff
@@ -475,36 +475,9 @@ describe('Step Custom repositories', () => {
       })
     );
 
-    await screen.findByText(/select all \(1016 items\)/i);
+    await screen.findByText(/select page \(10 items\)/i);
   });
 
-  test('filter works', async () => {
```
regexowl, line 481: I don't think we should be removing tests that don't pass even if it's after a big refactor. We'll still want to test if filtering repositories works.

When manually testing the filtering works well on real life data so in this case it's the MSW handlers/fixtures that will need to be updated for the mocked values to pass the tests as well. Since this might be a bigger task it should be okay to comment out the failing tests and split the fixtures updates to a separate PR

regexowl, line 481: I don't think we should be removing tests that don't pass even if it's after a big refactor. We'll still want to test if filtering repositories works.

When manually testing the filtering works well on real life data so in this case it's the MSW handlers/fixtures that will need to be updated for the mocked values to pass the tests as well. Since this might be a bigger task it should be okay to comment out the failing tests and split the fixtures updates to a separate PR

src/test/Components/CreateImageWizardV2/CreateImageWizard.content.test.tsx
```diff
@@ -475,36 +475,9 @@ describe('Step Custom repositories', () => {
       })
     );
 
-    await screen.findByText(/select all \(1016 items\)/i);
+    await screen.findByText(/select page \(10 items\)/i);
```
regexowl, line 478: I've discussed with @lucasgarfield previously that testing on 1000 of mocked repositories might be a bit too much. Maybe we could also update the fixture to generate less filler repos? Just a though, might not be completely in the scope of this PR :sweat_smile:

src/test/Components/CreateImageWizardV2/steps/Repositories/Repositories.test.tsx
```diff
@@ -102,28 +94,6 @@ describe('repositories request generated correctly', () => {
     },
   ];
 
-  const expectedNginxRepository: Repository = {
```
regexowl, line 105: Same here, as far as I'm aware this might be the only test with module_hotfixes. For the sake of not loosing coverage the MSW fixtures should be updated to accommodate the changes to the code and reflect how real data works.

Could be probably split into a separate PR, but the test should be commented out with a note rather than just removed.

src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
```diff
@@ -40,224 +43,173 @@ import {
...
+    isFetching,
+    isLoading,
+  } = useListRepositoriesQuery(
     {
       availableForArch: arch,
       availableForVersion: version,
       contentType: 'rpm',
       origin: 'external',
-      limit: 100,
+      limit: perPage,
```
regexowl, line 109: Previous double fetching method is not very ideal, but I wonder if we want to fetch based on perPage limit only as any changes cause more loading states. Will return to this in the main comment.

The pagination doesn't seem to work:
broken_pagination

regexowl (commented):
Thank you for working on the refactor! The code looks much cleaner :heart: 

Looking at this from the user point of view there seems to be considerably more loading as more actions trigger a new API call including search and removing repositories.

This is linked to the question of how many repositories should we initially fetch - previously we fetched all the repositories right from the start which looked a bit more seamless when going through the step. What I mean is this:

- previously the fetching happened at the initialization of the step, but then there was no further loading / skeletons:
repos-old

- now the loading state gets triggered on toggling between all/selected, searching, clearing search input, changing pagination perPage limit and removing a repo
repos-new

Which might not be necessarily bad, but it might be good to discuss which path to take from this point. What would be the advantages of handling the data via a series of API calls?

regexowl:
As for the feedback to removing the modal - I remember adding it for some reason, but can't really remember what the reason was :sweat_smile:

I guess it was there to make double sure that the user knows they can break their image with the change? Not sure if that's too much safeguarding though.

Any thoughts @mmenestr @kelseamann ?

Also to check - when removing a repository what would be the desired behaviour? Currently when in the Selected toggle the checkbox gets unchecked, but the repository stays there until the toggle gets toggled to All and retoggled to Selected again.

After the changes the repository would disappear from the Selected list as soon as it gets unchecked. I'm aware we're not currently consistent with the behaviour between repos / packages steps, so this might be an issue for a separate PR.

mmenestr:
I feel like we should keep the modal... this is probably a rare scenario anyway. Also, I think that when the user unselects the repo, the packages should automatically get removed from the "selected" toggle.

In general - my guidance when it comes to unselected items is this:

If the user is in the "selected" toggle, and unchecks something, then the items should remain in the "selected" list, unchecked, until the user toggles back to "All".

However, if the user is NOT on the selected toggle, and their actions impact the selected toggle, then the items should automatically go away from the selected toggle. 

Does that make sense? It basically just depends on whether the user is in the selected toggle while things are unselected or not.

regexowl:
> Also, I think that when the user unselects the repo, the packages should automatically get removed from the "selected" toggle.

Agree, alas that's currently not possible as the packages aren't linked to their origin repository.

src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
ezr-ondrej: Probably not caused by this PR: https://github.com/osbuild/image-builder-frontend/issues/2084 :)

regexowl: I believe those are two separate issues as currently the pagination on the Repositories step seems to work. The proposed changes in this PR would break it.

ezr-ondrej: Ah sorry! I've gotten confused, you're right, those are separate issues!

regexowl: No worries :)

Andrewgdewar:
> As for the feedback to removing the modal - I remember adding it for some reason, but can't really remember what the reason was 😅
> 
> I guess it was there to make double sure that the user knows they can break their image with the change? Not sure if that's too much safeguarding though.

> 
> Any thoughts @mmenestr @kelseamann ?
> 
> edit: it should be this modal if I'm correct image
> 

I spoke to @jlsherrill regarding this (SME for content) and he had basically said that the warning should be enough. 
But if you would like this to be re-added I can do!


> Also to check - when removing a repository what would be the desired behaviour? Currently when in the Selected toggle the checkbox gets unchecked, but the repository stays there until the toggle gets toggled to All and retoggled to Selected again.
> 
> After the changes the repository would disappear from the Selected list as soon as it gets unchecked. I'm aware we're not currently consistent with the behaviour between repos / packages steps, so this might be an issue for a separate PR.


Oh hmm, I see what you mean. 
That is the exact behaviour we have today in content templates when adding and removing a repository for reference. 
I may be able to add a debounce on the selected items to alleviate that extra loading.

src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
Andrewgdewar: > Ah sorry! I've gotten confused, you're right, those are separate issues!

Actually same issue, I keep forgetting both the page/perPage are needed references for useMemo/dependency arrays.

src/test/Components/CreateImageWizardV2/CreateImageWizard.content.test.tsx
regexowl: Aah, yeah, you're absolutely right. I didn't realize that this would now test API responses instead of filtering within the frontend. Mea culpa :sweat_smile:

src/test/Components/CreateImageWizardV2/CreateImageWizard.content.test.tsx
regexowl: Agree, let's stick with 10 here. The fixtures can be updated later if needed.

regexowl:
> The primary reason I am using the API in this way is consistency across our apps. As it was the API was built around this functionality.

Got it, sounds great!

> I understand the idea of liking to front-load all data, but in general, it is largely frowned upon as an architectural choice unless there is an inherit slowness with the API.
> 
> The usual reasons this is sited as not a good pattern in general is you can get to a point where with enough data it becomes a problem, there are also different data limits across browsers and devices, that can lock up/slow-down the web page dramatically (likely not an issue but I digress).

Fair enough! I don't know how many repositories users usually have, but I admit that relying on the idea that "there can't be that many" is not a good approach. :sweat_smile: 

We can try to figure out if there's a way to cache some data so the loading is less obvious, but I'd say that's out of the scope of this PR for now.

Thanks a lot for the explanation!

regexowl:
> But with the addition of the remove modal, it may be less obvious as the user will be slowed down further.

Hmm, when you're looking at the debounce would it be possible to make a short screencast with and without the modal?

I think there are two open questions regarding the modal:
- how to make it obvious to the user that removing previously used repositories is a really bad idea
- and in case the way is to display the modal, when should it be displayed?

Since we probably want the newly unchecked repos to stay in Selected for a bit, it may indeed clash with the modal popping out as soon as the repo gets unselected. 

Maybe we could show the modal just once when the Next button is clicked after the changes were applied? This would probably need to be handled via wizard footer then... :thinking: 

Or maybe we could display the alert only when a previously used repo was unselected so it pops out more?

What do you think @mmenestr?

regexowl:
Clicked around a bunch - this looks and works great, thank you! Current way of fetching/refreshing feels much better to me! The skeletons were a bit overwhelming :skull: 

I'd say let's double check the module_hotfixes test, decide on the modal and debounce and then it's good to get merged.

We'll need to update msw fixtures and handlers a bit if we want to keep making use of the in-browser mocking, but that's a separate problem. Opened an issue for that: https://github.com/osbuild/image-builder-frontend/issues/2107

regexowl:
Oh, one thing - could we make the search case insensitive?

image

image

Andrewgdewar:
> Oh, one thing - could we make the search case insensitive?
> 
> image
> 
> image

Good sanity check. 
Added a ticket for the API [here](https://issues.redhat.com/browse/HMS-4222).

Andrewgdewar:
Added the modal, made it so I only warn if on initial state. 
Also changed the wording a bit as with the warning in the background, it felt a bit redundant.
Screenshot 2024-05-29 at 3 00 22 PM

I also fixed some loading/state issues with the packages tab.

src/Components/CreateImageWizardV2/steps/Repositories/Repositories.tsx
```diff
@@ -299,418 +337,221 @@ const Repositories = () => {
...
             Remove anyway
           </Button>,
-          <Button key="back" variant="link" onClick={handleCloseModalToggle}>
+          <Button key="back" variant="link" onClick={onClose}>
             Back
           </Button>,
         ]}
       >
-        We do not recommend removing any repositories if you have added packages
+        You are removing a previously added repository
```
regexowl, line 393: A teeny tiny nitpick.

```suggestion
        You are removing a previously added repository.
```

src/Components/CreateImageWizardV2/steps/Repositories/components/BulkSelect.tsx
```diff
@@ -0,0 +1,86 @@
...
+  deselectAll: () => void;
+  perPage: number;
+  handleAddRemove: (
+    repo: ApiRepositoryResponseRead | ApiRepositoryResponseRead[],
+    selected: boolean
+  ) => void;
+  isDisabled: boolean;
+}
+
+export function BulkSelect({
```
regexowl, line 24: When removing the page of repos via bulk select the remove modal gets triggered even for new blueprint

repos-modal

regexowl (commented):
Found one bug and a tiny nitpick, but otherwise looks great to me!

src/Components/CreateImageWizardV2/steps/Repositories/components/BulkSelect.tsx
Andrewgdewar: Hmm will fix this.

Andrewgdewar:
Okay, so I fixed the bug above, and looked into cacheing [here](https://redux-toolkit.js.org/rtk-query/usage/cache-behavior). 
I then rewrote a few things and the cache/refetch functions now work a LOT better. 
I think the snappiness you are offer is largely back! 🎉

Andrewgdewar:
I have added the "search" field to the repositories mock function to get around the test failures.

src/test/Components/CreateImageWizardV2/steps/Repositories/Repositories.test.tsx
Andrewgdewar: Fixed, I mocked searching.

Andrewgdewar:
Fixed lint and flattened merge history.

regexowl (approved):
Looks awesome and works like magic! :mage: Thanks for all your work on this!!

PR #2053, opened 2024-05-03
Components/test: Migrate remaining files to TypeScript (HMS-4063)

regexowl:
This migrates remaining JavaScript files in Components and test directories to TypeScript.

MSW file `handlers.js` was left out of the migration since MSW will need to get manually bumped in the near future which means updating all handlers to new format.


src/test/Components/ImagesTable/ImagesTable.test.tsx
```diff
@@ -53,14 +53,15 @@ describe('Images Table', () => {
     const { getAllByRole } = within(table);
     const rows = getAllByRole('row');
     // remove first row from list since it is just header labels
-    const header = rows.shift();
+    const header = rows.shift() as HTMLElement;
```
lucasgarfield, line 56: Instead of using `as HTMLElement`, can we use `const header: HTMLElement = rows.shift()`?

`as` forces the Typescript linter to accept it as true, but the way I propose above allows the linter to do type checking as we want.

src/test/Components/ImagesTable/ImagesTable.test.tsx
```diff
@@ -127,7 +129,9 @@ describe('Images Table', () => {
 
     // No actual clicking because downloading is hard to test.
     // Instead, we just check href and download properties of the <a> element.
-    const downloadLink = await within(downloadButton).findByRole('link');
+    const downloadLink = (await within(downloadButton).findByRole(
+      'link'
+    )) as HTMLAnchorElement;
```
lucasgarfield, line 134: Again, let's use `const downloadLink: HTMLAnchorElement` instead of `as` please.

src/test/Components/ImagesTable/ImagesTable.test.tsx
```diff
@@ -275,11 +279,12 @@ describe('Clones table', () => {
     const cloneRows = within(clonesTable[0]).getAllByRole('row');
 
     // remove first row from list since it is just header labels
-    const header = cloneRows.shift();
+    const header = cloneRows.shift() as HTMLElement;
```
lucasgarfield, line 282: `header: HTMLElement`

lucasgarfield (approved):
Nice! Just a couple nitpicks. I wonder if we should forbid type assertions (using `as`) in our eslint?

regexowl:
> I wonder if we should forbid type assertions (using as) in our eslint?

Yeah, we probably should. I can take a look at it.

src/test/Components/ImagesTable/ImagesTable.test.tsx
regexowl: Done...

src/test/Components/ImagesTable/ImagesTable.test.tsx
regexowl: ... done...

lucasgarfield (approved):
Thank you!

PR #2059, opened 2024-05-06
V2Wizard: Fix repo/description in review table of package on Edit

regexowl:
Previously the repository in the table of packages on the Review step defaulted to "EPEL". This expands the condition to take into account packages added on Edit.

Description is handled in the same way.

The "Not available" text is consistent with how the packages table looks on the Packages step for pkgs populated from the request. A popover with more information was added also to the review table as it contains explanation about the "Not available" text.


jrusz (approved):
Checked in ephemeral and LGTM :) 
image

PR #2060, opened 2024-05-06
src/constants: Use major rhel distros by default

croissanne:
The tradeoff is not being able to discern between a minor version and a major version in the 2 weeks post GA if the images were built using the UI, but I think the tradeoff is worth it. The frontend will never need updating to switch to a new version.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -40,8 +40,8 @@ export const CREATE_RHEL_IMAGES_WITH_AUTOMATED_MANAGEMENT_URL =
 export const OSBUILD_SERVICE_ARCHITECTURE_URL =
   'https://osbuild.org/docs/service/architecture/';
 
-export const RHEL_8 = 'rhel-89';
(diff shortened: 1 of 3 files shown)


regexowl:
I wouldn't mind dropping the detail of minor version, we don't let the user choose between those anyway. And I like the idea of not having to update them after release. 

Should we add some kind of information about "image builder building always the latest and greatest" somewhere though?

If we don't specify minor version would the images automatically be built with the latest version? There are still out of date blueprints which I guess could be handled via a change to request mapper, but that would work only for editing blueprints. If I just click on "Build image" for rhel-93 blueprint, what version would it pick for a build?

regexowl:
DO NOT MERGE! Needs to pass pr_check which is currently broken.

regexowl:
pr_check passed! Let's rebase, make it pass again and then we can merge :)

PR #2063, opened 2024-05-07
V2Wizard: Make e-mail validation case insensitive

### LINKED ISSUE #1554: Improve e-mail validation
As discussed in https://github.com/RedHatInsights/image-builder-frontend/pull/1548#discussion_r1450629990, our e-mail validation (used for GCP only at the moment) needs some work.

regexowl:
Fixes #1554

Case validation for the e-mail field was previously case sensitive, disallowing uppercase letters. This updates the validation to accept both lower and upper case letters.

before:
image

after:
image


lucasgarfield (approved):
Nice catch.

PR #2064, opened 2024-05-07
LandingPage / ImagesTable: Set max width for the ImagesTable

### LINKED ISSUE #933: Images table - does it need a max width?
On widescreen monitors the Images Table looks s t r e t c h e d out.

Any thoughts @arburka?

image

regexowl:
Fixes #933

This applies `isWidthLimited` to the section containing ImagesTable, ensuring the table doesn't get stretched out too much.

The were also two sets of empty tags around ImagesTable return, this removes one of them.

before:
Screenshot from 2024-05-07 10-13-01

after:
Screenshot from 2024-05-07 10-13-19

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -180,66 +180,64 @@ const ImagesTable = () => {
 
   return (
     <>
-      <>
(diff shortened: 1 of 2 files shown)


lucasgarfield (approved):
Extremely nice!

PR #2065, opened 2024-05-07
src: Fix "Table headers must have an accessible name" warning

regexowl:
The tests output contained following warning:

```
console.warn
    Th: Table headers must have an accessible name. If the Th is intended to be visually empty,
pass in screenReaderText. If the Th contains only non-text, interactive content such as a checkbox
or expand toggle, pass in an aria-label.
```

This fixes the warning by adding an accesible `screenReaderText` text to every empty table heading cell. The text doesn't get rendered, but is available as an information for screen readers.

### DIFF
--- a/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
+++ b/src/Components/CreateImageWizard/formComponents/FileSystemConfiguration.js
@@ -379,7 +379,7 @@ const FileSystemConfiguration = ({ ...props }) => {
           >
             <Thead>
               <Tr>
-                <Th />
+                <Th aria-label="Drag mount point" />
                 <Th>Mount point</Th>
                 <Th>Type</Th>
                 <Th>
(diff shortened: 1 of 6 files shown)


lucasgarfield:
Hm, don't we usually use `aria-label`? I can't recall seeing `screenReaderText` anywhere else in the code base. :thinking:

regexowl:
> Hm, don't we usually use `aria-label`? I can't recall seeing `screenReaderText` anywhere else in the code base. 🤔

Great point. I've based the changes on the `If the Th is intended to be visually empty, pass in screenReaderText` part of the warning, but aria-label seems to be working as well and it's consistent with the rest of the code base. I've updated the PR

lucasgarfield:
> > Hm, don't we usually use `aria-label`? I can't recall seeing `screenReaderText` anywhere else in the code base. 🤔
> 
> Great point. I've based the changes on the `If the Th is intended to be visually empty, pass in screenReaderText` part of the warning, but aria-label seems to be working as well and it's consistent with the rest of the code base. I've updated the PR

I wonder what "intended to be visually empty" means? In any case, I'm happy with just using the aria-label! Thanks!

lucasgarfield (approved):
Thank you!

PR #2066, opened 2024-05-07
Status.tsx: show as many different error types as possible (HMS-1442)

schuellerf:
As far as I saw `error.details` rarely contains a `.reason` again.
It's sometimes directly a string and maybe there is `error.details.s` or it's even a list of strings
I'm really not sure if this is the right place to handle this but the `error.details` are somehow support to contain "arbitrary" stuff... (For this uncertainty it's a draft for now)

Also `\n` does not seem to work with react and `<p>` so I needed this `map()`

### DIFF
--- a/src/Components/ImagesTable/Status.tsx
+++ b/src/Components/ImagesTable/Status.tsx
@@ -4,11 +4,14 @@ import './ImageBuildStatus.scss';
 import {
   Alert,
   Button,
+  CodeBlock,
+  CodeBlockCode,
   Flex,
   Panel,
   PanelMain,
   Popover,
   Skeleton,
+  Text,
 } from '@patternfly/react-core';
 import {
   CheckCircleIcon,
@@ -288,18 +291,23 @@ type ErrorStatusPropTypes = {
 
 const ErrorStatus = ({ icon, text, error }: ErrorStatusPropTypes) => {
   let reason = '';
+  const detailsArray: string[] = [];
   if (typeof error === 'string') {
     reason = error;
   } else {
     if (error.reason) {
       reason = error.reason;
     }
-    if (error.details?.reason) {
-      if (reason !== '') {
-        reason = `${reason}\n\n${error.details?.reason}`;
-      } else {
-        reason = error.details.reason;
+    if (Array.isArray(error.details)) {
+      for (const line in error.details) {
+        detailsArray.push(`${error.details[line]}`);
       }
+    }
+    if (typeof error.details === 'string') {
+      detailsArray.push(error.details);
+    }
+    if (error.details?.reason) {
+      detailsArray.push(`${error.details.reason}`);
     }
   }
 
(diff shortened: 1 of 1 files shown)


schuellerf:
A total crash of osbuild should not happen but if it does it would look like this

NOTE: the forwarding of the last 20 lines of stderr is still to be committed if we want this - but it would look similar for other error details
Screenshot_20240508_110723

src/Components/ImagesTable/Status.tsx
```diff
@@ -288,19 +288,27 @@ type ErrorStatusPropTypes = {
...
+        detailsArray.push(`${error.details[line]}`);
       }
     }
+    if (typeof error.details === 'string') {
+      detailsArray.push(error.details);
+    }
+    if (error.details?.reason) {
+      detailsArray.push(`${error.details.reason}`);
+    }
+    if (error.details?.s) {
```
croissanne, line 309: Huh what's the `s` field?

src/Components/ImagesTable/Status.tsx
```diff
@@ -316,10 +324,22 @@ const ErrorStatus = ({ icon, text, error }: ErrorStatusPropTypes) => {
             <Panel isScrollable>
               <PanelMain maxHeight="25rem">
                 <div className="pf-u-mt-sm">
-                  <p>{reason}</p>
+                  <div className="pf-v5-c-code-block">
+                    <div className="pf-v5-c-code-block__header">{reason}</div>
+                    <div className="pf-v5-c-code-block__content">
+                      <pre className="pf-v5-c-code-block__pre">
+                        {detailsArray.join('\n')}
+                        <code className="pf-v5-c-code-block__code"></code>
```
croissanne, line 332: Not sure about using code blocks here, the result doesn't work well for more common errors:

DNF:
image

target errors:

image


We can maybe make the popover wider, and then have reason and details just separated by a vertical line or something?

If we can somehow tell when it's a stacktrace, putting that in a codeblock could be nice.

croissanne (commented):
Amazing! thank you

src/Components/ImagesTable/Status.tsx
schuellerf: totally agree - that's actually for someone else to decide - I'm far away from being a react or PatternFly expert!

src/Components/ImagesTable/Status.tsx
schuellerf: Update from our chats aside this issue:
It seems to stem from my confusion from the golang error type having the error in `.s`

https://github.com/golang/go/blob/1843464f014c946c1663de76249267486887626f/src/errors/errors.go#L67

but this seems to never land in the frontend anyway due to
https://github.com/golang/go/blob/1843464f014c946c1663de76249267486887626f/src/builtin/builtin.go#L309

not being able to serialize …
I'll remove from the frontend...

kelseamann:
Hey, I like that we have the english error terminology separated from the code itself. I think it looks great!

regexowl:
Awesome! Thanks for looking into this!

Checked out how the error is formatted - would it make sense to extract the reason outside of the code block? If I understand it correctly the reason would be a shorter summary and the possible json/code snippets would be contained in the details.

Would something like this work? 
image


```typescript
      <Popover
        data-testid="errorstatus-popover"
        position="bottom"
        minWidth="40rem"
        bodyContent={
          <>
            <Alert variant="danger" title={text} isInline isPlain />
            <Text className="pf-u-pt-md pf-u-pb-md">{reason}</Text>
            <Panel isScrollable>
              <PanelMain maxHeight="25rem">
                <CodeBlock>
                  <CodeBlockCode>{detailsArray.join('\n')}</CodeBlockCode>
                </CodeBlock>
              </PanelMain>
            </Panel>
            <Button
              variant="link"
              onClick={() =>
                navigator.clipboard.writeText(
                  reason + '\n\n' + detailsArray.join('\n')
                )
              }
              className="pf-u-pl-0 pf-u-mt-md"
            >
              Copy error text to clipboard <CopyIcon />
            </Button>
          </>
        }
      >
```

(I've set a wider minWidth and moved the things around a bit so only the code part would be scrollable)

regexowl (approved):
Looks great, thank you!

PR #2067, opened 2024-05-07
Bug - V2 Wizard: trim searchterm to prevent illegal characters

Andrewgdewar:
This was feedback from a user related to summit demos

- Removes extra lead/trailing whitespace for the api (not in the input)


regexowl (approved):
Nice, thank you!

PR #2078, opened 2024-05-13
FileSystemConfiguration: enable mouse selection in TextInput fields

schuellerf:
explicitly overriding the drag functionality of the parent TR/TD to enable mouse selection

Screenshot_20240513_125539

Do you think this is useful? Is there a simpler solution in react?

### DIFF
--- a/src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
+++ b/src/Components/CreateImageWizardV2/ValidatedTextInput.tsx
@@ -130,6 +130,11 @@ export const ValidatedTextInput = ({
     return validator(value) ? 'success' : 'error';
   };
 
+  const dontDragMe: TextInputProps['onDragStart'] = (evt) => {
+    evt.preventDefault();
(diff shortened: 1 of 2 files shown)


regexowl:
I see, yeah this is a great catch and definitely something we should look into! I guess the partition suffix should work like a normal input box, meaning the text should be selectable by dragging a mouse over it. Currently it starts dragging an entire row.

PatternFly doesn't seem to really like these changes, huh? To be honest I have no idea how to solve the bug now, can try to take a look at it later. :thinking:

regexowl:
From a quick look this might be a non-trivial issue as PF doesn't seem to have support for excluding cells from a draggable row and any attempt to overwrite the functionality will come with TS errors. We might need a custom component for this which would mean partially reimplementing things we currently source from PF.

Feel free to work on this further if you'd like. But if you'd prefer to just create an issue for this that's also perfectly fine as the current behaviour of the draggable row might be a bit annoying, but is not a blocker.

PR #2081, opened 2024-05-15
WizardV2: enable V2 by default on preview and ephemeral

ezr-ondrej:
This removes the need for a feature flag in Beta environments.

I see this as a natural first step - enable it in all Beta (including Ephemeral) environments without feature flag.
If we prefer to do default in Ephemeral only, let me know :)


ezr-ondrej:
@jrusz ^^ :)

ezr-ondrej:
@jrusz I think that now it's failing because it is trying the old workflow :)
Would you look into that, what is the strategy for this?

jrusz:
> @jrusz I think that now it's failing because it is trying the old workflow :) Would you look into that, what is the strategy for this?

Yes I'm looking into it, I can see the new wizard in ephemeral by default but I'm not sure why it's failing yet, having some trouble running tests locally against ephemeral.

jrusz:
Package search doesn't work in ephemeral, not sure why. I'll skip adding packages in ephemeral for now to unblock this.

regexowl (approved):
Makes sense to me, thanks!

PR #2083, opened 2024-05-16
Search against content sources for distro packages  (HMS-4094)

croissanne:
Switches over to content sources entirely. For package groups we'll need to use content sources as well, so I think it makes sense. This keeps the custom and distro packages query separated, because we do some heuristics there. 

I also noticed that distro packages are search when there's 2 characters, and custom/recommends when there's 3 or more. Could change this to be consistent.

I removed the v1 wizard test without concent sources entirely.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
@@ -54,7 +54,7 @@ import {
 import { useAppSelector } from '../../../../store/hooks';
 import {
   Package,
-  useGetPackagesQuery,
+  useGetArchitecturesQuery,
 } from '../../../../store/imageBuilderApi';
 import {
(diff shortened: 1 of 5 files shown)


src/test/Components/CreateImageWizardV2/steps/Packages/Packages.test.tsx
```diff
@@ -68,13 +68,13 @@ const searchForPackage = async () => {
 
 const selectFirstPackage = async () => {
   await userEvent.click(
-    await screen.findByRole('checkbox', { name: /select row 0/i })
+    await screen.findByRole('checkbox', { name: /select row 1/i })
```
mgold1234, line 71: just wander why do you start from row 1 and not 0?

regexowl (approved):
Excellent, thank you!

PR #2085, opened 2024-05-21
WizardV2: add 'some' word to image table toolbar title

mgold1234:
this commit add the word 'some' to the toolbar title
Screenshot 2024-05-21 at 12 34 38

### DIFF
--- a/src/Components/ImagesTable/ImagesTableToolbar.tsx
+++ b/src/Components/ImagesTable/ImagesTableToolbar.tsx
@@ -137,7 +137,7 @@ const ImagesTableToolbar: React.FC<imagesTableToolbarProps> = ({
                 '0 var(--pf-v5-c-toolbar__content--PaddingRight) 0 var(--pf-v5-c-toolbar__content--PaddingLeft)',
             }}
             isInline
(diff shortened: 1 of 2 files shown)


src/Components/ImagesTable/ImagesTableToolbar.tsx
```diff
@@ -137,7 +137,7 @@ const ImagesTableToolbar: React.FC<imagesTableToolbarProps> = ({
                 '0 var(--pf-v5-c-toolbar__content--PaddingRight) 0 var(--pf-v5-c-toolbar__content--PaddingLeft)',
             }}
             isInline
-            title={`The selected blueprint is at version ${selectedBlueprintVersion}, images are at version ${latestImageVersion}. Build images to synchronize with the latest version.`}
+            title={`The selected blueprint is at version ${selectedBlueprintVersion}, some images are at version ${latestImageVersion}. Build images to synchronize with the latest version.`}
```
regexowl, line 140: Was thinking about this previously - would it make sense to say: "The selected blueprint is at version ..., _**the latest images**_ are at version .... " since we're only getting the version of the latest images?

Might be clearer than "some" images which could from the message point of view apply even if there are up to date images.

mgold1234: great idea!!

regexowl (approved):
Thank you! LGTM

PR #2088, opened 2024-05-22
V2Wizard: Set pkg recommendations expandable to open by default (HMS-4187)

regexowl:
This makes the pkg recommendations expandable open by default and updates the tests accordingly.


mgold1234 (approved):
/lgtm

PR #2097, opened 2024-05-27
V2Wizard: Add "clear" to package search (HMS-4182)

regexowl:
This adds a "clear" button to the package search input. The button appears only when there is already a searchTerm in the input.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
@@ -34,6 +34,7 @@ import {
   HelpIcon,
   OptimizeIcon,
   SearchIcon,
+  TimesIcon,
 } from '@patternfly/react-icons';
 import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
 import { useDispatch } from 'react-redux';
@@ -613,6 +614,10 @@ const Packages = () => {
     setSearchTerm(selection);
   };
 
+  const handleClear = async () => {
+    setSearchTerm('');
+  };
+
   const handleSelect = (
     pkg: IBPackageWithRepositoryInfo,
     _: number,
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -886,6 +891,17 @@ const Packages = () => {
...
                   />
                 </InputGroupItem>
+                {searchTerm && (
+                  <InputGroupItem>
+                    <Button
+                      variant="control"
+                      aria-label="clear-package-search"
+                      onClick={handleClear}
+                    >
+                      <TimesIcon />
```
amirfefer, line 901: The Button component supports `icon` prop, could you use it?

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -886,6 +891,17 @@ const Packages = () => {
                     onChange={handleSearch}
                   />
                 </InputGroupItem>
+                {searchTerm && (
+                  <InputGroupItem>
+                    <Button
```
amirfefer, line 896: nitpick: it would be nice to add ouia-id for this behavior

amirfefer (commented):
Thanks @regexowl! Works as expected, just small nitpicks :)

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
regexowl: Right, I sure could! Fixed

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
regexowl: Great point, I've added one for the button component

amirfefer (approved):
Thanks @regexowl ! 👍

PR #2102, opened 2024-05-28
buildImageButton: add support of building image with mulitple targets <HMS-4066>

mgold1234:
this commit add support of building an image with mulitple targets, when user build an image, he can choose which targets he wants to build.

### DIFF
--- a/src/Components/Blueprints/BuildImagesButton.tsx
+++ b/src/Components/Blueprints/BuildImagesButton.tsx
@@ -15,42 +15,43 @@ import {
   ButtonProps,
   Button,
 } from '@patternfly/react-core';
+import { addNotification } from '@redhat-cloud-services/frontend-components-notifications/redux';
 import { skipToken } from '@reduxjs/toolkit/query';
 
+import { targetOptions } from '../../constants';
 import { selectSelectedBlueprintId } from '../../store/BlueprintSlice';
-import {useAppDispatch, useAppSelector} from '../../store/hooks';
+import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import {
   ImageTypes,
   useComposeBlueprintMutation,
   useGetBlueprintQuery,
 } from '../../store/imageBuilderApi';
-import {addNotification} from "@redhat-cloud-services/frontend-components-notifications/redux";
 
 export const BuildImagesButton = () => {
   const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
   const [deselectedTargets, setDeselectedTargets] = useState<ImageTypes[]>([]);
-  const [buildBlueprint, { isLoading: imageBuildLoading, isError: imageBuildError }] =
-      useComposeBlueprintMutation();
+  const [buildBlueprint, { isLoading: imageBuildLoading }] =
+    useComposeBlueprintMutation();
   const dispatch = useAppDispatch();
 
   const onBuildHandler = async () => {
     if (selectedBlueprintId) {
-      try{
-      await buildBlueprint({
-        id: selectedBlueprintId,
-        body: {
-          image_types: blueprintImageType?.filter(
-            (target) => !deselectedTargets?.includes(target)
-          ),
-        },
-      })}
-      catch(imageBuildError) {
+      try {
+        await buildBlueprint({
+          id: selectedBlueprintId,
+          body: {
+            image_types: blueprintImageType?.filter(
+              (target) => !deselectedTargets?.includes(target)
+            ),
+          },
+        });
+      } catch (imageBuildError) {
         dispatch(
-            addNotification({
-                  variant: 'warning',
-                  title: 'No blueprint was build',
-                })
-            );
+          addNotification({
+            variant: 'warning',
+            title: 'No blueprint was build',
+          })
+        );
       }
     }
   };
@@ -80,24 +81,6 @@ export const BuildImagesButton = () => {
     }
   };
 
-  const handleToggleEnvironment = (target: ImageTypes) => {
-    switch (target) {
-      case 'aws':
-        return 'Amazon Web Services (AWS)';
-      case 'gcp':
-        return 'Google Cloud Platform (GCP)';
-      case 'azure':
-        return 'Microsoft Azure';
-      case 'oci':
-        return 'Oracle Cloud Infrastructure';
-      case 'vsphere-ova':
-        return 'VMware vSphere';
-      case 'guest-image':
-        return 'Virtualization - Guest image (.qcow2)';
-      case 'image-installer':
-        return 'Bare metal - Installer';
-    }
-  };
   return (
     <Dropdown
       isOpen={isOpen}
@@ -105,6 +88,7 @@ export const BuildImagesButton = () => {
       toggle={(toggleRef: React.Ref<MenuToggleElement>) => (
         <MenuToggle
           variant="primary"
+          data-testid="blueprint-build-image-menu"
           ref={toggleRef}
           isFullWidth
           onClick={onToggleClick}
@@ -113,6 +97,7 @@ export const BuildImagesButton = () => {
             variant: 'action',
             items: [
               <MenuToggleAction
+                data-testid="blueprint-build-image-menu-option"
                 key="split-action"
                 onClick={onBuildHandler}
                 id="wizard-build-image-btn"
@@ -152,6 +137,7 @@ export const BuildImagesButton = () => {
           <MenuList>
             {blueprintImageType?.map((imageType, index) => (
               <MenuItem
+                data-testid="blueprint-menu-items"
                 key={imageType}
                 hasCheckbox
                 itemId={index}
@@ -159,7 +145,7 @@ export const BuildImagesButton = () => {
                   !deselectedTargets || !deselectedTargets.includes(imageType)
                 }
               >
-                {handleToggleEnvironment(imageType)}
+                {targetOptions[imageType]}
               </MenuItem>
             ))}
           </MenuList>
--- a/src/Components/Blueprints/BuildImagesButton.tsx
+++ b/src/Components/Blueprints/BuildImagesButton.tsx
@@ -1,20 +1,183 @@
-import React from 'react';
+import React, { useState } from 'react';
 
-import { Button, ButtonProps } from '@patternfly/react-core';
+import {
+  Dropdown,
+  MenuToggle,
+  MenuToggleElement,
+  Menu,
+  MenuContent,
+  MenuList,
+  MenuItem,
+  Flex,
+  FlexItem,
+  Spinner,
+  MenuToggleAction,
+  ButtonProps,
+  Button,
+} from '@patternfly/react-core';
+import { skipToken } from '@reduxjs/toolkit/query';
 
 import { selectSelectedBlueprintId } from '../../store/BlueprintSlice';
-import { useAppSelector } from '../../store/hooks';
-import { useComposeBlueprintMutation } from '../../store/imageBuilderApi';
+import {useAppDispatch, useAppSelector} from '../../store/hooks';
+import {
+  ImageTypes,
+  useComposeBlueprintMutation,
+  useGetBlueprintQuery,
+} from '../../store/imageBuilderApi';
+import {addNotification} from "@redhat-cloud-services/frontend-components-notifications/redux";
 
-type BuildImagesButtonPropTypes = {
+export const BuildImagesButton = () => {
+  const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
+  const [deselectedTargets, setDeselectedTargets] = useState<ImageTypes[]>([]);
+  const [buildBlueprint, { isLoading: imageBuildLoading, isError: imageBuildError }] =
+      useComposeBlueprintMutation();
+  const dispatch = useAppDispatch();
+
+  const onBuildHandler = async () => {
+    if (selectedBlueprintId) {
+      try{
+      await buildBlueprint({
+        id: selectedBlueprintId,
+        body: {
+          image_types: blueprintImageType?.filter(
+            (target) => !deselectedTargets?.includes(target)
+          ),
+        },
+      })}
+      catch(imageBuildError) {
+        dispatch(
+            addNotification({
+                  variant: 'warning',
+                  title: 'No blueprint was build',
+                })
+            );
+      }
+    }
+  };
+  const [isOpen, setIsOpen] = useState(false);
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+  const { data: blueprintDetails } = useGetBlueprintQuery(
+    selectedBlueprintId ? { id: selectedBlueprintId } : skipToken
+  );
+  const blueprintImageType = blueprintDetails?.image_requests.map(
+    (image) => image.image_type
+  );
+
+  const onSelect = (
+    _event: React.MouseEvent<Element, MouseEvent>,
+    itemId: number
+  ) => {
+    const imageType = blueprintImageType?.[itemId];
+
+    if (imageType && deselectedTargets?.includes(imageType)) {
+      setDeselectedTargets(
+        deselectedTargets.filter((target) => target !== imageType)
+      );
+    } else if (imageType) {
+      setDeselectedTargets([...deselectedTargets, imageType]);
+    }
+  };
+
+  const handleToggleEnvironment = (target: ImageTypes) => {
+    switch (target) {
+      case 'aws':
+        return 'Amazon Web Services (AWS)';
+      case 'gcp':
+        return 'Google Cloud Platform (GCP)';
+      case 'azure':
+        return 'Microsoft Azure';
+      case 'oci':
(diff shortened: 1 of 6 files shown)


src/Components/CreateImageWizardV2/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -39,7 +39,9 @@ export const CreateSaveAndBuildBtn = ({
       (await createBlueprint({
         createBlueprintRequest: requestBody,
       }).unwrap()); // unwrap - access the success payload immediately after a mutation
+
```
regexowl, line 42: Just to check - are the three new lines in `CreateDropdown` and `EditDropdown` files intentional?

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,145 @@
...
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+  const onSelect = (
+    _event: React.MouseEvent<Element, MouseEvent> | undefined,
+    value: string | number | undefined
+  ) => {
+    setIsOpen(false);
+  };
+  if (selectedBlueprintId !== undefined) {
```
ezr-ondrej, line 49: This won't work, hooks (useQuery is a hook) can't be called conditionally, see https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching on details.
I think we could use
```jsx
 } = useGetBlueprintQuery(!!selectedBlueprintId ? { id: selectedBlueprintId } : skipToken)
```

ezr-ondrej:
Do we have a mock for this, or will we get ack from UX on the PR screenshot/recording?

mgold1234:
@ezr-ondrej link to figma -https://www.figma.com/design/9t3owYlJi4M23fUXWpgXCE/HMS-Image-Builder?node-id=970-3284&t=8C1COFOncZHlDngQ-0

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,138 @@
...
+            >
+          </MenuToggle>
+        )}
+      >
+        <Menu onSelect={onSelect2} selected={selectedItems}>
+          <MenuContent>
+            <MenuList>
+              {blueprintImageType?.map((imageType, index) => (
+                <MenuItem
+                  key={index}
```
ezr-ondrej, line 125: could we use the `imageType` as key and itemId and have selectedItems contain the imageType directly, not an index? I'm just thinking out loud, but might be bit safer (the order of blueprintImageType can not be relied on 100% IMO :) )

regexowl:
Went through the mocks and what the PR looks like right now:
- the mocks seem to work on subtraction, meaning all the check boxes will be checked when the drop down menu is opened and you can uncheck environments to remove the target from the build queue. I feel like that makes sense as clicking on "Build images" directly should have the same functionality as it had before, that is building all targets included in the blueprint
image

- the targets have their full name in the drop down, which should be easy to update as we do have a target mapping object in `Target.tsx` which we could export and use
image
image

- when some of the targets are unchecked/removed from the build the text on the button should change from "Build images" to "Build selected", you can probably just compare the length of original list and length of selected list and render the name based on whether they match or something like that
image

src/Components/Blueprints/BuildImagesButton.tsx
regexowl: All for using imageType as key since the values are unique :+1:

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,138 @@
...
-type BuildImagesButtonPropTypes = {
-  variant?: ButtonProps['variant'];
-  children?: React.ReactNode;
-};
-
-export const BuildImagesButton = ({
-  variant,
-  children,
-}: BuildImagesButtonPropTypes) => {
+export const BuildImagesButton = () => {
```
regexowl, line 24: This component is used in two places:

image

When the props were changed a conflict between the attributes was introduced, causing this error:
image

Originally the component was called with variant (which should be "primary" and "link" from top to bottom) and children.

There are two ways to resolve this, either we can try to bend the current logic to be reusable with the other "Build images" button or the link "Build images" button can be added independent of the original component since now it wouldn't be a code duplication as the buttons are quite different.

regexowl: This is how the second button should look like:
image

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,138 @@
...
+    if (selectedBlueprintId) {
+      await buildBlueprint({
+        id: selectedBlueprintId,
+        body: {
+          image_types: selectedTargets
+        },
+      })};
+  };
+  const [selectedItems, setSelectedItems] = React.useState<number[]>([]);
+  const [isOpen, setIsOpen] = React.useState(false);
```
regexowl, line 40: Since you have `useState` in imports you can use just this here:

```suggestion
  const [selectedItems, setSelectedItems] = useState<number[]>([]);
  const [isOpen, setIsOpen] = useState(false);
```

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,138 @@
...
+    } =  useGetBlueprintQuery(selectedBlueprintId ? { id: selectedBlueprintId } : skipToken)
+    const blueprintImageType = blueprintDetails?.image_requests.map(
+        (image) => image.image_type
+    );
+
+  const onSelect2 = (
+      _event: React.MouseEvent<Element, MouseEvent> | undefined,
+      itemId: number | string | undefined
+  ) => {
+    const item = itemId as number;
```
regexowl, line 61: We should stay away from type assertion whenever possible as those only allow space for mistakes and future bugs.

The prop types say that `itemId` can be number, string or undefined, but then the const says it's always number. There is no exception handling so if itemId returns as undefined, the service could crash as this line would try to index with undefined:
```
const imageType = blueprintImageType?.[item];
```

But that being said I agree with Ondřej that we could and probably should index with the imageType, making the type string and going from there :)

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,138 @@
...
+    setIsOpen(false);
+  };
+    const {
+      data: blueprintDetails,
+    } =  useGetBlueprintQuery(selectedBlueprintId ? { id: selectedBlueprintId } : skipToken)
+    const blueprintImageType = blueprintDetails?.image_requests.map(
+        (image) => image.image_type
+    );
+
+  const onSelect2 = (
```
regexowl, line 57: Let's find out more descriptive name - maybe `onSelectTarget`? Or whatever you think will be more fitting :)

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,138 @@
...
+          image_types: selectedTargets
+        },
+      })};
+  };
+  const [selectedItems, setSelectedItems] = React.useState<number[]>([]);
+  const [isOpen, setIsOpen] = React.useState(false);
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+  const onSelect = (
```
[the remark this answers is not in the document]
regexowl: Or wait - do we even need `onSelect` here? Don't these lines do the same thing?

```
onSelect={onSelect}
onOpenChange={(isOpen: boolean) => setIsOpen(isOpen)}
```

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,36 +1,139 @@
...
-  children?: React.ReactNode;
-};
-
-export const BuildImagesButton = ({
-  variant,
-  children,
-}: BuildImagesButtonPropTypes) => {
+export const BuildImagesButton = () => {
   const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
+  const [selectedTargets, setSelectedTargets] =useState<ImageTypes[]>([]);
```
ezr-ondrej, line 25: IMO this means it's all unselected by default, IMO easier would be to
```suggestion
  const [selectedTargets, setSelectedTargets] =useState<ImageTypes[] | undefined>(undefined);
```

ezr-ondrej (commented):
I've left few comments I see we should fix to make the selection work. We have two states for it, I think we should  have just one.

src/Components/Blueprints/BuildImageButtonEmpytyState.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { selectSelectedBlueprintId } from '../../store/BlueprintSlice';
+import { useAppSelector } from '../../store/hooks';
+import { useComposeBlueprintMutation } from '../../store/imageBuilderApi';
+
+type BuildImagesButtonEmptyStatePropTypes = {
+    variant?: ButtonProps['variant'];
+    children?: React.ReactNode;
+};
+
+export const BuildImagesButtonEmptyState = ({
```
ezr-ondrej, line 14: IMO this component could live in the same file (BuildImagesButton), there is no strong need to have it separate.

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,20 +1,160 @@
...
+            {blueprintImageType?.map((imageType, index) => (
+              <MenuItem
+                key={imageType}
+                hasCheckbox
+                itemId={index}
+                isSelected={
+                  !selectedTargets || !selectedTargets.includes(imageType)
+                }
+              >
+                {imageType
```
ezr-ondrej, line 129: This might be better extracted to a separate function.
Might be ok to just have it in this file for now, but ideal would be some utilities/const file.
```js
export const humanReadableImageTarget = (target) =>
  switch(target) {
    case 'aws':
     return 'Amazon Web Services (AWS)';
  .....
```

I also wonder if we don't already have it, we need to have targets human translated.
EDIT: we do not have anything like it yet as far as I could find, so just having such function here localy makes sense :)

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,20 +1,160 @@
...
 
-type BuildImagesButtonPropTypes = {
+export const BuildImagesButton = () => {
+  const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
+  const [selectedTargets, setSelectedTargets] = useState<ImageTypes[]>([]);
+  const [buildBlueprint, { isLoading: imageBuildLoading }] =
+    useComposeBlueprintMutation();
+  const onBuildHandler = async () => {
+    if (selectedBlueprintId) {
+      await buildBlueprint({
```
ezr-ondrej, line 35: I'm not sure here, but where do we handle errors? :thinking:
@regexowl / @amirfefer ?

ezr-ondrej (commented):
I've not tested it, but the code looks quite reasonable, I've left few improvement ideas for the code readability and error handling, but apart of that looks quite ready :)

It would be nice to add at least some basic tests for it. Do you plan to do so in this PR?

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,20 +1,160 @@
...
+                        }
+                        isInline
+                        size="md"
+                      />
+                    </FlexItem>
+                  )}
+                  <FlexItem>
+                    {selectedTargets.length !== 0
+                      ? 'Build selected'
+                      : 'Build Image'}
```
regexowl, line 108:
```suggestion
                      : 'Build images'}
```

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,20 +1,160 @@
...
+          onClick={onToggleClick}
+          isExpanded={isOpen}
+          splitButtonOptions={{
+            variant: 'action',
+            items: [
+              <MenuToggleAction
+                key="split-action"
+                onClick={onBuildHandler}
+                id="wizard-build-image-btn"
+                isDisabled={!selectedBlueprintId}
```
regexowl, line 89: When no targets are selected the "Build images" option is still available and ends up in a toast notifications "Images are being built", but nothing else happens. Let's disable the button when all the targets are deselected.

```suggestion
                isDisabled={
                  !selectedBlueprintId ||
                  selectedTargets.length === blueprintImageType?.length
                }
```

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,20 +1,160 @@
...
+import {
+  ImageTypes,
+  useComposeBlueprintMutation,
+  useGetBlueprintQuery,
+} from '../../store/imageBuilderApi';
 
-type BuildImagesButtonPropTypes = {
+export const BuildImagesButton = () => {
+  const selectedBlueprintId = useAppSelector(selectSelectedBlueprintId);
+  const [selectedTargets, setSelectedTargets] = useState<ImageTypes[]>([]);
```
regexowl, line 30: Nit pick, but since the logic works on subtraction would it make sense to call this "deselectedTargets"?

regexowl (commented):
This looks really good, thanks for working on this! Added some comments :)

src/Components/Blueprints/BuildImagesButton.tsx
regexowl: That's the neat part - we don't... but in reality we probably should. It would be similar to the loading logic, what needs to be done is adding error to the mutation:

```typescript
  const [buildBlueprint, { isLoading: imageBuildLoading, isError: imageBuildError }] =
    useComposeBlueprintMutation();
```

And then do something like this probably?

```typescript
  try {
    await buildBlueprint({
      ...
  catch (imageBuildError) {
    dispatch(
      addNotification({
        <something's wrong>
```

Seems like we seldom properly handle errors in mutations though. :face_with_peeking_eye:

src/Components/Blueprints/BuildImagesButton.tsx
regexowl: Extracting the replace logic into separate function sounds good. I have an idea on how I'd like to deal with the mapping throughout the code base so I'd open a follow up PR once this is merged.

ezr-ondrej (commented):
Nice first test!
Should we also test, that when selecting only one target out of multiple targets that are available, the correct request is send to a server?

src/test/fixtures/blueprints.ts
```diff
@@ -13,10 +13,14 @@ export const mockBlueprintsCreation: CreateBlueprintResponse[] = [
...
     id: '677b010b-e95e-4694-9813-d11d847f1bfc',
   },
+  {
+    id: 'c1cfa347-4c37-49b5-8e73-6aa1d1746cfa',
+  },
 ];
 
 export const mockBlueprintIds = {
   darkChocolate: '677b010b-e95e-4694-9813-d11d847f1bfc',
+  multipleTargets:  'c1cfa347-4c37-49b5-8e73-6aa1d1746cfa',
```
ezr-ondrej, line 23: linter complaints.

src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -1,20 +1,185 @@
...
+      case 'oci':
+        return 'Oracle Cloud Infrastructure';
+      case 'vsphere-ova':
+        return 'VMware vSphere';
+      case 'guest-image':
+        return 'Virtualization - Guest image (.qcow2)';
+      case 'image-installer':
+        return 'Bare metal - Installer';
+    }
+  };
```
regexowl, line 100: https://github.com/osbuild/image-builder-frontend/pull/2135 got merged today - you can use `targetOptions` const to replace this function so the naming is consistent throughout the code base: https://github.com/osbuild/image-builder-frontend/blob/ee216857926bf1a628e74733bd4bf7942b7ad4e3/src/constants.ts#L51-L67

regexowl: But we can split it to a follow up as well, whatever works better :) This is just a heads up

src/test/Components/Blueprints/Blueprints.test.tsx
```diff
@@ -104,6 +107,43 @@ describe('Blueprints', () => {
...
+    await user.click(buildImageDropDown);
+    const awsOption = screen.getByText(/amazon web services \(aws\)/i);
+    await user.click(awsOption);
+    const buildSelectedBtn = screen.getByRole('button', {
+      name: /build selected/i,
+    });
+    expect(buildSelectedBtn).toBeDisabled();
+  });
+
+  test('uncheck one Target option and check that you can build an image', async () => {
```
ezr-ondrej, line 128: The test doesn't seem to test what it says it's testing, or does it?
I see expect on button to be disabled, that's probably not correct, right?

InsightsDroid:

InsightsDroid:

regexowl (approved):
Thank you! :heart:

InsightsDroid:

PR #2110, opened 2024-05-30
test: Fix `'mockPackagesResults' not found`

regexowl:
When switching to using content sources for distro packages an old fixture was left in the handlers. This fixes the issue.


amirfefer (approved):
Thanks @regexowl !

PR #2117, opened 2024-06-04
V2Wizard: add alert on snapshot step

croissanne:
Inform users that the snapshots only apply to RHEL content for now.

---

image

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Snapshot/index.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Snapshot/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { Button, Form, Grid, Text, Title } from '@patternfly/react-core';
+import { Alert, Button, Form, Grid, Text, Title } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizardV2/steps/Snapshot/index.tsx
```diff
@@ -13,6 +13,14 @@ export default function SnapshotStep() {
...
         Repository snapshot
+        <Alert
+          variant="info"
+          isInline
+          title="The snapshot date only applies to RHEL content."
+        >
+          Third party, custom repositories will always use the newest repository
+          state available.
+        </Alert>
       </Title>
```
regexowl, line 24: Can we put the alert after the title instead of inside it? The spacing looks better that way:

image


```suggestion
      <Title headingLevel="h1" size="xl">
        Repository snapshot
      </Title>
      <Alert
        variant="info"
        isInline
        title="The snapshot date only applies to RHEL content"
      >
        Third party, custom repositories will always use the newest repository
        state available.
      </Alert>
```

regexowl (approved):
Looks great, thanks! :sparkles:

PR #2118, opened 2024-06-04
pr_check: change --ref-env to insights-production

jrusz:
As that's our target eventually and it fixes issues with pulp/content-sources


regexowl (approved):
All green, thank you!!

PR #2140, opened 2024-06-11
test: Edit mode tests for repositories

regexowl:
This adds a new 'repositories' fixture to the blueprints as was done for 'oscap'.

Newly added unit tests check that repositories get parsed from and back to request correctly and that when removing previously used repositories an 'Are you sure?' modal pops up.


mgold1234 (approved):
looks perfect!

PR #2141, opened 2024-06-11
Wizard: Add Users step

ezr-ondrej:
Add users step that adds a single user.


ezr-ondrej:
Needs passwordless sudo to be useful

PR #2142, opened 2024-06-11
drop centos stream 8

ondrejbudai:
CentOS Stream 8 is EOL, thus we should no longer offer it in the
service, because it can pose a security risk.

See https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/


regexowl (approved):
Thank you!

PR #2164, opened 2024-06-18
V2Wizard: Hide package recommendations for CentOS

regexowl:
This hides the package recommendations expandable when CentOS is selected as a distribution.


InsightsDroid:

ezr-ondrej (approved):
Makes sense. Test for it would be amazing, but not strictly necessary I guess :)

InsightsDroid:

regexowl:
> Makes sense. Test for it would be amazing, but not strictly necessary I guess :)

True, I'll note it down. The recommendations tests should also include the relatively newly added descriptions so I'll create one common PR for both.

PR #2172, opened 2024-06-20
src / ESLint: Enable `forbid-pf-relative-imports` rule

regexowl:
This enables `forbid-pf-relative-imports` rule and updates affected imports.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -51,4 +51,3 @@ rules:
   "@typescript-eslint/ban-types": off
   # Temporarily disabled
   jsx-a11y/no-autofocus: off
-  rulesdir/forbid-pf-relative-imports: off
--- a/src/Components/Blueprints/BlueprintActionsMenu.tsx
+++ b/src/Components/Blueprints/BlueprintActionsMenu.tsx
@@ -1,13 +1,13 @@
 import React, { useState } from 'react';
 
+import { Dropdown } from '@patternfly/react-core/dist/dynamic/components/Dropdown';
+import { DropdownItem } from '@patternfly/react-core/dist/dynamic/components/Dropdown';
+import { DropdownList } from '@patternfly/react-core/dist/dynamic/components/Dropdown';
 import {
-  Dropdown,
-  DropdownItem,
-  DropdownList,
   MenuToggle,
   MenuToggleElement,
-} from '@patternfly/react-core';
-import { EllipsisVIcon } from '@patternfly/react-icons';
+} from '@patternfly/react-core/dist/dynamic/components/MenuToggle';
+import EllipsisVIcon from '@patternfly/react-icons/dist/dynamic/icons/ellipsis-v-icon';
 import { useFlag } from '@unleash/proxy-client-react';
 
 import { selectSelectedBlueprintId } from '../../store/BlueprintSlice';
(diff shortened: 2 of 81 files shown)


regexowl:
This will be needed for migrating from webpack configs to fec.config

ezr-ondrej:
> This will be needed for migrating from webpack configs to fec.config

I know this is for some reason a spec from shared eslint, but I'd like to understand why before we go for it, I didn't have the time, so I will not oppose doing it, but my uninformed opinion is: It's quite ugly xD

In provisioning, we are using fec config and we didn't migrated to the new eslint (https://github.com/RHEnVision/provisioning-frontend/pull/401), not sure if we can do the same here as the apps are quite different, just saying it was possible there.

regexowl:
> I know this is for some reason a spec from shared eslint, but I'd like to understand why before we go for it, I didn't have the time, so I will not oppose doing it, but my uninformed opinion is: It's quite ugly xD

Completely understandable, I very much agree it's on the ugly side :sweat_smile: 

Let me try to walk through why I think this will be needed. Also for now I've been putting the puzzle together on my own so notes and corrections are very welcome!

The main goal is to bump `@redhat-cloud-services/frontend-components-config` from version 5.x to 6.x so the service is up-to-date and can stay on consoledot. For this we will need to migrate from previously used webpack configs to currently used FEC config which I'm attempting to do here: https://github.com/osbuild/image-builder-frontend/pull/2183 (it's a very drafty draft right now).

The config runs when I'm on 5.x, but as soon as I bump the version to 6.x the build starts throwing following errors:
```bash
Module build failed (from ./node_modules/ts-loader/index.js):
Error: Unable to find source file for module <componentName>! The module likely does not have unique file
as is included within another file. Please add the entry into the COMPONENTS_CACHE in FEC repository
```

I traced back the errors and found out that when I'm trying to resolve the relative imports in this PR, some of the imports return 'false' as their origin by default, causing the error. I've looked around, found the mapper and for the 7 errors we get the imports get resolved as ESM which causes further problems because Jest. :melting_face: 

I guess we can probably not resolve the relative imports and address only the broken ones - either removing them from the code or trying to figure out a proper way to import them. Or maybe we can follow the advice and "add the entry into the COMPONENTS_CACHE in FEC repository" somehow :thinking:

ezr-ondrej:
Now seeing the error, I believe we've hit some issues upgrading to config 6.x

Here is a PR to unblock that: https://github.com/RHEnVision/provisioning-frontend/pull/370

Unfortunatelly I do not have the details, but Amir found out we needed to upgrade to msw 2.0 for some reason.

Also in provisioning we have https://github.com/RHEnVision/provisioning-frontend/blob/be963f681be95784a1d754626dd344631e8f3600/babel.config.js#L44

Which might fix the issue you're hitting, basically fixing the imports at compile time if I understand the block correctly.
That's also Amir's work, dunno where he got that :stuck_out_tongue:

regexowl:
I see, yeah, we currently have typescript blocked by msw 1.x... so that will need to be sorted out as well to bump config it seems, okay.

As for the imports - looks like you're resolving them as esm by default. Which means that I need to either make jest happy with ESM or somehow finish the Vitest migration :thinking:

regexowl:
Closing this one. Rebasing would be a nightmare and it may be possible to keep the previous format of imports. Either way this should be a part of https://github.com/osbuild/image-builder-frontend/pull/2183

PR #2182, opened 2024-06-25
CreateDropdown: change navigation to specific image <HMS-4258>

mgold1234:
this commit change navigation to specific image when creating and building new image

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Review/Footer/CreateDropdown.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Review/Footer/CreateDropdown.tsx
@@ -11,6 +11,8 @@ import {
   Button,
 } from '@patternfly/react-core';
 
+import { setBlueprintId } from '../../../../../store/BlueprintSlice';
+import { useAppDispatch } from '../../../../../store/hooks';
 import {
   CreateBlueprintRequest,
   useComposeBlueprintMutation,
@@ -32,7 +34,7 @@ export const CreateSaveAndBuildBtn = ({
   const [createBlueprint] = useCreateBlueprintMutation({
     fixedCacheKey: 'createBlueprintKey',
   });
-
+  const dispatch = useAppDispatch();
   const onSaveAndBuild = async () => {
     const requestBody = await getBlueprintPayload();
     setIsOpen(false);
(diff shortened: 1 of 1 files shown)


regexowl:
@mgold1234 Can you please resolve the lint errors?

InsightsDroid:

InsightsDroid:

InsightsDroid:

src/Components/CreateImageWizardV2/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -41,7 +43,19 @@ export const CreateSaveAndBuildBtn = ({
...
-    blueprint && buildBlueprint({ id: blueprint.id, body: {} });
+
+    if (blueprint) {
+      buildBlueprint({ id: blueprint.id, body: {} });
+      dispatch(setBlueprintId(blueprint.id));
+
+      setTimeout(() => {
+        const row = document.getElementById(`blueprint-row-${blueprint.id}`);
+        if (row) {
+          row.click();
```
ezr-ondrej, line 54: I wonder why do we need to click?
In my mind we should: invalidate the Blueprints data so we refetch and then dispatch the blueprint ID to select the blueprint, right?

Or is the click needed for some other reasons?

InsightsDroid:

InsightsDroid:

ezr-ondrej (dismissed):
Looks sane to me :)
Do you think we should add tests, or would it be disproportionatelly difficult to the simplicity of the patch?
@regexowl ? :)

regexowl:
Adding tests might be in this case far more complex than the patch so I'm perfectly fine with testing this one manually. Also this is a functionality that should trigger on every blueprint creation so it not working should be easy to catch.

That being said - when I create a blueprint it doesn't get automatically selected for me:

create-blueprint

mgold1234:
@regexowl I understand from the task itslef that it is only when user build an image, sohuld we add it in the creation as well?

regexowl:
I didn't finish my teapot of green tea yet :sweat_smile: But when you are building an image you already have a blueprint selected, right? Since the "Build images" button is visible only when `selectedBlueprintId !== undefined`

From what I understand looking at the task - when you create a blueprint, you get currently sent to "All images". Instead of this we want to send user directly to the "blueprint page" (automatically select the blueprint in the side bar) they just created so they don't need to click on it again to build images.

mgold1234:
that's make sense, thank you

regexowl:
> @regexowl the selection only happens when you "Create and build" Although I do not understand why 😛
> 
> Unless it was a request from someone, I'd select the blueprint even if we're just Creating the blueprint :)

I see! Overlooked that :sweat_smile: Okay, yeah then I think we need it for "Create" also.

ezr-ondrej:
Looks good, but the linter is unhappy :)

InsightsDroid:

ezr-ondrej (approved):
Nice! Works as expected! :+1:

PR #2191, opened 2024-06-27
src: Move content urls to const

regexowl:
This moves both preview and stable links to content to the `constants.ts` file.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
@@ -43,8 +43,7 @@ import CustomHelperText from './components/CustomHelperText';
 import PackageInfoNotAvailablePopover from './components/PackageInfoNotAvailablePopover';
 
 import {
-  CONTENT_BETA,
-  CONTENT_STABLE,
+  CONTENT_URL,
   EPEL_8_REPO_DEFINITION,
   EPEL_9_REPO_DEFINITION,
   RH_ICON_SIZE,
@@ -77,6 +76,7 @@ import {
 } from '../../../../store/wizardSlice';
 import useDebounce from '../../../../Utilities/useDebounce';
 import { useGetEnvironment } from '../../../../Utilities/useGetEnvironment';
+import { betaPath } from '../../utilities/betaPath';
 
 type PackageRepository = 'distro' | 'custom' | 'recommended' | '';
 
@@ -546,7 +546,7 @@ const Packages = () => {
                       target="_blank"
                       iconPosition="right"
                       icon={<ExternalLinkAltIcon />}
-                      href={isBeta() ? CONTENT_BETA : CONTENT_STABLE}
+                      href={betaPath(CONTENT_URL, isBeta())}
                     >
                       Manage your repositories and popular repositories
                     </Button>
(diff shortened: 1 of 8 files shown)


src/constants.ts
```diff
@@ -10,6 +10,8 @@ export const EDIT_BLUEPRINT = `${IMAGE_BUILDER_API}/blueprints`;
 
 export const CDN_PROD_URL = 'https://cdn.redhat.com/';
 export const CDN_STAGE_URL = 'https://cdn.stage.redhat.com/';
+export const CONTENT_BETA = '/preview/insights/content';
+export const CONTENT_STABLE = '/insights/content';
```
ezr-ondrej, line 14: I'm not pushing for this, but would one constant with a function betaUrl(path) => `/review/${path}` be something we would like to entertain?

ezr-ondrej (commented):
Just one nit, let me know what you think, apart of that definitely great idea to do so :orange_heart:

src/constants.ts
regexowl: Great idea! I've added a new commit, let me know if it looks like what you've had in mind.

Since BetaPath uses a `useGetEnvironment()` hook, it needs to be either a custom hook or a React component I think. And it's called conditionally in some places so I've made it a component as a custom hook made the linter unhappy.

InsightsDroid:

InsightsDroid:

src/Components/CreateImageWizardV2/steps/Packages/Packages.tsx
```diff
@@ -544,11 +544,7 @@ const Packages = () => {
                       target="_blank"
                       iconPosition="right"
                       icon={<ExternalLinkAltIcon />}
-                      href={
-                        isBeta()
-                          ? '/preview/insights/content'
-                          : '/insights/content'
-                      }
+                      href={BetaPath(CONTENT_URL)}
```
ezr-ondrej, line 547: Will this work? I know that we were doing the same already, but we have a hook call in a nested inline function.
I'm not sure it's what can do :thinking:

src/Components/sharedComponents/BetaPath.ts
```diff
@@ -0,0 +1,6 @@
+import { useGetEnvironment } from '../../Utilities/useGetEnvironment';
+
+export const BetaPath = (path: string) => {
+  const { isBeta } = useGetEnvironment();
+  return isBeta() ? `/preview/${path}` : path;
+};
```
ezr-ondrej, line 6: This definitelly works, but I was imagining something like:
```suggestion
export const consolePath = (path: string, beta: boolean) => {
  return beta ? `/preview/${path}` : path;
};
```

It would ease up the worries about using custom hook conditionally somewhere. We'd evaluate isBeta in the caller: `consolePath(path, isBeta())`.

Up to you, your solution also works ;)

regexowl: Aah, right! Yeah I've moved the hook call outside of the helper function. I agree treating it as a component was quite awkward :sweat_smile:

InsightsDroid:

ezr-ondrej (approved):
Amazing :orange_heart: great improvement.
Sorry for having nitpicked a lot here :disappointed:

regexowl:
The nit picks were very welcome! :rocket:

PR #2194, opened 2024-06-28
Wizard/Review: Rename from Image to blueprint

jrusz:
Where applicable renamed Image details or name to blueprint details or name.

### DIFF
--- a/src/test/Components/Blueprints/Blueprints.test.tsx
+++ b/src/test/Components/Blueprints/Blueprints.test.tsx
@@ -203,7 +203,9 @@ describe('Blueprints', () => {
         {},
         routes
       );
-      const blueprintDetails = await screen.findByText('Image details');
+      const blueprintDetails = await screen.findByTestId(
+        'image-details-expandable'
+      );
       await user.click(blueprintDetails);
       await screen.findByText(editedBlueprintName);
     });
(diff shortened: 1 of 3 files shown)


src/test/Components/Blueprints/Blueprints.test.tsx
```diff
@@ -203,7 +203,7 @@ describe('Blueprints', () => {
         {},
         routes
       );
-      const blueprintDetails = await screen.findByText('Image details');
+      const blueprintDetails = await screen.findByText('Details');
       await user.click(blueprintDetails);
       await screen.findByText(editedBlueprintName);
```
regexowl, line 208: "Details" appears twice on the Review step, once in a left side navigation and once as a heading of the expandable. This should do the trick :)

```suggestion
      const blueprintDetails = await screen.findAllByText('Details');
      await user.click(blueprintDetails[1]);
      await screen.findByText(editedBlueprintName);
```

regexowl: Or we can use `image-details-expandable` testId. Which might be better because then there's no need to "guess" the index of the correct "Details"

```suggestion
      const blueprintDetails = await screen.findByTestId(
        'image-details-expandable'
      );
      await user.click(blueprintDetails);
      await screen.findByText(editedBlueprintName);
```

regexowl (approved):
Nice catch! Thank you!

InsightsDroid:

PR #2207, opened 2024-07-03
wizard/compliance: re-enable stig profiles (HMS-4018)

kingsleyzissou:
The stig profiles have been fixed downstream, so we can re-enable these.

### DIFF
--- a/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
+++ b/src/Components/CreateImageWizardV2/steps/Oscap/Oscap.tsx
@@ -208,26 +208,15 @@ const ProfileSelector = () => {
         onFilter={(_event, value) => {
           if (profiles) {
             return [<OScapNoneOption key="oscap-none-option" />].concat(
-              profiles
-                // stig and stig_gui don't boot at the moment,
-                // so we should filter them out
-                .filter((profile_id) => {
-                  const brokenProfiles = [
-                    'xccdf_org.ssgproject.content_profile_stig',
-                    'xccdf_org.ssgproject.content_profile_stig_gui',
-                  ];
-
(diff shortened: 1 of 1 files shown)


ochosi (approved):
LGTM!

InsightsDroid:

InsightsDroid:

kingsleyzissou:
I need to re-think this, the fix only looks like it works for `9.4` and `8.10`, so we'll need to filter the profiles out for minor releases earlier than those two.

regexowl:
UI is by default creating blueprints with the latest major version of RHEL, so I guess only old blueprints would be affected. For those could we maybe just add a condition like `(distribution === RHEL_8 || distribution === RHEL_9) ? profiles : profiles.filter` ?

kingsleyzissou:
> UI is by default creating blueprints with the latest major version of RHEL, so I guess only old blueprints would be affected. For those could we maybe just add a condition like `(distribution === RHEL_8 || distribution === RHEL_9) ? profiles : profiles.filter` ?

Yeah I can do something like that just to be extra sure

kingsleyzissou:
@regexowl actually just tried to edit an old blueprint and it uses the latest release, so we don't need to filter and this PR should be fine :)

regexowl:
> @regexowl actually just tried to edit an old blueprint and it uses the latest release, so we don't need to filter and this PR should be fine :)

Yup, on edit it automatically bumps the RHEL version. I was thinking about old blueprints that won't be edited, but those should have stig profiles disabled which I didn't realize for some reason :sweat_smile:  so I guess yeah, we should be fine

regexowl (approved):
Thank you!

InsightsDroid:

PR #2233, opened 2024-07-15
Wizard: fix blueprint unique name validation in Edit mode

### LINKED ISSUE #2220: Details step is disabled on blueprint edit
When editing a blueprint I can't go to the Details step, I have to walk through the wizard to get there which I believe was not the case before. Also the tests are failing on this as when they get to this step the Next button is disabled.

disabled-step

ezr-ondrej:
Fixes #2220

In cases there is already a state in the wizardSlice, there might have been cases of flicker, when unique request could be started for edit mode, because wizards loads before requestMapper maps the new state of edited blueprints.

In such case, name is not empty (leftover from previous BP create/edit), but wizardMode was not yet set, so defaults to create.

This addresses it by just making sure the validation works as expected in Edit mode.
Now when the name is changed in Edit mode, we run the validation.
We now have the blueprintID in the state, which is empty on create state (might be just easier to recognize mode by id in the future :) )

### DIFF
--- a/src/test/fixtures/editMode.ts
+++ b/src/test/fixtures/editMode.ts
@@ -248,8 +248,8 @@ export const awsCreateBlueprintRequest: CreateBlueprintRequest = {
 
 export const awsBlueprintResponse: BlueprintResponse = {
   ...awsCreateBlueprintRequest,
-  id: mockBlueprintIds['registration'],
-  description: mockBlueprintDescriptions['registration'],
+  id: mockBlueprintIds['aws'],
+  description: mockBlueprintDescriptions['aws'],
 };
 
 export const gcpImageRequest: ImageRequest = {
--- a/src/Components/CreateImageWizardV2/utilities/requestMapper.ts
+++ b/src/Components/CreateImageWizardV2/utilities/requestMapper.ts
@@ -182,6 +182,7 @@ export const mapRequestToState = (request: BlueprintResponse): wizardState => {
   }
   return {
     wizardMode,
+    blueprintId: request.id,
     details: {
       blueprintName: request.name,
       blueprintDescription: request.description,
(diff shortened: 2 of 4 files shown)


ezr-ondrej:
Maybe I've actually broken it? we now fire a unique name request in edit mode. Maybe it needs to get mocked? :thinking:

InsightsDroid:

regexowl:
> Maybe I've actually broken it? we now fire a unique name request in edit mode. Maybe it needs to get mocked? 🤔

Maybe? But if we are firing a new request, why does it fail in AWS edit mode test only?

ezr-ondrej:
> Maybe? But if we are firing a new request, why does it fail in AWS edit mode test only?

No idea honestly xD but it passes on other PRs (e.g. https://github.com/osbuild/image-builder-frontend/pull/2237).
I'll try to dig a little

src/test/fixtures/editMode.ts
```diff
@@ -248,8 +248,8 @@ export const awsCreateBlueprintRequest: CreateBlueprintRequest = {
 
 export const awsBlueprintResponse: BlueprintResponse = {
   ...awsCreateBlueprintRequest,
-  id: mockBlueprintIds['registration'],
-  description: mockBlueprintDescriptions['registration'],
+  id: mockBlueprintIds['aws'],
+  description: mockBlueprintDescriptions['aws'],
```
ezr-ondrej, line 252: Hehe this was the culprit :stuck_out_tongue:
We loaded wrong id on the blueprint get request and then we were assuming the aws blueptintID is different then what the unique request was returning and failing to allow saving the changes :)

regexowl: That was probably my doing :face_with_peeking_eye: :sweat_smile: Thanks for digging into it!! :bug:

InsightsDroid:

regexowl (approved):
Thank you! Let's merge this next to unblock @jrusz

PR #2236, opened 2024-07-16
jest: Increase test timeout

regexowl:
This increases default test timeout from 10000 to 20000 and timeout for "image name invalid for more than 100 chars and description for 250" to 50000.

This should eliminate test flakes caused by timing out, later it will be "reverted" when the Vitest migration gets merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "setupFilesAfterEnv": [
       "./src/test/jest.setup.js"
     ],
-    "testTimeout": 10000
+    "testTimeout": 20000
   },
   "devDependencies": {
     "@babel/core": "7.24.7",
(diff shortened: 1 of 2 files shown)


InsightsDroid:

ezr-ondrej:
I've found many flakes in test Provisioning suite where we have been waitng for some network timeout because of improperly set msw handlers :)
Just saying, lets do this for now for sure, easy solutions are good solutions :sunglasses:

ezr-ondrej (approved):
Love the branch name! :stuck_out_tongue:

regexowl:
> Just saying, lets do this for now for sure, easy solutions are good solutions 😎

Cool! This should be "reverted" by the vitest migration, but should unblock merging for now

> Love the branch name! 😛

Forgot to update it :sweat_smile:

PR #2186, opened 2024-06-26
Bump msw

regexowl:
This bumps Mock Service Worker and updates the handlers.

Migrated following https://mswjs.io/docs/migrations/1.x-to-2.x

PR #2245, opened 2024-07-18
Bump msw

ezr-ondrej:
Build on top of https://github.com/osbuild/image-builder-frontend/pull/2186 and in the last commit:

* Add origin to have absolute path in MSW
* Fix the spy handler event to make sure it works as expected
* Also bumps few dependencies, but IMO that's not needed.

### DIFF
--- a/src/store/emptyImageBuilderApi.ts
+++ b/src/store/emptyImageBuilderApi.ts
@@ -6,7 +6,7 @@ import { IMAGE_BUILDER_API } from '../constants';
 export const emptyImageBuilderApi = createApi({
   reducerPath: 'imageBuilderApi',
   baseQuery: fetchBaseQuery({
-    baseUrl: IMAGE_BUILDER_API,
+    baseUrl: window.location.origin + IMAGE_BUILDER_API,
     prepareHeaders: (headers) => {
       // help the backend distinguish between requests from the UI and the API
       headers.set('X-ImageBuilder-ui', 'true');
--- a/src/test/Components/CreateImageWizard/wizardTestUtils.tsx
+++ b/src/test/Components/CreateImageWizard/wizardTestUtils.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 
 import { screen, waitFor } from '@testing-library/react';
 import { userEvent } from '@testing-library/user-event';
-import { MockedRequest } from 'msw';
 
 import ImageWizard from '../../../Components/CreateImageWizard';
 import { RHEL_9 } from '../../../constants';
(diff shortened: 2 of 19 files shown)


src/store/emptyImageBuilderApi.ts
```diff
@@ -6,7 +6,7 @@ import { IMAGE_BUILDER_API } from '../constants';
 export const emptyImageBuilderApi = createApi({
   reducerPath: 'imageBuilderApi',
   baseQuery: fetchBaseQuery({
-    baseUrl: IMAGE_BUILDER_API,
+    baseUrl: window.location.origin + IMAGE_BUILDER_API,
```
ezr-ondrej, line 9: this was the culprit of the errors.

src/test/Components/CreateImageWizard/wizardTestUtils.tsx
```diff
@@ -17,8 +16,9 @@ type RequestTypes = 'GET' | 'PUT' | 'POST' | 'DELETE';
 
 export function spyOnRequest(pathname: string, method: RequestTypes) {
   return new Promise((resolve) => {
-    const listener = async (req: MockedRequest) => {
-      if (req.url.pathname === pathname && req.method === method) {
+    const listener = async ({ request: req, requestId }: { request: Request, requestId: string }) => {
```
ezr-ondrej, line 19: this listener also needed to be fixed, as the structure was changed.

src/test/setup.ts
```diff
@@ -1,9 +1,9 @@
-import nodeFetch, { Request, Response } from 'node-fetch';
+// import nodeFetch, { Request, Response } from 'node-fetch';
 
 import { server } from './mocks/server';
 import 'vitest-canvas-mock';
 
-Object.assign(global, { fetch: nodeFetch, Request, Response });
+// Object.assign(global, { fetch: nodeFetch, Request, Response });
```
ezr-ondrej, line 6: I've removed this and it changed nothing, no idea why it's here, any idea @regexowl ?
But IMO it is irrelevat to the fix.

src/test/fixtures/composes.ts
```diff
@@ -26,10 +24,8 @@ export const mockComposesEmpty: ComposesResponse = {
 const currentDate = new Date();
 const currentDateInString = currentDate.toISOString();
 
-export const composesEndpoint = (
-  req: RestRequest<never, PathParams<string>>
-) => {
-  const params = req.url.searchParams;
+export const composesEndpoint = (url: URL) => {
+  const params = new URLSearchParams(url.search);
```
ezr-ondrej, line 28: This fixes the ImagesTable tests - by fixing the composes endpoint.

src/test/setup.ts
regexowl: I've added the polyfills during the vitest migration as there were errors outputted without them. But it doesn't seem like the case anymore with bumped msw, so I'd say we can get rid of them :+1:

ezr-ondrej:
https://github.com/osbuild/image-builder-frontend/pull/2186 is in :tada:

PR #2253, opened 2024-07-19
dependabot: Decrease open PR limit

regexowl:
The backlog of stuck dependencies is shrinking, meaning we can go back to previous limit of three open dependabot pull requests.


InsightsDroid:

ezr-ondrej (approved):
Would be nice to have a way to skip the pr_check :P or make it faster :)

PR #2254, opened 2024-07-19
devDeps: Fix one of `npm ci` warnings

regexowl:
`@babel/plugin-proposal-object-rest-spread` was merged to the ECMAScript standard and is no longer maintained, this replaces it with `@babel/plugin-transform-object-rest-spread`.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -10,7 +10,7 @@ module.exports = {
     // Put _extends helpers in their own file
         '@babel/plugin-transform-runtime',
         // Support for {...props} via Object.assign({}, props)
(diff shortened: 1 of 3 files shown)


InsightsDroid:

ezr-ondrej:
Ah this might not be needed if we do #2267, gotcha :)

regexowl:
> Ah this might not be needed if we do #2267, gotcha :)

True! Marking this as blocked for now

regexowl:
Since it looks like we'll be getting rid of Babel we won't need this. Closing for now, let's reopen it if needed.

PR #2262, opened 2024-07-22
devDeps, src: Bump ts-patch and update imports

regexowl:
This bumps ts-patch and updates PF imports where needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,7 +74,7 @@
         "stylelint": "15.11.0",
         "stylelint-config-recommended-scss": "13.1.0",
         "ts-node": "10.9.2",
-        "ts-patch": "3.2.0",
+        "ts-patch": "3.2.1",
         "typescript": "5.5.4",
         "uuid": "10.0.0",
         "vitest": "2.0.4",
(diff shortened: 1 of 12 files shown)


ezr-ondrej (approved):
I do not love the specific imports, but sure, if PF insists...
I guess it's for a reason I just don't understand :)

regexowl:
Rebased will need one more review :bow: 

> I do not love the specific imports, but sure, if PF insists...

I think there might be gaps in PF mapper for dynamic files which resolves some imports as `import <nameOfComponent> from 'false';`. This solves the problem on our side by pointing the component to its esm file.

ezr-ondrej (approved):
I do not understand one bit xD
If it works, I'm good, but might be good to get second pair of eyes given we are setting trend for these imports :shrug:

regexowl:
I think a solution would be to update mapper on PF side. I agree this is not ideal, but if it's okay I'd merge this to address the dep bump. But also created an issue as a reminder to revisit this later: https://github.com/osbuild/image-builder-frontend/issues/2330

ezr-ondrej:
No objections to merge on my side ;)
Just saying I do not understand the impact of esm vs non-esm, the nuances are lost on me, but it builds, so we should be good ;)

PR #2263, opened 2024-07-22
test: Improve stability of 'check error details' test

regexowl:
This adds awaiting to a few expects, hopefully improving the stability of the test.

PR #2267, opened 2024-07-23
Remove babel

regexowl:
After migrating to typescript and later to vitest, babel is no longer needed. This removes its config and all dependencies that are no longer necessary.


regexowl:
Added another commit with a non-solution to one of the failing tests. I've disabled it for now. :sweat: 

The thing is, what the test needs is to just add an `await` before `userEvent click()` method on line 224, but if I do that the output of the test gets plagued by tons of "need to be wrapped in act()" warnings. Which I mean fine - if wrapping in act solved the issue, but it does not. I'd love to test the particular part of the code, but I also really want clean test output, dunno :shrug:

ezr-ondrej:
Should we do #2263 and then this?
I've not looked in the act wrapping, no idea what's that about, but user.click should already be wrapping it, right?
Aren't we doing something weird in the code that quacks as a user action?

regexowl:
> Should we do https://github.com/osbuild/image-builder-frontend/pull/2263 and then this?

Let's!

> I've not looked in the act wrapping, no idea what's that about, but user.click should already be wrapping it, right?
Aren't we doing something weird in the code that quacks as a user action?

Yeah, userEvent methods should be already wrapped in `act()` AFAIK, that's what's weird about it.

> Aren't we doing something weird in the code that quacks as a user action?

We probably are, the question would be what and where. That's why I'm trying to go through it and remove what's probably not needed anymore :sweat_smile:

ezr-ondrej (dismissed):
I guess we won't be more ready then we are :thinking: xD

ezr-ondrej:
Lets get #2277 in and then merge this? :)

ezr-ondrej:
I guess we can open a draft PR with re-enabling the test as a reminder?

regexowl:
I'll rebase this after all the smaller PRs go in and we can merge this let's say on Monday just in case something goes wrong. I still believe the probability is small, but it's never zero :grimacing: 

> I guess we can open a draft PR with re-enabling the test as a reminder?

Good idea, I've opened an issue for now so we have it logged somewhere https://github.com/osbuild/image-builder-frontend/issues/2285

regexowl:
Rebased, this will need re-review pretty please and then let's merge first thing Monday morning.

Also good news - I've re-enabled the test and it runs smoothly now after we've added retry and bumped vitest version so that's nice :relaxed:

regexowl:
Can't reproduce the failure locally even when switched to node 18.0.0 sooo that's kinda annoying :thinking: It's not popping up when running the entire test suite nor when running the test file in isolation.

regexowl:
When in doubt, add more `await`s :upside_down_face: 

The failure seems to be resolved and this should be ready for review again.

ezr-ondrej (approved):
Shall we do this? :sunglasses:

PR #2268, opened 2024-07-23
imageTable: link images to parent blueprint <HMS-4353>

mgold1234:
this commit add link to parent blueprint from specific image

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -11,6 +11,7 @@ import {
   Spinner,
   Bullseye,
   Badge,
+  Button,
 } from '@patternfly/react-core';
 import {
   ActionsColumn,
@@ -22,6 +23,7 @@ import {
   Thead,
   Tr,
 } from '@patternfly/react-table';
+import { useDispatch } from 'react-redux';
 import { NavigateFunction, useNavigate } from 'react-router-dom';
 
 import './ImagesTable.scss';
@@ -51,6 +53,7 @@ import {
   selectLimit,
   selectOffset,
   selectSelectedBlueprintId,
+  setBlueprintId,
 } from '../../store/BlueprintSlice';
 import { useAppSelector } from '../../store/hooks';
 import {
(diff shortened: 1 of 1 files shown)


ezr-ondrej (changes requested):
I'm not sure we need to fetch all the data for this, I believe we do have all data we need.

Few thoughts:

1. I don't think we need to use Link component from react router, that has a purpose of redirecting to different ract-router managed page, which we don't do here, we are staying on the same page.
2. I believe we have the blueprint_id in compose here, we could use that instead of fetching it in the handleClick function.
3. We should ideally make it optional, so ad-hoc images (composes) that do not have blueprint_id, do not have this link, that would fail

Here is a rough example (not working code) what I mean:

```jsx
import { Button } from '@patternfly/core'
// ......

// might not even need to be a separate function and might be inlined
const handleClick = (blueprintId) => { dispatch(setBlueprintId(blueprintId)) }

// ..... 

{compose.blueprint_id && (
  <Button variant="link" onClick={(_evt) => handleClick(compose.blueprint_id)}>
    {compose.image_name || compose.id}
  </Button>{' '}
) || ({compose.image_name || compose.id})
```

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -452,6 +455,17 @@ const Row = ({
 }: RowPropTypes) => {
   const [isExpanded, setIsExpanded] = useState(false);
   const handleToggle = () => setIsExpanded(!isExpanded);
+  const dispatch = useDispatch();
+
+  const handleClick = ({
+    blueprintId,
+  }: {
+    blueprintId: string | undefined | null;
+  }) => {
```
regexowl, line 464: We already have a type for blueprintId, can we use that?

```suggestion
  const handleClick = ({
    blueprintId,
  }: {
    blueprintId: BlueprintItem['id'];
  }) => {
```

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -463,7 +477,14 @@ const Row = ({
             onToggle: () => handleToggle(),
           }}
         />
-        <Td dataLabel="Image name">{compose.image_name || compose.id}</Td>
+        <Td dataLabel="Image name">
+          <Button
+            variant="link"
+            onClick={() => handleClick({ blueprintId: compose.blueprint_id })}
```
regexowl, line 483: Maybe we should render the name as a button only when the image is connected to a blueprint? When the blueprint id is undefined, we should probably just do plain text. What do you think?

Currently when you have "orphan" image it's name is a link, but it doesn't do anything when clicking on it:
image

regexowl (commented):
Looks very nice! Added some nit picks :)

PR #2279, opened 2024-07-25
wizard/registration: Add missing article (typo)

ochosi:
When looking at the registration step I noticed that there was a missing article. So here's a tiny fix!


regexowl:
Great catch, thank you! Can you please fix the lint error?

ezr-ondrej:
Ah I can't squash on this repo! xD this would be bit better as one commit, but I guess it's not mandatory. It'll be forever known that Simon doesn't know that you always need to run `npm run lint:js -- --fix` because the linter is never happy even if you just add an article xD

ezr-ondrej (approved):
:orange_heart: attention to detail :)

ezr-ondrej:
:tada:

PR #2281, opened 2024-07-25
Firstboot: validate shebang is defined (HMS-4458)

ezr-ondrej:
Shebang needs to be defined in the script, otherwise it fails to execute with weird errors.
This makes sure we validate this for User, to surface this info to them.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -28,6 +28,7 @@ import Azure from './steps/TargetEnvironment/Azure';
 import Gcp from './steps/TargetEnvironment/Gcp';
 import {
   useFilesystemValidation,
+  useFirstBootValidation,
   useDetailsValidation,
 } from './utilities/useValidation';
 import {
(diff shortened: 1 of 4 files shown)


src/test/Components/CreateImageWizard/steps/FirstBoot/Firstboot.test.tsx
```diff
@@ -104,6 +104,22 @@ describe('First Boot step', () => {
     await goToFirstBootStep();
     await screen.findByText('First boot configuration');
   });
+
+  // test('should validate shebang', async () => {
```
ezr-ondrej, line 108: I've written the test, but it doesn't work because as below, vitest dom doesn't work well with the editor we use.
I'll try to find a follow-up.

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -203,6 +207,11 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
       (s) => s.index > step.index && s.isVisited
     );
 
+    // Only this code is different from the original
```
ezr-ondrej, line 210: I've pulled this up to make it clear what's the difference with the original code :)

PR #2286, opened 2024-07-26
test: Move compliance On Recreate test to On Edit

regexowl:
This migrates a test from previously used On Recreate to current On Edit.


regexowl:
This moves and enables one of the disabled tests. We'll need to re-enable and update more, but let's take it one step at a time for now.

src/test/Components/CreateImageWizard/steps/Oscap/Oscap.test.tsx
```diff
@@ -206,4 +191,30 @@ describe('OpenSCAP edit mode', () => {
...
+    });
+    user.click(packagesNavBtn);
+    await screen.findByRole('heading', {
+      name: /Additional packages/i,
+    });
+    const selectedBtn = await screen.findByRole('button', {
+      name: /Selected/i,
+    });
+    user.click(selectedBtn);
+    await screen.findByText('neovim');
```
ezr-ondrej, line 218: This changed from `nftables` to `neovim` :thinking:
I know nada about the oscap, so unsure what's happening, thus I figured I'd rather ask ;)

We should change the comment above at least I guess :)

regexowl: Yeah, might look a bit chaotic :sweat_smile: The test got disabled during work on Blueprints and in the meantime the mocks changed. This should be the object that gets returned as a mocked response from the commit that made the changes: https://github.com/osbuild/image-builder-frontend/commit/a1bdcaa54ff5e46e63b82c9c99f3a36fd523b4bc#diff-07b400aeb49d2de7ed7da9354500d974cacc41ea79688a0def4ac864b9bed1b0L17-L43

'nftables' were removed from packages together with other and replaced by 'neovim' only.

Great point about the comment, I'll update it right away!

ezr-ondrej: Ah, gotcha, thanks for explaining :orange_heart:

ezr-ondrej (approved):
Thanks! :orange_heart:

PR #2287, opened 2024-07-26
Wizard: disable non-snapshot repositories when needed (HMS-4481)

croissanne:
If a snapshot date is selected, all selected repositories need to have snapshotting enabled. So disable the ones which don't have snapshotting enabled.


mgold1234 (approved):
/lgtm

regexowl (approved):
I see so when snapshots are enabled incompatible repositories cannot be selected. Makes sense and looks good, thanks!

I think the PR will need a push to re-trigger the pr-best-practices bot with the updated description.

PR #2291, opened 2024-07-29
createImage: change resolution to be able to see buttons in create image wizard (HMS-4485)

mgold1234:
this commit change the resoluiton of wizard__outer-wrap to solve
the problem that user should scroll to see the next, backl and cancel buttons

https://github.com/user-attachments/assets/f54826fb-bf1a-4b2b-81db-516b1793d2c0

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.scss
+++ b/src/Components/CreateImageWizard/CreateImageWizard.scss
@@ -2,9 +2,6 @@
     padding-right: 0px;
 }
 
-.pf-v5-c-wizard__nav {
-    overflow-y: unset;
-}
 
 .pf-c-popover[data-popper-reference-hidden="true"] {
     font-weight: initial;
@@ -88,7 +85,6 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
     display: flex
 }
 
-// Fixes min height issues on the main wizard container
-.pf-v5-c-wizard__outer-wrap {
-    min-height: 602px;
+.pf-v5-c-wizard__main {
+    flex: 1 1 0
 }
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -8,6 +8,7 @@ import {
   WizardStep,
   WizardStepType,
   useWizardContext,
+  PageSection,
 } from '@patternfly/react-core';
 import { useFlag } from '@unleash/proxy-client-react';
 import { useNavigate, useSearchParams } from 'react-router-dom';
@@ -229,7 +230,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   return (
     <>
       <ImageBuilderHeader inWizard />
-      <section className="pf-l-page__main-section pf-c-page__main-section">
+      <PageSection isWidthLimited>
         <Wizard
           startIndex={startIndex}
           onClose={() => navigate(resolveRelPath(''))}
@@ -433,7 +434,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
             <ReviewStep snapshottingEnabled={snapshottingEnabled} />
           </WizardStep>
         </Wizard>
-      </section>
+      </PageSection>
     </>
   );
 };


regexowl:
Is this ready for review? Looked around and it seems to be working nicely :heart: can you please update the commit message so it doesn't start with [wip], add a PR description and rebase?

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -90,5 +90,5 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
 
 // Fixes min height issues on the main wizard container
 .pf-v5-c-wizard__outer-wrap {
-    min-height: 602px;
+   height: 50px;
```
lucasgarfield, line 93: This works but I wish we could explain why... without knowing why this is working it feels like a bit of a hack.

I'm worried that at some point some upstream css (for instance, in `insights-chrome`) is going to change and we're going to end up with a 50 pixel high wizard! Likely? Probably not. But possible? Maybe...

lucasgarfield (changes requested):

src/Components/CreateImageWizard/CreateImageWizard.scss
mgold1234: maybe the idea to change it to 50px is good if we remove the section and add pageSection like we have in the imageTable page WDYT @lucasgarfield

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -233,8 +234,8 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   return (
     <>
       <ImageBuilderHeader inWizard />
-      <section className="pf-l-page__main-section pf-c-page__main-section">
-        <Wizard
+      <PageSection>
+        <Wizard height={550}
```
lucasgarfield, line 238: Let's remove this height, it hard codes the wizard height to 550 but what we really want is for the wizard to be able to shrink/grow so the footer is always visible.

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -88,7 +88,4 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
     display: flex
 }
 
-// Fixes min height issues on the main wizard container
-.pf-v5-c-wizard__outer-wrap {
-    min-height: 602px;
-}
+
```
lucasgarfield, line 91: I spent a little time looking into this and it’s pretty tricky, you have to really understand flexbox behavior.

There are several nested flexboxes on the page, and we do need them as we want the wizard to grow or shrink with the viewport so the footer is always visible.

The actual element that is growing is this one: `<div class=”pf-v5-c-wizard__main”>`. It has `flex: 1 1 auto`. These values represent `flex-grow`, `flex-shrink`, and `flex-basis`, respectively. This flexbox is able to grow and is what is causing the wizard (its parent element) to grow and cause the entire page to become scrollable.

If we change the `flex-basis` from `1` to `0`, the element will no longer be able to grow and instead the content is rendered within a scrollable element within the wizard.
 
Although both this and setting a height work, I think this might make our intentions a little clearer for future readers of the code than setting a height, since it directly addresses the cause of the growth – the flexbox behavior.

To sum up, I think adding
```
.pf-v5-c-wizard__main {
    flex: 1 1 0
}
```
would work nicely for us.

lucasgarfield: I also think it is a good idea to remove 
```
// Fixes min height issues on the main wizard container
.pf-v5-c-wizard__outer-wrap {
    min-height: 602px;
}
```, so let's keep that! :+1: 

We also might want to think about removing this too:
```
.pf-v5-c-wizard__nav {
    overflow-y: unset;
}
```

Otherwise, at small viewport heights, the nav bar won't be scrollable.

File: src/Components/CreateImageWizard/CreateImageWizard.tsx [hunk 4]
```diff
@@ -233,8 +234,8 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   return (
     <>
       <ImageBuilderHeader inWizard />
-      <section className="pf-l-page__main-section pf-c-page__main-section">
-        <Wizard
+      <PageSection>
```
lucasgarfield, line 237: Love this! Doesn't the `<LandingPage>` have an `isWidthLimited` prop? Would we want to put that here?

src/Components/CreateImageWizard/CreateImageWizard.scss
mgold1234: work perfect!! thank you

mgold1234: where did you find the  .pf-v5-c-wizard__main {
    flex: 1 1 0
} in patternFly? I looked here - https://www.patternfly.org/components/wizard

lucasgarfield (approved):
Nice work, thank you!!!

PR #2292, opened 2024-07-29
Export: Use export endpoint (HMS-4459)

avitova:
We have a new export endpoint, that I used here.
I also incorporated some kind of change for centos-8 that was pulled from the API, and is not relevant. :thinking:

### DIFF
--- a/src/Components/Blueprints/ImportBlueprintModal.test.tsx
+++ b/src/Components/Blueprints/ImportBlueprintModal.test.tsx
@@ -63,7 +63,7 @@ const IGNORE_SUBSCRIPTION_BLUEPRINT = `{
   "name": "Blueprint test"
 }`;
 
-const IGNORE_TARGETS_BLUEPRINT = `{
+const INVALID_ARCHITECTURE_JSON = `{
   "customizations": {
     "files": [
     ],
@@ -169,18 +169,15 @@ describe('Import modal', () => {
     await waitFor(() => expect(helperText).toBeInTheDocument());
   });
 
-  test('should enable button and ignore targets in blueprint file', async () => {
+  test('should show alert on invalid blueprint incorrect architecture', async () => {
     await setUp();
-    await uploadFile(`blueprints.json`, IGNORE_TARGETS_BLUEPRINT);
+    await uploadFile(`blueprints.json`, INVALID_ARCHITECTURE_JSON);
     const reviewButton = screen.getByTestId('import-blueprint-finish');
-    await waitFor(() => expect(reviewButton).not.toHaveClass('pf-m-disabled'));
-    user.click(reviewButton);
-
-    await waitFor(async () =>
-      expect(
-        await screen.findByText('Image output', { selector: 'h1' })
-      ).toBeInTheDocument()
+    expect(reviewButton).toHaveClass('pf-m-disabled');
+    const helperText = await screen.findByText(
+      /not compatible with the blueprints format\./i
     );
+    await waitFor(() => expect(helperText).toBeInTheDocument());
   });
 
   test('should enable button and ignore subscription in blueprint file', async () => {
--- a/src/Components/Blueprints/ImportBlueprintModal.test.tsx
+++ b/src/Components/Blueprints/ImportBlueprintModal.test.tsx
@@ -4,106 +4,122 @@ import userEvent from '@testing-library/user-event';
 import { renderCustomRoutesWithReduxRouter } from '../../test/testUtils';
 
 const BLUEPRINT_JSON = `{
-    "customizations": {
-      "files": [
-      ],
-      "kernel": {
-      },
-      "openscap": {
-      },
-      "packages": [
-        "aide",
-        "sudo",
-        "audit",
-        "rsyslog",
-        "firewalld",
-        "nftables",
-        "libselinux"
-      ],
-      "services": {
-        "enabled": [
-          "crond",
-          "firewalld",
-          "systemd-journald",
-          "rsyslog",
-          "auditd"
-        ]
-      },
-      "subscription": {
-      }
-    },
-    "description": "Tested blueprint",
-    "distribution": "rhel-93",
-    "id": "052bf998-7955-45ad-952d-49ce3573e0b7",
-    "image_requests": [
-      {
-        "architecture": "aarch64",
-        "image_type": "aws",
-        "upload_request": {
-          "options": {
-            "share_with_sources": [
-              "473980"
-            ]
-          },
-          "type": "aws"
-        }
-      }
+  "customizations": {
+    "packages": [],
+    "subscription": {
+      "activation-key": "",
+      "base-url": "",
+      "insights": false,
+      "organization": 0,
+      "server-url": ""
+    }
+  },
+  "description": "Lorem ipsum dolor 2211 sit amet, consectetur adipiscing elit. Pellentesque malesuada ultricies diam ac eleifend. Proin ipsum ante, consequat vel justo vel, tristique vestibulum lorem. Vestibulum sit amet pulvinar orci. Vivamus vel ipsum.",
+  "distribution": "rhel-8",
+  "metadata": {
+    "exported_at": "2024-07-29 17:26:51.666952376 +0000 UTC",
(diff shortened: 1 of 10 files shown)


src/test/Components/Blueprints/Blueprints.test.tsx
avitova, line 1: This is basically just a test that there exists a button for export. I might delete it if you feel like it is not needed.
The current implementation of the Export is that we use "handleClick", where we create an "artificial" link, that we "artificially" click and invoke the download operation. This is quite hard to test with unit tests.
Afaik, if we wanted to do this in a more simple way, we'd have to call the `export` endpoint even at times when we would not want to export the blueprint.

src/test/Components/Blueprints/Blueprints.test.tsx
```diff
@@ -167,22 +168,6 @@ describe('Blueprints', () => {
     ).not.toBeInTheDocument();
   });
 
-  test('CentOS 8 Stream renders', async () => {
```
ezr-ondrej, line 170: Shouldn't we keep this test? That should still work, right?
If there is a blueprint saved with centos-8, we should make sure we render some warning.

src/test/fixtures/blueprints.ts
```diff
@@ -322,7 +322,7 @@ export const mockCentosBlueprintComposes: GetBlueprintComposesApiResponse = {
       created_at: '2021-04-27T12:31:12Z',
       blueprint_version: 1,
       request: {
-        distribution: 'centos-8',
+        distribution: 'rhel-10-nightly',
```
ezr-ondrej, line 325: Is this why you needed to delete the test?
I think we could convince the handler to return special case centos-8 (copy paste it into the centos-8 test) in the test itself. MSW can update single mock in test itself

ezr-ondrej (commented):
The change itself seems straight forward, but I wonder why we are changing so many tests. Especially the deleted test worries me a bit.

src/Components/Blueprints/ImportBlueprintModal.test.tsx
```diff
@@ -24,106 +24,122 @@ vi.mock('@unleash/proxy-client-react', () => ({
...
+      "insights": false,
+      "organization": 0,
+      "server-url": ""
+    }
+  },
+  "description": "Lorem ipsum dolor 2211 sit amet, consectetur adipiscing elit. Pellentesque malesuada ultricies diam ac eleifend. Proin ipsum ante, consequat vel justo vel, tristique vestibulum lorem. Vestibulum sit amet pulvinar orci. Vivamus vel ipsum.",
+  "distribution": "rhel-8",
+  "metadata": {
+    "exported_at": "2024-07-29 17:26:51.666952376 +0000 UTC",
+    "parent_id": "b3385e6d-ecc4-485c-b33c-f65131c46f52"
```
avitova, line 41: I just changed this valid blueprint to add the metadata and test it. We also should not have image_requests and id there. They would not matter, however, this is a test for success with a correct blueprint.

avitova:
@ezr-ondrej I deleted image_requests field, because it is not necessary there, it would go through, but I do not see a reason to test it. I also chose to add metadata to the blueprint that is supposed to end up with success.
Some of the deleted test blueprints are just a git being git, and showing incorrect changes. :thinking: That is the case, for example, with `INVALID_JSON`.

ezr-ondrej:
Gotcha! Makes sense :)
The PR LGTM, I'll try to get around to test it tomorrow :)

avitova:
Sorry, this testing with await and flags won over me a few times. PR should be ready to review now.

ezr-ondrej (commented):
Great work, thank you! :+1:
I've got just one DRY concern, let me know if we can do better, if not, we will need to rely on unit tests :)

ezr-ondrej:
Ah you are hitting issues with `FileSystemPartitionMode` rename. See https://github.com/osbuild/image-builder-frontend/pull/2343

avitova:
Meh, CI pipeline was failing, let's try again after new rebase.

ezr-ondrej (dismissed):
Cool, looks good to me :+1: I've not tested yet.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -137,6 +143,70 @@ const getLatestRelease = (distribution: Distributions) => {
...
+    firstBoot: {
+      script: getFirstBootScript(request.customizations.files),
+    },
+    distribution: getLatestRelease(request.distribution),
+    repositories: {
+      customRepositories: request.customizations.custom_repositories || [],
+      payloadRepositories: request.customizations.payload_repositories || [],
+      recommendedRepositories: [],
+    },
+    registration: {
```
ezr-ondrej, line 182: Given we [filter out the registration](https://github.com/osbuild/image-builder/blob/767cb9ce2496146e193d5c58ebbafe68956d5ab3/internal/v1/handler_blueprints.go#L254), it should probably be ommited for export, right (meaning not be in the common section)?

ezr-ondrej (commented):
Nice job! :orange_heart: 
I've only found one issue with now. When you import exported Blueprint, it has automatically `register-later`.
I belive we want to default to register-now-rhc, unless set otherwise.  I think it's because we are ommiting registration, we default to register-later, thus IMO we should ommit registration from the common function and maybe just default to `register-now-rhc` for imported (I don't know what happens if we ommit it completely, but if it has the same effect, then I'd ommit completely :))

avitova:
I addressed your comment @ezr-ondrej thanks for catching that. Tbh I was kinda afraid to break the "create" workflow, that would be unfortunate. But yeah, we can definitely use some of the common parts.

ezr-ondrej (dismissed):
Nice! Works well :orange_heart:

ezr-ondrej:
Hehe, you have broken it! :stuck_out_tongue:
@jrusz we're checking image_requests, we've now removed that from the "Download blueprint", I guess that's the culprit here? :)

jrusz:
> Hehe, you have broken it! 😛 @jrusz we're checking image_requests, we've now removed that from the "Download blueprint", I guess that's the culprit here? :)

Yup :) 
```
21:59:42 >       assert expected_values["image_requests"] == blueprint_details["image_requests"]
21:59:42 E       KeyError: 'image_requests'
```
I'll update the test.

avitova:
Thank you guys for the cooperation!:)

src/test/Components/Blueprints/Blueprints.test.tsx
```diff
@@ -11,6 +11,30 @@ import { emptyGetBlueprints } from '../../fixtures/blueprints';
...
+      case 'image-builder.snapshots.enabled':
+        return true;
+      case 'image-builder.import.enabled':
+        return true;
+      default:
+        return false;
+    }
+  }),
+}));
+
```
regexowl, line 37: A late FYI - mocking chrome and unleash should no longer be needed within test files. It's included in the test setup file now. https://github.com/osbuild/image-builder-frontend/blob/f60e80b4e88fbb9604b213818bfa5acd7404c0ba/src/test/setup.ts#L18-L64

We can remove the local mocks in a small follow-up

avitova: Okay, sure! If we can do that in a follow-up, that would be nice. I will start working on it, and open a PR once we merge this. :)

regexowl:
Wrote a short update to frontend channel - the issue we were facing with eph last week should be resolved! But now there's a gitlab outage with a few hours ETA for the fix :upside_down_face: so if the tests fail with bad gateway that's why

avitova:
/retest
I think that the fix for the pipeline was just merged. :pray:

ezr-ondrej:
This still looks related:

```
        imported_bp_details = imported_bp.details
>       imported_bp_details.pop("id")
E       KeyError: 'id'

/iqe_venv/lib/python3.11/site-packages/iqe_image_builder/tests/test_blueprints.py:418: KeyError
```

ezr-ondrej (approved):
LGTM :orange_heart: Thank you all for figuring this one out together! :tada:

regexowl:
Sooo green :green_heart:

avitova:
Oh, nice! Thank you guys!

PR #2298, opened 2024-07-30
templates/testing: Increase timeout for selenium session

jrusz:
I think it's causing issues with test_compose_clone so I want to try increasing the timeout.

### DIFF
--- a/deploy/iqe-trigger-integration.yml
+++ b/deploy/iqe-trigger-integration.yml
@@ -106,6 +106,8 @@ objects:
               value: ${SELENIUM_JAVA_OPTS}
             - name: VNC_GEOMETRY
               value: ${VNC_GEOMETRY}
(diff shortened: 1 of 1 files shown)


mgold1234 (approved):
/lgtm

deploy/iqe-trigger-integration.yml
```diff
@@ -177,3 +179,5 @@ parameters:
   value: '1920x1080'
 - name: IQE_PARALLEL_ENABLED
   value: "false"
+- name: SE_NODE_SESSION_TIMEOUT
+  value: 600
```
BumbleFeng, line 183: `error processing template: [None]: error: failed to read input object (not a Template?): unable to decode "STDIN": json: cannot unmarshal number into Go struct field Parameter.parameters.value of type string
`
```suggestion
  value: "600"
```

PR #2299, opened 2024-07-30
requestMapper: Automatically replace CentOS 8 with CentOS 9

regexowl:
CentOS 8 was deprecated and is no longer used, but 'On Edit' for a no longer valid blueprint with 'centos-8' as their distribution defaulted to RHEL 9.


avitova:
Thank you, Klara!

PR #2300, opened 2024-07-30
api: Update image-builder API

regexowl:
This updates image-builder API by running `npm run api` and addresses problems caused by the update.

### DIFF
--- a/api/schema/imageBuilder.yaml
+++ b/api/schema/imageBuilder.yaml
@@ -242,6 +242,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPErrorList'
+  /blueprints/{id}/export:
+    parameters:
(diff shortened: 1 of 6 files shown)


regexowl:
Is using type assertions much of a cop out in this case? :face_with_peeking_eye:

ezr-ondrej:
> Is using type assertions too much of a cop out in this case? 🫣 I mean 'centos-8' should be in the place of distribution in the blueprint but is not in the Distributions type anymore.

Hehe interesting case :stuck_out_tongue: But looks like you've convinced it to work :+1:

ezr-ondrej:
/retest

Not exactly sure what went wrong there xD It looks 100% passed :face_exhaling:

ezr-ondrej (approved):

PR #2301, opened 2024-07-30
Wizard: disable bulk select when appropriate

croissanne:
If repositories are disabled, don't select them via bulk select.

https://github.com/user-attachments/assets/d64bf5ca-62cb-441c-a59f-d642d9c6ed92


regexowl (approved):
Excellent, thank you! :sparkles:

PR #2306, opened 2024-07-31
change the value of SE_NODE_SESSION_TIMEOUT from number to string

mgold1234:
this commit change the value of SE_NODE_SESSION_TIMEOUT 
from number to string becuase it is blocking auto-promotion test


BumbleFeng (approved):

jrusz (approved):
Thank you!

PR #2308, opened 2024-08-01
Wizard: "Always blue" update to the Registration step (HMS-4484)

regexowl:
This updates the Registration step as per recent mocks.

There is now only one checkbox which switches between (register everything now / register later). When the checkbox is not checked "register later" option gets selected, activation key dropdown gets disabled and information about activation keys is hidden to support the fact that no activation key will be used.

When the checkbox is checked it either loads previously used activation key in the case of "on edit" scenario, looks for a recently used activation key in localStorage or in a case there are no activationKeys automatically creates and selects one.

The Next button gets disabled only when "register-now" is selected, but there is no activationKey selected. As the option to clear the dropdown got removed this should occur only when fetching/creating activation key.


regexowl:
@jrusz This will most probaly have an effect on pr_check as Registration step pretty much completely changes. Could you please take a look at what would need to be updated?

This is how the step looks like now based on [mocks](https://www.figma.com/design/9t3owYlJi4M23fUXWpgXCE/HMS-Image-Builder?node-id=367-8927&t=fBR8LsGXDX63Jcmg-0):
image

The idea is that the Next button should be always enabled (it gets disabled for "register now" with missing activation key, but as the key is populated automatically and can't be removed it should take 1s at max to re-enable the button).

Checked checkbox means register now with everything (rhsm, insights, rhc), unchecked means register later.

The key gets populated a few different ways:
- if there was an activation key previously set it's loaded from the state
- if there was a recently used activation key for a different image this gets loaded from local storage
- if there was no previously chosen key, the first key in API response gets selected
- if there is no key at all, a default-activation-key is created and automatically selected.

regexowl:
I gave up and disabled `fileParallelism` again for now. I'll take a look at it in a separate PR, didn't want to block this one on a test failing because of async.

Gotta love locally non-reproducible Node 18 only tests :upside_down_face:

lucasgarfield:
This is awesome Klara!

But ugh… I just realized something…  I’m very, very sorry I didn’t think of this before but I don’t think we can go straight to eliminating the radio buttons for the other registration types yet. :facepalm: 

The reason is backwards compatibility… how will we handle blueprints that were previously created with `register-now-insights` or `register-now-rhc`? Either in “edit” mode or “import” mode?

@ezr-ondrej  and I recently discussed the need for blueprint versions and the ability to gracefully handle version changes like this – but we don’t have a solution yet. This makes me realize we should prioritize that work, but for now, I’d still like to get this in sooner rather than later. 

Maybe I’m wrong and missing some obvious solution? That'd be nice...

I think we could also add some additional test coverage here. I can also think of three new tests we can add: 

1. The user has no activation keys. Ensure a key gets created. 
2. The user has activation keys, but does not have an entry in the browser’s local storage. Ensure the first key gets added and ends up in the request correctly.
3. The user does have activation keys, and does have an entry in the browser’s local storage. Ensure that key gets added and ends up in the request correctly.

Note that for test (1) I don’t think we should ensure the key makes it into the request, that’s something that would be better suited to the IQE tests – mocking the Activation Keys API (GET empty array of keys, POST request to make keys, GET array of keys again with newly created key) would just be too complicated.

Of course Activation Keys doesn’t launch in Ephemeral, as we recently learned discussing this with @jrusz, but maybe we can work on getting that fixed.

regexowl:
I've added tests for a key stored in localStorage and for no key in localStorage. But got really stuck on the third case - creating a default key. I might be doing something wrong (which is very much a possibility), but I can't seem to make the case work.

What I've tried is creating a new handler for intercepting the POST request:
```typescript
  http.post(`${RHSM_API}/activation_keys`, async () => {
    return HttpResponse.json({
      body: {
        id: '2',
        name: 'activation-key-default',
        serviceLevel: 'Self-Support',
      },
    });
  }),
```

and mocking an empty response to activation_keys GET request to trigger the POST one within a test:
```typescript
    server.use(
      http.get(`${RHSM_API}/activation_keys`, () => {
        return HttpResponse.json({ body: [] });
      })
    );
```

The POST request triggers, but looking at the network events in a non-mocked env it looks like after the POST request there should be another GET request to the activation_keys end point which fetches new activation_keys now containing the 'activation-key-default'. Even with the "newly created key" added to the mocked response (which by itself seems like it's slightly against the tests) the new GET posts doesn't seem to do what it should :thinking:

ochosi:
@regexowl I agree with the functionality as described above by @lucasgarfield - that's what we discussed in previous UX meetings to my understanding.

So if the PR works as advertised (the tests seem to suggest it does?) I'd be good with merging this as is.

ezr-ondrej (approved):
Code LGTM.
Just one nit, that actually should not be in scope of this PR anyway :)

src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
```diff
@@ -136,7 +95,7 @@ const ManageKeysButton = () => {
       isInline
       href={isProd() ? ACTIVATION_KEYS_PROD_URL : ACTIVATION_KEYS_STAGE_URL}
     >
-      Activation keys page
+      Customer portal
```
ezr-ondrej, line 98: I don't know if this comes from mocks, but it's wrong.
Customer portal is where AKs were managed in the past, before we've got Activation keys page.
If we'd want to link  to customer portal, we should link https://access.redhat.com/
But if I'm correct, AKs can't even be managed there anymore.

ezr-ondrej (commented):
Small caveat while testing, but no idea how to solve it xD

https://github.com/user-attachments/assets/aef5e45e-f8c1-41ed-9169-a9834c66ebb0

ezr-ondrej:
I think for the above ^ if I check the check-box it should automatically enable `register-now-rhc`.

src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
```diff
@@ -145,6 +104,9 @@ const ActivationKeysList = () => {
   const dispatch = useAppDispatch();
 
   const activationKey = useAppSelector(selectActivationKey);
+  const registrationType = useAppSelector(selectRegistrationType);
+
+  const defaultActivationKeyName = 'activation-key-default';
```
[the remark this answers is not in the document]
regexowl: Great point, yeah, I think we need to address that in the follow up then. Random suffix sounds like a way to go. Created an issue for now: https://github.com/osbuild/image-builder-frontend/issues/2353

regexowl:
> I think for the above ^ if I check the check-box it should automatically enable `register-now-rhc`.

Agree, I'll update the code. Thanks!

src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
regexowl: Oh, good to know. Yeah, it's from the mocks, but this part might not have been updated then. I'll double check and update the copy. Thanks for the heads up!

regexowl:
Alright, the PR is updated. I've added the two fixes to a new commit to make it easier to review :)

regexowl:
The rebasing hurts :smiling_face_with_tear: but the tests should be passing once more

ezr-ondrej (approved):
Amazing work :orange_heart:

jrusz:
Apparently the unique name validation can fail in ephemeral as well. Give it like ~10 minutes, new plugin version where the test is disabled will be available.

regexowl:
/retest

Hmm, failed on `[clowdapp/content-sources-backend] timed out waiting for resource to be ready, details:   clowdapp/content-sources-backend not ready` [again](https://ci.int.devshift.net/job/osbuild-image-builder-frontend-pr-check/3047/)

PR #2309, opened 2024-08-01
Wizard: Replace `isBeta()` with conditional flag (HMS-4513)

regexowl:
This replaces `isBeta()` with a constraint added to the 'image-builder.pkgrecs.enabled' unleash flag.


lucasgarfield (approved):
Works nicely!

PR #2310, opened 2024-08-01
createImage : add default name in details step <HMS-4494>

mgold1234:
this commit add default name for blueprint name field to be able
to create an image without filling any name
format name: "$OS $ARCH $TARGETS $DATE".

### DIFF
--- a/src/Components/CreateImageWizard/steps/Details/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Details/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import {
   Form,
@@ -14,9 +14,13 @@ import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
 import {
   changeBlueprintDescription,
   changeBlueprintName,
+  selectArchitecture,
   selectBlueprintDescription,
   selectBlueprintName,
+  selectDistribution,
+  selectImageTypes,
 } from '../../../../store/wizardSlice';
+import { generateDefaultName } from '../../utilities/generateDefaultName';
 import { useDetailsValidation } from '../../utilities/useValidation';
 import { HookValidatedInput } from '../../ValidatedTextInput';
 
@@ -24,6 +28,19 @@ const DetailsStep = () => {
   const dispatch = useAppDispatch();
   const blueprintName = useAppSelector(selectBlueprintName);
   const blueprintDescription = useAppSelector(selectBlueprintDescription);
+  const distribution = useAppSelector(selectDistribution);
+  const arch = useAppSelector(selectArchitecture);
+  const targetEnvironments = useAppSelector(selectImageTypes);
+
+  useEffect(() => {
+    if (!blueprintName) {
+      dispatch(
+        changeBlueprintName(
+          generateDefaultName(distribution, arch, targetEnvironments)
+        )
+      );
+    }
+  }, []);
   const handleNameChange = (
     _event: React.FormEvent<HTMLInputElement>,
     name: string
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
@@ -395,7 +395,6 @@ describe('Step Upload to AWS', () => {
     await clickNext();
     await clickNext();
     await clickNext();
-    await clickNext();
     await enterBlueprintName();
     await clickNext();
     // informational modal pops up in the first test only as it's tied
--- a/src/test/Components/CreateImageWizard/steps/Details/Details.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Details/Details.test.tsx
@@ -25,7 +25,6 @@ const goToDetailsStep = async () => {
   await clickNext();
   await clickNext();
   await clickNext();
-  await clickNext();
 };
 
 const enterBlueprintDescription = async () => {
@@ -53,7 +52,7 @@ describe('validates name', () => {
     await clickRegisterLater();
     await goToDetailsStep();
     const nextButton = await getNextButton();
-    expect(nextButton).toBeDisabled();
+    expect(nextButton).toBeEnabled();
     await enterBlueprintName(' ');
     await waitFor(() => expect(nextButton).toBeDisabled());
   });
--- a/src/test/Components/CreateImageWizard/steps/Repositories/Repositories.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Repositories/Repositories.test.tsx
@@ -43,7 +43,6 @@ const goToReviewStep = async () => {
   await clickNext(); // Additional packages
   await clickNext();
   await clickNext(); // First Boot
-  await clickNext(); // Details
   await enterBlueprintName();
   await clickNext(); // Review
 };
--- a/src/test/Components/CreateImageWizard/steps/Snapshot/Snapshot.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Snapshot/Snapshot.test.tsx
@@ -38,7 +38,6 @@ const goToReviewStep = async () => {
   await clickNext(); // Repositories step
   await clickNext(); // Additional packages
   await clickNext();
-  await clickNext(); // Details
   await enterBlueprintName();
   await clickNext();
 };
--- a/src/Components/CreateImageWizard/utilities/generateDefaultName.ts
+++ b/src/Components/CreateImageWizard/utilities/generateDefaultName.ts
@@ -1,0 +1,15 @@
+import { ImageTypes } from '../../../store/imageBuilderApi';
+
+export const generateDefaultName = (
+  distribution: string,
+  arch: string,
+  targetEnvironments: ImageTypes[]
+) => {
+  const date = new Date();
+  const day = date.getDate().toString().padStart(2, '0');
+  const month = date.toLocaleString('en-US', { month: 'long' });
+  const year = date.getFullYear();
+  return `${distribution} ${arch} ${targetEnvironments.join(
+    ' '
+  )} ${day} ${month} ${year}`;
+};
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -105,7 +105,7 @@ export function useDetailsValidation(): StepValidation {
           setUniqueName(true);
         });
     }
-  }, [blueprintId, name, setUniqueName, trigger]);
+  }, [blueprintId, name, setUniqueName, trigger, nameValid]);
 
   let nameError = '';
   if (!nameValid) {
--- a/src/test/Components/CreateImageWizard/wizardTestUtils.tsx
+++ b/src/test/Components/CreateImageWizard/wizardTestUtils.tsx
@@ -126,6 +126,9 @@ export const enterBlueprintName = async (name: string = 'Red Velvet') => {
   const blueprintName = await screen.findByRole('textbox', {
     name: /blueprint name/i,
   });
+
+  await waitFor(() => user.clear(blueprintName));
+  await waitFor(() => expect(blueprintName).toHaveValue(''));
   await waitFor(() => user.type(blueprintName, name));
 };
 


src/Components/CreateImageWizard/validators.ts
```diff
@@ -48,8 +48,8 @@ export const isMountpointMinSizeValid = (minSize: string) => {
 export const isBlueprintNameValid = (blueprintName: string) =>
   blueprintName !== undefined &&
   blueprintName.length >= 2 &&
-  blueprintName.length <= 100 &&
-  /\w+/.test(blueprintName);
+  blueprintName.length <= 300 &&
+  /[a-zA-Z0-9]{2,}/.test(blueprintName);
```
ezr-ondrej, line 52: This changes the non-asci characters from being valid to not being valid. Was that an intention?

src/Components/CreateImageWizard/steps/Details/index.tsx
```diff
@@ -24,13 +27,32 @@ const DetailsStep = () => {
...
   const blueprintDescription = useAppSelector(selectBlueprintDescription);
+  const distribution = useAppSelector(selectDistribution);
+  const arch = useAppSelector(selectArchitecture);
+  const targetEnvironments = useAppSelector(selectImageTypes);
+
+  useEffect(() => {
+    if (!blueprintName) {
+      dispatch(changeBlueprintName(generateDefaultName()));
+    }
+  }, [dispatch, blueprintName]);
```
ezr-ondrej, line 38: Won't this generate a name when I type a name and then delete it?
I think that using empty dependency array is correct in this case.

src/Components/CreateImageWizard/steps/Details/index.tsx
```diff
@@ -24,13 +27,32 @@ const DetailsStep = () => {
...
   const handleNameChange = (
     _event: React.FormEvent<HTMLInputElement>,
     name: string
   ) => {
     dispatch(changeBlueprintName(name));
   };
 
+  const generateDefaultName = () => {
```
ezr-ondrej, line 46: I think this could be a function in a helper file and take `distribution`, `arch` `environments` as arguments :)
It's not a big deal, but it would be slightly cleaner.

src/Components/CreateImageWizard/steps/Details/index.tsx
```diff
@@ -24,13 +27,32 @@ const DetailsStep = () => {
...
+    const date = new Date();
+
+    const day = date.getDate().toString().padStart(2, '0');
+    const month = date.toLocaleString('en-US', { month: 'long' });
+    const year = date.getFullYear();
+    const environments = targetEnvironments.join(' ');
+    return `${distribution} ${arch} ${environments} ${day} ${month} ${year}`;
```
ezr-ondrej, line 53: This name does not pass the validation, does it?

ezr-ondrej (commented):
I left few comments, some nits, but the changed validation worries me, is that an intention to have strictrer rules?

Also we could now change all the tests to not fill the blueprint name, but go with the default :stuck_out_tongue:

src/Components/CreateImageWizard/steps/Details/index.tsx
mgold1234: thanks

src/Components/CreateImageWizard/steps/Details/index.tsx
mgold1234: good idea

src/Components/CreateImageWizard/steps/Details/index.tsx
mgold1234: should it? if we know for sure that the default name is valid, why should check the validation?

ezr-ondrej: I was just thinking we disabled spaces, but we didn't, so this passes the validation.
My concern was this would be judged as invalid name and wouldn't allow user to proceed, but I was wrong ;)
Disregard! :)

src/test/Components/CreateImageWizard/CreateImageWizard.content.test.tsx
```diff
@@ -158,10 +158,12 @@ describe('Step Packages', () => {
     await clickNext();
   };
 
-  test('clicking Next loads Image name', async () => {
+  test('pclicking Next loads Image name', async () => {
```
ezr-ondrej, line 161:
```suggestion
  test('clicking Next loads Image name', async () => {
```

src/test/Components/CreateImageWizard/CreateImageWizard.content.test.tsx
```diff
@@ -158,10 +158,12 @@ describe('Step Packages', () => {
...
     await setUp();
 
-    await clickNext();
+    // skip packages
+    //  await clickNext();
+    // skip repositories
```
ezr-ondrej, line 166: if we don't need it, lets remove it, tho I do not understand why we have less steps now :thinking:

ezr-ondrej (commented):
Hmm I do not understand why there is less clicks on next in all the steps? :thinking:
I'm not saying it's necessarily wrong, I just don't understand :)
Also, if it's really not needed, lets remove it, instead of commenting it out.

src/test/Components/CreateImageWizard/CreateImageWizard.content.test.tsx
mgold1234: actually I didnt understand too, and I saw that First Boot step didnt exist when I check it locally, is it possible?

src/Components/CreateImageWizard/validators.ts
```diff
@@ -51,6 +53,20 @@ export const isBlueprintNameValid = (blueprintName: string) =>
...
+) => {
+  const date = new Date();
+  const day = date.getDate().toString().padStart(2, '0');
+  const month = date.toLocaleString('en-US', { month: 'long' });
+  const year = date.getFullYear();
+  return `${distribution} ${arch} ${targetEnvironments.join(
+    ' '
+  )} ${day} ${month} ${year}`;
+};
+
```
regexowl, line 69: I don't think this function should be in validators. We can maybe create a separate `generateDefaultName.ts` file for it in `src/CreateImageWizard/utilities`?

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -423,10 +423,10 @@ describe('Step Upload to AWS', () => {
...
     await clickNext();
     await clickNext();
-    await clickNext();
-    await clickNext();
-    await enterBlueprintName();
-    await clickNext();
+    // await clickNext();
+    // await clickNext();
+    // await enterBlueprintName();
+    //  await clickNext();
```
regexowl, line 429: Is there a reason to leave the commented lines in the test code?

regexowl (commented):
Looks good! I've added one comment and a question.

I don't see any reason to leave commented lines in the code aside from a code we know we'll need in the future.

src/test/Components/CreateImageWizard/CreateImageWizard.content.test.tsx
regexowl: Some of it could be flags, but also did anyone reeeally count the steps before? If we got to Details step, clicked on disabled Next button three times and then filled in a name and continued with the button now enabled, the tests would pass, right? :sweat_smile:

src/Components/CreateImageWizard/validators.ts
mgold1234: thank you I think it is a good solution!

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -636,9 +632,9 @@ describe('Step Registration', () => {
     await clickNext();
     await clickNext();
     await clickNext();
-    await clickNext();
-    await enterBlueprintName();
-    await clickNext();
+    //  await clickNext();
+    //  await enterBlueprintName();
+    // await clickNext();
```
regexowl, line 637: Sorry for not being more specific, I've meant we can remove all of the commented lines :sweat_smile: So these...

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -686,10 +682,10 @@ describe('Step Registration', () => {
...
     await clickNext();
     await clickNext();
-    await clickNext();
-    await clickNext();
-    await enterBlueprintName();
-    await clickNext();
+    //   await clickNext();
+    //  await clickNext();
+    // await enterBlueprintName();
+    //   await clickNext();
```
regexowl, line 688: ... and these...

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -772,10 +765,10 @@ describe('Step Registration', () => {
...
     await clickNext();
     await clickNext();
-    await clickNext();
-    await clickNext();
-    await enterBlueprintName();
-    await clickNext();
+    //  await clickNext();
+    // await clickNext();
+    //  await enterBlueprintName();
+    //  await clickNext();
```
regexowl, line 771: ... and these...

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -1060,9 +1053,9 @@ describe('Step Review', () => {
     await clickNext();
     // skip packages
     await clickNext();
-    await clickNext();
+    // await clickNext();
```
regexowl, line 1056: ... this can go...

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -1060,9 +1053,9 @@ describe('Step Review', () => {
...
     // skip firstboot
-    await clickNext();
+    //await clickNext();
```
regexowl, line 1058: ... this can also go...

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -1124,11 +1117,10 @@ describe('Step Review', () => {
     await clickNext();
     // skip repositories
     await clickNext();
+    // skip details
     await clickNext();
     // skip First boot
-    await clickNext();
-    await enterBlueprintName();
-    await clickNext();
+    //await clickNext();
```
regexowl, line 1123: ... we don't need this line...

src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
```diff
@@ -1314,7 +1306,7 @@ describe('Keyboard accessibility', () => {
     await clickNext();
     await clickNext();
     // TODO: Focus on textbox on Details step
-    await clickNext();
+    //   await clickNext();
```
regexowl, line 1309: ... neither this line...

src/test/Components/CreateImageWizard/steps/Details/Details.test.tsx
```diff
@@ -58,7 +58,7 @@ const goToDetailsStep = async () => {
   await clickNext();
   await clickNext();
   await clickNext();
-  await clickNext();
+  // await clickNext();
```
regexowl, line 61: ... and this line...

src/test/Components/CreateImageWizard/steps/Repositories/Repositories.test.tsx
```diff
@@ -76,7 +76,7 @@ const goToReviewStep = async () => {
   await clickNext(); // Additional packages
   await clickNext();
   await clickNext(); // First Boot
-  await clickNext(); // Details
+  // await clickNext();// Details
```
regexowl, line 79: ... this line...

src/test/Components/CreateImageWizard/steps/Snapshot/Snapshot.test.tsx
```diff
@@ -71,7 +71,7 @@ const goToReviewStep = async () => {
   await clickNext(); // Repositories step
   await clickNext(); // Additional packages
   await clickNext();
-  await clickNext(); // Details
+  // await clickNext(); // Details
```
regexowl, line 74: ... and this line. :D

PR #2312, opened 2024-08-01
Wizard: Migrate WSL from `isBeta()` to unleash flag (HMS-4513)

regexowl:
This gates WSL target behind an 'image-builder.wsl.enabled' unleash flag instead of `isBeta()`.


ezr-ondrej (dismissed):
LGTM :+1:
~Tho this would expose it everywhere, do we want to add the strategy in the flag to expose it only in Preview?~ :)
EDIT: Sorry, safe to merge.

PR #2314, opened 2024-08-01
Wizard: Fix death by thousand reviews (HMS-4513)

regexowl:
When switching between preview and non-preview the Wizard nav got messy. This should fix the problem.


lucasgarfield (approved):
Thanks for the quick fix! Looks good to me. I'm glad we caught the edit mode bug.

PR #2315, opened 2024-08-01
Wizard: Add `isHidden` to snapshotting step

regexowl:
Replace conditional rendering with `isHidden` for snapshotting step.


lucasgarfield (approved):
Makes good sense. Hopefully there is no more conditional rendering lurking in the wizard...

lucasgarfield (approved):
Perfetto!

PR #2321, opened 2024-08-05
devDeps: bump vitest and @vitest/coverage-v8

regexowl:
This bumps:
- vitest from 2.0.4 to 2.0.5
- @vitest/coverage-v8 from 2.0.4 to 2.0.5

and removes reporter from the coverage setting, leaving only default output for now (console output equal to jest).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "@typescript-eslint/eslint-plugin": "8.1.0",
         "@typescript-eslint/parser": "8.0.1",
         "@vitejs/plugin-react": "4.3.1",
-        "@vitest/coverage-v8": "2.0.4",
+        "@vitest/coverage-v8": "2.0.5",
         "chart.js": "4.4.3",
         "chartjs-adapter-moment": "1.0.1",
         "chartjs-plugin-annotation": "3.0.1",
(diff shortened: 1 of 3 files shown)


ezr-ondrej:
Flaky or related?
```
10:25:20  FAIL  src/Components/Blueprints/ImportBlueprintModal.test.tsx > Import model > should show alert on invalid blueprint
10:25:20 Error: Timed out in waitFor.
```

regexowl:
> Flaky or related?
> 
> ```
> 10:25:20  FAIL  src/Components/Blueprints/ImportBlueprintModal.test.tsx > Import model > should show alert on invalid blueprint
> 10:25:20 Error: Timed out in waitFor.
> ```

I'm around 75 % sure it's just a flake. FileParallelism is great when it works, but in our case it proves to be quite flakey. I've added a commit that would disable it again for the time being. It's a part of the registration step update PR, but maybe I should just put it in a separate PR...

ezr-ondrej (approved):
Dunno why it won't go green tho xD

regexowl:
> Dunno why it won't go green tho xD

Only in pr_check though, that's the weird thing. And it's not like the tests are failing consistently. They're timing out randomly.

I wonder if it could be a similar scenario to enabled fileParallelism. The unit tests were passing perfectly fine in the devChecks actions, but bunch of them would randomly timeout in pr_check.

I have absolutely not idea what I'm talking about, but could it be affected by resources that are available to pr_check?

PR #2322, opened 2024-08-05
Bluperint: Allow selecting targets from diff modal

ezr-ondrej:
image


src/Components/Blueprints/BuildImagesButton.tsx
```diff
@@ -91,7 +91,6 @@ export const BuildImagesButton = () => {
           variant="primary"
           data-testid="blueprint-build-image-menu"
           ref={toggleRef}
-          isFullWidth
```
ezr-ondrej, line 94: This took me forever to find and it's actually useless as far as I can tell :face_exhaling:
The new Dropdown implementation is VERY confusing to debug xD

regexowl (approved):
Thank you!

PR #2325, opened 2024-08-06
test: Move util function to wizardTestUtils

regexowl:
This moves `selectCustomRepo` to `wizardTestUtils.tsx`

Since the function was originally imported into `Packages.test.tsx` from `CreateImageWizard.content.test.tsx` the entire content suite was re-run under Packages suite each time. This solves the issue.


mgold1234 (approved):
/lgtm

regexowl:
This is incorrect run when running Packages test suite separately, including both tests from the content test suite (highlighted) and the Packages test suite:

image

and this is correct output:

image

PR #2327, opened 2024-08-07
Wizard: Add analytics tracking for package recommendations

regexowl:
This adds a code snippet to track package recommendations via Amplitude.


mgold1234 (dismissed):
/lgtm

src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
```diff
@@ -213,7 +215,12 @@ const PackageRecommendations = () => {
                           <Button
                             variant="link"
                             component="a"
-                            onClick={() => addRecommendedPackage(pkg)}
+                            onClick={() => {
+                              analytics.track('recommended-package-added', {
+                                pkgName: pkg,
```
chenrocky, line 220:
```suggestion
                                pkgName: pkg,
                                selectedPackages: packages.map(pkg => pkg.name),
                                shownRecommendations: data.packages
```
@regexowl not sure if what I suggested works as expected, but I'd like to also track the packages that the user has already selected prior to adding a recommended package as well as the recommendations that are displayed to them prior to adding the recommended package.

regexowl: Works great, thank you! The PR is updated, the full list of previously selected packages and the list of recommendations should be tracked as well now :)

mgold1234:
can you squash the second commit please?

mgold1234:
and then we merge

mgold1234:
is it possible somehow to add test for that analystics tracking?

regexowl:
I don't think we can add tests on our side, we even mock a flat value in the test suite. The only way to confirm this works and works correctly is probably via Amplitude only. :thinking:

mgold1234 (approved):
/lgtm

PR #2328, opened 2024-08-07
Wizard: Fix information rendered in package recommendations

regexowl:
When a package got selected and deselected again, landing on 0 for selected packages, the information shown in the package recommendations was "Select packages to generate recommendations.No recommendations found for the set of selected packages", this updates the logic to separate the messages.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
@@ -164,7 +164,7 @@ const PackageRecommendations = () => {
                 again by changing your selected packages.
               </Alert>
             )}
-            {isSuccess && !data?.packages?.length && (
+            {isSuccess && !data?.packages?.length && packages.length > 0 && (
               <>No recommendations found for the set of selected packages</>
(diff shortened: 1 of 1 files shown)


ezr-ondrej (commented):
Shouldn't we disable the query when `packages.length == 0` by [skip](https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching) and return early?
Sounds like extra query for no reason :)

regexowl:
> Shouldn't we disable the query when `packages.length == 0` by [skip](https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching) and return early? Sounds like extra query for no reason :)

The package recommendations query should be triggered only when packages.length > 0, the condition is included in this `useEffect`
https://github.com/osbuild/image-builder-frontend/blob/c39bde963aff4f459f5d250cc3bdb635cc53929f/src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx#L67-L78

ezr-ondrej:
Gotcha! I've missed that :+1:

ezr-ondrej (approved):

PR #2329, opened 2024-08-07
test: Move mocks to setup file

regexowl:
This moves all the mocks used throughout our test files into the test setup file.

### DIFF
--- a/src/Components/Blueprints/ImportBlueprintModal.test.tsx
+++ b/src/Components/Blueprints/ImportBlueprintModal.test.tsx
@@ -3,26 +3,6 @@ import userEvent from '@testing-library/user-event';
 
 import { renderWithReduxRouter } from '../../test/testUtils';
 
-vi.mock('@redhat-cloud-services/frontend-components/useChrome', () => ({
-  useChrome: () => ({
-    isBeta: () => true,
-    isProd: () => true,
-    getEnvironment: () => 'stage',
-  }),
-}));
-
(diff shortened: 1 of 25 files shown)


ezr-ondrej (commented):
Love it, I hate setups in all the files, but one question:

What do I do if I need my test to disable flag?
I'm generally behind the idea of enabling all by default, maybe even defaulting to true for all. But there might be cases I want flag disabled :thinking:

regexowl:
> Love it, I hate setups in all the files, but one question:
> 
> What do I do if I need my test to disable flag? I'm generally behind the idea of enabling all by default, maybe even defaulting to true for all. But there might be cases I want flag disabled 🤔

Great question! AFAIK you can unmock in the context of any test and re-mock with values you'd like to use. Took one of the smaller files - for example in `test/edge/EdgeImagesTable.test.tsx` you could do something like:
```typescript
    vi.unmock('@unleash/proxy-client-react');
    vi.mock('@unleash/proxy-client-react', () => ({
      useUnleashContext: () => vi.fn(),
      useFlag: vi.fn((flag) =>
        flag === 'image-builder.firstboot.enabled' ? true : false
      ),
    }));
```

and the test should start failing as it's switching the boolean values.

mgold1234:
@regexowl it is looking very clean, amzaing!

mgold1234 (approved):
/lgtm

ezr-ondrej (approved):
I'm good, now for the hard part - convincing CI xD

PR #2332, opened 2024-08-07
vitest: Disable `fileParallelism`

regexowl:
This re-disables `fileParallelism` again as it's still proving too flaky.


avitova:
Thanks!

ezr-ondrej (approved):

PR #2337, opened 2024-08-09
test: Move clicking helper functions to `wizardTestUtils`

regexowl:
As clicking helper functions are used only in the context of Wizard, this moves them to an appropriate place.


ezr-ondrej (approved):
:tada: test cleanup! :tada:

PR #2338, opened 2024-08-09
testUtils: Merge `renderCustomRoutesWithReduxRouter` and `renderWithReduxRouter`

regexowl:
`renderCustomRoutesWithReduxRouter` and `renderWithReduxRouter`are functionally the same apart from routes they use to render components.

We could just use `renderCustomRoutesWithReduxRouter` and use the routes from `renderWithReduxRouter` as a default value for an optional parameter `routes`.


regexowl:
Ok, this is just an idea so feel free to throw tomatoes my way if it doesn't make sense, but it slightly bothers me that we have two separate functions with the same functionality apart from an optional parameter. :sweat_smile:

mgold1234 (approved):
/lgtm

PR #2341, opened 2024-08-12
Wizard: Add eslint-disable to useEffect

regexowl:
This disables `react-hooks/exhaustive-deps` rule for dependency arrays that should run only once on mount and therefore need to stay empty.


mgold1234 (approved):
/lgtm

mgold1234:
amazing!

PR #2342, opened 2024-08-12
Wizard: add partitioning modes into file system config

ondrejbudai:
The image builder API supports partitioning modes for a while. Currently, the frontend doesn't specify one, which makes the backend default to "auto-lvm". This partitions the image as lvm if the request contains an extra mountpoint that isn't specified in the base partition table. Otherwise, it's raw-partitioned. In other words, it's confusing.

This commit gives users an option to specify the mode explicitly: either to lvm, or raw. This should reduce confusion and give users more power.

The partitioning mode is now enforced by the frontend. This means that old blueprints are "converted" when edited: A simple heurestic is in charge of that: if there's a custom mountpoint in the original blueprint, the mode is set to lvm. Otherwise, it defaults to raw.

TODO:

- [ ] Discuss this with the UX folks.
- [ ] Add some tests (I will need help with this part :sweat_smile: ).

There's also one extra unfortunate thing: There's already a property called `FileSystemPartitionMode` in the code used to control whether the partitioning is manual, or automatic. Unfortunately, the API property for lvm/raw is called a partitioning mode, so this creates a lot of confusion. I guess we should rename the original "partition mode" to prevent this clash. Suggestions are welcome!

image

### DIFF
--- a/src/Components/CreateImageWizard/steps/FileSystem/FileSystemConfiguration.tsx
+++ b/src/Components/CreateImageWizard/steps/FileSystem/FileSystemConfiguration.tsx
@@ -3,6 +3,8 @@ import React, { useState } from 'react';
 import {
   Alert,
   Button,
+  FormGroup,
+  Radio,
   Text,
   TextContent,
   TextInput,
@@ -31,6 +33,8 @@ import {
   removePartition,
   selectPartitions,
   changePartitionUnit,
+  selectFileSystemPartitioningMode,
+  changeFileSystemPartitioningMode,
 } from '../../../../store/wizardSlice';
 import UsrSubDirectoriesDisabled from '../../UsrSubDirectoriesDisabled';
 import { useFilesystemValidation } from '../../utilities/useValidation';
@@ -48,6 +52,7 @@ export type Partition = {
 const FileSystemConfiguration = () => {
   const partitions = useAppSelector(selectPartitions);
   const environments = useAppSelector(selectImageTypes);
+  const partitioningMode = useAppSelector(selectFileSystemPartitioningMode);
 
   const dispatch = useAppDispatch();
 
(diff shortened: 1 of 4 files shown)


regexowl:
> There's also one extra unfortunate thing: There's already a property called `FileSystemPartitionMode` in the code used to control whether the partitioning is manual, or automatic. Unfortunately, the API property for lvm/raw is called a partitioning mode, so this creates a lot of confusion. I guess we should rename the original "partition mode" to prevent this clash. Suggestions are welcome!

We can do that for sure! The hardest part will be coming up with a new name, but something like `FileSystemConfigurationMode` might work? Or just `fscMode`, it's a value only for us so the key will be finding something that won't get mixed up with the partitioning mode :thinking:

regexowl:
Getting back to this a bit later, sorry about that :sweat: How can we move this forward?

A rebase to resolve the conflicts will be needed, previously used `FileSystemPartitionMode` was renamed to `FileSystemConfigurationType` which should make the naming less confusing. Some linting errors will need to be resolved as well.

I agree some tests would be nice, I can put something together for sure and try to add them in a new commit here. Or walk you through the tests if you'd like. Whatever works best.

src/Components/CreateImageWizard/steps/FileSystem/FileSystemConfiguration.tsx
```diff
@@ -74,9 +79,34 @@ const FileSystemConfiguration = () => {
...
-          device type.
+          sizes.
         </Text>
+        <FormGroup>
+          <Radio
+            id="raw partitioning radio"
+            ouiaId="raw-partitioning-radio"
+            label="Use raw partitioning"
+            name="fsc-radio-raw"
+            description="Filesystems will be created directly on top of physical partitions"
```
regexowl, line 90: I believe this was feedback on the UX call.

```suggestion
            description="File systems will be created directly on top of physical partitions"
```

regexowl (changes requested):
@ondrejbudai This looks great! Thank you! Could you please rebase and resolve lint errors? `npm run lint:js:fix` should be able to do all the work. I can then take over and either try to add some tests in this PR or better open a follow up.

Should we maybe gate this behind a feature flag for some time? To make this available in beta only? What do you think @lucasgarfield

ondrejbudai:
Closing this, we decided to re-implement the partitioning on the backend, so let's wait for it, and then implement it in the frontend.

PR #2348, opened 2024-08-13
Wizard: Optional steps (HMS-4589)

regexowl:
This adds an "Optional steps" section to the Wizard and moves all the non-mandatory customisation options there. "Review and finish" button was also added to allow jumping through the optional section.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -71,14 +71,17 @@ type CustomWizardFooterPropType = {
   disableBack?: boolean;
   disableNext: boolean;
   beforeNext?: () => boolean;
+  optional?: boolean;
 };
 
 export const CustomWizardFooter = ({
   disableBack: disableBack,
   disableNext: disableNext,
   beforeNext,
+  optional: optional,
 }: CustomWizardFooterPropType) => {
-  const { goToNextStep, goToPrevStep, close } = useWizardContext();
+  const { goToNextStep, goToPrevStep, goToStepById, close } =
+    useWizardContext();
   return (
     <WizardFooterWrapper>
       <Button
@@ -99,6 +102,18 @@ export const CustomWizardFooter = ({
       >
         Back
       </Button>
+      {optional && (
+        <Button
+          ouiaId="wizard-review-and-finish-btn"
+          variant="tertiary"
+          onClick={() => {
+            if (!beforeNext || beforeNext()) goToStepById('step-details');
+          }}
+          isDisabled={disableNext}
+        >
+          Review and finish
+        </Button>
+      )}
       <Button ouiaId="wizard-cancel-btn" variant="link" onClick={close}>
         Cancel
       </Button>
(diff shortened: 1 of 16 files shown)


mgold1234 (approved):
/lgtm it looks very good, when it will be ready :)

regexowl:
Ok, this is now ready for review. The tests are currently just blindly fixed, clean up is in progress, but I didn't want to include it in this PR.

@jrusz This could probably confuse pr_check as we're changing the order of steps in the Wizard. There is now new "Optional" section which can be completely skipped by clicking on "Review and finish" button on the first optional step (Registration) and would land the user on the Details step. Jumping between steps using the nav and not the Next/Back buttons should still work. No nav items names were changed.

Can you please take a look at what changes will be needed when you have time? It's probably not super urgent, but it would be nice to merge this some time next week or so. It would probably go into stage first and to prod after one week of soaking, not sure if we'd want to do early release because of this.

Here are [mocks](https://www.figma.com/proto/9t3owYlJi4M23fUXWpgXCE/HMS-Image-Builder?page-id=367%3A8927&type=design&node-id=370-12856&viewport=45703%2C50354%2C0.49&t=2SH3uw422cJAOXhU-1&scaling=min-zoom&starting-point-node-id=370%3A12856) and this is how the Wizard left side nav looks like now:

image

jrusz:
> Ok, this is now ready for review. The tests are currently just blindly fixed, clean up is in progress, but I didn't want to include it in this PR.
> 
> @jrusz This could probably confuse pr_check as we're changing the order of steps in the Wizard. There is now new "Optional" section which can be completely skipped by clicking on "Review and finish" button on the first optional step (Registration) and would land the user on the Details step. Jumping between steps using the nav and not the Next/Back buttons should still work. No nav items names were changed.
> 
> Can you please take a look at what changes will be needed when you have time? It's probably not super urgent, but it would be nice to merge this some time next week or so. It would probably go into stage first and to prod after one week of soaking, not sure if we'd want to do early release because of this.
> 
> Here are [mocks](https://www.figma.com/proto/9t3owYlJi4M23fUXWpgXCE/HMS-Image-Builder?page-id=367%3A8927&type=design&node-id=370-12856&viewport=45703%2C50354%2C0.49&t=2SH3uw422cJAOXhU-1&scaling=min-zoom&starting-point-node-id=370%3A12856) and this is how the Wizard left side nav looks like now:
> 
> image

Oh yeah this is going to be mighty confusing for the pr_check :sweat_smile: however I could just click on the "Review and Finish" step to skip all this for now like you suggest and add proper testing sometime later. I'll try to prepare it by the end of the week :)

regexowl:
@jrusz Good news, pr_check won't need an update. Sorry about the confusion :disappointed:

regexowl:
I've removed the "Content" subcategory from the Optional category and spread the steps as Wizard doesn't deal well with nesting substeps within a substep.

Also we've talked about an option to generate Wizard programatically from a json schema, I've looked into it a bit more and here are some reasons why I'm not in favor of that:
- there is no existing json -> ts generator we could utilise, meaning we would need to write a custom one from scratch expanding possibe bug surface
- it would add another layer of abstraction which might be harder to debug, there was a similar situation with V1Wizard and DDF as DDF generated Wizard using an in-built schema which wasn't very pleasant to work with
- conditionals in json would be a problem
- also sometimes we need to do stuff specific for just one step which we would need to allow in the generator, loosening potential safeguards

What do you think? Does that make sense?

lucasgarfield:
> I've removed the "Content" subcategory from the Optional category and spread the steps as Wizard doesn't deal well with nesting substeps within a substep.
> 
> Also we've talked about an option to generate Wizard programatically from a json schema, I've looked into it a bit more and here are some reasons why I'm not in favor of that:
> 
>     * there is no existing json -> ts generator we could utilise, meaning we would need to write a custom one from scratch expanding possibe bug surface
> 
>     * it would add another layer of abstraction which might be harder to debug, there was a similar situation with V1Wizard and DDF as DDF generated Wizard using an in-built schema which wasn't very pleasant to work with
> 
>     * conditionals in json would be a problem
> 
>     * also sometimes we need to do stuff specific for just one step which we would need to allow in the generator, loosening potential safeguards
> 
> 
> What do you think? Does that make sense?

Makes good sense to me! 

At some point we will need to create some sort of abstraction so we can have differentiated Wizard experiences - not only for Image Builder on-prem which will replace Cockpit Composer but also for Image Mode for RHEL. But that doesn't mean we need to use json, we can find a good solution using React components I'm sure.

lucasgarfield:
This looks really good! 

I noticed a little :bug:  with the validation, though. To reproduce, create an invalid file system (e.g., two `/home` mount points). You can still skip to the review step – the “Create” button is disabled, but no information as to why is given.

I think it would also be good to add some test coverage for this. Off the top of my head, it would be good to:
1. Verify the “Skip to review” button takes you to the correct step
2. Verify the “Skip to review” button is disabled if the wizard state is invalid (possible with both FSC and details).

lucasgarfield:
This is out of scope as this PR matches the mocks perfectly, but I do wonder if we want to consider two things, possibly in this PR or in a future improvement:

1. Skipping directly to the Review step since we now generate a default blueprint name... right now, it doesn't _actually_ skip to the review step, it skips to the details step...
2. Putting the "Skip to review" button on the last non-optional step rather than the first optional step - i.e. if no cloud images are selected, the "Image output" step, if AWS and Azure are selected then Azure, etc...

lucasgarfield:
One other little :bug: I noticed - in some cases there may be unfetched data (for instance, the list of activation keys) and it is possible to click "Skip to review" while the request is in flight. To reproduce, go to the "Register" step and then very quickly click "Skip to review". The "Activation key" section on the review step will be empty, and clicking "Create blueprint" doesn't do anything (an error is logged in the console about the activation key being undefined).

lucasgarfield:
Oh, I now realize the tests are in #2387... could we rebase that onto this one instead of having separate PRs? I suspect it will be easier to stay disciplined about writing tests if we keep the tests in the same PR as the features... that way we can't say "oh, I'll write tests _later_"... not that I would ever do that. :wink:

regexowl:
> This is out of scope as this PR matches the mocks perfectly, but I do wonder if we want to consider two things, possibly in this PR or in a future improvement:
> 
>     1. Skipping directly to the Review step since we now generate a default blueprint name... right now, it doesn't _actually_ skip to the review step, it skips to the details step...
> 
>     2. Putting the "Skip to review" button on the last non-optional step rather than the first optional step - i.e. if no cloud images are selected, the "Image output" step, if AWS and Azure are selected then Azure, etc...

Both make sense! I'd put them on a list of future improvements and address both in a follow up since the implementation will be a bit more complex. Currently we populate activation key and blueprint name on init of their corresponding step, meaning at least that will require a re-write.

regexowl:
> Oh, I now realize the tests are in #2387... could we rebase that onto this one instead of having separate PRs? I suspect it will be easier to stay disciplined about writing tests if we keep the tests in the same PR as the features... that way we can't say "oh, I'll write tests _later_"... not that I would ever do that. 😉

For sure! Done! Originally I've opened a new PR since I've noticed we've overlooked the missing tests in the past independently on the optional section feature, but since those are connected after all it makes perfect sense to have them in one PR.

A heads up - the tests don't pass for Node 18 (the async weirdness I've mentioned at a standup), I've opened a PR to migrate all our tests to Node 20 only: https://github.com/osbuild/image-builder-frontend/pull/2404

regexowl:
> One other little 🐛 I noticed - in some cases there may be unfetched data (for instance, the list of activation keys) and it is possible to click "Skip to review" while the request is in flight. To reproduce, go to the "Register" step and then very quickly click "Skip to review". The "Activation key" section on the review step will be empty, and clicking "Create blueprint" doesn't do anything (an error is logged in the console about the activation key being undefined).

The "Review and finish" button will mirror Next button behaviour for now. This deserves another look and possibly a rework when we decide to skip from the Image Output step or directly to Review step - completely skipping optional section as mentioned in your suggestions in https://github.com/osbuild/image-builder-frontend/pull/2348#issuecomment-2322311784

regexowl:
> This looks really good!
> 
> I noticed a little 🐛 with the validation, though. To reproduce, create an invalid file system (e.g., two `/home` mount points). You can still skip to the review step – the “Create” button is disabled, but no information as to why is given.
> 
> I think it would also be good to add some test coverage for this. Off the top of my head, it would be good to:
> 
>     1. Verify the “Skip to review” button takes you to the correct step
> 
>     2. Verify the “Skip to review” button is disabled if the wizard state is invalid (possible with both FSC and details).

Solved and added some tests so hopefully everything addressed should be addressed now :sweat_smile:

lucasgarfield (approved):
All bugs are fixed and test coverage looks good! Fantastic work as always. Thank you!

PR #2350, opened 2024-08-13
Wizard: Confirm Image builder is using origin for contentList call

Andrewgdewar:
- This ticket just adds a new enum and ensures that Origin is being set in the useListRepositoriesQuery.
Related ticket in HMS [here](https://issues.redhat.com/browse/HMS-4426)

Other changes: 
Added one comment to remind myself to address a performance issue at some point. 
Will address in another ticket at a later time.

- [x] Confirm that we only want External and not Custom for the majority of image-builders usage with @jlsherrill

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -44,6 +44,7 @@ import PackageInfoNotAvailablePopover from './components/PackageInfoNotAvailable
 
 import {
   CONTENT_URL,
+  ContentOrigin,
   EPEL_8_REPO_DEFINITION,
   EPEL_9_REPO_DEFINITION,
   RH_ICON_SIZE,
@@ -135,6 +136,7 @@ const Packages = () => {
   const { data: epelRepo, isSuccess: isSuccessEpelRepo } =
     useListRepositoriesQuery({
       url: epelRepoUrlByDistribution,
+      origin: ContentOrigin.EXTERNAL,
     });
 
   const [isRepoModalOpen, setIsRepoModalOpen] = useState(false);
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStepTables.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStepTables.tsx
@@ -11,6 +11,7 @@ import {
 } from '@patternfly/react-core';
 import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
 
+import { ContentOrigin } from '../../../../constants';
 import {
   ApiSnapshotForDate,
   useListRepositoriesQuery,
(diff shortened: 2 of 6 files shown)


src/Components/CreateImageWizard/steps/Packages/Packages.tsx
```diff
@@ -135,6 +136,7 @@ const Packages = () => {
   const { data: epelRepo, isSuccess: isSuccessEpelRepo } =
     useListRepositoriesQuery({
       url: epelRepoUrlByDistribution,
+      origin: ContentOrigin.CUSTOM,
```
regexowl, line 139: Sorry, got a bit lost in the discussion so just to check - should this be external only or is this a different case? :sweat_smile:

regexowl (commented):
Looks good, thank you! Added one question.

Could you also please remove the merge commit so we can just rebase?

mgold1234:
maybe stupid question, but wander from where we know that:
packageRecommendations -> ContentOrigin.REDHAT
packages -> ContentOrigin.CUSTOM
repositories -> ContentOrigin.EXTERNAL

src/Components/CreateImageWizard/steps/Review/ReviewStepTables.tsx
```diff
@@ -128,7 +129,7 @@ export const SnapshotTable = ({
 }) => {
   const { data, isSuccess, isLoading, isError } = useListRepositoriesQuery({
     uuid: snapshotForDate.map(({ repository_uuid }) => repository_uuid).join(),
-    origin: 'red_hat,external', // Make sure to show both redhat and custom
+    origin: ContentOrigin.REDHAT + ',' + ContentOrigin.EXTERNAL, // Make sure to show both redhat and custom
```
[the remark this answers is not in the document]
Andrewgdewar: ahh yes that would be less confusing to read! 
Custom used to mean only external. 😕

src/Components/CreateImageWizard/steps/Packages/Packages.tsx
Andrewgdewar: You are correct, not sure why I put custom here, with the filter it shouldn't matter but I'll update regardless.

Andrewgdewar:
> maybe stupid question, but wander from where we know that: packageRecommendations -> ContentOrigin.REDHAT packages -> ContentOrigin.CUSTOM repositories -> ContentOrigin.EXTERNAL

For packages -> ContentOrigin.CUSTOM: 
Right now nothing should be .CUSTOM (this would include the new Upload repositories, my mistake adding to that confusion, upload is for a later usage).  I've updated that to EXTERNAL, but all we are really doing with the "useListRepositoriesQuery" call in Packages.tsx is grabbing a single associated repository (epel 8 or 9) and adding it in certain cases.  Epel is an external package so that's why we are using "external" there. 

External are what we currently know of today as repositories "external" to redhat. (custom repositories or repositories are those that have been added by the user mostly, and are external to redhat's maintained repositories).

When we make recommendations, we generally only do so for repositories Redhat owns. 

Hope this helps. 

The crossover between "custom" and the naming on the enum isn't great, but that will match what we will soon have in the repositories view.

regexowl (approved):
Excellent, thank you! :sparkles:

PR #2362, opened 2024-08-16
Wizard: Unique default key name

### LINKED ISSUE #2353: Wizard: Unique default key name
Default keys should have less predictable names: https://github.com/osbuild/image-builder-frontend/pull/2308#discussion_r1715769255

regexowl:
Fixes #2353

This adds a random suffix to the default activation key name, making it less predictable.


ezr-ondrej (approved):
Oh I'm sorry to push for it, I assumed there is a library we can use, but very elegant solution! :+1: 
Nice! :orange_heart:

regexowl:
Ah, right, there might be a library able to do this :thinking: well, this works as well I guess and we have control over char set :sweat_smile:

PR #2183, opened 2024-06-25
config: Replace webpack configs with FEC config (HMS-4448)

regexowl:
This migrates from previously used webpack configs to one fec.config.js.

This will allow us to bump `frontend-components-config` dependency to version 6.x and stay supported in the future.

PR #2363, opened 2024-08-16
Local dev improvements

schuellerf:
I did an upstream patch so `--clouddotEnv=stage` acutally works now for non-interactive local startup
(with `@redhat-cloud-services/frontend-components-config` v6.2.6 or higher)

and I think [`JSON.stringify()`](https://github.com/osbuild/image-builder-frontend/pull/2363/files#diff-d0f06d7cd724a6add60ee1bac8183344dc66faae9dac3ee6ee920f5b7ac3a88eR18) is needed to get `process.env.MSW` to the client
but still MSW doesn't really work but maybe it's part of the solution?

I can also split them up into separate PRs if needed


package.json
```diff
@@ -85,7 +85,11 @@
     "lint:js": "eslint src",
     "lint:js:fix": "eslint src --fix",
     "start": "fec dev",
-    "start:msw": "MSW=TRUE fec dev",
+    "start:stage": "fec dev --clouddotEnv=stage",
+    "start:prod": "fec dev --clouddotEnv=prod",
+    "start:msw": "NODE_ENV=development MSW=TRUE fec dev",
+    "start:msw:stage": "NODE_ENV=development MSW=TRUE fec dev --clouddotEnv=stage",
+    "start:msw:prod": "NODE_ENV=development MSW=TRUE fec dev --clouddotEnv=prod",
```
regexowl, line 92: Just a note - previously the MSW in browser was allowed for stage only. I think it was based on webpack configs being split into two separate files for prod and stage so it might work also for prod with the new config

https://github.com/osbuild/image-builder-frontend/pull/2183/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L99

regexowl (approved):
Thank you for looking into it and patching up frontend-components-config! Even with MSW in browser not working yet this feels like a step in the right direction! :sparkles:

package.json
schuellerf: I can easily remove if it doesn't make sense! Just thumb up here and I'll adapt the PR.

regexowl: Okay, let's remove `start:msw` and `start:msw:prod` for now, we can always expand the options later :) Thank you!

regexowl (approved):
:sparkles::rocket:

PR #2365, opened 2024-08-19
Blueprints: change diff modal button label

ezr-ondrej:
For the diff modal, we want to change the button label.
In this context Synchronize images is more appropriate label.

Why: Kelsea asked me to! :)

image


regexowl (approved):
Nice, thank you! :sparkles:

PR #2370, opened 2024-08-20
test: Cleanup Packages tests (HMS-4658)

regexowl:
This cleans up Packages test file after relevant tests were moved there from `CreateImageWizard.content.test.tsx`.

In these cleanups nothing gets deleted, the parts only get moved around to make the flow of the test file less chaotic and where possible blocks of code get replaced by helper functions.


mgold1234 (approved):
/lgrm

PR #2373, opened 2024-08-20
test: Re-enable a test

regexowl:
This re-enables "Repositories edit mode modal pops up when deselecting previously used repository" test.

This works reliably locally, not sure what will pr_check think.


regexowl:
Ok, nevermind, the tests is very flakey...

PR #2305, opened 2024-07-31
Wizard: Validate snapshot date with useValidation

ezr-ondrej:
This uses the concept of useValidation for snapshot date.
It is the last validation that doesn't use this.

PR #2374, opened 2024-08-20
Wizard: Debounce unique validation

ezr-ondrej:
Debounce the validation, so we don't validate every character. Debounce also ensures the right ordering.
Only one request runs at a time, we cancel the previous before we fire a new.


regexowl:
The debounce works great, but found one edge case - if the user is fast enough they can input already existing name and skip to Review before the validation error shows:

name-validation

They get the error on the next step and need to fix it before creating a blueprint so all's safe. But would it be possible to disable the Next button until validation finishes somehow maybe?

ezr-ondrej:
> They get the error on the next step and need to fix it before creating a blueprint so all's safe. But would it be possible to disable the Next button until validation finishes somehow maybe?

Sure :) here we go ;)
It just blocks the user for a bit if the request is slow, but I guess it should be fine in 90% of cases.

regexowl:
> Sure :) here we go ;)
It just blocks the user for a bit if the request is slow, but I guess it should be fine in 90% of cases and bit of lag in 10% :)

Beautiful, thank you! 

This has similar problem with pristinity as https://github.com/osbuild/image-builder-frontend/pull/2305, see https://github.com/osbuild/image-builder-frontend/pull/2305#issuecomment-2317561629. But I think we should take a look at it separately. Even though we have control over pristinity in this case I think we should keep the way we communicate validation issue consistent throughout wizard.

PR #2378, opened 2024-08-22
Wizard: Update Amplitude tracking

regexowl:
This updates snippets used for event tracking in Amplitude, adding `imageBuilder` to the name of the event and as a `module` property.

Two more typed events were also added to track number of created blueprints. Name for both events is the same `blueprintCreated`, the differentiation between Create and Create and build images is captured in the `type` property.

These new events will allow to calculate percentage of users that were shown recommendations and of those who added a recommended package while succesfully finishing blueprint creation.


mgold1234 (approved):
/lgtm

PR #2379, opened 2024-08-26
createImageWizard: change the default name of bllueprint

mgold1234:
this commit change the default name to different format 

`<release>-<arch>-<date>-<time>`

### DIFF
--- a/src/Components/CreateImageWizard/utilities/generateDefaultName.ts
+++ b/src/Components/CreateImageWizard/utilities/generateDefaultName.ts
@@ -1,15 +1,17 @@
-import { ImageTypes } from '../../../store/imageBuilderApi';
+import { Distributions, ImageRequest } from '../../../store/imageBuilderApi';
 
 export const generateDefaultName = (
-  distribution: string,
-  arch: string,
-  targetEnvironments: ImageTypes[]
+  distribution: Distributions,
+  arch: ImageRequest['architecture']
 ) => {
   const date = new Date();
   const day = date.getDate().toString().padStart(2, '0');
-  const month = date.toLocaleString('en-US', { month: 'long' });
-  const year = date.getFullYear();
-  return `${distribution} ${arch} ${targetEnvironments.join(
-    ' '
-  )} ${day} ${month} ${year}`;
+  const month = (date.getMonth() + 1).toString().padStart(2, '0');
+  const year = date.getFullYear().toString();
+  const hours = date.getHours().toString().padStart(2, '0');
+  const minutes = date.getMinutes().toString().padStart(2, '0');
+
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
I'm ok with it, but Klara asked for full year, WDYT?

src/Components/CreateImageWizard/utilities/generateDefaultName.ts
```diff
@@ -1,15 +1,12 @@
...
+export const generateDefaultName = (distribution: string, arch: string) => {
   const date = new Date();
   const day = date.getDate().toString().padStart(2, '0');
-  const month = date.toLocaleString('en-US', { month: 'long' });
-  const year = date.getFullYear();
-  return `${distribution} ${arch} ${targetEnvironments.join(
-    ' '
-  )} ${day} ${month} ${year}`;
+  const month = (date.getMonth() + 1).toString().padStart(2, '0');
+  const year = date.getFullYear().toString().slice(-2);
```
regexowl, line 5: Ah yeah, sorry, jumping all over the place this morning so getting to this a bit later. This would be the only change and I believe it would improve readability of the date, making it a bit more obvious.

```suggestion
  const year = date.getFullYear().toString();
```

Otherwise looks and works great!

src/Components/CreateImageWizard/utilities/generateDefaultName.ts
```diff
@@ -1,15 +1,12 @@
-import { ImageTypes } from '../../../store/imageBuilderApi';
-
-export const generateDefaultName = (
-  distribution: string,
-  arch: string,
-  targetEnvironments: ImageTypes[]
-) => {
+export const generateDefaultName = (distribution: string, arch: string) => {
```
regexowl, line 1: Ah, one more thing. We have proper types for those, let's use them:

```suggestion
export const generateDefaultName = (
  distribution: Distributions,
  arch: ImageRequest['architecture']
) => {
```

regexowl (changes requested):

src/Components/CreateImageWizard/utilities/generateDefaultName.ts
mgold1234: I think it is good idea to add the year.
I added it like that `${month}/${day}/${year}-${hours}:${minutes}` to be more clear

ezr-ondrej: I'd not do that, that will confuse every European out there, that's why it's not common to use that format.
YYYYMMDD-HHMM is most widely understood without spaces IMO

regexowl:
Can you please resolve the linting errors? After those are gone I'd say we're good to go! :rocket:

regexowl (approved):
Looks great! Thank you! :heart:

PR #2387, opened 2024-08-28
test: Add revisit step button checks

regexowl:
This adds tests to check that the revisit buttons on the Review step are working correctly.

Also some minor cleanup, namely making case used in the names of describe and test more consistent and moving render functions outside of functions selecting targets to make the structure easier to go through.


regexowl:
I nearly unknowingly broke Revisit buttons on the Review step while working on https://github.com/osbuild/image-builder-frontend/pull/2348 so figured adding a simple test to each step won't hurt.

regexowl:
The locally non-reproducible Node 18 only async flakes will be the end of me :roll_eyes:

regexowl:
Cherry-picked into https://github.com/osbuild/image-builder-frontend/pull/2348

PR #2392, opened 2024-08-29
Wizard: Add beforeNext conditions to Details step footer

regexowl:
This adds beforeNext condition to the custom Details step footer, allowing to validate after clicking Next. That way the button is enabled, but gets disabled in the case validation errors and the helper text with the error renders immediately.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -203,6 +203,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   // Firstboot
   const firstBootValidation = useFirstBootValidation();
   // Details
+  const [detailsPristine, setDetailsPristine] = useState(true);
   const detailsValidation = useDetailsValidation();
 
   let startIndex = 1; // default index
@@ -453,7 +454,14 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
             status={detailsValidation.disabledNext ? 'error' : 'default'}
             footer={
               <CustomWizardFooter
-                disableNext={detailsValidation.disabledNext}
+                beforeNext={() => {
+                  if (detailsValidation.disabledNext) {
+                    setDetailsPristine(false);
+                    return false;
+                  }
+                  return true;
+                }}
+                disableNext={!detailsPristine && detailsValidation.disabledNext}
               />
             }
           >
--- a/src/test/Components/CreateImageWizard/steps/Details/Details.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Details/Details.test.tsx
@@ -63,6 +63,7 @@ describe('Step Details', () => {
     const nextButton = await getNextButton();
     expect(nextButton).toBeEnabled();
     await enterBlueprintName(' ');
+    await clickNext();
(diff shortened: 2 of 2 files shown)


regexowl:
@ezr-ondrej This is what I had in mind in https://github.com/osbuild/image-builder-frontend/pull/2305#issuecomment-2317561629, what do you think? Would this make sense?

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -187,6 +187,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   // Firstboot
   const firstBootValidation = useFirstBootValidation();
   // Details
+  const [detailsPristine, setDetailsPristine] = useState(true);
```
mgold1234, line 190: maybe change the name to 'detailsValidation'?

mgold1234:
looks good, just small suggestion

src/Components/CreateImageWizard/CreateImageWizard.tsx
regexowl: This is following the naming we have going on for `filesystemPristine` a few lines above. Validation is handled via the `useValidation` hook, this variable holds information about the pristinity of the step only.

regexowl:
To continue the discussion we had with @ezr-ondrej https://github.com/osbuild/image-builder-frontend/pull/2305#issuecomment-2318051064 - does this make sense or not for our use case?

@lucasgarfield @mgold1234 With the way things are set up in PatternFly validation we have two options of showing the user validation errors. This will be specifically for the Details step's Blueprint name input, but there are several places where similar problem occurs (snapshot validation comes to mind):

1. when a non-valid name is entered, we currently disable Next, but an error doesn't get rendered unless the user clicks outside of the input (`onBlur`). That means user can have the name filled in, can't continue because of disabled button, but doesn't see why.
2. similarly to FSC step we could let the Next button be enabled, but disable it and render an error when the Next button with non-valid name filled in is clicked, which would be this PR.

A fair point is that the always enabled Next even with incorrect values might feel like a bait. On the other hand I feel it's a bit more transparent, but to be honest I'm very much undecided on this. What do you think?

ezr-ondrej (approved):
Sorry for the delay here, I think this is a sane aproach :+1:
Slight disadvantage is that we are still incorrect in some cases (you can't go next, but the button is blue). But it's better then the alternative today - keeping the button disabled even if you could continue and there is just lag.

regexowl:
Aah right, this will break the IQE tests I guess. :thinking: 

The original behaviour of Details step when entering an invalid blueprint name was that the Next button got disabled, but no error was rendered.

The new behaviour should be that the Next button is enabled, but when clicking on it the validation kicks in, disables the button and the error gets immediately rendered (when the form is pristine only).

@jrusz @tkoscieln would you have time to take a look at this please? It's not urgent, we'll also probably need to support both scenarios while the changes are soaking in the stage.

jrusz:
I see a bug with this. When I type an already existing blueprint name, the error is shown but the "Next" button is still enabled. It only gets disabled once clicked on. The test sees that it's enabled, clicks on it and then fails because it expects the Review step to appear. 
image

regexowl:
This is a bit stale. Might revisit how we're doing validation and informing users about errors later, but closing for now.

PR #2393, opened 2024-08-29
codecov: enable codecov in our repo

mgold1234:
this commit add the configuration for codecov,
to be able to see the coverage of vitest in codecov website.

### DIFF
--- a/.github/workflows/dev-checks.yml
+++ b/.github/workflows/dev-checks.yml
@@ -28,4 +28,10 @@ jobs:
     - name: Run lint check
       run: npm run lint
     - name: Run unit tests
-      run: npm run test
+      run: npm run test:coverage
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v4
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        files: ./coverage/junit.xml
+        verbose: true
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -17,6 +17,10 @@ const config = {
     testTimeout: 10000,
     fileParallelism: false,
   },
+  reporters: ['default', 'junit'],
+  outputFile: {
+    junit: './coverage/junit.xml',
+  },
   resolve: {
     mainFields: ['module'],
   },


schuellerf (approved):
/LGTM

package.json
```diff
@@ -90,7 +90,7 @@
     "start:msw:stage": "NODE_ENV=development MSW=TRUE fec dev --clouddotEnv=stage",
     "test": "TZ=UTC vitest run",
     "test:watch": "TZ=UTC vitest",
-    "test:coverage": "TZ=UTC vitest run --coverage",
+    "test:coverage":  "TZ=UTC vitest run --coverage --reporter=junit --outputFile=./coverage/junit.xml",
```
regexowl, line 93: Looks like there is one extra space, but otherwise looks good to me!

```suggestion
    "test:coverage": "TZ=UTC vitest run --coverage --reporter=junit --outputFile=./coverage/junit.xml",
```

.gitignore
```diff
@@ -24,3 +24,4 @@ coverage
 
 *~
 bots
+./coverage/junit.xml
```
regexowl, line 27: We already have `coverage` in .gitignore, shouldn't that be enough to remove the coverage folder from files to commit? :thinking:

mgold1234: It should be enough, I will remove it

schuellerf (approved):
lgtm

regexowl:
One more question - sorry, always getting late to it :sweat_smile: - shouldn't we add another reporter as well? What I mean is when you look at the Run unit tests section in a devCheck https://github.com/osbuild/image-builder-frontend/actions/runs/10667183406/job/29564199819?pr=2393 there are results for the coverage, but no test outputs. I've tried locally breaking a test and I can see the failures in `junit.xml`, but didn't get anything in the console output. Would this mean we'll need to got to the coverage report to see if there are any broken tests each time?

image

regexowl:
Aha! Quickly looked into it a bit and currently the junit is overriding the default reporters. What could be done is adding junit next to them and setting up an output file for it separately. We could move both settings to the vitest.config.ts to have all configuration in one place.

So what I'd propose is adding something like this to the "coverage" section of `vitest.config.ts`:
```
      reporters: ['default', 'junit'],
      outputFile: {
        junit: './coverage/junit.xml',
      },
```

and removing the reporter and the outputFile from the `package.json` script. This way we should keep both the default set of reporters and also have the junit set up with an outputFile. What do you think?

I'm basing everything on https://vitest.dev/guide/reporters#junit-reporter so it's possible I've missed something :sweat_smile:

regexowl (changes requested):
Requesting changes so it doesn't get merged by accident :sweat_smile:

regexowl (approved):
Works great! Thank you :sparkles:

PR #2398, opened 2024-08-30
Wizard: Align revisit buttons

regexowl:
This aligns Revisit step buttons on the Review step with the second column of description lists.

The formatting of the expandable's toggleContent is formatted the same way as the main content. That way if the value of min-width gets updated, the spacing will stay consistent.

The duplicate code was moved to `composeExpandable` function that takes two arguments - name of the expandable/step and a stepId that links to the correct step in the Wizard.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
@@ -5,6 +5,10 @@ import {
   ExpandableSection,
   Text,
   TextContent,
+  TextList,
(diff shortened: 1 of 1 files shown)


regexowl:
@lucasgarfield Tried to align the Revisit button with the second column, what do you think?

lucasgarfield:
> @lucasgarfield Tried to align the Revisit button with the second column, what do you think?

I love it!

lucasgarfield (approved):
Nice! This is actually a lot cleaner, too!

PR #2404, opened 2024-09-02
test: Test with Node 20 only

regexowl:
This updates the test env to test with Node 20 only, pr_check is also updated to use Node 20.

The changes shouldn't affect users in any way, please let me know if you think it might.

There were common async flakes when using testing-library with Node 18, those often surfaced in gh actions / pr_check only and did not reproduce locally unless resources got limited (I can reproduce some flakes in Power Saver mode only, leading me to conclusion, that it's just async being mischievous). Switching to testing with Node 20 only should also allow us to re-enable some tests that got disabled specifically because them running with Node 18 ended up in a time out.

### DIFF
--- a/.github/workflows/dev-checks.yml
+++ b/.github/workflows/dev-checks.yml
@@ -5,19 +5,15 @@ on:
     branches: [ "main" ]
 
 jobs:
-  tests:
+  dev-check:
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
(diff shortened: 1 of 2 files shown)


regexowl:
Btw the required checks will be a mess until they're updated in repo settings so please ignore the "tests" actions :sweat_smile:

lucasgarfield (approved):
Makes good sense to me! I can't think of any compelling reason to run our unit tests on old versions of Node.

PR #2412, opened 2024-09-04
src: Integrate image-builder-frontend as a Cockpit package with webpack support

mgold1234:
Screenshot 2024-09-04 at 18 05 58

- Added image-builder-frontend repository as a Cockpit package.     
- Configured webpack with necessary loaders and plugins for handling JS, TS, CSS, and SCSS files.
- Ensured compatibility with Cockpit by externalizing cockpit dependencies and updating the webpack configuration.
- Provided initial setup with image-builder-frontend.


public/manifest.json
```diff
@@ -0,0 +1,18 @@
+{
+  "requires": {
+    "cockpit": "134"
+  },
+  "dashboard": {
+    "index": {
+      "label": "Images",
```
croissanne, line 10: I'd make this Image Builder, it's set to replace cockpit-composer after all.

src/AppCockpit.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  Image builder for RPM-DNF documentation
+                </Button>
+              </EmptyStateActions>
+            </EmptyStateFooter>
+          </>
+        </EmptyState>
+      </Bullseye>
+      <NotificationsPortal />
+      <BrowserRouter>
+        <Routes>
```
croissanne, line 90: I think we should reuse the routes from `Router.tsx`, and just have the BrowserRouter component in here.

package.json
```diff
@@ -92,6 +92,7 @@
     "test:watch": "TZ=UTC vitest",
     "test:coverage": "TZ=UTC vitest run --coverage",
     "build": "fec build",
+    "build:dev": "webpack --config webpack.config.js",
```
croissanne, line 95: Can we rename cockpit related commands so it's obvious? Like "cockpit:build" etc.

croissanne (changes requested):
Neat! Thank you.

I wonder if the directory structure can be nicer if we put the cockpit related files into it's own directory somehow, like src/cockpit/AppCockpit.tsx, and `cockpit/public / cockpit/webpack.config.js` ?

Also would it be good to reuse a single existing component as a poc in this first version? Like anything, the landing page or something that doesn't require any api interaction.
I

public/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+  "requires": {
+    "cockpit": "134"
+  },
+  "version": 0,
+  "name": "composer",
```
KKoukiou, line 6: Decide what URL you want to use. If it's images in the path you want to change that.

public/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+  "priority": 50,
```
KKoukiou, line 7: You can switch to priority: 0 if you use a new name. Aka not composer.

KKoukiou (commented):
You will need to create a Fedora package, aka you 're missing a spec file at least here.

In that you want to obsolete the old package. https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages

src/AppCockpit.tsx
mgold1234: I am sure its douable in that stage, becuase Router.txs require any api interaction.
@lucasgarfield WDYT?

PR #2415, opened 2024-09-08
api: add latest changes in api

mgold1234:
this commit adds the latest changes in api schema


mgold1234:
need to fix some tests before merging

regexowl:
There are build errors that will need to be addressed as well after the API update.

PR #2418, opened 2024-09-10
devDeps: Manual bump msw from 2.4.1 to 2.4.4

regexowl:
Dependabot doesn't touch the `mockServiceWorker.js` file that gets automatically updated by `npm install` which causes problem with "manual changes". We'll need to bump this manually at least for the time being.


mgold1234 (approved):
/lgtm

PR #2423, opened 2024-09-11
Wizard: Add `isPreview` property to analytics tracking

regexowl:
This adds `isPreview` property to the analytics tracking with the use of chrome's `isBeta()`.


regexowl:
This needs to wait for the next release of chrome that will address a bug when API didn't propagate changes in the UI env to the scalprum instance. As soon as that's out I'll add the dependency bump to this PR.

mgold1234 (approved):
/lgtm

regexowl:
pr_check fails on teardown only, merging

PR #2424, opened 2024-09-11
Added image-builder-frontend repository as a Cockpit package (COMPOSER-2378)

mgold1234:
Align Insights IB Service and IB On-Prem UI with Unified Blueprint and Image Functionality

- Implemented identical UI components for both Insights IB service and IB on-prem.
- Standardized blueprint functionalities, allowing creation, editing, and import/export operations.
- Unified image-building capabilities based on blueprints.
- Maximized code sharing and reuse by abstracting common functionalities.
- Utilized RTK Query (RTKQ) for efficient communication with two distinct APIs: `image-builder` (service) and `cloud-api` (on-prem).


croissanne (commented):
Thank you! Just a general first comment, to make this a little easier to review, could you maybe split it out into several commits? Having a good description for each one?  The current commit message doesn't describe all the changes made.

src/Components/LandingPage/BlueprintEmpty.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import { Title, EmptyState, EmptyStateBody } from '@patternfly/react-core';
+
+const BlueprintsEmpty = () => (
```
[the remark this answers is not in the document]
mgold1234: I change this compoent to empty image table

regexowl:
Is this a duplicate of https://github.com/osbuild/image-builder-frontend/pull/2412? Can you please close a PR that is not needed anymore?

mgold1234:
@regexowl I moved https://github.com/osbuild/image-builder-frontend/pull/2412 to draft becuase I still need it, hope its ok

regexowl:
@mgold1234 Yup, perfectly fine, thank you!

src/Router.tsx
```diff
@@ -13,23 +15,51 @@ const ImportImageWizard = lazy(
...
+  flagHook: (flagName: string) => boolean
+) => {
+  // If it's the Cockpit app (on-premise), always return false
+  if (isOnPremise === 'cockpit') {
+    return false;
+  }
+
+  // Otherwise, return the actual flag value
+  return flagHook(flagName);
+};
```
croissanne, line 30: Can we move this fn out to `src/Utilities/useGetEnvironment` ? This will be useful everywhere we call out to useFlag.

src/Router.tsx
```diff
@@ -13,23 +15,51 @@ const ImportImageWizard = lazy(
   () => import('./Components/CreateImageWizard/ImportImageWizard')
 );
 const CreateImageWizard = lazy(() => import('./Components/CreateImageWizard'));
+const isOnPremise = process.env.APP_TYPE_ENV;
+const useFlagWithAppTypeCheck = (
```
croissanne, line 19: Maybe a shorter name, like `getFlag` or `getFeatureFlag`? This name is kinda unwieldy.

src/Router.tsx
```diff
@@ -13,23 +15,51 @@ const ImportImageWizard = lazy(
   () => import('./Components/CreateImageWizard/ImportImageWizard')
 );
 const CreateImageWizard = lazy(() => import('./Components/CreateImageWizard'));
+const isOnPremise = process.env.APP_TYPE_ENV;
```
croissanne, line 18: Let's wrap this in a utility function in `src/Utilities/useGetEnvironment` as well?

croissanne (dismissed):
Cool that we can reuse the router :) Mostly style comments.

src/Components/LandingPage/ImagesEmpty.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import { Title } from '@patternfly/react-core';
+import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
+
+const ImagesEmpty = () => (
```
croissanne, line 6: This isn't really adding anything, I think it would be best to just leave it out. The real challenge is reusing existing components, and as such, it's unlikely this will ever be used.

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -80,19 +82,28 @@ const ImagesTable = () => {
   const blueprintVersionFilter = useAppSelector(selectBlueprintVersionFilter);
   const blueprintsOffset = useAppSelector(selectOffset);
   const blueprintsLimit = useAppSelector(selectLimit);
-
-  const { selectedBlueprintVersion } = useGetBlueprintsQuery(
+  const { selectedBlueprintVersion } = (
+    isOnPremise === 'cockpit'
+      ? useGetBlueprintsOnPremQuery
+      : useGetBlueprintsQuery
```
lucasgarfield, line 88: We really want the hooks to have identical names so that we don't have to have conditionals like this in the code where we differentiate between IB service and IB on-prem.

I think we should look into conditional importing. Then we can import the hook from the service store if it is the service, and import the hook from the on-prem store if it is on-prem.

lucasgarfield (changes requested):

src/Components/ImagesTable/ImagesTable.tsx
lucasgarfield: I think this should be possible:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#importing_different_modules_based_on_environment

src/store/cloudApi.ts
```diff
@@ -0,0 +1,33 @@
...
+      name: 'Blueprint 2',
+      version: '1.1',
+      description: '70% Dark Chocolate Blueprint 2',
+      last_modified_at: '2021-09-09T14:38:00.000Z',
+    },
+  ],
+};
+
+// initialize an empty api service that we'll inject endpoints into later as needed
+export const cloudApi = api.injectEndpoints({
```
lucasgarfield, line 23: We should probably call this the `blueprintsReducer`. We'll add the `cloudApi` later but we will need to handle all the blueprints related queries ourselves using Cockpit and files.

lucasgarfield (approved):
A very promising start! :heart: 

Can you please add a commit that updates the README with instructions on how to get this running locally?

I tried the two new `npm run` commands, but couldn't get this to appear inside of Cockpit on a fresh Fedora install.

cockpit/webpack.config.js
```diff
@@ -0,0 +1,68 @@
...
+module.exports = {
+  entry: './src/AppCockpit.tsx',
+  output,
+  mode,
+  devtool,
+  plugins,
+  externals: { cockpit: 'cockpit' },
+  resolve: {
+    modules: ['node_modules'],
+    extensions: ['.js', '.jsx', '.ts', '.tsx'],
```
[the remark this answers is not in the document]
mgold1234: if I remove it I recieve this error
ERROR in ./node_modules/moment/locale/zh-tw.js 8:50-70
Module not found: Error: Can't resolve '../moment' in '/Users/michalgold/git/edge-management/image-builder-project/image-builder-frontend/node_modules/moment/locale'
 @ ./node_modules/moment/locale/ sync ^\.\/.*$ ./zh-tw.js
 @ ./node_modules/moment/moment.js 2165:16-50

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -153,6 +171,81 @@ const ImagesTable = () => {
...
 
+  let composes = data?.data;
+  if (selectedBlueprintId && blueprintVersionFilter === 'latest') {
+    composes = composes?.filter((compose) => {
+      return compose.blueprint_version === selectedBlueprintVersion;
+    });
+  }
+  const itemCount = data?.meta.count || 0;
+
+  if (isOnPremise === 'cockpit') {
```
amirfefer, line 182: For a better maintainable code, env logic should be separated from component logic as far as we can. I think it would be better to use abstract environment specific logic, separating the component logic from the env, by using dedicated config files for each env (i.e for API specification),  HOC (high order components), React Context for propagating the `isOnPrem`, dynamic importing and more.

cockpit/webpack.config.ts
```diff
@@ -0,0 +1,71 @@
+const path = require('path');
```
amirfefer, line 1: The `isOnPrem` env var  is constant during the entire a, so we can leverage webpack to build the relevant modules, this also ensure that the environment-specific modules are code-split and loaded only when necessary

i.e
```js
module.exports = {
// ...
  plugins: [
    new webpack.DefinePlugin({
      // This replaces `process.env.IS_ON_PREMISE` with `true` or `false` at build time
      'process.env.IS_ON_PREMISE': JSON.stringify(isOnPremise),
    }),
    // ... other plugins
  ],
  resolve: {
    alias: {
      // Use different modules based on the environment
      // For example, `api` will resolve to `cloudApi` or `imageBuilderApi` accordingly
      api: isOnPremise
        ? path.resolve(__dirname, 'src/store/imageBuilderApi.ts')
        : path.resolve(__dirname, 'src/store/cloudApi.ts'),
    },
  },
```

amirfefer (commented):
@mgold1234,

Great start! 👍

Implementing a standalone UI that supports both hosted micro frontends and on-premise environments is certainly challenging. Here are some key points to consider to enhance your approach:

#### Separation of Environment-Specific Logic

- For better maintained codebase, it's important to Isolate environment-specific logic from component logic to improve maintainability.
- Consider using Higher-Order Components (HOCs)  to manage environment differences effectively and keep the common codebase as whole as much as we can.

#### Reduce Code Duplication

- Conditional Rendering: Only render parts of the UI that differ between environments conditionally, minimizing duplication.
- Consider using react context, utilize React Context to propagate environment state, allowing specific changes without repetitive code.

#### Leverage Webpack and Configuration Files

- Build-Time Constants: Since isOnPremise is a build-time constant, use Webpack and configuration files to handle environment separation.
- Configure Webpack to build the appropriate modules (like API modules)  based on the environment, ensuring seamless integration.
- Swap out modules based on the environment, ensuring that only relevant code is included in the final bundle, ensures that environment-specific modules are code-split and loaded only when necessary

#### Hosted env via on-prem env
In the hosted environment, the app interacts with various services like sources, provisioning, and RBAC.
Consider use ConsoleDot’s [standalone solution](https://github.com/RedHatInsights/frontend-components/tree/master/packages/config#standalone) for building frontend services in a standalone environment. Although it's no longer maintained, its key concepts could offer valuable insights.

src/AppCockpit.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { HashRouter } from 'react-router-dom';
+
+import { Router } from './Router';
+import { store } from './store';
+
+const Application = () => {
+  return (
+    <React.Fragment>
+      <NotificationsPortal />
+      <HashRouter>
```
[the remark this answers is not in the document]
mgold1234: HashRouter allows you to manage client-side routing using URL fragments (hashes) without needing any special server setup.
HashRouter relies on URL hashes, it does not require the server to handle different routes, making it a more straightforward solution for environments where server-side routing cannot be easily configured.
we unable to configure URL rewriting on the server in an on-premise environment, so  using HashRouter is a practical solution. It avoids the need for server-side routing configuration and ensures that client-side routing works correctly without relying on server support.

src/Utilities/useGetEnvironment.ts
```diff
@@ -26,3 +26,18 @@ export const useFlagWithEphemDefault = (
   const { getEnvironment } = useChrome();
   return (getEnvironment() === 'qa' && ephemDefault) || getFlag;
 };
+
+export const isOnPremise = process.env.APP_TYPE_ENV;
```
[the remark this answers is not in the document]
mgold1234: I changed it to boolean

src/Utilities/useGetEnvironment.ts
```diff
@@ -26,3 +26,18 @@ export const useFlagWithEphemDefault = (
...
+
+export const getFeatureFlag = (
```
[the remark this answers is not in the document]
mgold1234: I think the correct location of isOnPremise in the constant file and not insdie the hook

src/AppCockpit.tsx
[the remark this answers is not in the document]
amirfefer: Both `HashRouter` and `BrowserRouter` seem to work here. The hash router enforces all routes to add the `#` prefix.

package.json
```diff
@@ -84,14 +90,15 @@
...
-    "start": "fec dev",
     "start:stage": "fec dev --clouddotEnv=stage",
     "start:prod": "fec dev --clouddotEnv=prod",
     "start:msw:stage": "NODE_ENV=development MSW=TRUE fec dev --clouddotEnv=stage",
     "test": "TZ=UTC vitest run",
     "test:watch": "TZ=UTC vitest",
     "test:coverage": "TZ=UTC vitest run --coverage",
     "build": "fec build",
+    "cockpit:build": "webpack --config cockpit/webpack.config.ts && npm run cockpit:buildln",
+    "cockpit:buildln": "ln -s $(pwd)/cockpit/public ~/.local/share/cockpit/image-builder-frontend",
```
amirfefer, line 101: nitpick: the symlink can be added in a first setup, i.e `cockpit:setup`.  it's not needed for every build.

amirfefer (commented):
I tested this PR with cockpit, the landing page was rendered as expected in empty mode! 👍  The routing (like opening the wizard) doesn't work though, we need to adjust the prefix `insights/image-builder` for links.

package.json
mgold1234: is it really metter if its build or setup definition?

src/Utilities/useGetEnvironment.ts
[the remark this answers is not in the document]
amirfefer: I agree, `isOnPremise` should stay constant, the added example is for custom hooks usage.

package.json
amirfefer: setup is for initial/one-time tasks like the host patch for adding consoleDot DNS records. Building the cockpit bundle is a repetitive task, which also will be triggered from CI automation, so I think it would better to keep the logic separated.  Keep in mind that with no webpack dev server, a build is required for every change during development.

lucasgarfield: Do we want to do this all in `package.json`? The Cockpit project has a "starter kit" repo (https://github.com/cockpit-project/starter-kit/) which uses a Makefile to set up the development environment - would it make sense to use a similar pattern here? 

See here:
https://github.com/cockpit-project/starter-kit/blob/4d8fc78a3f32e0c7b7abc6961c2f54e8ccd47514/Makefile#L105

src/AppCockpit.tsx
```diff
@@ -0,0 +1,38 @@
...
+    </React.Fragment>
+  );
+};
+const ImageBuilder = () => (
+  <Provider store={store}>
+    <Application />
+  </Provider>
+);
+
+const main = async () => {
```
lucasgarfield, line 30: This seems very similar to what we do in https://github.com/osbuild/image-builder-frontend/blob/main/src/bootstrap.tsx, could we re-use that code?

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -81,6 +83,14 @@ const ImagesTable = () => {
   const blueprintsOffset = useAppSelector(selectOffset);
   const blueprintsLimit = useAppSelector(selectLimit);
 
+  let useGetBlueprintsQuery;
```
lucasgarfield, line 86: Can we try doing a dynamic import at the top level instead of this? Something like this:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#importing_different_modules_based_on_environment

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -153,6 +163,81 @@ const ImagesTable = () => {
     );
   }
```
lucasgarfield, line 165: Can we apply the DRY ("Don't repeat yourself") principle to lines 166 through 240? This is a lot of copy/pasted code. The goal is to not conditionally render anything and to re-use/share as much code as possible between the service and on-prem.

lucasgarfield (changes requested):

package.json
amirfefer: Creating a makefile just for cockpit setup?
Also, notice that the starter-kit switched to esbuild from webpack.

mgold1234: I think that it is make sense to leave it at package.json,
all the scripts in one place- we build Saas env there and also build the onPrem env

src/AppCockpit.tsx
lucasgarfield: Okay, I think it makes sense to have separate entry points!

src/store/cockpitApi.ts
```diff
@@ -0,0 +1,24 @@
+import { emptyCockpitApi as api } from './emptyCockpitApi';
+import {
+  GetBlueprintsApiArg,
+  GetBlueprintsApiResponse,
+} from './imageBuilderApi';
+
+// Injects API endpoints into the service for querying blueprints
+export const blueprintsReducer = api.injectEndpoints({
```
lucasgarfield, line 10: This looks good, but we need to teach it how to actually return a GetBlueprintsApiResponse object.

You'll want to use a queryFn, check the RTKQ docs here:
https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#customizing-queries-with-queryfn

You might want to use the Cockpit Files API to read in the blueprints from a directory, docs here: https://cockpit-project.org/guide/latest/cockpit-file

We haven't 100% decided on the directory yet, but perhaps `~/.local/share/image-builder/blueprints`.

cockpit/public/index.html
```diff
@@ -0,0 +1,15 @@
...
+
+    <!-- js dependencies -->
+    <script type="text/javascript" src="../base1/cockpit.js"></script>
+    <link href="main.css" rel="stylesheet">
+</head>
+<body>
+<div id="main"></div>
+<script defer src="main.js"></script>
+</body>
+</html>
```
croissanne, line 15: Can you indent this file properly?

src/store/cockpitApi.ts
```diff
@@ -16,17 +19,47 @@ export const blueprintsReducer = api.injectEndpoints({
...
-      query: (queryArg) => ({
-        url: `/blueprints`,
-        params: {
-          name: queryArg.name,
-          search: queryArg.search,
-          limit: queryArg.limit,
-          offset: queryArg.offset,
-        },
-      }),
+      queryFn: async () => {
```
croissanne, line 22: this is great, very neat!

src/Utilities/useGetEnvironment.ts
```diff
@@ -26,3 +28,23 @@ export const useFlagWithEphemDefault = (
   const { getEnvironment } = useChrome();
   return (getEnvironment() === 'qa' && ephemDefault) || getFlag;
 };
+
+export const useGetFeatureFlag = (flagName: string) => {
+  // If it's the Cockpit app (on-premise), always return false
+  if (isOnPremise) {
+    return false;
+  }
+  if (
```
croissanne, line 37: I wonder if a switch statement makes it easier to read here.

croissanne (commented):
This is looking quite good I think, some compiler errors however, but I like the usage of queryFn!

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -150,10 +153,12 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
 
   // =========================TO REMOVE=======================
```
croissanne, line 154: hmm is this comment still valid now? or should it move down?

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -10,7 +10,12 @@ import {
   PageSection,
 } from '@patternfly/react-core';
 import { WizardStepType } from '@patternfly/react-core/dist/esm/components/Wizard';
-import { useFlag } from '@unleash/proxy-client-react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
lucasgarfield, line 14: Why do we need to ignore this?

src/Utilities/useGetEnvironment.ts
```diff
@@ -26,3 +28,24 @@ export const useFlagWithEphemDefault = (
...
+  if (isOnPremise) {
+    return false;
+  }
+  switch (flagName) {
+    case 'edgeParity.image-list':
+    case 'image-builder.snapshots.enabled':
+    case 'image-builder.firstboot.enabled':
+    case 'image-builder.wsl.enabled':
+    case 'image-builder.compliance.enabled':
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
lucasgarfield, line 43: React hooks can't be used inside of conditionals, that's what was probably causing the problem with react-hooks/rules-of-hooks... I don't think we should violate this rule. Let's try to find a pattern that doesn't require disabling this rule.

src/Components/CreateImageWizard/CreateImageWizard.tsx
kingsleyzissou: I wonder if importing this from a relative path would fix this and get rid of the need for the `ts-ignore`

cockpit/public/blueprints/sample-blueprint.json
lucasgarfield, line 1: I think it might be too early for this, maybe we should hold off until the next PR when we add the Cockpit files API. For this PR, we just want to get the empty table to display.

cockpit/public/index.html
```diff
@@ -0,0 +1,15 @@
+<!doctype html>
+<html lang="en-us" class="layout-pf pf-m-redhat-font">
+<head>
+    <meta charset="utf-8">
+    <title>Image-Builder-Frontend</title>
```
lucasgarfield, line 5: Image Builder for title?

fec.config.js
```diff
@@ -99,6 +100,18 @@ module.exports = {
     }),
   },
   plugins: plugins,
+  resolve: {
+    modules: ['node_modules'],
+    extensions: ['.js', '.jsx', '.ts', '.tsx'],
+    alias: {
+      api: resolve(__dirname, './src/store/imageBuilderApi.ts'),
+      pathRes: resolve(__dirname, './src/Utilities/path.ts'),
+      getFeatureFlag: resolve(
```
lucasgarfield, line 109: Feels strange to me, should this be useGetFeatureFlag i.e. a React hook?

fec.config.js
lucasgarfield, line 1: Since FEC is for Insights specifically it feels a bit strange that we need to modify this file.

src/store/cockpitApi.ts
lucasgarfield, line 1: Let's take the file management out and punt it to the next PR. This PR is already complicated enough. :sweat_smile:

src/store/index.ts
```diff
@@ -19,13 +20,26 @@ import wizardSlice, {
...
   [rhsmApi.reducerPath]: rhsmApi.reducer,
   [provisioningApi.reducerPath]: provisioningApi.reducer,
   [complianceApi.reducerPath]: complianceApi.reducer,
+  [blueprintsReducer.reducerPath]: blueprintsReducer.reducer,
+  notifications: notificationsReducer,
+  wizard: wizardSlice,
+  blueprints: blueprintsSlice.reducer,
+});
+
+export const onPremReducer = combineReducers({
```
lucasgarfield, line 38: Let's remove the Insights related reducers and use only what we need for now, bare minimum to get empty blueprints table.

src/store/index.ts
```diff
@@ -97,10 +111,29 @@ export const middleware = (getDefaultMiddleware: Function) =>
...
+      complianceApi.middleware,
+      blueprintsReducer.middleware
     );
 
-export const store = configureStore({ reducer, middleware });
+// eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+export const onPremMiddleware = (getDefaultMiddleware: Function) =>
+  getDefaultMiddleware()
+    .prepend(listenerMiddleware.middleware)
+    .concat(
```
lucasgarfield, line 120: See comment above about Insights reducers/middleware.

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -151,10 +154,12 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
 
   // =========================TO REMOVE=======================
 
-  // Feature flags
```
lucasgarfield, line 154: I wonder if we could just have separate useFlag hooks for on-prem and the service and then import conditionally?

lucasgarfield: Maybe something like the example here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#importing_different_modules_based_on_environment

src/constants.ts
```diff
@@ -8,7 +8,7 @@ export const PROVISIONING_API = '/api/provisioning/v1';
 export const COMPLIANCE_API = '/api/compliance/v2';
 export const CREATE_BLUEPRINT = `${IMAGE_BUILDER_API}/blueprints`;
 export const EDIT_BLUEPRINT = `${IMAGE_BUILDER_API}/blueprints`;
-
+export const isOnPremise = process.env.IS_ON_PREMISE;
```
lucasgarfield, line 11: I wonder if we need this? It doesn't seem like a constant to me, since it can be either true or false. :thinking:

README.md
```diff
@@ -217,6 +217,29 @@ see the [osbuild-getting-started project](https://github.com/osbuild/osbuild-get
...
+#### 1. Build the Cockpit Frontend
+
+Runs Webpack with the specified configuration (cockpit/webpack.config.ts) to build the frontend assets.
+Use this command whenever you need to compile the latest changes in your frontend code.
+
+Creates the necessary directory in the user's local Cockpit share (~/.local/share/cockpit/).
+Creates a symbolic link (image-builder-frontend) pointing to the built frontend assets (cockpit/public).
+Use this command after building the frontend to install it locally for development purposes.
+The symbolic link allows Cockpit to serve the frontend assets from your local development environment, making it easier to test changes in real-time without deploying to a remote server.
+```bash
```
lucasgarfield, line 237: Let's update this for the makefile.

Makefile
```diff
@@ -17,3 +17,28 @@ install:
...
 
+
+
+all: devel-install build devel-uninstall
+
+# this requires a built source tree and avoids having to install anything system-wide
+devel-install:
+	mkdir -p ~/.local/share/cockpit
+	ln  -sfn $(shell pwd)/cockpit/public $(INSTALL_DIR)
+	mkdir -p pkg/lib
```
lucasgarfield, line 28: Hm, this is a bit different than the Cockpit starter kit makefile... do we really need this?
https://github.com/cockpit-project/starter-kit/blob/main/Makefile

src/store/backendApi.ts
```diff
@@ -0,0 +1,6 @@
+import { useGetBlueprintsQuery as useImageBuilderGetBlueprintsQuery } from './imageBuilderApi';
+import { useGetBlueprintsQuery as useCockpitGetBlueprintsQuery } from './cockpitApi';
+
+export const useGetBlueprintsQuery = process.env.IS_ON_PREMISE
```
kingsleyzissou, line 4: Again, same concept here, we can just import `useGetBlueprints` query from this file.
And then depending on what environment we're in, we should get the one we need.

Makefile
kingsleyzissou: Haven't had a chance to look at this yet, will come back to it tomorrow :)

Makefile
```diff
@@ -17,3 +17,28 @@ install:
...
+	git clone --depth 1 --branch main https://github.com/cockpit-project/cockpit.git temp_cockpit_repo
+	cp -r temp_cockpit_repo/pkg/lib/* pkg/lib/
+	rm -rf temp_cockpit_repo
+
+build:
```
kingsleyzissou, line 33: @schuellerf you had some suggestions about the makefile, right?
Do you mind having a look and adding your comments here please
🙏

src/store/backendApi.ts
lucasgarfield: I love this, it's so simple.

kingsleyzissou:
Screenshot attached 😄 

image

src/Utilities/useGetEnvironment.ts
```diff
@@ -22,7 +22,16 @@ export const useFlagWithEphemDefault = (
...
+    return false;
+  }
+  const getFlag = useUnleashFlag(flag);
   const { getEnvironment } = useChrome();
   return (getEnvironment() === 'qa' && ephemDefault) || getFlag;
 };
+
+// Since some of these flags are only relevant to
+// the service, we need a way of bypassing this for on-prem
+export const useFlag = !process.env.IS_ON_PREMISE
```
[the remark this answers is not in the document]
lucasgarfield: Great idea!

Makefile
schuellerf: My "comment" is more a PR, I thought that's easier to get in line like this
https://github.com/mgold1234/image-builder-frontend/pull/12/commits

https://github.com/mgold1234/image-builder-frontend/pull/12/commits/496cd6607006afed0ae8bbad9879e530c16d3a40

kingsleyzissou: Okay, thanks I will have a look. I might end up cherry-picking that commit.

I just need to refer to the cockpit boilerplate [here](https://github.com/cockpit-project/starter-kit) and see what directories we will actually need. I don't think we need `pkg/lib` any more.

Makefile
```diff
@@ -17,3 +17,28 @@ install:
...
+
+
+all: devel-install build devel-uninstall
+
+# this requires a built source tree and avoids having to install anything system-wide
+devel-install:
+	mkdir -p ~/.local/share/cockpit
+	ln  -sfn $(shell pwd)/cockpit/public $(INSTALL_DIR)
+	mkdir -p pkg/lib
+	git clone --depth 1 --branch main https://github.com/cockpit-project/cockpit.git temp_cockpit_repo
```
kingsleyzissou, line 29: I'm also not convinced we need to do this

src/store/index.ts
```diff
@@ -100,9 +114,34 @@ export const middleware = (getDefaultMiddleware: Function) =>
...
+
+export const serviceStore = configureStore({
+  reducer: serviceReducer,
+  middleware: serviceMiddleware,
+});
+
+// we don't need to export these for now, they are just helpers
+// for some of the functions in this file
+type onPremState = ReturnType<typeof onPremStore.getState>;
+type serviceState = ReturnType<typeof serviceStore.getState>;
```
kingsleyzissou, line 138: This was needed to get the `startAppListening` functions to work properly

src/store/index.ts
```diff
@@ -68,10 +79,13 @@ startAppListening({
...
-      imageBuilderApi.endpoints.getArchitectures.select({
-        distribution: distribution,
-      })(state);
+    const architecturesResponse = process.env.IS_ON_PREMISE
+      ? cockpitApi.endpoints.getArchitectures.select({
+          distribution: distribution,
+        })(state as onPremState)
+      : imageBuilderApi.endpoints.getArchitectures.select({
+          distribution: distribution,
+        })(state as serviceState);
```
kingsleyzissou, line 88: Had to force the type here

src/store/backendApi.ts
kingsleyzissou: Thanks 😄

Makefile
kingsleyzissou: Also I feel like github doesn't like the umlaut in your last name. Had to remove it to recognise you as a co-author 😅

cockpit/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "dashboard": {
+    "index": {
+      "label": "Image Builder",
+      "icon": "pficon-build",
+      "docs": [
+        {
+          "label": "Creating system images",
+          "url": "https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/creating_customized_rhel_images_using_the_image_builder_service"
```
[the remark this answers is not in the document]
kingsleyzissou: I just used our public docs for now until we have something more official.

lucasgarfield (approved):
lgtm

PR #2425, opened 2024-09-12
src: Prefetch list of targets on hover of Create blueprint button

regexowl:
This prefetches list of possible targets based on distribution for `onMouseEnter` event of Create blueprint button.


mgold1234 (commented):
/lgtm
its looking very good! thank you

mgold1234 (approved):
/lgtm

PR #2426, opened 2024-09-12
test: Update Edit wizard test

regexowl:
The content steps were surfaced from previously used Content section and we're now testing with all flags set to true. Updating the list of step nav items that should be enabled.


croissanne (approved):

croissanne (approved):

mgold1234 (approved):
/lgtm

PR #2432, opened 2024-09-13
src/store: Remove server store hooks

croissanne:
Removing the server store makes the way we handle data going in and out of the wizard state more consistent. Each customisation is mapped into the wizard state and pulled out when generating the blueprint payload.

When the services and kernel customisations are implemented, this information will need to be stored inside of the wizard state anyway.

Lastly this will make implementing a compliance step easier for edit mode, removing the need to write to the wizard state from within the server store when only a compliance policy id is available (on the review page), which would be used to fetch the profile ref id, which would in turn be used to fetch the customisations not stored in the wizard state.


regexowl (approved):
Looks good! Thanks :sparkles:

regexowl:
/retest

pr_check is very temperamental today ;_;

mgold1234 (approved):
/lgtm

PR #2433, opened 2024-09-13
Wizard: Validate registration with useValidation

regexowl:
This migrated Registration step validation to `useValidation`.

The scenarios in which the Next button and Create/Save button should be disabled are:
- any of "register now" options is chosen, but no activation key is selected
- activation key is selected, but it's invalid (information about it cannot be fetched)


avitova:
Nice! I just rebased your solution on my latest local changes for import, and works well. :clap: Thank you, that was very quick!

avitova:
\retest

mgold1234 (approved):
looks good!

PR #2434, opened 2024-09-13
HMS-4667: Blueprints: Allow importing without subscription

avitova:
This requires a small change on our backend - on export endpoint, we would need to export "subscription: {}" in the customizations field, so that we can detect that a subscription was there, when we import the blueprint back. :)
Let's not merge before we merge the backend change.

All information is in [the ticket](https://issues.redhat.com/browse/HMS-4667). But TLDR; we omit subscription now, and because of that, we can not detect if the blueprint was subscribed or not.

### DIFF
--- a/src/Components/CreateImageWizard/utilities/requestMapper.ts
+++ b/src/Components/CreateImageWizard/utilities/requestMapper.ts
@@ -206,32 +206,38 @@ function commonRequestToState(
     distribution:
       getLatestRelease(request.distribution) || initialState.distribution,
     imageTypes: request.image_requests.map((image) => image.image_type),
-    azure: {
-      shareMethod: (azureUploadOptions?.source_id
-        ? 'sources'
-        : 'manual') as AzureShareMethod,
-      source: azureUploadOptions?.source_id || '',
-      tenantId: azureUploadOptions?.tenant_id || '',
-      subscriptionId: azureUploadOptions?.subscription_id || '',
-      resourceGroup: azureUploadOptions?.resource_group,
-    },
-    gcp: {
-      shareMethod: (gcpUploadOptions?.share_with_accounts
-        ? 'withGoogle'
-        : 'withInsights') as GcpShareMethod,
-      accountType: gcpUploadOptions?.share_with_accounts?.[0].split(
-        ':'
-      )[0] as GcpAccountType,
-      email: gcpUploadOptions?.share_with_accounts?.[0].split(':')[1] || '',
-    },
-    aws: {
-      accountId: awsUploadOptions?.share_with_accounts?.[0] || '',
-      shareMethod: (awsUploadOptions?.share_with_sources
-        ? 'sources'
-        : 'manual') as AwsShareMethod,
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -206,32 +206,38 @@ function commonRequestToState(
...
+      ? {
+          shareMethod: (azureUploadOptions?.source_id
+            ? 'sources'
+            : 'manual') as AzureShareMethod,
+          source: azureUploadOptions?.source_id || '',
+          tenantId: azureUploadOptions?.tenant_id || '',
+          subscriptionId: azureUploadOptions?.subscription_id || '',
+          resourceGroup: azureUploadOptions?.resource_group,
+        }
+      : initialState.azure,
```
amirfefer, line 219: Makes sense to use default values when no uploadOption exists. 👍

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -311,8 +318,34 @@ export const mapExportRequestToState = (
...
+              ? 'register-now-rhc'
+              : 'register-now-insights'
+            : initialState.registration.registrationType,
+          activationKey:
+            request.customizations.subscription?.['activation-key'] ||
+            initialState.registration.activationKey,
+        }
+      : {
+          registrationType: 'register-later',
+          activationKey: '',
```
amirfefer, line 347: Just to make this more structure and readable, we could first create a `subscriptionState` object, using the `subscribed` flag just once.

```ts
  const subscriptionState = subscribed
    ? {
        serverUrl: request.customizations.subscription?.['server-url'] || initialState.env.serverUrl,
        baseUrl: request.customizations.subscription?.['base-url'] || initialState.env.baseUrl,
        registrationType: request.customizations.subscription
          ? request.customizations.subscription.rhc
            ? 'register-now-rhc'
            : 'register-now-insights'
          : initialState.registration.registrationType,
        activationKey: request.customizations.subscription?.['activation-key'] || initialState.registration.activationKey,
      }
    : {
        serverUrl: '',
        baseUrl: '',
        registrationType: 'register-later' as const,
        activationKey: '',
      };
```

and return 
```ts
  return {

    // ...metadata...

    env: {
      serverUrl: subscriptionState.serverUrl,
      baseUrl: subscriptionState.baseUrl,
    },
    registration: {
      registrationType: subscriptionState.registrationType,
      activationKey: subscriptionState.activationKey,
    },
    ...commonRequestToState(blueprintResponse),
```

amirfefer (commented):
Thanks, @avitova! 👍

I have a quick thought: Would it make sense to open the wizard directly on the subscription step when no subscription is present? This could help guide users more intuitively.

I've also left a few minor nitpicks.

avitova:
After discussions about the customer value of this solution, I am closing this pull request. Any details can be found in the backend pull request that is related to this one: [ib backend #1343](https://github.com/osbuild/image-builder/pull/1343#issuecomment-2369231289)

PR #2439, opened 2024-09-16
Api: generate RTK Query for CloudAPI

mgold1234:
This commit adds the Cloud API to RTK Query in preparation of porting cloudapi source code to image-builder-frontend
* paste entire cloudApi schema
* setup cloudApi in api/config file
* create new emptyCludApi.ts file
* add the new end point to the store
* npm run api that generate the cloudApi.ts file with all the endpoints

### DIFF
--- a/api/codegen.sh
+++ b/api/codegen.sh
@@ -6,7 +6,7 @@ npx @rtk-query/codegen-openapi ./api/config/rhsm.ts &
 npx @rtk-query/codegen-openapi ./api/config/contentSources.ts &
 npx @rtk-query/codegen-openapi ./api/config/provisioning.ts &
 npx @rtk-query/codegen-openapi ./api/config/edge.ts &
-
+npx @rtk-query/codegen-openapi ./api/config/cloudapi.ts &
 # Wait for all background jobs to finish
 wait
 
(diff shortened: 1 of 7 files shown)


api/codegen.sh
```diff
@@ -6,7 +6,7 @@ npx @rtk-query/codegen-openapi ./api/config/rhsm.ts &
 npx @rtk-query/codegen-openapi ./api/config/contentSources.ts &
 npx @rtk-query/codegen-openapi ./api/config/provisioning.ts &
 npx @rtk-query/codegen-openapi ./api/config/edge.ts &
-
+npx @rtk-query/codegen-openapi ./api/config/cloudapi.ts &
```
regexowl, line 9: Let's keep the newline in this case

```suggestion
npx @rtk-query/codegen-openapi ./api/config/cloudapi.ts &

```

regexowl:
I've tried to run `npm run api` and there were some changes generated, maybe running it once more is needed to make the "Check for manual changes to API" check happy?

croissanne:
> I've tried to run `npm run api` and there were some changes generated, maybe running it once more is needed to make the "Check for manual changes to API" check happy?

To avoid this I think cloudApi.ts needs to be added to .eslintignore, otherwise lint:js:fix will conflict with api:generate.

regexowl:
> To avoid this I think cloudApi.ts needs to be added to .eslintignore, otherwise lint:js:fix will conflict with api:generate.

Ah right! That explains the changes.

lucasgarfield:
Superseded by #2424

PR #2440, opened 2024-09-16
Insights Compliance integration (HMS-3829)

croissanne:
The current state: 

[webm](https://github.com/user-attachments/assets/6b7d0973-473e-4bc5-9748-67d598999552)

One thing worth discussing is how we want to expose the compliance step. Should we show it alongside the existing openscap step? Should the steps be merged?

Since it would only be in preview, how about we just show both steps for now, knowing that they might conflict?

### DIFF
--- a/src/Components/CreateImageWizard/steps/Oscap/OscapProfileInformation.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/OscapProfileInformation.tsx
@@ -124,7 +124,10 @@ export const OscapProfileInformation = ({
               >
                 Reference ID:
               </TextListItem>
-              <TextListItem component={TextListItemVariants.dd}>
+              <TextListItem
+                data-testid="oscap-profile-info-ref-id"
+                component={TextListItemVariants.dd}
+              >
                 {oscapProfile?.profile_id}
               </TextListItem>
               <TextListItem
--- a/src/Components/CreateImageWizard/steps/Oscap/OscapProfileInformation.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/OscapProfileInformation.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import {
   CodeBlock,
@@ -12,16 +12,30 @@ import {
 } from '@patternfly/react-core';
 
 import { RELEASES } from '../../../../constants';
-import { useAppSelector } from '../../../../store/hooks';
-import { useGetOscapCustomizationsQuery } from '../../../../store/imageBuilderApi';
+import { PolicyRead, usePolicyQuery } from '../../../../store/complianceApi';
+import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
 import {
+  OpenScapProfile,
+  useGetOscapCustomizationsQuery,
+} from '../../../../store/imageBuilderApi';
+import {
+  changeCompliance,
+  selectCompliancePolicyID,
+  selectComplianceProfileID,
   selectDistribution,
-  selectProfile,
 } from '../../../../store/wizardSlice';
 
-export const OscapProfileInformation = (): JSX.Element => {
+type OscapProfileInformationOptionPropType = {
+  allowChangingCompliancePolicy?: boolean;
+};
+
+export const OscapProfileInformation = ({
+  allowChangingCompliancePolicy = false,
+}: OscapProfileInformationOptionPropType): JSX.Element => {
+  const dispatch = useAppDispatch();
   const release = useAppSelector(selectDistribution);
-  const oscapProfile = useAppSelector(selectProfile);
+  const compliancePolicyID = useAppSelector(selectCompliancePolicyID);
+  const complianceProfileID = useAppSelector(selectComplianceProfileID);
 
   const {
     data: oscapProfileInfo,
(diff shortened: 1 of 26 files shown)


ochosi:
What if we showed the OpenSCAP step as a fallback, if a customer hasn't set up a policy yet, and add a huge throbbing notice that they can easily create a policy?

src/Components/CreateImageWizard/steps/Compliance/Compliance.tsx
```diff
@@ -0,0 +1,303 @@
...
+  };
+
+  return (
+    <FormGroup
+      isRequired={true}
+      data-testid="policies-form-group"
+      label={
+        isFetchingPolicies ? (
+          <>
+            Policiy <Spinner size="md" />
```
regexowl, line 213:
```suggestion
            Policy <Spinner size="md" />
```

src/test/Components/CreateImageWizard/steps/Compliance/Compliance.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const user = userEvent.setup();
+  const policyMenu = await screen.findByRole('button', {
+    name: /options menu/i,
+  });
+  await user.click(policyMenu);
+
+  const cisl2 = await screen.findByRole('option', {
+    name: /CIS Red Hat Enterprise Linux 8 Benchmark for Level 2 - Workstation/i,
+  });
+  await user.click(cisl2);
```
regexowl, line 63: This should solve the `was not wrapped in act(...)` hell in test output

```suggestion
  await waitFor(() => user.click(policyMenu));

  const cisl2 = await screen.findByRole('option', {
    name: /CIS Red Hat Enterprise Linux 8 Benchmark for Level 2 - Workstation/i,
  });
  await waitFor(() => user.click(cisl2));
```

regexowl (commented):
This looks awesome so far! Added a few comments that should help with the noise in test outputs.

"radio buttoning" between compliance and Oscap will be great in eliminating the overlap

src/Components/CreateImageWizard/steps/Oscap/OscapProfileInformation.tsx
```diff
@@ -126,4 +175,8 @@ export const OscapProfileInformation = (): JSX.Element => {
   );
 };
 
+OscapProfileInformation.defaultProps = {
+  allowChangingCompliancePolicy: false,
+};
+
```
regexowl, line 181: This can be removed as the default value of the prop is set directly on line 33. The warning in the test outputs then should go away.

src/Components/CreateImageWizard/steps/Oscap/index.tsx
```diff
@@ -26,24 +49,78 @@ const OscapStep = () => {
...
-      </Text>
+      {complianceEnabled && (
+        <FormGroup>
+          <Radio
+            id="openscap radio openscap type"
+            label="OpenSCAP"
+            name="oscap-radio-openscap"
+            isChecked={complianceType === 'openscap'}
+            onChange={() =>
+              dispatch(changeComplianceType('openscap' as ComplianceType))
```
regexowl, line 62: Should the data selected in the other option also get dropped when switching between compliance/openscap? Currently it's possible to select compliance with openscap profile selected.

src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
```diff
@@ -244,6 +252,23 @@ const Review = ({ snapshottingEnabled }: { snapshottingEnabled: boolean }) => {
           <OscapList />
         </ExpandableSection>
       )}
+      {compliancePolicy && complianceType === 'compliance' && (
+        <ExpandableSection
+          toggleContent={composeExpandable(
+            'Compliance',
+            'revisit-compliance',
+            'step-oscap'
```
regexowl, line 260:
```suggestion
            'step-compliance'
```

regexowl (commented):
Looks great with the radio! Added a few nitpicks.

src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
croissanne: It's the same step now, so I think `step-oscap` is correct.

src/Components/CreateImageWizard/steps/Oscap/index.tsx
croissanne: Yea so I wondered about that and I'm not sure. Like currently it'll grab the profile that the policy was based on when switching from openscap to compliance. When switching back without changing the profile the policy shuold still be there. I'm not sure, for the file system customisation we do something where we keep the mountpoints, so when the user switches back and forth they're preserved.

src/Components/CreateImageWizard/steps/Oscap/OscapProfileInformation.tsx
croissanne: ohhh ok

src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
regexowl: Right, sorry, was still thinking in separate steps :bow:

src/Components/CreateImageWizard/steps/Oscap/OscapProfileInformation.tsx
croissanne: Hmm I don't see how it compiles without it ._.

croissanne: oh nvm found it, was missing a `?`

src/Components/CreateImageWizard/steps/Oscap/index.tsx
regexowl: Hmm, I see. Why I'm asking is this case - I've picked an Oscap profile and switched to Compliance. The Oscap profile information is still there, I guess what confuses me is - are we handling oscap and compliance as two separate things or are we looking at compliance like more of a complement to originally selected oscap profile? Hope that's not a super stupid question, still trying to figure out how it works :sweat_smile: 

image

regexowl: @croissanne Just this question, otherwise I'd say all good :+1:

croissanne: Ah right, yea that is a tricky one, the oscap profile information simply takes whatever profile id is present. Maybe it's best to clear between states for now?

regexowl: Perfect, makes sense to me. At least for now, we can refine things as we go. One last thing - when clearing between states we'll need to also clear oscap packages, fsc, kernel and services so those don't get left behind ig.

So something like `handleClear()` in `src/Components/CreateImageWizard/steps/Oscap/Oscap.tsx` maybe?

croissanne: oh right, good point

croissanne: hmm there's definitely some duplication going on now, but moving the radio into oscap would require more restructuring.

regexowl: A bit of duplication is okay, we can revisit it in a follow up :+1:

regexowl (approved):
Thank you! Looks and works great :sparkles:

PR #2441, opened 2024-09-16
Wizard: Fix Registration bug

regexowl:
The second condition in `useRegistrationValidation` needs to be triggered only for `register-now` options, otherwise the Next button stays disabled on fetching and error even for `register-later`.


croissanne (approved):

mgold1234 (approved):
/kgtm

PR #2456, opened 2024-09-20
Wizard: Temporarily disable fixed footer

regexowl:
There's a new bug that's at least partially caused by styling commented out in this PR: when zooming in or out while in Wizard, the Wizard shrunks it's height to an unusable value.


croissanne (approved):

mgold1234 (approved):
/lgtm

PR #2464, opened 2024-09-24
requestMapper: Fix enabled services with firstboot

regexowl:
Previously used `.push()` method was limited by non-writable length of the enabled services array, ending up in the following error:
```
 Uncaught (in promise) TypeError: can't define array index property past the end of an array with non-writable length
```

This fixes the problem by updating the array via spread operator.

Before: when creating an image with Oscap profile with enabled serviced and adding a firstboot script, the Create blueprint button at the Review step threw an error on click.
After fix: it's possible to create an image with Oscap profile that includes enabled services and a firstboot script.


jrusz:
Tests fail on teardown, will be resolved soon in separate PR. I was able to create the blueprint manually. Integration tests will be ran in stage to verify, let's merge this please :)

croissanne (approved):

PR #2478, opened 2024-09-27
CreateImageWizard/requestMapper: fix enabling firstboot

schuellerf:
when there none of `service.{enabled|masked|disabled}` firstboot also is not enabled (if requested)
this patch fixes this bug

### DIFF
--- a/src/Components/CreateImageWizard/utilities/requestMapper.ts
+++ b/src/Components/CreateImageWizard/utilities/requestMapper.ts
@@ -497,7 +497,8 @@ const getServices = (state: RootState): Services | undefined => {
   if (
     services.enabled.length === 0 &&
     services.masked.length === 0 &&
-    services.disabled.length === 0
+    services.disabled.length === 0 &&
+    !selectFirstBootScript(state)
   ) {
     return undefined;
   }
--- a/src/Components/CreateImageWizard/utilities/requestMapper.ts
+++ b/src/Components/CreateImageWizard/utilities/requestMapper.ts
@@ -494,21 +494,19 @@ const getCustomizations = (state: RootState, orgID: string): Customizations => {
 
 const getServices = (state: RootState): Services | undefined => {
   const services = selectServices(state);
-  if (
-    services.enabled.length === 0 &&
-    services.masked.length === 0 &&
-    services.disabled.length === 0 &&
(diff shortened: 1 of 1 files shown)


mgold1234 (approved):
/lgtm

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -497,7 +497,8 @@ const getServices = (state: RootState): Services | undefined => {
   if (
     services.enabled.length === 0 &&
     services.masked.length === 0 &&
-    services.disabled.length === 0
+    services.disabled.length === 0 &&
+    !selectFirstBootScript(state)
```
amirfefer, line 501: nit: instead of making this condition more complex, we can change the order and use the `enabledSvcs` in the condition:

```suggestion
  const enabledSvcs = services.enabled || [];
  const includeFBSvc: boolean =
    !!selectFirstBootScript(state) &&
    !services.enabled?.includes(FIRST_BOOT_SERVICE);
  if (includeFBSvc) {
    enabledSvcs.push(FIRST_BOOT_SERVICE);
  }
  if (
    enabledSvcs.length === 0 &&
    services.masked.length === 0 &&
    services.disabled.length === 0
  ) {
    return undefined;
  }
```

amirfefer (commented):
Thanks @schuellerf ! great catch 👍 
As a regression, could you please add a test case?

src/Components/CreateImageWizard/utilities/requestMapper.ts
schuellerf: Thanks @amirfefer! I also though about this, but first wanted to check if I solved the problem in the first place!
I'll revise

schuellerf:
I'll check if I can fix the test regression …
https://github.com/osbuild/image-builder-frontend/commit/144832f2

regexowl (approved):
Works great! Thank you for the fix!

The tests will be a bit more complicated as vitest doesn't like the editor component the first boot step is based on. Looking into it, hoping to open a follow up with at least some coverage of the functionality soon.

For now let's merge the fix.

regexowl:
And I'll admin merge it for good measure so we don't have to wait for ci weekly maintenance and then pr_check :upside_down_face:

PR #2482, opened 2024-09-30
vitest/test: Inline `@monaco-editor` and enable "First boot request generated correctly" test

regexowl:
This inlines a dependency that was causing problem in test runs and re-enables a test checking request generated when adding first boot script to the blueprint.


mgold1234 (approved):
/lgtm

PR #2489, opened 2024-10-02
test: Update test naming

regexowl:
The incorrect "registration" name is bothering me. Also updated the name of the first describe block to match the pattern we have in other test files.


mgold1234 (approved):
/lgtm

PR #2495, opened 2024-10-07
Wizard: Fix repos toggle height

regexowl:
The repos toggle height was slightly bigger than the Available/Selected toggle. This was caused by the popover button. Changes its size to small to fix the issue.

image


mgold1234 (approved):
/lgtm

PR #2497, opened 2024-10-07
Wizard: Move VMware radios into VMware checkbox

regexowl:
This moves the VMware radios into the body of VMware checkbox, following PatternFly's pattern of parent and child checkboxes / form elements: https://www.patternfly.org/components/forms/checkbox#controlled


mgold1234 (approved):
/lgtm

PR #2498, opened 2024-10-07
Wizard: Fix VMware select and unselect

regexowl:
This updates the behaviour of VMware checkbox to make it possible to unselect VMware as a target.

How to reproduce:
1. select VMware vSphere checkbox
2. click on Virtual disk radio
3. try to unselect VMware vSphere checkbox

Previous behaviour - the checkbox can't be unselected New behaviour - the checkbox can be unselected, removing relevant type of VMware target from imageTypes

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
@@ -216,6 +216,7 @@ const TargetEnvironment = () => {
                       name="vsphere-radio"
                       aria-label="VMware vSphere radio button OVA"
                       id="vsphere-radio-ova"
+                      data-testid="radio-vsphere-ova"
                       label={
                         <>
                           Open virtualization format (.ova)
@@ -264,6 +265,7 @@ const TargetEnvironment = () => {
                     name="vsphere-radio"
                     aria-label="VMware vSphere radio button VMDK"
                     id="vsphere-radio-vmdk"
+                    data-testid="radio-vsphere-vmdk"
                     label={
                       <>
                         Virtual disk (.vmdk)
--- a/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 
 import {
   Button,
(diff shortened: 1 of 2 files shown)


mgold1234 (dismissed):
/lgtm

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
```diff
@@ -191,13 +192,18 @@ const TargetEnvironment = () => {
...
+                if (!hasVsphere) {
+                  handleToggleEnvironment('vsphere-ova');
+                } else {
+                  if (environments.includes('vsphere')) {
+                    handleToggleEnvironment('vsphere');
+                  }
+                  if (environments.includes('vsphere-ova')) {
+                    handleToggleEnvironment('vsphere-ova');
+                  }
+                }
```
avitova, line 206: This block is a bit confusing to me, and I had to think twice. I think in this context, maybe this will be a bit more readable:
!environments.includes('vsphere') && !environments.includes('vsphere-ova')

Could you explain a bit more what is this supposed to do - why do we call handleToggleEnvironment('vsphere-ova') if !hasSphere?

avitova (commented):
Thanks! I just have one nitpick because the code was not easily readable.

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
lucasgarfield: The `if (!hasVsphere)` handles the case where a vSphere environment has not been selected, and then selects vsphere-ova as the default.

The `else` clause handles the case where a vSphere environment has been selected, and deselects and also deselects whichever radio (ova or vmdk) has been selectged.

I think you're right, @avitova, that we should make this more obvious. I think the right way to do this is to refactor handleToggleEnvironment into two separate functions, `addEnvironment` and `removeEnvironment`. I'll create an issue to track this refactoring.

lucasgarfield: Fixed it up a bit @avitova, lmk if you still find it confusing and we can do some additional refactoring.

lucasgarfield (approved):
lgtm

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
avitova: No, it looks good, thank you!:)

PR #2503, opened 2024-10-08
Wizard/test: Implement keydown for target tiles

regexowl:
This implements onKeyDown functionality for the Public cloud tiles, allowing selecting them with a Space key.

Corresponding test was also re-enabled.


mgold1234 (approved):
/lgtm

mgold1234:
can you please squash the commits before merging

regexowl:
I would like to keep the commits separate in this case as the first is connected only to implementing, re-enabling and updating the test for the key down functionality and the second commit cleans up the CreateImageWizard.test.tsx file.

PR #2505, opened 2024-10-08
update compliance and content sources api

avitova:
Some changes are still missing, but let's get this in so that we will not have a huge PR later. :)


regexowl (approved):
Looks good! Thank you for adding content to the pull script!

PR #2510, opened 2024-10-10
src: Resolve type-only imports causing circular dependencies

regexowl:
There were some type-only imports that caused circular dependencies. Those will be resolved in this PR.

~~Two circular dependencies are still remaining and will need to be addressed later. When that's done we should add the check to the dev-checks so we don't introduce new problems again.~~
Done as well.


croissanne (approved):
thank you :heart:

croissanne (approved):

PR #2519, opened 2024-10-15
src: Add on prem blueprints import support (HMS-4683)

avitova:
This pull request relates to [#4683](https://issues.redhat.com/browse/HMS-4683).
We want to support import on prem blueprints in our import feature. This enables importing toml blueprints that get automatically detected, and parsed.
A lot of fields we need to ignore, a lot of fields from our blueprints are not in the on-prem blueprint structure. This is our first attempt to allow on-prem blueprint, and it will get us more information whether users would use it or not. We also need to add a method to measure the usage of this feature. This is not here yet, and I would like to do it as a follow up, as it can raise discussions, and we do not want to block this feature on it.

We will bring a different solution, so eventually, we will get rid of this mapper. But the other solution that should be more "bulletproof" will take time, in the meantime we can use this one. :)

I know, that on-prem blueprints can also come in json, this will be quite easy to do, we will adjust the UI a bit for that, so that users would be able to import even on-prem json.

Feel free to ask any questions, and find bugs!

### DIFF
--- a/src/Components/Blueprints/ImportBlueprintModal.test.tsx
+++ b/src/Components/Blueprints/ImportBlueprintModal.test.tsx
@@ -1,6 +1,7 @@
 import { screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
+import { clickNext } from '../../test/Components/CreateImageWizard/wizardTestUtils';
 import { renderCustomRoutesWithReduxRouter } from '../../test/testUtils';
 
 const BLUEPRINT_JSON = `{
@@ -121,6 +122,58 @@ const INVALID_JSON = `{
   "name": "Blueprint test"
 }`;
 
+const ONPREM_BLUEPRINT_TOML = `
+name = "tmux"
+description = "tmux image with openssh"
+version = "1.2.16"
+distro = "rhel-93"
+
+[[packages]]
+name = "tmux"
+version = "*"
+
+[[packages]]
+name = "openssh-server"
+version = "*"
+
+[[groups]]
+name = "anaconda-tools"
+
+[customizations]
+hostname = "baseimage"
+fips = true
+
+[[customizations.sshkey]]
+user = "root"
+key = "PUBLIC SSH KEY"
+
+[customizations.services]
+enabled = ["sshd", "cockpit.socket", "httpd"]
+disabled = ["postfix", "telnetd"]
+masked = ["rpcbind"]
+
+[[customizations.files]]
+data = "W1VuaXRdCkRlc2NyaXB0aW9uPVJ1biBmaXJzdCBib290IHNjcmlwdApDb25kaXRpb25QYXRoRXhpc3RzPS91c3IvbG9jYWwvc2Jpbi9jdXN0b20tZmlyc3QtYm9vdApXYW50cz1uZXR3b3JrLW9ubGluZS50YXJnZXQKQWZ0ZXI9bmV0d29yay1vbmxpbmUudGFyZ2V0CkFmdGVyPW9zYnVpbGQtZmlyc3QtYm9vdC5zZXJ2aWNlCgpbU2VydmljZV0KVHlwZT1vbmVzaG90CkV4ZWNTdGFydD0vdXNyL2xvY2FsL3NiaW4vY3VzdG9tLWZpcnN0LWJvb3QKRXhlY1N0YXJ0UG9zdD1tdiAvdXNyL2xvY2FsL3NiaW4vY3VzdG9tLWZpcnN0LWJvb3QgL3Vzci9sb2NhbC9zYmluL2N1c3RvbS1maXJzdC1ib290LmRvbmUKCltJbnN0YWxsXQpXYW50ZWRCeT1tdWx0aS11c2VyLnRhcmdldAo="
+data_encoding = "base64"
+ensure_parents = true
+path = "/etc/systemd/system/custom-first-boot.service"
+
+[[customizations.files]]
+data = "IyEvYmluL2Jhc2gKZmlyc3Rib290IHNjcmlwdCB0byB0ZXN0IGltcG9ydA=="
+data_encoding = "base64"
+ensure_parents = true
+mode = "0774"
+path = "/usr/local/sbin/custom-first-boot"
+
+[[customizations.filesystem]]
+mountpoint = "/var"
+minsize = 2147483648
+
+[customizations.installer]
+unattended = true
+sudo-nopasswd = ["user", "%wheel"]
+`;
+
 const uploadFile = async (filename: string, content: string): Promise<void> => {
   const user = userEvent.setup();
   const fileInput: HTMLElement | null =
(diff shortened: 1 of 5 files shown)


ezr-ondrej (commented):
What's the status here? should I test it out?

src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
```diff
@@ -77,7 +77,7 @@ const PackageRecommendations = () => {
           },
         });
 
-        if (response && response.data && response.data.packages.length > 0) {
+        if (response && response.data && response.data.packages && response.data.packages.length > 0) {
```
avitova, line 80: This is not related, but I was getting some errors, and I believe that we need to check for this too.

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -50,52 +53,63 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+        const isToml = filename.endsWith('.toml');
+        const isJson = filename.endsWith('.json');
+        if (isToml) {
+          const tomlBlueprint = parse(fileContent);
+          const blueprintFromFile = mapOnPremToHosted(tomlBlueprint);
+          const importBlueprintState = mapExportRequestToState(
+            blueprintFromFile,
+            []
+          );
+          setImportedBlueprint(importBlueprintState);
```
ezr-ondrej, line 73: Should we show some warning, that this is a best effort import?

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -50,52 +53,63 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+            blueprintFromFile.image_requests || []
+          );
+          setImportedBlueprint(importBlueprintState);
+        }
+      } catch (error) {
+        setIsInvalidFormat(true);
+        dispatch(
+          addNotification({
+            variant: 'warning',
+            title: 'No blueprint was build',
```
ezr-ondrej, line 94: I know this is just a copy, but the warning is probably incorrect, we should say something like "File is not a valid blueprint"

ezr-ondrej: Mocks are not always 100% tought through, this case I feel like it could have been copy pasted from other mock.
What I'm interested in is whether my comment makes sense to you, or if I just don't get why this might be actually correct message :)

avitova:
Okay, I addressed all the comments, thanks for the review!
/retest

src/Components/Blueprints/ImportBlueprintModal.tsx
avitova: Well, I feel like the notification might make sense. We already inform in the warning at the bottom of the text field that the format is not compatible with our blueprints. So this one just made sense to me, so that users know that we did not create any blueprint just from the fact that they uploaded the file. We can discuss this with Kelsea, but I think this is a minor thing:)

src/Components/Blueprints/ImportBlueprintModal.tsx
avitova: Let me know how you like that one I came up with. We can adjust it. :)

src/Components/Blueprints/ImportBlueprintModal.tsx
ezr-ondrej: Yeah, totaly minor, if you feel this is good, lets keep it :)

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -33,69 +36,83 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
   };
-  const [jsonContent, setJsonContent] = React.useState('');
+  const [fileContent, setFileContent] = React.useState('');
   const [importedBlueprint, setImportedBlueprint] =
     React.useState<wizardState>();
   const [isInvalidFormat, setIsInvalidFormat] = React.useState(false);
   const [filename, setFilename] = React.useState('');
   const [isLoading, setIsLoading] = React.useState(false);
   const [isRejected, setIsRejected] = React.useState(false);
+  const [isOnPrem, setIsOnPrem] = React.useState(false);
```
ezr-ondrej, line 50: You might want to reset this flag in `onClear` handler.

ezr-ondrej (approved):
I'm good here, but merging with the current CI state might be difficult xD

FTR: I've tested this with toml and seems fine, just one nit :)

src/Components/Blueprints/ImportBlueprintModal.tsx
avitova: Good catch, thank you! I added more resets to the default value, you are right.

ezr-ondrej (dismissed):
Thanks, looks great! :+1:

src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
```diff
@@ -77,7 +77,12 @@ const PackageRecommendations = () => {
           },
         });
 
-        if (response && response.data && response.data.packages.length > 0) {
+        if (
```
mgold1234, line 80: is that change related to the pr?

mgold1234 (commented):
this support of import on Prem is amazing!

src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
avitova: This is not related. However, we access the `response.data.packages`. This was a bug, that probably was not tested, but with import, I found out that this was failing. And I think we should not really use methods that are not safe for empty arrays before checking `response.data.packages`.

regexowl (commented):
This looks great, thank you! 

Found one problem, the Wizard crashes for me when I get to the packages step because of the PackageRecommendations, adding a screenshot of the error:

image

Tried to temporarily comment out PackageRecommendations to see other imported values and everything else looks amazing :heart:

regexowl:
I think adding a new format to the import options shouldn't have effect on existing IQE tests, but not completely sure - what do you think @jrusz?

regexowl:
Ah, this seems to help with the crashes:
image

`PackageRecommendations.tsx`, line 112

regexowl:
Will address the package recommendations boo boos in https://github.com/osbuild/image-builder-frontend/pull/2606

jrusz:
> I think adding a new format to the import options shouldn't have effect on existing IQE tests, but not completely sure - what do you think @jrusz?

Yeah it should be fine as long as the old workflow and format works.

ezr-ondrej:
@avitova now there are merge conflicts with the fix, can you please rebase, it seems apart of that we are ready :)

src/Components/Blueprints/ImportBlueprintModal.test.tsx
```diff
@@ -207,4 +260,79 @@ describe('Import modal', () => {
...
+    await screen.findByRole('textbox', {
+      name: 'Select activation key',
+    });
+    await clickNext();
+
+    // OpenScap
+    await clickNext();
+
+    //File system configuration
+    await clickNext();
```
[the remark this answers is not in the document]
avitova: Yeah, we might. :thinking: And I think it could show to us more, what values we are able to import from on prem. :) But would you be okay if we just did that in a followup maybe? :thinking:

avitova:
I rebased and pinned the version. Thank you for your reviews. :)

regexowl (approved):
Looks great, thank you!

PR #2523, opened 2024-10-17
.github/workflows: run dev checks on main

croissanne:
The codecov comments don't make much sense as there's no reference point currently.


ochosi (approved):
LGTM!

PR #2525, opened 2024-10-18
src: Add custom repositories export <HMS-3597>

avitova:
:wave: Export of custom repositories is here!


avitova:
Meh, this should definitely be done on the backend side. 🤦‍♀️ Tgif.

ezr-ondrej:
Well we don't have any import endpoint at the backend, right? Would that mean we are about to add it?

avitova:
No, I do not think that we need the import backend to implement this, we can just call the endpoint on frontend to import it. Depends if we want to introduce it.

PR #2526, opened 2024-10-18
Wizard: switch snapshot date in wizard to RFC3339

xbhouse:
The changes here should format the requested date from YYYY-MM-DD to RFC3339 in the snapshot step of the wizard. This helps content-sources clean up some tech debt and allows us to use RFC3339 for all date formats in our template-related APIs.

Background:
This [PR](https://github.com/content-services/content-sources-backend/pull/839) removed support for YYYY-MM-DD in the request for `/snapshots/for_date/` and broke things. The snapshot step in the wizard needed to be converted to use RFC3339 also. This commit has been reverted and things are good again.

Without these changes in the wizard, errors like this would be seen when calling `/snapshots/for_date/` from the content sources API with an unsupported date format:
```
parsing time \"2024-10-18\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \"\" as \"T\", internal=parsing time \"2024-10-18\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \"\" as \"T\" \n
```

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
@@ -478,7 +478,7 @@ export const ContentList = ({
       case useLatest:
         return 'Use latest';
       case !!snapshotDate:
-        return `State as of ${snapshotDate}`;
+        return `State as of ${yyyyMMddFormat(new Date(snapshotDate))}`;
       default:
         return '';
     }
@@ -505,7 +505,9 @@ export const ContentList = ({
                   headerContent={
                     useLatest
                       ? 'Repositories as of today'
-                      : `Repositories as of ${snapshotDate}`
+                      : `Repositories as of ${yyyyMMddFormat(
+                          new Date(snapshotDate)
+                        )}`
                   }
                   hasAutoWidth
                   minWidth="60rem"
--- a/src/Components/CreateImageWizard/steps/Snapshot/Snapshot.tsx
+++ b/src/Components/CreateImageWizard/steps/Snapshot/Snapshot.tsx
@@ -18,10 +18,6 @@ import {
   changeUseLatest,
   changeSnapshotDate,
 } from '../../../../store/wizardSlice';
-import {
-  parseYYYYMMDDToDate,
-  yyyyMMddFormat,
-} from '../../../../Utilities/time';
 import { isSnapshotDateValid } from '../../validators';
 
 export default function Snapshot() {
@@ -75,12 +71,10 @@ export default function Snapshot() {
               <DatePicker
                 id="snapshot-date-picker"
                 name="pick-snapshot-date"
-                value={snapshotDate}
+                value={snapshotDate ? snapshotDate.split('T')[0] : ''}
                 required
                 requiredDateOptions={{ isRequired: true }}
                 placeholder="YYYY-MM-DD"
-                dateParse={parseYYYYMMDDToDate}
-                dateFormat={yyyyMMddFormat}
                 validators={[
                   (date: Date) => {
                     if (!isSnapshotDateValid(date)) {
--- a/src/store/wizardSlice.ts
+++ b/src/store/wizardSlice.ts
@@ -663,7 +663,13 @@ export const wizardSlice = createSlice({
       state.snapshotting.useLatest = action.payload;
     },
     changeSnapshotDate: (state, action: PayloadAction<string>) => {
-      state.snapshotting.snapshotDate = action.payload;
+      const yyyyMMDDRegex = /^\d{4}-\d{2}-\d{2}$/;
+      const date = new Date(action.payload);
+      if (action.payload === '') {
+        state.snapshotting.snapshotDate = action.payload;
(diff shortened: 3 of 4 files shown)


xbhouse:
issue [here](https://issues.redhat.com/browse/HMS-4863)

croissanne (changes requested):
Thank you! 

Could you edit the commit message? The PR title is good. https://osbuild.org/docs/developer-guide/general/workflow#preferred-commit-message-format

src/Utilities/time.ts
```diff
@@ -2,9 +2,9 @@ export const parseYYYYMMDDToDate = (val: string) =>
   val ? new Date(`${val}T00:00:00`) : new Date('');
 
 export const yyyyMMddFormat = (date: Date) =>
-  `${date.getFullYear()}-${(date.getMonth() + 1)
+  `${date.getUTCFullYear()}-${(date.getUTCMonth() + 1)
```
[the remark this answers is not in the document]
xbhouse: i was seeing some weird behavior where i'd select a date but it was always showing the day before as selected 🤔 converting explicitly to UTC here fixed that

xbhouse:
@croissanne this is ready for another review when you have a chance!

also, you mentioned image-builder needing an update to move from YYYY-MM-DD to RFC3339 too? let me know if there is anywhere else that will need a change as well, tracking them all [here](https://issues.redhat.com/browse/HMS-4863)

ochosi:
Looks like this has been stale for a bit, but it also seems to me that this should still be addressed, no?

xbhouse:
@ochosi agreed, this lets us remove some tech debt from content-sources. i believe there are some changes needed in image-builder and maybe IQE tests too? i'm not sure which would need to be merged first. CC @croissanne

src/Utilities/time.ts
```diff
@@ -2,9 +2,9 @@ export const parseYYYYMMDDToDate = (val: string) =>
   val ? new Date(`${val}T00:00:00`) : new Date('');
 
 export const yyyyMMddFormat = (date: Date) =>
-  `${date.getFullYear()}-${(date.getMonth() + 1)
```
regexowl, line 5: The only problem I see is the same behaviour you've mentioned in https://github.com/osbuild/image-builder-frontend/pull/2526/files#r1815188998, but the other way around. Without the UTC the picked date seems accurate, but with UTC it's a always set as the day before. Could it be some weird time zoney thing? :thinking:

regexowl (commented):
Hey @xbhouse this looks great, thank you! pr_check is passing so IQE should be hopefully fine. @jrusz what do you think? Could changing the snapshot date format break something outside of pr_check? As for API, it looks like we're all set there: https://github.com/osbuild/image-builder/pull/1252

jrusz:
> Hey @xbhouse this looks great, thank you! pr_check is passing so IQE should be hopefully fine. @jrusz what do you think? Could changing the snapshot date format break something outside of pr_check? As for API, it looks like we're all set there: [osbuild/image-builder#1252](https://github.com/osbuild/image-builder/pull/1252)

We have a specific function to convert between the api and ui format in IQE plugin so it will most likely need some adjustment, that's going to be an easy change though

src/Utilities/time.ts
xbhouse: thank you for the review @regexowl ! 

ah i see what you mean, when i changed my system time to Brno i see the same issue. since dateParse and dateFormat have [defaults](https://www.patternfly.org/components/date-and-time/date-picker/#datepicker) that look similar to the helper functions, i removed those fields in the DatePicker and formatted the date for the value in the picker to just remove the timestamp. it looks ok now on my end for both Brno and New York timezones. let me know what you think 🤔

regexowl: Excellent, that did the trick! Thank you!

regexowl:
@xbhouse Could you please squash the commits and rebase? The changes look great :sparkles: 

@jrusz @tkoscieln This will be good to merge, could you please take a look at what changes will be needed and let me know when they're ready?

xbhouse:
> @xbhouse It seems this is not ready to merge yet. I tested it out and got a 500 code when trying to start the composes with a following error:
> 
> ```
> "errors": [
>     {
>       "detail": "Unable to build customizations: code=400, message=Snapshot date 2024-12-03T00:00:00.000Z is not in DateOnly (yyyy-mm-dd) format",
>       "title": "500"
>     }
>   ]
>   
> ```
> 
> @regexowl produced a fix for it [here in backend](https://github.com/osbuild/image-builder/pull/1429).

ooh ok, nice catch! thank you ❤️

regexowl:
The backend fix got merged! Trying to compose an image and all looks good so far. @tkoscieln could you please take a look at the IQE test adjustment?

tkoscieln:
@regexowl I tested it out, the compose works fine and I tried running the integration tests on it (as those are the ones that test the snapshotting) and they passed fine, so it seems this change is good to go.

regexowl:
@tkoscieln Excellent, thank you! I'd say we're ready to merge then.

regexowl (approved):
Thank you! :heart:

regexowl:
Thank you! Please let me know if anything seems amiss.

Issue #2467, opened 2024-09-25
The next button has a different vertical position on the repo screen


ondrejbudai:
![output-file](https://github.com/user-attachments/assets/fde8090e-a792-413a-91b1-c4c399b06225)

rissh:
Hey,
I want to give it a try and work on this issue.

PR #2527, opened 2024-10-22
Wizard: Ensure wizard footer remains visible in repo step

rissh:
- Improved the layout of the repository step in the wizard to prevent the "Next" button from being hidden behind a scroll.
- Adjusted the wizard size to ensure a better user experience without needing to scroll in the repository selection step.
- Ensured consistent button positioning across varying screen resolutions.
- Attached is a video demo showcasing how this change resolves it.
- This is my approach to solving the issue, and I'm happy to consider alternative suggestions if anyone has a different idea.

[Screencast from 2024-10-22 11-59-24.webm](https://github.com/user-attachments/assets/c93ce198-bc93-4b7a-8bc4-c87e5f9db26b)

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.scss
+++ b/src/Components/CreateImageWizard/CreateImageWizard.scss
@@ -76,16 +76,13 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
 }
 
 .pf-v5-c-wizard__main {
-    /* TO DO: This part of the code is responsible for the Wizard footer being fixed
-    at the bottom of the page. Unfortunately there's a new bug that's at least
-    partially caused by this styling: when zooming in or out while in Wizard,
-    the Wizard shrunks it's height to an unusable value.
+    flex: 1 1 auto;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+}
 
-    Temporarily commenting out this code, until we figure out how the handle
-    this bug.
-    */
-    // flex: 1 1 0
-    
-    // TO DO: Remove after the fix
-    min-height: 800px;
+.pf-v5-c-wizard__footer {
+    position: sticky;
+    bottom: 0;
 }


rissh:
Fixes #2467

ezr-ondrej:
Hi! This now is in fixed height, but it always scrolls for me even on zoomed out screen where there is a lot of room blank room, the buttons are hidden.

https://github.com/user-attachments/assets/65bc2f99-3ce7-48fc-97ee-3f78bb1c8f6f

rissh:
Hi! 
Thanks for your feedback. I've made the necessary adjustments to address the issue, and the buttons should now remain visible without unnecessary scrolling. 

I've also added a video for reference to show how it’s working fine now.

Let me know if you have any further suggestions!
[Screencast from 2024-10-23 14-56-46.webm](https://github.com/user-attachments/assets/fac96349-2d0c-40fd-a9fd-d5531308736f)

avitova (commented):
Thank you for looking into this. :partying_face: 
If I zoom out now, the button is at the bottom. However, when zooming in, now I have to scroll. Is that the intended behaviour?

[Screencast from 2024-10-24 11-02-46.webm](https://github.com/user-attachments/assets/5a366154-26f0-42d9-b766-8102499657fe)

rissh:
> Thank you for looking into this. 🥳 If I zoom out now, the button is at the bottom. However, when zooming in, now I have to scroll. Is that the intended behaviour?
> 
>  Screencast.from.2024-10-24.11-02-46.webm

Hey,
I’m aware of this behavior, and I kept it that way since Ondrej suggested the buttons should always be at the bottom when zoomed out. For zooming in, the buttons stay at the bottom up to 110%, but after that, you’ll need to scroll down.

If you'd prefer the buttons to always be visible regardless of zoom, I can make that change! The only thing holding me back is that if you zoom in too much, the buttons can become annoyingly large for users. But if we want to move forward with it, I’m happy to tweak it—just say the word! 😊

avitova:
Yes, I thought the scrolling problem was the initial reason for the [issue](https://github.com/osbuild/image-builder-frontend/issues/2467).
My first thoughts for solving this are a bit different approach, so I thought it could be interesting to share. We could also put the contents of the custom repositories step into something like a fixed-height size "flexbox" (not sure how to describe this). That way, it would not extend and not push the button down. But honestly, that is just my two cents and different view on things.

rissh:
> Yes, I thought the scrolling problem was the initial reason for the [issue](https://github.com/osbuild/image-builder-frontend/issues/2467). My first thoughts for solving this are a bit different approach, so I thought it could be interesting to share. We could also put the contents of the custom repositories step into something like a fixed-height size "flexbox" (not sure how to describe this). That way, it would not extend and not push the button down. But honestly, that is just my two cents and different view on things. Whatever works for you!

Okay, this could definitely be one approach to try. I’m happy to explore it—thanks for the suggestion! 😊

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -97,5 +103,16 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
...
+    justify-content: space-between;
+    min-height: 800px;  /* Always go with pixels */
+}
+
+.pf-v5-c-wizard__footer {
+    position: sticky;
+    bottom: 0;
+    padding: 1rem;
+    background-color: white;  // Ensure background covers content below
+    z-index: 10;  // Ensure the footer is above the content
```
ezr-ondrej, line 117: This is 100 by default:
according to [code](https://github.com/patternfly/patternfly/blob/7d6a2437f14461c32222e6a63ffdca1b122cf0ca/src/patternfly/components/Wizard/wizard.scss#L169) it's using `--pf-v5-global--ZIndex--xs` which has value `100` according to [docs](https://v5-archive.patternfly.org/developer-resources/global-css-variables/#component-variables)

Why are you overriding it to 10, which is less then default?

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -97,5 +103,16 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
...
+    flex-direction: column;
+    justify-content: space-between;
+    min-height: 800px;  /* Always go with pixels */
+}
+
+.pf-v5-c-wizard__footer {
+    position: sticky;
+    bottom: 0;
+    padding: 1rem;
+    background-color: white;  // Ensure background covers content below
```
ezr-ondrej, line 116: Was this an issue? this seem to be quite strict, PF supports changing themes, so I would avoid setting colors directly unless strictly required.

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -97,5 +103,16 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
...
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    min-height: 800px;  /* Always go with pixels */
+}
+
+.pf-v5-c-wizard__footer {
+    position: sticky;
+    bottom: 0;
+    padding: 1rem;
```
ezr-ondrej, line 115: Why are you adding the padding here?

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -97,5 +103,16 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
...
     // TO DO: Remove after the fix
-    min-height: 800px;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    min-height: 800px;  /* Always go with pixels */
+}
+
+.pf-v5-c-wizard__footer {
+    position: sticky;
```
ezr-ondrej, line 113: I'm not sure I get how this works with flex. But sounds like we can do one or the other, or how does this work? Can you elaborate a bit pls? I'm bit lost in css positioning :thinking:

ezr-ondrej (commented):
I'm still not sure how this is supposed to work, I'm lost in css positioning, could you elaborate a bit? I've left few inline comments on lines I'm mostly lost on.

src/Components/CreateImageWizard/CreateImageWizard.scss
rissh: I set z-index: 10 instead of the default 100 to ensure the footer stays above the content without conflicting with other UI elements that might have higher z-index values. I tested it with the default z-index: of 100, but it didn’t give us the expected result.

ezr-ondrej: Isn't higher z-index more on top?

rissh:
> I'm still not sure how this is supposed to work, I'm lost in css positioning, could you elaborate a bit? I've left few inline comments on lines I'm mostly lost on.

Sure, I'll go through each of your comments one by one and clarify the CSS positioning as best I can. I'll make any necessary adjustments along the way if needed. Thanks for the detailed feedback!

src/Components/CreateImageWizard/CreateImageWizard.scss
rissh: Good point! The direct color setting isn’t strictly necessary here, and I’ve removed it to keep it theme-compatible with PatternFly.

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -100,19 +94,13 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
     Temporarily commenting out this code, until we figure out how the handle
```
ezr-ondrej, line 94: Looks like we are re adding the flex, so this comment and the above TO-DO doesn't make sense any more. Should we remove it?

src/Components/CreateImageWizard/CreateImageWizard.scss
```diff
@@ -94,8 +94,13 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
...
+    flex: 1 1 auto;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
 }
+
+.pf-v5-c-wizard__footer {
+    position: sticky;
+    bottom: 0;
+}
```
ezr-ondrej, line 106: You're missing empty line at the end of the line, can you please add it for consitency? :)

rissh:
I have attached a video demonstrating the current behavior. The video showcases how the footer behaves with the latest changes and how it resolves the previous positioning issues.

Feel free to reach out if you have any questions or need further clarification! I'm still open to making any changes if needed—just let me know, and I'll be happy to update it.


https://github.com/user-attachments/assets/181dac96-45bd-4191-8b65-02ce8f67b83b

ezr-ondrej (commented):
Please remove the now unecessary comments and add an empty line at EOF, apart of that it looks good to me :+1:

src/Components/CreateImageWizard/CreateImageWizard.scss
rissh: Whoops! I wouldn't want to mess with consistency. 😅

src/Components/CreateImageWizard/CreateImageWizard.scss
rissh: Sure, No more confusing TODOs! 😄

avitova: If we need any followup in the future, let's just create a ticket right now, so that we do not forget though! :)

avitova:
Good, I am happy with this solution. Feel free to merge after Ondra approves. :)

ezr-ondrej (approved):
Thanks :+1: Works nicely on my end.

ezr-ondrej (changes requested):
Doesn't work on Firefox

regexowl:
Hmm, not sure what's up with the pr_check failure

/retest

ezr-ondrej:
Yeah, doesn't seem to break stuff :shrug: And I don't assume we'll get @rissh to rebase this, or continue working on it.
Can you admin push to squash? or squash on merge?

rissh:
> Tried to revisit this PR - not sure if anything changed directly in PF, but the sticky footer now seems to be working in Firefox as well most of the time (there are like two out of ten re-renders when the css seems to be ignored for some reason). I'd say let's merge this, what do you think @ezr-ondrej?
> 
> @rissh Would it be possible to squash the commits please? Sorry this took so long.

@regexowl Sure, I just rebased and squashed the commits. Let me know if anything else is needed. Thanks for following up!

regexowl:
Thank you! @tkoscieln is looking into the failing check, otherwise looks good to me :+1:

regexowl (approved):
This works great! We're no longer blocked on the failing check, so I'd say we can finally merge. Thank you so much for looking into this and also thanks for your patience :heart:

PR #2528, opened 2024-10-23
Wizard: only show compliance radio in preview

croissanne:
Let's expose compliance in preview while we're ironing out the creases.


mgold1234 (approved):
/lgtm

PR #2529, opened 2024-10-23
Wizard: fix editing RHEL 9 beta blueprints

croissanne:
It would default to RHEL 9 instead of the beta.


jrusz (commented):
The change seems ok, but it's failing on unit tests...

jrusz (approved):
Works, LGTM.

PR #2531, opened 2024-10-24
src/constants: fix FIRST_BOOT_SERVICE_DATA

jrusz:
There was a wrong file encoded as the custom-first-boot.service systemd unit. Fixing with the correct one forom aux/custom-first-boot.service.


jrusz:
Closing in favor of #2532

PR #2532, opened 2024-10-24
Fix first-boot script

ondrejbudai:
I don't think there's a value in having the unit encoded in base64. It looks fishy, and it's error prone. We discovered that the custom-first-boot.service file in aux was different than the base64- encoded string. There was nothing malicious, but we apparently forgot to rerun the Makefile target.
    
Instead, we can just put the raw script into constants.js and base64- encode it on the fly. The cost is close to zero.
    
This way, we can also get rid of the aux dir, and of the Makefile target. 🧹

I've also removed `osbuild-first-boot.target` because I'm pretty sure it doesn't exist.


jrusz (approved):
Seems to work just fine:
```
[jrusz@local-system /]$ systemctl status custom-first-boot.service
● custom-first-boot.service - Custom first boot script
     Loaded: loaded (/etc/systemd/system/custom-first-boot.service; enabled; preset: disabled)
     Active: active (exited) since Thu 2024-10-24 17:02:39 EDT; 43s ago
    Process: 1011 ExecStart=/usr/local/sbin/custom-first-boot (code=exited, status=0/SUCCESS)
    Process: 1017 ExecStartPost=/usr/bin/touch /var/local/.custom-first-boot-done (code=exited, status=0/SUCCESS)
   Main PID: 1011 (code=exited, status=0/SUCCESS)
        CPU: 6ms

Oct 24 17:02:39 local-system systemd[1]: Starting Custom first boot script...
Oct 24 17:02:39 local-system systemd[1]: Finished Custom first boot script.
[jrusz@local-system /]$ cat /etc/systemd/system/custom-first-boot.service 
[Unit]
Description=Custom first boot script
ConditionFileIsExecutable=/usr/local/sbin/custom-first-boot
ConditionPathExists=!/var/local/.custom-first-boot-done
Wants=network-online.target
After=network-online.target
After=osbuild-first-boot.service

[Service]
Type=oneshot
ExecStart=/usr/local/sbin/custom-first-boot
ExecStartPost=/usr/bin/touch /var/local/.custom-first-boot-done
RemainAfterExit=yes

[Install]
WantedBy=basic.target
WantedBy=multi-user.target
WantedBy=graphical.target
```
And the script created a folder like I wanted to.

lzap:
Apologies, we need to add a CI check that will verify that `make` does not perform any code change. I can add that.

lzap:
Ah I see the target from Makefile was dropped, then it is rendered useless:

https://github.com/osbuild/image-builder-frontend/pull/2533

PR #2537, opened 2024-10-28
'api: update content-sources spec

mgold1234:
run `npm run api` to update the content-sources spec


ezr-ondrej (approved):
The commit message doesn't follow guidelines, but not so important in generated files.
LGTM

PR #2541, opened 2024-10-29
Wizard: add Users step to the wizard <HMS-4902>

mgold1234:
this commit add new step 'Users' that enabled only in stage-preview
Screenshot 2024-10-29 at 14 42 57


mgold1234:
need to fix all the unit tests that fail

src/Components/CreateImageWizard/steps/Users/component/Empty.tsx
```diff
@@ -0,0 +1,28 @@
...
+  Button,
+  EmptyState,
+  EmptyStateFooter,
+  EmptyStateHeader,
+  EmptyStateIcon,
+  EmptyStateVariant,
+} from '@patternfly/react-core';
+import UserIcon from '@patternfly/react-icons/dist/esm/icons/user-icon';
+
+const Empty = () => {
```
lucasgarfield, line 13: Maybe `EmptyUserState` would be a bit more descriptive? `Empty` is fairly generic.

src/Components/CreateImageWizard/steps/Users/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import Empty from './component/Empty';
+
+const UsersStep = () => {
+  return (
+    <Form>
+      <Title headingLevel="h1" size="xl">
+        Users
+      </Title>
+      <Text>Add a user to your image.</Text>
+      <Empty />
```
lucasgarfield, line 14: This looks great. Just a hint for your next PR, you'll want to use conditional rendering to display the empty state only if there is no user.
```javascript
user ? <UserInfo /> : <UserEmptyState />
```

mgold1234: will do it in the next task thanks

avitova (commented):
Just a small comment, I do not know the reason for comments though.

src/test/Components/CreateImageWizard/steps/FileSystemConfiguration/FileSystemConfiguration.test.tsx
```diff
@@ -221,6 +222,7 @@ describe('Step File system configuration', () => {
     await selectGuestImage();
     await goToFileSystemConfigurationStep();
     await goToReviewStep();
+    await openAndDismissSaveAndBuildModal();
```
lucasgarfield, line 225: This seems really strange to me. I'd like to understand why we need to call this function in some tests, and remove it from other tests. I don't see how adding a new user step should have any impact on whether or not the save and build modal appears and needs to be dismissed. :thinking:

lucasgarfield (changes requested):

src/test/Components/CreateImageWizard/steps/FileSystemConfiguration/FileSystemConfiguration.test.tsx
regexowl: The `openAndDismissSaveAndBuildModal()` function should be called only once per test file, always in the first test that clicks on the "Create blueprint" button. When the popup gets dismissed, it won't appear again. We store something like "popupSeen" in the local storage. This should persist throughout test file.

That being said - I've tried to revert the changes and the tests finished perfectly fine. What problem was it causing @mgold1234? I mean it's functionally the same, this way we just click on "Create blueprint" and dismiss the popup "prematurely" as we don't interact with the create button further in "revisit step button on Review works".

lucasgarfield (approved):

regexowl:
Konflux checks seem to be back up

/retest

PR #2551, opened 2024-11-03
<WIP> User info

mgold1234:


mgold1234:
@regexowl forgot to delete the last commit that I tried yesterday to fix the requestMapper, can you please take a look now?

lucasgarfield:
@mgold1234 could you squash your commits? That would make this a bit easier to review.

src/store/wizardSlice.ts
```diff
@@ -87,6 +88,14 @@ export type wizardState = {
...
     snapshotDate: string;
   };
+  users: User[];
+  userInfo: {
+    userName: string;
+    password: string;
+    confirmPassword: string;
+    sshKey: string;
+    administrator: boolean;
+  };
```
regexowl, line 98: Already asked a bit offline - if I understand correctly the UI will support just one user for now and I understand that the API type of User doesn't include all the data we want to keep in the store.

I think we should either have a single user and later expand it to an array, or we could make the structure a bit more future proof and create an array right away:
```
users: UserWithAdditionalInfo[];
```

With the type looking something like:
```
export type UserWithAdditionalInfo = {
  name: User['name'];
  ssh_key: User['ssh_key'];
  randomOtherInfo: string;
  ...
```

What do you think?

src/store/wizardSlice.ts
```diff
@@ -179,6 +188,14 @@ export const initialState: wizardState = {
...
   },
   firstBoot: { script: '' },
+  users: [],
+  userInfo: {
+    userName: '',
+    password: '',
+    confirmPassword: '',
+    sshKey: '',
+    administrator: false,
+  },
```
regexowl, line 198: So here it would be just:
```
users: [],
```
as an initial value.

src/store/wizardSlice.ts
```diff
@@ -336,6 +353,29 @@ export const selectFirstBootScript = (state: RootState) => {
...
+  return state.wizard.userInfo?.confirmPassword;
+};
+export const selectUserSshKey = (state: RootState) => {
+  return state.wizard.userInfo?.sshKey;
+};
+
+export const selectUserAdministrator = (state: RootState) => {
+  return state.wizard.userInfo?.administrator;
+};
+
```
regexowl, line 378: We wouldn't need any of these apart `selectUsers` as we would only work with the entire Users object at once.

src/store/wizardSlice.ts
```diff
@@ -656,6 +696,22 @@ export const wizardSlice = createSlice({
...
+    setConfirmUserPassword: (state, action: PayloadAction<string>) => {
+      state.userInfo.confirmPassword = action.payload;
+    },
+    setUserSshKey: (state, action: PayloadAction<string>) => {
+      state.userInfo.sshKey = action.payload;
+    },
+    changeUserAdministrator: (state, action: PayloadAction<boolean>) => {
+      state.userInfo.administrator = action.payload;
+    },
```
regexowl, line 714: These would then be replaced by `addUser`, `editUser` and `removeUser`.

- `addUser` would dispatch a new entry, so it would just push a new object to `users` array.
- `removeUser` would filter through the `users` array and remove an item by probably a user name. For now it could also just rewrite the array with an empty one if we're working with just one user.
- `editUser` should be able to check whether the user is already there and update the values in the object. So it would probably try to find the correct user by name and then update existing values with new ones.

src/store/wizardSlice.ts
```diff
@@ -722,6 +778,11 @@ export const {
   changeBlueprintDescription,
   loadWizardState,
   setFirstBootScript,
+  setUserName,
+  setUserPassword,
+  setConfirmUserPassword,
+  setUserSshKey,
+  changeUserAdministrator,
```
regexowl, line 785: ... and here it would be just: add, edit and removeUser, would that make sense?

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -194,6 +195,27 @@ function commonRequestToState(
...
+          sshKey: request.customizations.users[0].ssh_key || '',
+          administrator: false,
+        }
+      : {
+          userName: '',
+          password: '',
+          confirmPassword: '',
+          sshKey: '',
+          administrator: false,
+        },
```
regexowl, line 218: Here we have some duplicity as well. I believe it would be better to either work with a single user, removing the `users` array, or better be prepared for multiple users in the future, removing the `userInfo` and mapping all values into `users[]` using a custom type that extends User type and includes all extra info we need.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -533,6 +556,41 @@ const getOpenscap = (state: RootState): OpenScap | undefined => {
...
+
+const getUsers = (state: RootState): User[] => {
+  const currentUser = getUserInfo(state);
+
+  if (currentUser) {
+    return [...state.wizard.users, currentUser];
+  }
+
+  return state.wizard.users;
+};
```
regexowl, line 592: It's a bit hard for me to get oriented with the `users` and `userInfo`, but what we want to do here is take the values from the store and shape them to fit the request.

So we should probably check if there are any users store in the `users` array in the store and if there are any items, we will want to iterate through them and for each push smth like this to the `users` array of the request:
```
{
  name: currentUser.name,
  password: currentUser.password || '',
  ssh_key: currentUser.ssh_key || '',
}
```

regexowl (commented):
This looks like a great start! I've managed to go through the `wizardStore` and `requestMapper` for now and wanted to add some comments as I think some changes to the base data structure could make things easier.

Let me know what you think and also let me know if I don't make sense :sweat_smile:

avitova (commented):
This looks amazing! I have some comments on this, I saw Klara already put hers, so it is not much more than she commented:)

lucasgarfield:
I'm not sure the SSH key regex is correct. For instance, consider this public key:
`ssh-ed25519 AAAAC3NzaC1lZDI1NTE5ZZZZIPz/IFF6bpPm2qPpDIqByoPne+KifkBWEaRZGvPaE5rY`
It fails validation:
image
The same key passes validation in Edge:
image

lucasgarfield:
Right now the user can _either_ set a password _or_ an SSH key.

Should we allow both to be set? I don’t know how common it would be, but if osbuild will build an image using both (disclaimer: have not confirmed this but believe it is possible) then I think we can allow it in the UI.

src/store/wizardSlice.ts
lucasgarfield: `Administrator` adds the user to sudoers. It's just a boolean.

ochosi:
Love how this looks and how it's coming along - thanks for working on this, @mgold1234 !

src/Components/CreateImageWizard/validators.ts
```diff
@@ -58,6 +58,54 @@ export const isSnapshotValid = (dateString: string) => {
   return !isNaN(date.getTime()) && isSnapshotDateValid(date);
 };
 
+export const isUserNameValid = (userName: string) => {
+  const isLengthValid =
+    userName !== undefined && userName.length >= 2 && userName.length <= 32;
+  const isPatternValid = /^[a-zA-Z0-9_.][a-zA-Z0-9_.-]*[a-zA-Z0-9_.$-]?$/.test(
```
lucasgarfield, line 64: I'm looking at the man pages for useradd and this is what it says:
>Usernames may contain only lower and upper case letters, digits, underscores, or dashes. They can end with a dollar sign. Dashes are not allowed at the beginning of the username. Fully numeric usernames and usernames . or .. are also disallowed. It is not recommended to use usernames beginning with . character as their home directories will be hidden in the ls output. Usernames may only be up to 32 characters long.

It didn't occur to me before to check how useradd works but I realized that's what gets called by the osbuild stage :facepalm: : https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.users

Let's just follow the man page rules exactly. So the minimum length should be 1. `..` and `.` would be special cases and forbidden.

lucasgarfield (changes requested):

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -133,6 +141,40 @@ export function useFirstBootValidation(): StepValidation {
...
+
+  const passwordConfirmMatchValid = isConfirmPasswordValid(
+    userPassword,
+    userConfirmPassword
+  );
+  const sshKeyValid = shouldValidateSshKey ? isSshKeyValid(userSshKey) : true;
+  return {
+    errors: {
+      userName: !userNameValid ? 'Invalid user name' : '',
+      userPassword: !passwordValid ? 'Invalid user password' : '',
```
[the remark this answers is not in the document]
lucasgarfield: I mentioned in a comment below that I think we should just follow the exact guidelines found in the useradd man pages. We could also use those here. Let's only show them when the username is invalid?
image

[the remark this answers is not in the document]
avitova: Okay, no problem. :)

avitova:
There is a bug now - if you fill in the user information, click next, visit the Users step again, and change some of the fields, all of the contents of the other fields get deleted. 

https://github.com/user-attachments/assets/2dde1f3d-58de-4961-98cb-6dbae95e2a2e

Other than that, it looks good, I think. :)

src/store/wizardSlice.ts
mgold1234: I added the userState structure to make it easy to fill every field and validate it 
```
  userState: {
    users: [],
    userInfo: {
      index: 0,
      name: '',
      password: '',
      confirmPassword: '',
      ssh_key: '',
      administrator: false,
    },
  },
```

src/store/wizardSlice.ts
mgold1234: I actually think that we need those fields to simplified the process

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -0,0 +1,168 @@
...
+import {
+  HookValidatedInput,
+  HookValidatedInputWithButton,
+  HookValidatedTextArea,
+} from '../../../ValidatedTextInput';
+
+const UserInfo = () => {
+  const dispatch = useAppDispatch();
+  const users = useAppSelector(selectUsers);
+  const [newUser, setNewUser] = useState<UserWithAdditionalInfo>({
```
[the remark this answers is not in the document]
mgold1234: I changed this behavior, thank you for mention that.

avitova:
> There is a bug now - if you fill in the user information, click next, visit the Users step again, and change some of the fields, all of the contents of the other fields get deleted.
> 
>  Screencast.From.2024-11-21.17-44-57.mp4 
> Other than that, it looks good, I think. :)

This still happens in case that I:
1) Fill in the user info.
2) Go to the next steps.
3) Return back, and try to change any of the fields.

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -0,0 +1,199 @@
...
+          stepValidation={stepValidation}
+          fieldName="userPassword"
+          togglePasswordVisibility={togglePasswordVisibility}
+          isPasswordVisible={isPasswordVisible}
+          isEmpty={userPassword === ''}
+        />
+        <FormHelperText>
+          <HelperText>
+            <HelperTextItem>
+              Can contain at least 3 different character classes (uppercase
```
regexowl, line 124: I guess we require three different character classes, I'd change the wording a bit.

```suggestion
              Must contain at least 3 different character classes (uppercase
```

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -0,0 +1,199 @@
...
+    };
+    // Call specific setter action if provided
+    if (setterAction) {
+      dispatch(setterAction(value));
+    }
+    // Optional additional logic
+    if (additionalLogic) {
+      additionalLogic(updatedUser);
+    }
+    // Determine if user exists by index, not name
```
regexowl, line 77: Why not name? Shouldn't the user name be unique? :thinking:

src/Components/CreateImageWizard/steps/Users/component/Empty.tsx
```diff
@@ -10,19 +10,23 @@ import {
...
+const EmptyUserState = ({ onAddUserClick }: EmptyUserStateProps) => {
   return (
     <EmptyState variant={EmptyStateVariant.lg}>
       <EmptyStateHeader
         icon={<EmptyStateIcon icon={UserIcon} />}
         headingLevel="h4"
       />
       <EmptyStateFooter>
-        <Button variant="secondary" onClick={() => {}}>
+        <Button variant="secondary" onClick={onAddUserClick}>
```
regexowl, line 24: Switching between the empty and user creation states could be handled in a similar way like FSC I think, we could introduce a helper into the users structure called `userMode` or smth like that and then switch what renders on the step based on a store value.

But this might also be a follow up.

src/store/wizardSlice.ts
```diff
@@ -87,6 +94,11 @@ export type wizardState = {
     useLatest: boolean;
     snapshotDate: string;
   };
+
+  userState: {
+    users: UserWithAdditionalInfo[];
+    userInfo: UserWithAdditionalInfo;
+  };
```
regexowl, line 101: What blocks us from doing just:
```suggestion
  users: UserWithAdditionalInfo[];
```
and accessing the users by their name value?

This way we have a complex structure that includes an array and then just one item of the array that is stored outside of it

regexowl (commented):
Added some questions, the step is looking really good :sparkles:

src/test/fixtures/editMode.ts
```diff
@@ -165,12 +166,20 @@ export const baseImageRequest: ImageRequest = {
...
+  },
+];
+
 export const baseCreateBlueprintRequest: CreateBlueprintRequest = {
   name: 'Red Velvet',
   description: '',
   distribution: RHEL_9,
   image_requests: [baseImageRequest],
-  customizations: {},
+  customizations: { users },
```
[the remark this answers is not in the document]
mgold1234: yes, removed that :)

src/store/wizardSlice.ts
mgold1234: it becomes very complicated to implement the change of every field. and like in this way, the verifaition of every filed stay simple

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
regexowl: Ah right, yeah, that makes sense :+1:

regexowl:
Okay, I've spent a bit too long trying to debug the behaviour we're seeing now (every new key input means a new user) and I'm still getting lost to be honest. Tried to think about it a bit, let me know if some of the things were already discussed/decided.

There are two modes we'll care about: creation and edit, starting with edit:
- if `users.length > 1` I'd propose we grey out the "Add user" button and show a text informing the user, that multi users are not supported for now rather than randomly picking one to make changes to
- if `users.length === 1` we should just parse the user object from the request via the requestMapper and pre-fill the inputs

For creation: 
- if `users.length === 0` it starts to get slightly more complicated with data management. Since we're currently dealing with only one user at any time we will need two actions - create an empty user structure with index 0 and edit already existing user, is that correct?
- for edit we currently don't need to check the index as it will edit the only existing user

What I also wonder - should we keep ephemeral data in the step state or do we want to push every keystroke to the store? Nothing wrong with the second option ig, but maybe we could also update the store from the states?

Was thinking something like:
```typescript
// creating a state
const [userName, setUsername] = useState('');

// updating the store
useEffect(() => {
	dispatch(editUser(<data-structure-that-includes-state-values>))
}, [userName])

// handling the changes
const handleNameChange = (_event: React.KeyboardEvent, value: string) => {
	setUserName(value);
};

// input element
<HookValidatedInput
	ariaLabel="blueprint user name"
	value={userName}
	onChange={(_event, value) => handleNameChange(_event, value)}
	placeholder="Enter username"
	fieldName="userName"
/>
```

or a dispatch button that would edit all data at the same time. Something like "Confirm user".

Would something like this even make sense? I need a sanity check. I'd like to simplify store operations, but dunno, my head hurts ;_;

regexowl:
Ok, gave it another look - thinking within what we already have, just trying to simplify things a bit. One day we'll need to have multi user support, but today is not that day so I'd say let's embrace it.

Is my thinking right? We have only one user per blueprint support. As mentioned in the comment above I'd ignore multi user blueprints and just grey out the add button with "Sorry, no multi user for you" message. Meaning we have two more options - one user in a blueprint or no users in a blueprint. Both mean that their `id === 0` as when we're creating a new user we start indexing from 0 and when we read a user from the request there is no id, meaning there's no reason to not start at 0 also.

So we can expect that the index we're currently working with is always 0. Focusing on `userName` only so I'm not jumping from one place to another this could probably mean something like:

```typescript
  const handleNameChange = (
    _e: React.FormEvent<HTMLInputElement>,
    value: string
  ) => {
    if (users.length === 1) { // in case we have exactly one user
      dispatch(setUserNameById(value)); // we can make a change directly to the userName field
    } else if (users.length === 0) { // in case we have no users
      dispatch(addUser(updatedUser)); // we will first dispatch an `addUser` function to initialize the user structure
    }
  };
```

BUT!! We already have a specialized "Add user" button that should get rendered only when there are no pre-existing users in the blueprint, so why not use that one? The `addUser()` which would add an empty user structure to the store could be simply:
```typescript
    addUser: (state) => {
      const emptyUser: UserWithAdditionalInfo = {
        id: 0,
        name: '',
        password: '',
        confirmPassword: '',
        ssh_key: '',
        administrator: false,
      };

      state.users.push(emptyUser);
    },
``` 

So a requestMapper looks at the request, sees there is exactly one user and on a User step skips the empty state directly to prefilled fields. Or requestMapper looks at the request and sees there is no user at all and renders an empty state on the Users step.

In both cases when we're in the user form, we can just edit the structure (either edit existing data or overwrite initial data via the same mechanism), this is where the `setUser<insert-a-field-name>ById` functions come in. Again focusing on name only:
```typescript
    setUserNameById: (state, action: PayloadAction<string>) => {
      state.users[0].name = action.payload; // we know there is no other index than 0
    },
```
and we can get the existing userName from the state like so:
```typescript
export const selectUserName = (state: RootState) => {
  return state.wizard.users.find((user) => user.id === 0).name;
};
```

What would that mean for the `UserInfo.tsx` fields then? The `handleNameChange` function would handle only the name change:
```typescript
  const handleNameChange = (
    _e: React.FormEvent<HTMLInputElement>,
    value: string
  ) => {
    dispatch(setUserNameById(value));
  };
```

And the field itself would look like this:
```typescript
      <FormGroup isRequired label="Username" fieldId="blueprint-user-name">
        <HookValidatedInput
          ariaLabel="blueprint user name"
          value={userName}
          placeholder="Enter username"
          onChange={(_e, value) => handleNameChange(_e, value)}
          fieldName="userName"
        />
      </FormGroup>
```

This still might not be the most optimal solution, there is a lot going on in the step. But maybe it could make things easier? What do you think? What could be possible setbacks?

mgold1234:
@regexowl wooo that's sounds really better way, I will give it a try, thank you again!!!!

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -448,7 +457,16 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 key="wizard-users"
                 isHidden={!isUsersEnabled}
                 footer={
-                  <CustomWizardFooter disableNext={false} optional={true} />
+                  <CustomWizardFooter
+                    disableNext={
+                      !(
+                        userName === '' &&
+                        userPassword === '' &&
+                        userSshKey === ''
```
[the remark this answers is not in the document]
mgold1234: I checked it in the userValidation - if there is password or ssh key

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
mgold1234: I remove this function

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -526,6 +534,24 @@ const getServices = (state: RootState): Services | undefined => {
...
 
+const getUsers = (state: RootState): User[] | undefined => {
+  const users = selectUsers(state);
+  if (!users || users.length === 0) {
+    return undefined;
+  }
+  return users.map((user) => ({
+    name: user.name,
+    ...(user.password !== undefined && { password: user.password }),
+    ...(user.confirmPassword !== undefined && {
```
lucasgarfield, line 545: I'm a bit confused here. This function, `getUsers` gets called by `getCustomizations` which in turn is called by `mapRequestFromState`. It is responsible for generating the request that is sent to the image-builder API.

It returns an array of `Users`. A `User` has the following fields:

```javascript
export type User = {
  name: string;
  /** List of groups to add the user to. The 'wheel' group should be added explicitly, as the
    default value is empty.
     */
  groups?: string[];
  ssh_key?: string;
  /** Plaintext passwords are also supported, they will be hashed and stored using the SHA-512 algorithm.
    The password is never returned in the response.
    Empty string can be used to remove the password during update but only with ssh_key set.
     */
  password?: string;
};
```

It does not have a `confirmPassword` field or an `administrator` field.

Note that to make a user an administrator, you need to add "wheel" to the user's `groups` field as shown in this image-builder API example:
https://github.com/osbuild/image-builder/blob/9bcc1043e0b661bb34dbbe9929e29068e787d029/internal/v1/api.yaml#L1929

Unfortunately I don't think Typescript can flag this because of how its type compatibility works:
https://www.typescriptlang.org/docs/handbook/type-compatibility.html

lucasgarfield (changes requested):
Getting closer! Nice work so far, this is really coming along. I do still think we can simplify this a bit and I've left some comments describing how.

Regarding handling multiple users - this is something that we will need to do relatively soon as it is important for parity with Cockpit Composer. While generally I support the YAGNI principle ("you aren't going to need it"), in this case we can definitely say that we will need it and I would go ahead and engineer it in now by storing the users in an array and having the actions take the index of the user as part of the payload. When we _dispatch_ the actions, we can hardcode a 0 index into the payload. But I think that ought to happen at the dispatch site.

src/Components/CreateImageWizard/utilities/requestMapper.ts
mgold1234: if I remove the confirmPassword and Administrator it will not appear in the post request, when I create the blueprint 
Screenshot 2024-12-03 at 22 30 38

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -194,12 +194,20 @@ function commonRequestToState(
...
       blueprintDescription: request.description || '',
     },
+    users:
+      request.customizations?.users?.map((user, id) => ({
+        id,
+        name: user.name,
+        password: user?.password || '',
+        ssh_key: user?.ssh_key || '',
+        confirmPassword: undefined,
+        administrator: false,
```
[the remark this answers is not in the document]
mgold1234: why 'wheel' and not every group name?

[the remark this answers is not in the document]
mgold1234: I changed it to ' isAdministrator: user.groups?.includes('wheel') || false,'
if we will not find the 'wheel' it will be false

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -526,6 +534,24 @@ const getServices = (state: RootState): Services | undefined => {
...
 };
 
+const getUsers = (state: RootState): User[] | undefined => {
+  const users = selectUsers(state);
+  if (!users || users.length === 0) {
+    return undefined;
+  }
+  return users.map((user) => ({
+    name: user.name,
+    ...(user.password !== undefined && { password: user.password }),
```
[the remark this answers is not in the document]
mgold1234: I removed the user.password. becuase the password will be hashed and stored using the SHA-512 algorithm and the password is never returned in the response.

avitova (commented):
I found some issues, that I think should not be happening in the first version of this feature:
a) If I fill in name and SSH key, go to the next step, go back and make the ssh key invalid, it is still possible to create the user (using left side navigation).
b) If you fill in username, and both fields for password correctly, proceed to the next step, but then go back, delete both password fields, and put invalid password, and only to one of the password fields, you would still be able to create the user if you use the navigation on the left.
Basically, both are related with the validation not happening during the "Create blueprint" button onClick action.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -137,3 +233,60 @@ export const ValidatedTextInput = ({
     </>
   );
 };
+
+export const HookValidatedTextArea = ({
```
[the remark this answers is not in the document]
mgold1234: HookValidatedTextAreaPropTypes extend the TextAreaProps and its different from TextInputProps 
so I am not sure we can reuser here

src/store/wizardSlice.ts
```diff
@@ -719,6 +776,60 @@ export const wizardSlice = createSlice({
...
+      state.users[action.payload.index].ssh_key = action.payload.sshKey;
+    },
+
+    setUserAdministratorByIndex: (
+      state,
+      action: PayloadAction<UserAdministratorPayload>
+    ) => {
+      state.users[action.payload.index].isAdministrator =
+        action.payload.isAdministrator;
+      if (action.payload.isAdministrator) {
```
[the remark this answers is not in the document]
lucasgarfield: I realize that a big part of the reason this is such a complicated conditional is because these fields can be undefined. What if we used the Typescript utility function `Required` on UserWithAdditionalInfo so that these fields can't be undefined?

https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype

lucasgarfield (changes requested):

src/store/wizardSlice.ts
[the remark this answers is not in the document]
mgold1234: I like the solution to force groups field to be required, 
I fixed that with 
  extends Omit<User, 'groups'>,
    Required<Pick<User, 'groups'>>
 Omit - removes groups field temporarily.
Pick- picks only the groups field and makes it required with Required.
all other field stay as they are.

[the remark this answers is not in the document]
mgold1234: maye the simplest solution will be at the backend side? to change groups to to be mandatory?

lucasgarfield:
Validation error appears immediately upon creating a user:
image

lucasgarfield:
Password validation does not match validation message, according to validation message this is a valid password:
image

I think password validation is kind of a pain, especially since I would *hope* most users immediately change the password. We should see if a simple password (e.g. "pw") that violates these rules results in a successful build/bootable image. 

Generally for validation I think we should always balance ease of use (minimize drop off for first time users) while making sure we don't build images that can't boot.

src/store/wizardSlice.ts
```diff
@@ -719,6 +778,53 @@ export const wizardSlice = createSlice({
...
+        state.users?.[action.payload.index].groups?.push('wheel');
+      } else {
+        state.users[0].groups = state.users?.[0].groups?.filter(
```
lucasgarfield, line 823: I don't think we need to hard code a 0 index here since the index comes in with the payload.

lucasgarfield:
@mgold1234 please review my additional commits and squash them as you see fit! Also please update the commit message/PR description to be accurate.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
lucasgarfield, line 1: We'll need to add a test that confirms the request to image-builder is generated correctly for create and edit mode. You can use another step as inspiration. 

For instance, for locale, create mode tests:
https://github.com/osbuild/image-builder-frontend/blob/ae795c3b6a9f93a5dd301b7adcf70c76e6b87ef6/src/test/Components/CreateImageWizard/steps/Locale/Locale.test.tsx#L164

Edit mode tests:
https://github.com/osbuild/image-builder-frontend/blob/ae795c3b6a9f93a5dd301b7adcf70c76e6b87ef6/src/test/Components/CreateImageWizard/steps/Locale/Locale.test.tsx#L236

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -51,16 +51,40 @@ interface HookValidatedTextInputWithButtonPropTypes
...
+  };
+
+  return {
+    handleBlur,
+    getValidationState,
+  };
+};
+
+const renderValidationHelperText = (
+  validationState: 'error' | 'default' | 'success',
```
lucasgarfield, line 78: There are three possible validation states but only one of them ('error') is actually used. This seems like a good time to apply the "You aren't going to need it" (YAGNI) principle and simplify the component. Why not make this something like ErrorHelperText? Then the only prop needed would be errorMessage.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -51,16 +51,40 @@ interface HookValidatedTextInputWithButtonPropTypes
...
-  isPristine: boolean,
-  isEmpty: boolean | undefined,
-  stepValidation: StepValidation,
-  fieldName: string
-): 'default' | 'error' | 'success' => {
-  if (isEmpty) return 'default';
-  if (isPristine) return 'default';
-  if (stepValidation.errors[fieldName] === 'default') return 'default';
-  return stepValidation.errors[fieldName] ? 'error' : 'success';
+const useValidatedInput = (value: string, isEmpty?: boolean) => {
```
lucasgarfield, line 54: I think a hook is overkill here. Why not just use a function?

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -51,16 +51,40 @@ interface HookValidatedTextInputWithButtonPropTypes
...
+  const [isPristine, setIsPristine] = useState(!value);
+
+  const handleBlur = () => {
```
lucasgarfield, line 57: Why put handleBlur inside this hook?

mgold1234:
will split the work on that pr to small prs

PR #2563, opened 2024-11-05
Wizard: Add min size check

avitova:
Partition's minsize is uint64 in backend, which maps to `any`. Therefore, it is sometimes possible that we let in different types there if we import the blueprint with such value.
This should fix it, let me know what you think. We can also put an empty string as a default.

### DIFF
--- a/src/Components/CreateImageWizard/utilities/parseSizeUnit.ts
+++ b/src/Components/CreateImageWizard/utilities/parseSizeUnit.ts
@@ -2,7 +2,7 @@ import { UNIT_GIB, UNIT_KIB, UNIT_MIB } from '../../../constants';
 import { Units } from '../steps/FileSystem/FileSystemTable';
 
 export const parseSizeUnit = (bytesize: string) => {
-  let size;
+  let size: number = 10;
   let unit: Units = 'GiB';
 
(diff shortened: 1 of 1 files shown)


ezr-ondrej (commented):
I'd probably try to fix it during the import (in the mapper) as this will run a dispatch on render. Not a big deal, but it's a bit weird to do IMO

avitova:
I agree, there are better solutions than this. We can solve this in the import part, but the fact that we pass a value to the component and do not check its type is a bit worrying. So I still think we should aim to fix this even here so that we do not give a chance to any upcoming bugs. Or do you think it would not be a problem in the future?

ezr-ondrej:
The fact it can get feeded bad value is worrying for sure I'm not denying that :) I was just brainstorming what we can do about it... We could even solve it in the store itself IMO, so there would be no chance to keep such a value.

avitova:
Okay, I talked about this with Klara, and we will try a different approach here. I will make sure to fix this in a sensible way.
If someone creates a blueprint using API, with min_size that does not get parsed by `parseSizeUnit()`, we would face the same problem, but we can not use this solution. ^ The solution is on the way.

PR #2571, opened 2024-11-10
Wizard: add support for other input type in`HookValidatedInput`

mgold1234:
adds `type` prop to `HookValidatedInput` with a default of 'text', to be able to support other types input like 'password'


regexowl (approved):
Looks good!

PR #2575, opened 2024-11-11
Wizard: Basic Timezone step (HMS-5009)

regexowl:
This adds bare bones Timezone step with appropriate file structure and basic unit tests.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -27,6 +27,7 @@ import SnapshotStep from './steps/Snapshot';
 import Aws from './steps/TargetEnvironment/Aws';
 import Azure from './steps/TargetEnvironment/Azure';
 import Gcp from './steps/TargetEnvironment/Gcp';
+import TimezoneStep from './steps/Timezone';
 import UsersStep from './steps/Users';
 import {
   useFilesystemValidation,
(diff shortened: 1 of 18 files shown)


avitova (approved):
As far as I understand, there is supposed to be a placeholder on that step?:)

regexowl:
> As far as I understand, there is supposed to be a placeholder on that step?:)

Exactly, this is just a bare bone step, timezone drop down and ntp server input / drop down will come later.

mgold1234:
hope its ok that I enable the flag for dew second just to see this step locally

regexowl:
> it looks good, I wander if for next step we should see dropdown with different timezone?

Yup, that's the plan!

> hope its ok that I enable the flag for dew second just to see this step locally

Sure! I've kept it disabled, but since this isn't merged yet we can toggle the flag freely.

lucasgarfield:
Can we set the Unleash flag up so it is only exposed in stage-preview, similar to what we're doing for the Users step?

regexowl:
> Can we set the Unleash flag up so it is only exposed in stage-preview, similar to what we're doing for the Users step?

The constraint is there, I kept the flag disabled by default - not sure what the plan is when the changes get released to stage, should the empty, not working step be visible?

Created a disabled flag with the constraint for prod also.

lucasgarfield:
pr_check failing, going to admin merge
`18:11:48 2024-11-19 18:11:48 [   ERROR] [           thread-89] [clowdapp/rhsm-system-profile-bridge] timed out waiting for resource to be ready, details:   clowdapp/rhsm-system-profile-bridge not ready, status conditions:`

lucasgarfield (approved):
Thanks!

PR #2576, opened 2024-11-12
Wizard: handle dos scripts in first boot (HMS-4966)

croissanne:
Scripts in which lines are terminated by CRLF break the first boot service, so let's make sure the line termination is correct.


regexowl (approved):
Looks good!

regexowl:
> It looks good. I did not test manually due to the outage. Just a heads-up - the test in either this PR or in #2575 will have to be changed, as far as I understand, so let's just make sure not to merge both at once.

Good point, this PR has higher priority. I can rebase https://github.com/osbuild/image-builder-frontend/pull/2575 when it's merged.

PR #2577, opened 2024-11-12
Wizard: Add bytes to min_size partitioning <HMS-4981>

avitova:
Some people might define a minsize of their blueprints that is not a certain number of KiB, MiB, GiB. In case we import such blueprint, we would prefill "undefined" string, which is not a valid value.
I think that the same would eventually happen if we created a blueprint using API, I do not think we check for minimal value.
Because we probably do not want to suggest that users should create blueprints with partitioning with minimal size in bytes, I did the following:
If the value is such, that we cannot display it in KiB, MiB or GiB, we have a new unit - Bytes - in the dropdown, and if it is selected, the text field is not editable. In case users click on different units, they can change it, but if they click back on 'B', they have their initial value in bytes again. Does that make sense?
In case we parse an incorrect value from the imported file, we default to 10 GiB. This would be good to mention in the future, but that is not in the blueprints MVP.
image

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedTextInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedTextInput.tsx
@@ -34,6 +34,7 @@ export const HookValidatedInput = ({
   ouiaId,
   ariaLabel,
   value,
+  isDisabled,
   placeholder,
   onChange,
   stepValidation,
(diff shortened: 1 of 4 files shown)


regexowl:
Looks good to me!

I'm thinking about reverting to original value when re-selecting bytes, but I guess it makes sense since we probably want to allow the original value if the user changes their mind about selecting different units :thinking: 

The undefined is gone which is really nice. Also - great catch! :heart:

regexowl (approved):

avitova:
@regexowl Yeah. Exactly, I thought it would be kinda annoying if users were not able to choose the previous value. They would have to exit the wizard, open it again, and edit all of the stuff they previously changed. That just seems like unwanted behaviour, almost like a bug.

PR #2582, opened 2024-11-14
Wizard: Prefetch the list of activation keys on Image output step

regexowl:
This prefetches the list of activation keys upon entering the Wizard, making the loading times on the Registration step better for the user.

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import {
   Button,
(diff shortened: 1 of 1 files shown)


regexowl:
This is just a small part of activation keys prefetching, another improvement would be prefetching info about the selected activation key as well, which would required moving things between steps a bit. Which are changes that are a part of https://github.com/osbuild/image-builder-frontend/pull/2504

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
```diff
@@ -49,6 +50,11 @@ const TargetEnvironment = () => {
 
   const dispatch = useAppDispatch();
   const prefetchSources = provisioningApi.usePrefetch('getSourceList');
+  const prefetchActivationKeys = rhsmApi.usePrefetch('listActivationKeys');
+
+  useEffect(() => {
```
lucasgarfield, line 55: I was trying to figure out whether we needed a useEffect or not for this. The RTKQ docs seem to use a similar pattern for this scenario so I think it makes sense!
https://redux-toolkit.js.org/rtk-query/usage/prefetching#recipe-prefetch-immediately

lucasgarfield (approved):
lgtm, thank you!

ochosi:
Weeee, looking forward to this landing! 😍

PR #2595, opened 2024-11-18
deps: Revert frontend-components bump, install from zero

regexowl:
This reverts #2556 and installs all dependencies from complete zero so there are no leftovers.


croissanne (approved):
thanks!

PR #2597, opened 2024-11-18
Rhel 10 beta (HMS-5011)

croissanne:
Adds rhel 10 beta support :)

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/ReleaseSelect.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/ReleaseSelect.tsx
@@ -16,6 +16,7 @@ import {
   RHEL_9_BETA,
   RHEL_9_FULL_SUPPORT,
   RHEL_9_MAINTENANCE_SUPPORT,
+  RHEL_10_BETA,
 } from '../../../../constants';
 import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
 import { Distributions } from '../../../../store/imageBuilderApi';
@@ -45,7 +46,7 @@ const ReleaseSelect = () => {
   };
 
   const setDescription = (key: Distributions) => {
-    if (key === RHEL_9_BETA) {
+    if (key === RHEL_9_BETA || key === RHEL_10_BETA) {
       return '';
     }
 
--- a/src/Components/CreateImageWizard/steps/Oscap/Oscap.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/Oscap.tsx
@@ -18,7 +18,12 @@ import { v4 as uuidv4 } from 'uuid';
 
 import OscapProfileInformation from './OscapProfileInformation';
 
-import { RHEL_9_BETA, RHEL_9 } from '../../../../constants';
+import {
+  RHEL_9_BETA,
+  RHEL_9,
+  RHEL_10_BETA,
+  RHEL_10,
+} from '../../../../constants';
 import { usePoliciesQuery, PolicyRead } from '../../../../store/complianceApi';
 import { useAppDispatch, useAppSelector } from '../../../../store/hooks';
 import {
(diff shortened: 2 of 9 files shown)


regexowl:
Changes from `npm run api` will need to be added as well I think.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -133,7 +134,11 @@ const convertFilesystemToPartition = (filesystem: Filesystem): Partition => {
  * @param distribution blueprint distribution
  */
 const getLatestRelease = (distribution: Distributions) => {
-  return distribution === RHEL_9_BETA
+  return distribution === RHEL_10_BETA
+    ? (RHEL_10_BETA as Distributions)
+    : distribution.startswith('rhel-10')
+    ? RHEL_10
```
regexowl, line 140: There's no RHEL 10 option in API, should there be?

src/Components/CreateImageWizard/steps/Oscap/Oscap.tsx
```diff
@@ -460,6 +465,8 @@ const ComplianceSelectOption = ({ policy }: ComplianceSelectOptionPropType) => {
 // so just use the ones from the major release.
 export const removeBetaFromRelease = (dist: Distributions): Distributions => {
   switch (dist) {
+    case RHEL_10_BETA:
+      return RHEL_10;
```
regexowl, line 469: Same here, no RHEL_10 in Distributions.

croissanne: ugh right, hmm how's that going to work :confused:

src/Components/CreateImageWizard/utilities/requestMapper.ts
croissanne: right I see what you mean, hmmm

src/Components/CreateImageWizard/steps/Oscap/Oscap.tsx
croissanne: actually I think we can just return the constant and cast it?
bit hacky but should work until rhel 10 actually exists.

regexowl:
And `ImagesTable/Release.tsx` will need `'rhel-10-beta': 'RHEL 10 Beta',` added as well.

regexowl:
Hmm, repo filtering will need and update for RHEL 10

image

Oscap is probably also confused by the RHEL 10

image

regexowl:
Tried to create a blueprint

image

src/Components/CreateImageWizard/steps/ImageOutput/ReleaseSelect.tsx
```diff
@@ -45,7 +46,7 @@ const ReleaseSelect = () => {
   };
 
   const setDescription = (key: Distributions) => {
-    if (key === RHEL_9_BETA) {
+    if (key === RHEL_9_BETA || key == RHEL_10_BETA) {
```
regexowl, line 49:
```suggestion
    if (key === RHEL_9_BETA || key === RHEL_10_BETA) {
```

src/Components/ImagesTable/Release.tsx
```diff
@@ -26,6 +26,7 @@ const Release = ({ release }: ReleaseProps) => {
     'rhel-93': 'RHEL 9.3',
     'rhel-94': 'RHEL 9.4',
     'rhel-10-nightly': 'RHEL 10',
+    'rhel-10-beta': 'RHEL 10 Beta',
```
avitova, line 29: Amazing, I just ran into this problem after updating the image builder API. Thanks for addressing this!:)

regexowl (commented):
Looks good, the only problem is the custom repos not being filtered by release for RHEL 10. 

OpenSCAP doesn't currently work for RHEL 10, but it might not be a blocker as it's just a "missing" cutomization option that could be added in a follow up.

regexowl:
We could also probably hide the repos step for RHEL 10 Beta or add some kind of alert that it's "under construction" if this needs to go out quickly.

croissanne:
konflux fails because rhsm is down.

PR #2599, opened 2024-11-19
Wizard: Update description fetching for recommended packages (HMS-4950)

regexowl:
contentSources updated the `searchRpm` call to allow passing an array of package names as an argument. This allows us to fetch all package recommendations' descriptions at once instead of a separate call for each of the recommendations.

The updates contentSources API is currently available in stage.


ochosi:
Yay, looking forward to this!

lucasgarfield (approved):
Excellent! Thank you!

PR #2600, opened 2024-11-19
deps: Bump PatternFly dependencies to version 5.4.2

regexowl:
While trying to resolve the release blocking issue PF versions were reverted to 5.4.1 which was one patch version larger jump than needed.

This uses the latest 5.4.2 release, we should stick to it and not update to version 6.x before the platform migrates to avoid possible problems.


regexowl:
Closing this one, the bumps are better addressed one by one.

PR #2605, opened 2024-11-20
Konflux: Compare rpms-signature-scan with generated pipeline

regexowl:
I've noticed there was an auto-generarated rpms-signature-scan for the Konflux pipelines and it was disabled as it didn't work at the time of migration: https://github.com/osbuild/image-builder-frontend/pull/2524/commits/985c5d2b0f0add8b01fcda69180e2a94d608f326

Comparing it to what we have now, there are some small changes. Not sure if those are completely necessary, but thought they might alleviate hypothetical unpredictable behaviour in the future. :shrug: 

This PR is intentionally split into two commits, one just moves the rpms-signature-scan block lower in the pipeline and the second includes changes to the blocks, so it's hopefully nicer to review.

### DIFF
--- a/.tekton/image-builder-frontend-pull-request.yaml
+++ b/.tekton/image-builder-frontend-pull-request.yaml
@@ -593,11 +593,11 @@ spec:
     - name: rpms-signature-scan
       params:
       - name: image-digest
-        value: $(tasks.build-container.results.IMAGE_DIGEST)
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Cool! thank you for that dilligency! :orange_heart:

Note: I'd really love to figure out a way how we don't need to keep those pipelines defined directly in the repo, it's not very DRY and prone to not getting latest changes.

PR #2606, opened 2024-11-20
Wizard: Fix package recommendations

regexowl:
Previously the Wizard crashed when looking for recommendations for packages that had none.

How to reproduce the crash:
- go to the Packages step
- search for and select `tmux` package

After fix:
- when selecting a `tmux` package or any other package without recommendations, the recommendations stay empty, but the Wizard doesn't crash


jrusz (approved):
I can no longer reproduce the bug with this change, so LGTM :)

PR #2613, opened 2024-11-22
Wizard: TIMEZONE 1 - Timezone drop down and NTP servers input (HMS-5028)

regexowl:
This adds a functional time zones drop down and a "chipping" input for NTP servers.

~~Let's take this as a preview. Still need to figure out, what to do with the `Property 'supportedValuesOf' does not exist on type 'typeof Intl'.` error, AFAIK the type should have been updated since Typescript 5.1 which we've already left far behind. Any tips on aternative ways of getting a complete list of timezones welcome.~~

Brought to you by black tea and '80s Japanese city pop.

TO DO in follow ups:
- NTP servers validation probably
- adding timezone and NTP servers to the on prem -> hosted import mapper

image

Typeahed select:
image

Filled in:
image

Split into four commits:
1. basic data structure
2. drop down and input
3. timezone expandable on the Review step
4. tests


mgold1234:
@regexowl can you please look on this 
```
ERROR in ./src/Components/CreateImageWizard/steps/Timezone/components/TimezoneDropDown.tsx:22:31
TS2339: Property 'supportedValuesOf' does not exist on type 'typeof Intl'.
    20 | } from '../../../../../store/wizardSlice';
    21 |
  > 22 | const defaultTimezones = Intl.supportedValuesOf('timeZone');
       |                               ^^^^^^^^^^^^^^^^^
    23 |
    24 | const TimezoneDropDown = () => {
    25 |   const timezone = useAppSelector(selectTimezone)
```;

**Comment by mgold1234** [comment]
@regexowl it is looking very good, want to ask is there any validation to check the NLP server field? I tried to check and didnt see any validation

**Comment by regexowl** [comment]
> @regexowl it is looking very good, want to ask is there any validation to check the NLP server field? I tried to check and didnt see any validation

Correct, no validation of NTP servers for now. I've mentioned it in description - I want to leave validation and including timezone values in on prem mapper for a follow up.

**Comment by lucasgarfield** [comment]
Extremely good stuff here.

One small red flag for me is that we have to provide instructions about how to enter the NTP server (`Confirm the NTP server by pressing space, comma or enter.`). Ideally we wouldn’t have to tell the user what to do, they would just be able to figure it out.

If we keep it as is, I think we also might want to add the NTP server on-blur (although that could be annoying too, depends on the users workflow I guess).

Would you be willing to consider using a separate input field with a ‘+’ button that would add the NTP server (if valid) to a text area below? The text area below could still use chips if you want.

image

File: src/Components/CreateImageWizard/steps/Timezone/components/TimezoneDropDown.tsx [hunk 1]
```diff
@@ -1,7 +1,152 @@
...
+} from '@patternfly/react-core';
+import TimesIcon from '@patternfly/react-icons/dist/esm/icons/times-icon';
+
+import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
+import {
+  changeTimezone,
+  selectTimezone,
+} from '../../../../../store/wizardSlice';
+
+const defaultTimezones = Intl.supportedValuesOf('timeZone');
```
lucasgarfield, line 23: I was wondering how you would populate the list of timezones! I'm not sure this is the right approach though. If you look at the blueprint reference in our documentation, there is some guidance:

>The values supported by timezone can be listed by running the command:
>```bash
>$ timedatectl list-timezones
>```
https://osbuild.org/docs/user-guide/blueprint-reference#timezone

I cross referenced this with the output of Intl.suppertedValuesOf('timeZone') and they are in fact different!

lucasgarfield: We might need to consult with some of the backend folks about this to be sure about the best way to do this.

The simplest option would be to just run `timedatectl` and put the output into an array we keep in `constants.ts`. But even then, could that be different for different versions of RHEL? Seems a bit excessive to add an endpoint for this to image-builder.

I've been thinking through how to figure this out ourselves and will just put some of my investigation here since the approach might be helpful for future customizations where we have questions like this.

One thing you can do to get a better idea about how these customizations are actually used during the build is to check the osbuild stages.

https://github.com/osbuild/osbuild/blob/f4b899873b1f108edbf12d81ee5dbed037238a7e/stages/org.osbuild.timezone

So here's how the timezone actually gets added in the osbuild stage:
```python
subprocess.run(["systemd-firstboot", f"--root={tree}", f"--timezone={zone}"], check=True)
```

If you run `man systemd-firstboot` there is some guidance:
image

Then if you do some additional research, you'll realize that `timedatectl` is installed alongside `systemd`. So it makes sense that we use `timedatectl` as the source of truth for the timezones.

lucasgarfield: Quick example of how the two outputs are different - `timedatectl list-timezones` has a 'Zulu' timezone. Intl.supportedValuesOf('timeZone') does not.

lucasgarfield:
Overall I think this looks really good. Let's sort out the UX and the list of timezones and I'll give it a second pass, but I don't anticipate requesting any other major changes! :+1:

regexowl:
>One small red flag for me is that we have to provide instructions about how to enter the NTP server (Confirm the NTP server by pressing space, comma or enter.). Ideally we wouldn’t have to tell the user what to do, they would just be able to figure it out.
>
>If we keep it as is, I think we also might want to add the NTP server on-blur (although that could be annoying too, depends on the users workflow I guess).
>
>Would you be willing to consider using a separate input field with a ‘+’ button that would add the NTP server (if valid) to a text area below? The text area below could still use chips if you want.

I see, thanks for the mock! I've updated the way NTP servers get added. Left the changes in a separate commit so it's easier to review.

image

src/Components/CreateImageWizard/steps/Timezone/components/TimezoneDropDown.tsx
regexowl: Oh no... started a thread on the main channel. Looks like the lists change based on release

regexowl:
Alright, the list of timezones generated with `Intl.supportedValuesOf('timeZone')` was replaced by tzdata-2024b. :scissors:

regexowl:
Although I guess we could add some kind of a script that would fetch the list directly from the python/tzdata repo and format it into an array :thinking: Can we do that? Fetch the list directly from the repo?

lucasgarfield:
Love the new method of adding NTP servers. We can re-use that pattern for things like kernel args and services, too!

I would even drop the "Added NTP servers" text, imho it is obvious, but feel free to push back against that if you disagree.

Slight issue on alignment on review step:
image

regexowl:
>I would even drop the "Added NTP servers" text, imho it is obvious, but feel free to push back against that if you disagree.

Sure! I'm no set on it. I kinda liked the grey background that comes with it being a category, but it looks just fine without it. Without hacking the styles in we can decide between category name + grey background and no name + no background.

image

>Slight issue on alignment on review step

It was the missing version slug, fixed!

>Hm, also seems like there is some kind of validation bug, I can't create a blueprint

Interesting, do you have steps to reproduce it? I've tried creating blueprints both with and without timezone values set and all seems to be working:

image

lucasgarfield (approved):
Heck yes!

PR #2614, opened 2024-11-22
Blueprints: Add flag to track on prem import usage

avitova:
The flag can be used later to track the usage of on prem import feature.

### DIFF
--- a/src/Components/Blueprints/helpers/onPremToHostedBlueprintMapper.tsx
+++ b/src/Components/Blueprints/helpers/onPremToHostedBlueprintMapper.tsx
@@ -150,6 +150,7 @@ export const mapOnPremToHosted = (
     metadata: {
       parent_id: null,
       exported_at: '',
+      is_on_prem: true,
     },
   };
 };
--- a/api/schema/compliance.json
+++ b/api/schema/compliance.json
@@ -113,124 +113,124 @@
                     "value": {
                       "data": [
                         {
-                          "id": "1e8b2d9f-0b52-471a-84c6-f27a8c132b08",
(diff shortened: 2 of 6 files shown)


ezr-ondrej (commented):
Good preparation, tho we could set it in this PR directly, or is that harder then I assume?

avitova:
@ezr-ondrej Not sure what you mean - in the first PR with on prem, we did not yet have the field, and I wanted to make sure that we actually merge on-prem into frontend, only then to have it also on backend (since it shows in API). So this is why the feature is added only now.

ezr-ondrej:
I mean we should set the flag somewhere in import, right? Something should set this to true if I import on-prem blueprint :)

src/Components/Blueprints/helpers/onPremToHostedBlueprintMapper.tsx
```diff
@@ -150,6 +150,7 @@ export const mapOnPremToHosted = (
     metadata: {
       parent_id: null,
       exported_at: '',
+      is_on_prem: true,
```
avitova, line 153: @ezr-ondrej

ezr-ondrej: Ah, missed that, sorry! Ignore me in that case! :)

ezr-ondrej (approved):

regexowl:
One more try...

/retest

regexowl:
Alright, sneaky merge time. This shouldn't have any effect on IQE

avitova:
Thank you!

PR #2620, opened 2024-11-25
ImagesTable: Fix image name spacing

regexowl:
This adds a component value to the image name, making the spacing work correctly.

Before:
image

After:
image


mgold1234 (approved):
/lgtm

PR #2622, opened 2024-11-25
store/enhancedImageBuilderApi: handle empty error data (HMS-5049)

croissanne:
There's a case where the IB api returns an error without more data in it, though it might be coming from a component which treats the request before the image builder api.

### DIFF
--- a/src/store/enhancedImageBuilderApi.ts
+++ b/src/store/enhancedImageBuilderApi.ts
@@ -5,7 +5,7 @@ import { imageBuilderApi } from './imageBuilderApi';
 /* eslint-disable @typescript-eslint/no-explicit-any */
 const errorMessage = (err: any) => {
   let msg = err.error.statusText;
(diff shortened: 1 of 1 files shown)


mgold1234 (approved):
/lgtm

src/store/enhancedImageBuilderApi.ts
```diff
@@ -5,7 +5,7 @@ import { imageBuilderApi } from './imageBuilderApi';
 /* eslint-disable @typescript-eslint/no-explicit-any */
 const errorMessage = (err: any) => {
   let msg = err.error.statusText;
-  if (err.error.data?.errors[0]?.detail) {
+  if (err.error.data?.errors.length > 0 && err.error.data?.errors[0]?.detail) {
```
regexowl, line 8: Can the length also be equal to 0?

regexowl: Let's say I'm tired x_x

PR #2629, opened 2024-11-27
Wizard: hide compliance radio on stable

croissanne:
Currently only the title is correctly changed between preview and stable. On stable, the radio should obviously be hidden as well.

---
stable
afbeelding

preview
afbeelding


ochosi (dismissed):

regexowl:
Could we gate the beta by setting up unleash constraint for preview only? Let me try.

croissanne:
> Could we gate the beta by setting up unleash constraint for preview only? Let me try.

oh right, i forgot you could do that!

regexowl:
Done, I think it worked as needed, can anyone check that the visibility of the radio works as it should?

PR #2631, opened 2024-11-27
ImagesTable: add RHEL 9.5 display value

### LINKED ISSUE #2632: RHEL 9.5 doesn't have a display value
When explicitly building rhel-95, it doesn't show up in the UI

image

ezr-ondrej:
SSIA

Fixes #2632

### DIFF
--- a/api/schema/imageBuilder.yaml
+++ b/api/schema/imageBuilder.yaml
@@ -1117,6 +1117,7 @@ components:
         - rhel-92
         - rhel-93
         - rhel-94
+        - rhel-95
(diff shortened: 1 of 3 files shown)


regexowl:
Absolutely yes! But it's weird that we didn't get any errors on build. :thinking:

regexowl:
> I've explicitly updated only the IB schema, there is A LOT of changes in compliance, I did not want to pollute this PR xD

Fair enough! This is perfect, thanks!

regexowl (approved):

PR #2633, opened 2024-11-27
wizard: add  Users to review step (HMS-4906)

mgold1234:
this commit add Users to review step - (HMS-4906)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
@@ -68,6 +68,8 @@ import {
   selectLanguages,
   selectKeyboard,
   selectHostname,
+  selectUserNameByIndex,
+  selectUserPasswordByIndex,
 } from '../../../../store/wizardSlice';
 import { toMonthAndYear, yyyyMMddFormat } from '../../../../Utilities/time';
 import {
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -774,6 +777,45 @@ export const TimezoneList = () => {
...
+              User name
+            </TextListItem>
+            <TextListItem component={TextListItemVariants.dd}>
+              {userName}
+            </TextListItem>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-v5-u-min-width"
+            >
+              Passwords
```
regexowl, line 807: Can there be more passwords or should it be just "Password"?

src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -774,6 +777,45 @@ export const TimezoneList = () => {
...
+  return (
+    <TextContent>
+      <TextList component={TextListVariants.dl}>
+        {users.length > 0 && (
+          <>
+            <TextListItem
+              component={TextListItemVariants.dt}
+              className="pf-v5-u-min-width"
+            >
+              User name
```
regexowl, line 798:
```suggestion
              Username
```

src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -774,6 +777,45 @@ export const TimezoneList = () => {
...
+  const index = 0;
+  const userNameSelector = selectUserNameByIndex(index);
+  const userName = useAppSelector(userNameSelector);
+  const userPasswordSelector = selectUserPasswordByIndex(index);
+  const userPassword = useAppSelector(userPasswordSelector);
+
+  return (
+    <TextContent>
+      <TextList component={TextListVariants.dl}>
+        {users.length > 0 && (
```
regexowl, line 792: This will need a tiny tweak as when there's no username and/or password the item descriptions are rendered empty:

image

That's because when clicking "Add user" the object gets added and stays in the store even when there are no values, thus the array has length of 1:

image

We're currently showing the expandables even when they're empty so we could simply remove this condition and in the descriptions do something like `{userName || 'None'} so we'd have

image

We can hide the entire expandable for sure if there are no values, but that would have to be done on the `ReviewStep.tsx` level.

regexowl (commented):
Looking good! Added a few comments.

regexowl (approved):
Thank you!

PR #2625, opened 2024-11-26
Wizard: LOCALE 1 - Basic Locale step (HMS-5055)

regexowl:
This adds Locale step basics and file structure.

PR #2636, opened 2024-11-28
Wizard: LOCALE 2- Add locale structure to store and mapper (HMS-5097)

regexowl:
This creates structure and actions for the locale customization in the wizardSlice and requestMapper.

Based on and blocked by https://github.com/osbuild/image-builder-frontend/pull/2625

### DIFF
--- a/src/store/wizardSlice.ts
+++ b/src/store/wizardSlice.ts
@@ -7,6 +7,7 @@ import type {
   Distributions,
   ImageRequest,
   ImageTypes,
+  Locale,
   Repository,
 } from './imageBuilderApi';
 import type { ActivationKeys } from './rhsmApi';
@@ -106,6 +107,7 @@ export type wizardState = {
   kernel: {
     append: string;
   };
+  locale: Locale;
   details: {
     blueprintName: string;
     blueprintDescription: string;
@@ -176,6 +178,10 @@ export const initialState: wizardState = {
   },
   kernel: {
     append: '',
+  },
+  locale: {
+    languages: [],
+    keyboard: '',
   },
   details: {
     blueprintName: '',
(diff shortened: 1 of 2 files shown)


src/store/wizardSlice.ts
```diff
@@ -660,6 +674,23 @@ export const wizardSlice = createSlice({
         1
       );
     },
+    addLanguage: (state, action: PayloadAction<string>) => {
+      if (!state.locale.languages?.some((lang) => lang === action.payload)) {
+        state.locale.languages?.push(action.payload);
```
lucasgarfield, line 679: Ultra nitpick, do we need the optional chaining here (`?`)? I think we should only enter the `if` if state.locale.languages exists.

lucasgarfield (approved):
I love your approach of breaking this down into bite sized, super easy to review pull requests. They’re simple enough even for a manager to review. 😉

Just to play devil’s advocate on this specific PR – since we aren’t actually using any of the new actions or selectors it is hard to be confident that they will work as intended. It would be nice to see test coverage for them going in at the same time.

But to play devil’s advocate to my devil’s advocate – these are also just following the bog standard pattern that we use everywhere else in the project. So in this case, I don’t think the lack of test coverage is a blocker as long as we make sure we write test for the actual forms in the subsequent PRs.

That’s just some food for thought. But again, I’m really enjoying reviewing these small PRs and I think it is helping us go a lot faster overall (and I’m curious to hear if you think so, too). 

Not going to merge since you still have it marked as draft but as far as I'm concerned this can go in - maybe fix the optional chaining nitpick I pointed out if you get a chance.

regexowl:
Excellent, I'll try to split the steps into PRs when possible. Glad those are nicer to review. And yep, fully agree that tests will be necessary as soon as there is at least some functionality. :+1: 

Blocker was resolved, moving outside of draft

src/store/wizardSlice.ts
regexowl: Great point! Added the condition both in `addLanguage` and `removeLanguage` to get rid of the optional chaining.

lucasgarfield (approved):
Nice!

PR #2637, opened 2024-11-28
Azure: Add support for selecting HyperV generation

croissanne:
HyperV generation will change the uploaded image into a V1/V2 image. V2 supports more instance types, and has UEFI boot.

image


regexowl (approved):
Looks great! We could add the HyperV gen to the Azure expandable on Review step as well. Would be also perfectly fine in a follow-up.

regexowl (approved):
Thank you!

PR #2648, opened 2024-12-04
Wizard: Conditionally hide package recommendations

regexowl:
This shows package recommendations only when RHEL 8 or RHEL 9 are selected as a distribution.


lucasgarfield (approved):
Makes sense!

PR #2651, opened 2024-12-06
Wizard: LOCALE - Add Keyboard drop down (HMS-5096)

regexowl:
This adds a drop down for keyboard selection. The options are populated with an output of `localectl list-keymaps` and sorted in the order of:
- exact match
- starting with the search input alphabetically
- not starting with search input alphabetically

New tests are also added.

image


mgold1234:
I like it!

mgold1234 (approved):
/lgtm

mgold1234:
small question, if there is no keyboard result, the 'Next' button should become disable?

src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
```diff
@@ -0,0 +1,184 @@
...
+      }
+    }
+    setSelectOptions(filteredKeyboards.sort((a, b) => sortfn(a, b)));
+
+    // This useEffect hook should run *only* on when the filter value changes.
+    // eslint's exhaustive-deps rule does not support this use.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [filterValue]);
+
+  const sortfn = (a: string, b: string) => {
```
mgold1234, line 54: what do you think about maybe writing all the condtion shorter? something like -
```suggestion

const sortfn = (a: string, b: string) => {
  const aKeyboard = a.toLowerCase();
  const bKeyboard = b.toLowerCase();
  
  if (aKeyboard === filterValue || aKeyboard < bKeyboard) return -1;
  if (bKeyboard === filterValue || bKeyboard < aKeyboard) return 1;

  const aStartsWithFilter = aKeyboard.startsWith(filterValue);
  const bStartsWithFilter = bKeyboard.startsWith(filterValue);

  if (aStartsWithFilter && !bStartsWithFilter) return -1;
  if (bStartsWithFilter && !aStartsWithFilter) return 1;

  return aKeyboard.localeCompare(bKeyboard);
};
```

lucasgarfield:
I went ahead and rebased on main and resolved the conflicts.

src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
lucasgarfield, line 54: This is the same behavior as the package sort function as far as I can tell. What if we extracted this into a utility function? It would need to take a 3rd argument, the search term, but that would help DRY out the code.

lucasgarfield (changes requested):

src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
lucasgarfield: Would this have the same behavior? :thinking:

lucasgarfield: I added a commit for this little refactor so we don't have to go through a full edit/review cycle on Monday.

lucasgarfield:
> small question, if there is no keyboard result, the 'Next' button should become disable? Screenshot 2024-12-06 at 13 06 22

This is an optional customization so we don't need to disable Next.

lucasgarfield:
Noticed a weird bug where if you hit "enter" while using the dropdown and you haven't made a selection it causes the wizard to refresh. This bug affects other fields, too, such as the OpenSCAP so let's not block on it.

https://github.com/osbuild/image-builder-frontend/issues/2655

lucasgarfield (approved):
Looks good to me. Since the bug I noticed (hitting enter while the typeahead is focused can cause a refresh) is also present in other typeaheads, I'm going to say we defer the responsibility to fixing that to another PR that fixes all typeahead behavior.

lucasgarfield:
Failing on unit tests in pr_check. Dev checks are passing, so I'm merging.

PR #2665, opened 2024-12-10
test: Packages selected on import (HMS-5166)

regexowl:
This checks if specified packages get selected when importing a toml blueprint.


lucasgarfield (approved):
Nice!

PR #2666, opened 2024-12-10
test: Clean up comments left over after rebasing

regexowl:
This removes two comments left over after rebasing, both tests are already implemented a bit higher in the file.


lucasgarfield (approved):
Good catch. Might be my fault, I left those in when I resolved a conflict on another PR.

PR #2672, opened 2024-12-11
Wizard: Add Hostname functionality (HMS-5175)

regexowl:
This adds a validated hostname input and new tests.<br/><br/>JIRA: [HMS-5175](https://issues.redhat.com/browse/HMS-5175)

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -18,6 +18,7 @@ import {
   selectUseLatest,
   selectActivationKey,
   selectRegistrationType,
+  selectHostname,
 } from '../../../store/wizardSlice';
 import {
   getDuplicateMountPoints,
@@ -25,6 +26,7 @@ import {
   isBlueprintDescriptionValid,
   isMountpointMinSizeValid,
   isSnapshotValid,
+  isHostnameValid,
 } from '../validators';
 
 export type StepValidation = {
@@ -38,12 +40,14 @@ export function useIsBlueprintValid(): boolean {
   const registration = useRegistrationValidation();
   const filesystem = useFilesystemValidation();
   const snapshot = useSnapshotValidation();
+  const hostname = useHostnameValidation();
   const firstBoot = useFirstBootValidation();
   const details = useDetailsValidation();
   return (
     !registration.disabledNext &&
     !filesystem.disabledNext &&
     !snapshot.disabledNext &&
+    !hostname.disabledNext &&
     !firstBoot.disabledNext &&
     !details.disabledNext
   );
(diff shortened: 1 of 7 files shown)


regexowl:
Alright, let's give this a try -> /jira-epic HMS-5164

regexowl:
What am I doing wrong with the epic? :thinking:

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -133,6 +137,20 @@ export function useFirstBootValidation(): StepValidation {
   };
 }
 
+export function useHostnameValidation(): StepValidation {
+  const hostname = useAppSelector(selectHostname);
+
+  if (!isHostnameValid(hostname)) {
+    return {
+      errors: {
+        hostname: 'Invalid hostname',
```
lucasgarfield, line 146: Let's provide some additional guidance to users about what constitutes a valid hostname.

Had to look this up myself, of course. Run `man 5 hostname` and you get this:
```
HOSTNAME(5)                                                                         hostname                                                                         HOSTNAME(5)

NAME
       hostname - Local hostname configuration file

SYNOPSIS
       /etc/hostname

DESCRIPTION
       The /etc/hostname file configures the name of the local system. Unless overridden as described in the next section, systemd(1) will set this hostname during boot using
       the sethostname(2) system call.

       The file should contain a single newline-terminated hostname string. Comments (lines starting with a "#") are ignored. The hostname should be composed of up to 64 7-bit
       ASCII lower-case alphanumeric characters or hyphens forming a valid DNS domain name. It is recommended that this name contains only a single label, i.e. without any
       dots. Invalid characters will be filtered out in an attempt to make the name valid, but obviously it is recommended to use a valid name and not rely on this filtering.

       You may use hostnamectl(1) to change the value of this file during runtime from the command line. Use systemd-firstboot(1) to initialize it on mounted (but not booted)
       system images.
```

So I would suggest the error say: "Invalid hostname. The hostname should be composed of up to 64 7-bit
       ASCII lower-case alphanumeric characters or hyphens forming a valid DNS domain name. It is recommended that this name contains only a single label, i.e. without any
       dots. Invalid characters will be filtered out in an attempt to make the name valid, but obviously it is recommended to use a valid name and not rely on this filtering."

src/Components/CreateImageWizard/validators.ts
```diff
@@ -90,3 +90,12 @@ export const isNtpServerValid = (ntpServer: string) => {
     /^([a-z0-9-]+)?(([.:/]{1,3}[a-z0-9-]+)){1,}$/.test(ntpServer)
   );
 };
+
+export const isHostnameValid = (hostname: string) => {
+  if (hostname !== '') {
+    return /^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]*[A-Za-z0-9])$/.test(
```
lucasgarfield, line 96: I think this expression fits what I'm proposing in my other comment. We just need to also set the max length to 64.

lucasgarfield:
I added a commit with my suggested changes, hope you don't mind. Just want to keep the review cycle to a minimum so we can ship these before end of Q4.

lucasgarfield (approved):
:teapot:

lucasgarfield:
Unit tests failing in pr_check but passing in dev-check. Admin merging.

18:51:03  FAIL  src/test/Components/CreateImageWizard/steps/ImageOutput/ImageOutput.test.tsx > Step Image output > revisit step button on Review works****

PR #2673, opened 2024-12-11
Wizard: Add popover to suggest importing on-prem

avitova:
In order to hint users that on-premise blueprints are accepted in import, we are adding a popover with a message.


avitova:
https://github.com/user-attachments/assets/e4df2b6e-eb19-41cb-b87c-215468caf530

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -140,7 +142,23 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+            bodyContent={
+              <div>
+                You can import the blueprints you created by using the Red Hat
+                image builder into Insights images to create customized images.
+              </div>
+            }
+          >
+            <OutlinedQuestionCircleIcon
+              style={{ fontSize: 'var(--pf-v5-global--icon--FontSize--md)' }}
+            />
```
regexowl, line 158: We're using `<HelpIcon />` for popover buttons throughout the code base so even though the icons looks the same I'd vote for using the help one for the sake of consistency :sweat_smile: 

This is more of a suggestion so I'll leave the final decision up to you - we could use the PF Icon sizing to get rid of the inlined style like so:

```suggestion
            <Icon size="md">
              <HelpIcon />
            </Icon>
```

regexowl (commented):
Thank you! Looks good, added one suggestion

lucasgarfield:
I think we should also update the "Upload a JSON file" text since on-prem blueprints are more likely to be toml. We can reach out to copy team for help if we need to.

avitova:
Good point, @lucasgarfield! I updated the description! 
image

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -140,7 +142,28 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+            bodyContent={
+              <div>
+                You can import the blueprints you created by using the Red Hat
+                image builder into Insights images to create customized images.
+              </div>
+            }
+          >
+            <Button
+              variant="plain"
+              aria-label="About regions"
```
regexowl, line 158:
```suggestion
              aria-label="About import"
```

src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -140,7 +142,28 @@ export const ImportBlueprintModal: React.FunctionComponent<
...
+              className="pf-v5-u-pl-sm header-button"
```
regexowl, line 159: The `header-button` class updates right padding only, might not be needed here.

```suggestion
              className="pf-v5-u-pl-sm"
```

regexowl (commented):
Looks great, thank you! Two small nitpicks and then I'd say it's good to go :rocket:

regexowl:
I mean... the changes shouldn't break IQE. I'll try my luck, let's admin merge.

PR #2675, opened 2024-12-11
sentry: use webpack plugin for injecting (HMS-5179)

croissanne:
Rationale is the same as in  #2662, but we need the module metadata for correct routing. None of these values are actually secret, only the auth token is. And turns out that the webpack plugin justs injects and doesn't upload in case we don't specify the auth token, so that's nice.<br/><br/>JIRA: [HMS-5179](https://issues.redhat.com/browse/HMS-5179)

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -13,7 +13,6 @@ if [ "$IS_PR" == true ]; then
 else
     export BETA=false
     build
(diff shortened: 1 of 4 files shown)


regexowl (approved):
:rocket:

PR #2690, opened 2024-12-17
Wizard: Add Kernel name input (HMS-5204)

regexowl:
This adds a kernel name input.

JIRA: [HMS-5204](https://issues.redhat.com/browse/HMS-5204)

### DIFF
--- a/src/Components/CreateImageWizard/validators.ts
+++ b/src/Components/CreateImageWizard/validators.ts
@@ -91,6 +91,17 @@ export const isHostnameValid = (hostname: string) => {
   );
 };
 
+export const isKernelNameValid = (kernelName: string) => {
+  if (!kernelName) {
+    return true;
+  }
+
+  return (
+    kernelName.length < 65 &&
+    /^[a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9]$/.test(kernelName)
+  );
+};
+
 export const isPortValid = (port: string) => {
   return /^(\d{1,5}|[a-z]{1,6})(-\d{1,5})?:[a-z]{1,6}$/.test(port);
 };
--- a/src/Components/CreateImageWizard/validators.ts
+++ b/src/Components/CreateImageWizard/validators.ts
@@ -91,6 +91,17 @@ export const isHostnameValid = (hostname: string) => {
   );
 };
 
+export const isKernelNameValid = (kernelName: string) => {
+  if (!kernelName) {
+    return true;
+  }
+
+  return (
+    kernelName.length < 65 &&
+    /^[a-z0-9]|[a-z0-9][a-z0-9-_.+]*[a-z0-9]$/.test(kernelName)
+  );
+};
+
 export const isPortValid = (port: string) => {
   return /^(\d{1,5}|[a-z]{1,6})(-\d{1,5})?:[a-z]{1,6}$/.test(port);
 };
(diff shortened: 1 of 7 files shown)


src/Components/CreateImageWizard/validators.ts
```diff
@@ -103,3 +103,14 @@ export const isHostnameValid = (hostname: string) => {
...
+
+export const isKernelNameValid = (kernelName: string) => {
+  if (!kernelName) {
+    return true;
+  }
+
+  return (
+    kernelName.length < 65 &&
+    /^[a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9]$/.test(kernelName)
+  );
```
regexowl, line 115: Not completely sure about the validation pattern. Didn't find much information in blueprint reference or stages :thinking:

lucasgarfield: What I've learned is that when a kernel is specified, DNF is actually used to install it. So it would need to follow whatever the rules for package names in DNF are. I'm sure there must be a standard for that we can find, although my Google-fu has failed me. I did find something for Ubuntu:

https://askubuntu.com/questions/1262528/what-are-usable-characters-are-for-package-names#:~:text=Package%20names%20(both%20source%20and,start%20with%20an%20alphanumeric%20character.

I think if we could find something similar for RH we would have our answer.

mgold1234:
@regexowl it is looking awsome!
one small thing, I am little confusing why I see the Append without any box, maybe we can remove it from now for that pr?
Screenshot 2024-12-20 at 11 50 26

lucasgarfield:
debug, rt, and rt-debug all fail with the same error:
image

I wonder if something in the backend is broken?

ochosi:
> debug, rt, and rt-debug all fail with the same error:
> image
> 
> I wonder if something in the backend is broken?

What happens if you try the API directly?

regexowl:
That's weird, I've tried to build an image with "kernel-debug" and the build seems successful (RHEL 9 .qcow2)

kernel-name

regexowl:
Hmm, but rt fails immediately...

regexowl:
Tried to create a blueprint via API (below) and rt failed immediately, same as via UI. 

```json
{
  "name": "kernel-name",
  "description": "",
  "distribution": "rhel-9",
  "image_requests": [
  {
    "architecture": "x86_64",
    "image_type": "guest-image",
    "upload_request": {
      "options": {},
      "type": "aws.s3"
    }
  }
  ],
  "customizations": {
    "kernel": {
      "name": "kernel-debug"
    }
  }
}
```

regexowl:
After discussion on main channel, I've updated the PR to allow only kernel and kernel-debug packages for now as rt and rt-debug are in separate repositories.

regexowl:
@lucasgarfield I've removed the validator for now as it isn't used anywhere.

regexowl:
Adding custom name option.

regexowl:
Allows to add a custom kernel name:
image
image

regexowl:
Wasn't sure if we want the alert or not. There's a validation for the kernel name format, but even if that passes the build might fail with kernel package that's not included.

croissanne (approved):
lgtm, though I wonder if something like a text input with a suggestion autocomplete exists? so we could suggest the default options?

regexowl:
> lgtm, though I wonder if something like a text input with a suggestion autocomplete exists? so we could suggest the default options?

Thanks for the review! The default options are there when you click on the input
image

The create option gets rendered when there's a text input that doesn't match any of the default options
image

PR #2691, opened 2024-12-17
devDeps: Manually bump msw from 2.6.6 to 2.7.0 (HMS-5205)

regexowl:
This bumps msw from 2.6.6 to 2.7.0

JIRA: [HMS-5205](https://issues.redhat.com/browse/HMS-5205)


lucasgarfield:
All the tests pass so I'm tempted to just rubberstamp and merge but I have to admit I don't understand the conditional around the acceptHeaders... would you be so kind as to add an explanation to the commit message so that future us can use `git blame` to figure out why we need this?

regexowl:
All I did was manually bump the version of MSW, so really I've just changed the version in package.json and ran `npm i`.

The thing with MSW bumps is - `npm i` often makes changes to the `src/mockServiceWorker.js` file that's fetched from the dependency.

BUT dependabot doesn't want to touch anything outside of package.json and package-lock.json. Meaning the dependabot PRs always fail on the "manually changing files" check, because the changes are contained withing the package and package-lock files only, but `npm i` tries to refresh the `mockServiceWorker.js` file as well, creating a mismatch.

PR #2695, opened 2024-12-18
api: add latest changes in api

mgold1234:
this commit adds the latest changes in api schema


regexowl:
Please update the commit message if this PR is just an API regen. Two things come to mind:
- not sure why this would trigger a security check
- there's always a lot of changes in the compliance file it seems. @lucasgarfield would you have any idea what's up with that? Just wondering if we need to bump the Compliance API schema on more regular basis, the Latin really confuses me :sweat_smile:

regexowl (approved):

PR #2696, opened 2024-12-18
test: Add create mode test for User step

mgold1234:
This adds test that check the functionality of blueprint request with user.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -225,7 +225,6 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   const firstBootValidation = useFirstBootValidation();
   // Details
   const detailsValidation = useDetailsValidation();
-
   let startIndex = 1; // default index
   if (isEdit) {
     startIndex = 20;
--- a/src/Components/CreateImageWizard/ValidatedTextInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedTextInput.tsx
@@ -1,11 +1,15 @@
-import React, { useState } from 'react';
+import React, { ChangeEvent, useState } from 'react';
 
 import {
   HelperText,
   HelperTextItem,
   TextInput,
   TextInputProps,
+  Button,
+  TextAreaProps,
+  TextArea,
 } from '@patternfly/react-core';
+import { EyeSlashIcon, EyeIcon } from '@patternfly/react-icons';
 
 import type { StepValidation } from './utilities/useValidation';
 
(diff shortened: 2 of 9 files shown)


src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -225,7 +225,6 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   const firstBootValidation = useFirstBootValidation();
   // Details
   const detailsValidation = useDetailsValidation();
-
```
lucasgarfield, line 228: Mistake?

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,62 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+  isPristine: boolean,
+  isEmpty?: boolean
+): 'default' | 'error' | 'success' => {
+  if (isEmpty) return 'default';
+  if (isPristine) return 'default';
+  if (stepValidation.errors[fieldName] === 'default') return 'default';
+  return stepValidation.errors[fieldName] ? 'error' : 'success';
+};
+
+const ErrorHelperText = ({
```
lucasgarfield, line 67: To be consistent with the rest of the code base, please create a new type for the props instead of defining them inline.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,62 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+  errorMessage,
+}: {
+  errorMessage?: string | undefined;
```
lucasgarfield, line 70: I think we can really simplify this component by making errorMessage mandatory (remove `?`) and also remove undefined.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,62 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+}) => {
+  if (!errorMessage) return null;
```
lucasgarfield, line 72: This wouldn't be needed anymore if we do what I've suggested above.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,62 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+  return (
+    <HelperText>
+      <HelperTextItem variant="error" hasIcon>
+        {errorMessage}
+      </HelperTextItem>
+    </HelperText>
+  );
+};
+
+const usePristineState = (initialValue: string) => {
```
lucasgarfield, line 83: I think that this is a bit confusing because it goes against the convention of useState hooks being at the top level of the component.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,62 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+  const [isPristine, setIsPristine] = useState(!initialValue);
+
+  const handleBlur = () => {
```
lucasgarfield, line 86: Having a hook return a handler also goes against the existing conventions in the project.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -90,6 +145,57 @@ export const HookValidatedInput = ({
   );
 };
 
+export const HookValidatedInputWithPasswordVisibilityButton = (
```
lucasgarfield, line 148: image

I think we should try reusing the <HookValidatedInput> component inside of the <Password> component, extending <HookValidatedInput> if necessary.

PR #2697, opened 2024-12-18
wizard: add user information step (only user name and password) (HMS-4903)

mgold1234:
This commit introduces the user information step with the following fields: 
(*) `userName` field
(*) `password` field

starting this step with very small steps and continue the work in other prs 

Screenshot 2024-12-18 at 17 44 12


lucasgarfield:
I wonder why this is failing snyk?

lucasgarfield:
> I wonder why this is failing snyk?

Ah, it's a false positive related to where we define a password field for the tests:
image

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const userNameSelector = selectUserName(0);
+  const userName = useAppSelector(userNameSelector);
+  const userPasswordSelector = selectUserPassword(0);
+  const userPassword = useAppSelector(userPasswordSelector);
+
+  const handleNameChange = (
+    _e: React.FormEvent<HTMLInputElement>,
+    value: string
+  ) => {
+    const index = 0;
```
lucasgarfield, line 25: Even though this is temporary, we can DRY the code out by moving this outside of this handler and re-using it in handlePasswordChange.

lucasgarfield: We can also re-use it in userNameSelector and userPasswordSelector. Right now the 0 we specify is sort of a "magic number", and it is a good practice to avoid those.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -558,6 +569,30 @@ const getOpenscap = (state: RootState): OpenScap | undefined => {
   return undefined;
 };
 
+const getUsers = (state: RootState): User[] | undefined => {
+  const users = selectUsers(state);
+
+  if (!users || users.length === 0) {
```
lucasgarfield, line 575: It’s good to always to think really carefully about the types and their implications.

What’s the return type of selectUsers? It is <User>[].

So what does checking for `!users` do?

Well, users is either going to be an empty array or an array of <User> objects.

In both cases, `!users` returns false.

If you’re ever confused about what something will return, an easy way to check is using the developer console in your web browser:
Screenshot from 2024-12-18 14-21-06

src/store/wizardSlice.ts
```diff
@@ -43,6 +44,18 @@ export type RegistrationType =
...
+export type UserWithAdditionalInfo = Required<User>;
+
+type UserPayload = {
+  index: number;
+  name: string;
+};
+
+type UserPasswordPayload = {
+  index: number;
+  password: string | undefined;
```
lucasgarfield, line 56: Here I think that we need to think about what the empty state for the <TextInput> we will use for the password is. Is it `''` or `undefined`?

That makes working with the <TextInput> component in the React component code easier since we don't have to handle the special empty case by switching the value to undefined. On the other hand, it does mean  that when we generate the request we send to the image-builder API that we need to take care to convert the `''` to `undefined`. 

I don't think that one way is necessarily right or wrong, so here we should just let consistency guide us.

To be consistent with the rest of the code base, we should use `''` as the empty state (assuming things haven't changed too drastically since we made this decision over a year ago...).

So I think that the type for password should just be `string`.

src/store/wizardSlice.ts
```diff
@@ -337,6 +352,18 @@ export const selectServices = (state: RootState) => {
   return state.wizard.services;
 };
 
+export const selectUsers = (state: RootState) => {
+  return state.wizard.users;
+};
+
+export const selectUserName = (userIndex: number) => (state: RootState) => {
```
lucasgarfield, line 359: Can we rename this selectUserNameByIndex?

src/store/wizardSlice.ts
```diff
@@ -337,6 +352,18 @@ export const selectServices = (state: RootState) => {
...
+  return state.wizard.users[userIndex]?.name;
+};
+
+export const selectUserPassword = (userIndex: number) => (state: RootState) => {
```
lucasgarfield, line 363: Can we rename this selectUserPasswordByIndex?

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+  await waitFor(() => expect(nextButton).toBeEnabled());
+};
+
+describe('Step Users', () => {
+  beforeEach(async () => {
+    vi.clearAllMocks();
+    router = undefined;
+  });
+
+  test('without adding user loads First Boot', async () => {
```
lucasgarfield, line 64: I think this test needs to be renamed to reflect what it's doing - it does add a user.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
lucasgarfield, line 1: Can we also please add a test for edit mode?

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
mgold1234: this test when user get to the 'add a user' screen, he press Next and actually doesnt add a new user. `
just make sure, that user can skip this step and doesnt add a new user

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
mgold1234: yes, but I think to do it in other pr, as we agree very small step

lucasgarfield: Even though we're working iteratively via small prs, we still need to be sure that the functionality those prs adds works. New functionality should always be accompanied by tests so that we have confidence everything is working as we continue developing. 

This pull request adds support for create mode, and tests it. That's good.

This pull request also adds support for edit mode. Therefore, we should test it in this pr.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
lucasgarfield: Ah I see, you're right. But I think the name of the test still needs some work, this says "without adding user loads First Boot" but the next step is actually timezone.

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const userName = useAppSelector(userNameSelector);
+  const userPasswordSelector = selectUserPasswordByIndex(index);
+  const userPassword = useAppSelector(userPasswordSelector);
+
+  const handleNameChange = (
+    _e: React.FormEvent<HTMLInputElement>,
+    value: string
+  ) => {
+    dispatch(setUserNameByIndex({ index: index, name: value }));
+  };
```
lucasgarfield, line 26: Can we add some whitespace between these functions to improve readability?

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+          ],
+        },
+      };
+
+      await waitFor(() => {
+        expect(receivedRequest).toEqual(expectedRequest);
+      });
+    });
+  });
+  describe('Users edit mode', () => {
```
lucasgarfield, line 140: Please add whitespace in-between describe blocks.

src/Components/CreateImageWizard/steps/Users/component/Empty.tsx
```diff
@@ -10,19 +10,28 @@ import {
 } from '@patternfly/react-core';
 import UserIcon from '@patternfly/react-icons/dist/esm/icons/user-icon';
 
+import { useAppDispatch } from '../../../../../store/hooks';
+import { addUser } from '../../../../../store/wizardSlice';
+
 const EmptyUserState = () => {
+  const dispatch = useAppDispatch();
```
lucasgarfield, line 17: Please add white space in-between the hook block and the handler.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+    router = undefined;
+  });
+
+  test('without adding user loads timezone', async () => {
+    const user = userEvent.setup();
+    await renderCreateMode();
+    await goToRegistrationStep();
+    await clickRegisterLater();
+    await goToUsersStep();
+    const addUser = await screen.findByRole('button', { name: /add a user/i });
```
lucasgarfield, line 74: One goal for these tests is to limit their scope to test one specific thing and avoid duplicate test coverage to the extent possible. This will hopefully make them easier to manage and work with.

This test makes sure that the "next" button works and that the appropriate step loads when it is clicked. That's a good thing to test, and we have similar tests in all the other steps.

However, do we really need to check and make sure that the add user button is enabled? I think that the test where we actually add a user already covers that case.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+    await waitFor(() => user.click(nextButton));
+    await screen.findByText('Select a timezone for your image.');
+  });
+
+  test('clicking Back loads Additional packages', async () => {
+    await renderCreateMode();
+    await goToRegistrationStep();
+    await clickRegisterLater();
+    await goToUsersStep();
+    const addUser = await screen.findByRole('button', { name: /add a user/i });
```
lucasgarfield, line 87: Same thing here. Making sure that the back button works is a good test and something we do on the other steps. I don't think we need to check that the add user button is enabled during this specific test since that is covered in another test.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+    await screen.findByRole('heading', { name: 'Additional packages' });
+  });
+
+  test('clicking Cancel loads landing page', async () => {
+    await renderCreateMode();
+    await goToUsersStep();
+    await verifyCancelButton(router);
+  });
+
+  test('with valid name', async () => {
```
lucasgarfield, line 99: This test is identical to the 'with valid name and password' test that appears in the 'User request generated correctly' block up to (and including) goToReviewStep().

Keep in mind that every test takes time to run, and the faster the test suite runs the better the developer experience. We want coverage to be high, but we also want to avoid duplication.

We will want to add tests that check the behavior for an _invalid_ name, of course, once we add validation.

src/test/fixtures/editMode.ts
```diff
@@ -439,6 +439,28 @@ export const timezoneBlueprintResponse: BlueprintResponse = {
...
+  ...baseCreateBlueprintRequest,
+  name: mockBlueprintNames['users'],
+  description: mockBlueprintDescriptions['users'],
+  customizations: {
+    users: [
+      {
+        name: 'best',
+        password: '',
+        groups: [],
+        ssh_key: '',
```
lucasgarfield, line 452: I don't think these should be here yet. In the User[] that we put together for the image-builder API, these are optional fields. Only the name should appear in the request for now.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -558,6 +569,30 @@ const getOpenscap = (state: RootState): OpenScap | undefined => {
...
+
+  if (users.length === 0) {
+    return undefined;
+  }
+
+  return users.map((user) => {
+    const result: User = {
+      name: user.name,
+    };
+    if (user.password !== undefined) {
```
lucasgarfield, line 583: Please see what I wrote here:
https://github.com/osbuild/image-builder-frontend/pull/2697#discussion_r1890818172

The password will be stored as a string. We will use `''` as the empty state, not undefined. Therefore, we don't need to check whether or not it is undefined - we need to check whether it is `''` or not.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -558,6 +569,30 @@ const getOpenscap = (state: RootState): OpenScap | undefined => {
...
+      result.password = user.password;
+    }
+    if (user.ssh_key !== undefined) {
+      result.ssh_key = user.ssh_key;
+    }
+    if (user.groups !== undefined) {
```
lucasgarfield, line 589: Same as the comment above here. However, since we don't support groups yet, let's just remove this for now and save it for the isAdmin pr.

src/store/wizardSlice.ts
```diff
@@ -43,6 +44,18 @@ export type RegistrationType =
 
 export type ComplianceType = 'openscap' | 'compliance';
 
+export type UserWithAdditionalInfo = Required<User>;
```
lucasgarfield, line 47: I just realized that the User type also allows the fields to be `undefined`. Since we will be using `''` as the empty state for the form fields, we can tighten this up even a bit more by using the `NonNullable` utility function in addition to `Required`.

https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype

lucasgarfield: image

lucasgarfield:
pr_check failing on a timeout, was green previously before I rebased, admin merging

17:17:06 2024-12-20 17:17:06 [   ERROR] [          MainThread] hit timeout error: Could not do 'wait for all deployed resources to be ready' at /var/lib/jenkins/jenkins-osbuild-image-builder-frontend-pr-check-4899-w05iBB/.bonfire_venv/lib64/python3.8/site-packages/bonfire/openshift.py:85 in time

PR #2603, opened 2024-11-19
On-prem: read blueprints from disk (COMPOSER-2413)

kingsleyzissou:
We can use the cockpit files api to read in blueprints from disk.

This PR adds commits to bring in the required dependencies, some convenience scripts and an initial commit using an RTK query function to read blueprints from file and pull the details into the store.

PR #2701, opened 2024-12-19
Add image-builder-frontend specfile (HMS-5221)

kingsleyzissou:
This PR adds the initial specfile to the project and convenience scripts in the Makefile to build rpms and srpms

/jira-epic COMPOSER-2411

JIRA: [HMS-5221](https://issues.redhat.com/browse/HMS-5221)


kingsleyzissou:
This depends on https://github.com/osbuild/image-builder-frontend/pull/2603

cockpit/image-builder-frontend.spec
```diff
@@ -0,0 +1,39 @@
+# XXX: package name?
+Name:           image-builder-frontend
```
croissanne, line 2: To be able to release it in time, I think we decided to keep the name `cockpit-composer`.

cockpit/image-builder-frontend.spec
```diff
@@ -0,0 +1,39 @@
...
+Version:        0
+Release:        1
```
croissanne, line 4: should be 1%?{dist} I think

kingsleyzissou: Thanks

Makefile
```diff
@@ -50,3 +71,36 @@ cockpit/build: cockpit/download
...
+
+RPM_SPEC=cockpit/$(PACKAGE_NAME).spec
+RPM_TARBALL=rpmbuild/SOURCES/$(PACKAGE_NAME)-$(VERSION).tar.gz
+NODE_MODULES_TEST=package-lock.json
+
+$(RPM_SPEC): $(RPM_SPEC) $(NODE_MODULES_TEST)
+	provides=$$(npm ls --omit dev --package-lock-only --depth=Infinity | grep -Eo '[^[:space:]]+@[^[:space:]]+' | sort -u | sed 's/^/Provides: bundled(npm(/; s/\(.*\)@/\1)) = /'); \
+	awk -v p="$$provides" '{gsub(/%{VERSION}/, "$(VERSION)"); $(SUB_NODE_ENV) gsub(/%{NPM_PROVIDES}/, p)}1' $< > $@
+
+$(RPM_TARBALL): export NODE_ENV ?= production
```
croissanne, line 87: something is wrong here I think, if NODE_ENV is set to production `npm ci` doesn't install the dev dependencies. I think NODE_ENV should be set to production after npm ci is run, but before the build is run

croissanne: We could just remove `npm ci` from the `cockpit/build` target.

kingsleyzissou: Oh right good catch

kingsleyzissou: Okay done

Makefile
```diff
@@ -45,8 +66,40 @@ cockpit/download: Makefile
...
+	touch cockpit/public/*
+	tar czf $(RPM_TARBALL) --transform 's,^,image-builder-frontend/,' \
+		--exclude node_modules \
+		$$(git ls-files) $(RPM_SPEC) $(NODE_MODULES_TEST) cockpit/public/ cockpit/README.md
+	realpath $(RPM_TARBALL)
+
+.PHONY: srpm
+srpm: $(RPM_TARBALL)
+	rpmbuild -bs \
+		--define "_topdir $(CURDIR)/rpmbuild" \
```
croissanne, line 98: `--define "_sourcedir `pwd`" \`  avoids having to juggle with rpmbuild/SOURCES.

Makefile
```diff
@@ -45,8 +66,40 @@ cockpit/download: Makefile
...
+		$(RPM_SPEC)
+
+.PHONY: rpm
+rpm: $(RPM_TARBALL)
+	rpmbuild -bb \
+		--define "_topdir $(CURDIR)/rpmbuild" \
```
croissanne, line 104: same here, _sourcedir.

Makefile
```diff
@@ -45,8 +66,40 @@ cockpit/download: Makefile
...
 
 .PHONY: cockpit/devel
 cockpit/devel: cockpit/devel-uninstall cockpit/build cockpit/devel-install
+
+#
+# Building packages
+#
+
+RPM_SPEC=cockpit/$(PACKAGE_NAME).spec
+RPM_TARBALL=rpmbuild/SOURCES/$(PACKAGE_NAME)-$(VERSION).tar.gz
```
croissanne, line 79: just `$(PACKAGE_NAME)-$(VERSION).tar.gz`, but package_name needs to be `cockpit-composer` now too.

Makefile
```diff
@@ -45,8 +66,40 @@ cockpit/download: Makefile
...
+
+$(RPM_SPEC): $(RPM_SPEC) $(NODE_MODULES_TEST)
+	provides=$$(npm ls --omit dev --package-lock-only --depth=Infinity | grep -Eo '[^[:space:]]+@[^[:space:]]+' | sort -u | sed 's/^/Provides: bundled(npm(/; s/\(.*\)@/\1)) = /'); \
+	awk -v p="$$provides" '{gsub(/%{VERSION}/, "$(VERSION)"); $(SUB_NODE_ENV) gsub(/%{NPM_PROVIDES}/, p)}1' $< > $@
+
+$(RPM_TARBALL): export NODE_ENV ?= production
+$(RPM_TARBALL): cockpit/build
+	touch -r package.json package-lock.json
+	touch cockpit/public/*
+	tar czf $(RPM_TARBALL) --transform 's,^,image-builder-frontend/,' \
```
croissanne, line 90: `image-builder-frontend` can just be `$(PACKAGE_NAME)`

cockpit/image-builder-frontend.spec
```diff
@@ -0,0 +1,37 @@
+Name:           cockpit-composer
+Version:        0
```
croissanne, line 2: Should be version 53, that's the version cockpit-composer is at. Once we release it should be 54. Could just start on 54 to be on the safe side.

croissanne (commented):
Nice! A few comments, also we could add a packit config to start building rpms:

```
upstream_project_url: https://github.com/osbuild/image-builder-frontend
specfile_path: cockpit/image-builder-frontend.spec
upstream_package_name: cockpit-composer
downstream_package_name: cockpit-composer
# use the nicely formatted release description from our upstream release, instead of git shortlog
copy_upstream_release_description: true

actions:
  create-archive: make dist

srpm_build_deps:
  - make
  - npm

jobs:
  - job: copr_build
    trigger: pull_request
    targets: &build_targets
      - centos-stream-9
      - centos-stream-10
      - fedora-all
 ```

File: Makefile [hunk 9]
```diff
@@ -45,8 +66,46 @@ cockpit/download: Makefile
...
 .PHONY: cockpit/devel
 cockpit/devel: cockpit/devel-uninstall cockpit/build cockpit/devel-install
+
+#
+# Building packages
+#
+
+RPM_SPEC=cockpit/$(PACKAGE_NAME).spec
+NODE_MODULES_TEST=package-lock.json
+RPM_NAME=cockpit-composer
```
croissanne, line 80: since rpm_name and package_name are the same now, we might as well use package_name everywhere I think.

croissanne (dismissed):

kingsleyzissou:
Ah right, I put it in the cockpit dir. I'll move it to the top level

thozza:
On a general note, wouldn't it make more sense to ship this as `image-builder-frontend` or `cockpit-image-builder` RPM? Sure, a bit more work to add a new package to Fedora, but it would look much cleaner to me. The new RPM can then `Obsolete` `cockpit-composer`, so the upgrade path would be preserved. :thinking:

lucasgarfield:
> On a general note, wouldn't it make more sense to ship this as `image-builder-frontend` or `cockpit-image-builder` RPM? Sure, a bit more work to add a new package to Fedora, but it would look much cleaner to me. The new RPM can then `Obsolete` `cockpit-composer`, so the upgrade path would be preserved. 🤔

This makes good sense to me. I'm not sure that either myself or @kingsleyzissou have enough experience packing in Fedora to make this happen in a reasonable amount of time... we'd need to rely on your help pretty heavily @thozza, would you mind helping us get this into Rawhide?

ondrejbudai:
It slightly depends on the timeframe. If we want to ship it in RHEL 10.0, the cockpit-composer shortcut (while being ugly) might be the only way.

supakeen:
> On a general note, wouldn't it make more sense to ship this as `image-builder-frontend` or `cockpit-image-builder` RPM? Sure, a bit more work to add a new package to Fedora, but it would look much cleaner to me. The new RPM can then `Obsolete` `cockpit-composer`, so the upgrade path would be preserved. 🤔

This does make sense for Fedora though we might run across issues during review as it's not clear which `image-builder` this is. The name would be consistent with the other package we want to submit for review in Fedora `image-builder-cli` so I assume we'd get comments like that there first.

@ondrejbudai That might work for RHEL but it'd be very weird to do in Fedora, is there a way we can split this or is it too much maintenance?

croissanne (approved):

supakeen (approved):

packit.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - make
+  - npm
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    targets: &build_targets
+      - centos-stream-9
+      - centos-stream-10
+      - fedora-all
```
thozza, line 23: Nitpick for potential followup: would it make sense to build RPMs for all architectures? These aliases will ensure builds only for x86_64.

thozza (approved):
:+1:

packit.yaml
croissanne: Yup, will followup immediately.

PR #2704, opened 2024-12-20
on-prem: add ability to delete blueprints (HMS-5224)

kingsleyzissou:
Subject says it all, this PR adds the ability to delete blueprints using the on-prem version of the frontend.

/jira-epic COMPOSER-2411

JIRA: [HMS-5224](https://issues.redhat.com/browse/HMS-5224)

### DIFF
--- a/src/store/cockpitApi.ts
+++ b/src/store/cockpitApi.ts
@@ -1,3 +1,5 @@
+import path from 'path';
+
 // Note: for the on-prem version of the frontend we have configured
 // this so that we check `node_modules` and `pkg/lib` for packages.
 // To get around this for the hosted service, we have configured
@@ -14,6 +16,8 @@ import {
   GetArchitecturesApiArg,
   GetBlueprintsApiArg,
   GetBlueprintsApiResponse,
+  DeleteBlueprintApiResponse,
+  DeleteBlueprintApiArg,
   BlueprintItem,
 } from './imageBuilderApi';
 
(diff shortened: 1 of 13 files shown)


src/store/enhancedCockpitApi.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const enhancedApi = cockpitApi.enhanceEndpoints({
+  addTagTypes: ['Blueprints'],
+  endpoints: {
+    getBlueprints: {
+      providesTags: () => {
+        return [{ type: 'Blueprints' }];
+      },
+    },
+    deleteBlueprint: {
```
kingsleyzissou, line 22: This is duplicate code, but it was kind of difficult to extract this because of all of the different types.

src/store/cockpitApi.ts
```diff
@@ -89,8 +91,34 @@ export const cockpitApi = emptyCockpitApi.injectEndpoints({
...
       }),
+      deleteBlueprint: builder.mutation<
+        DeleteBlueprintApiResponse,
+        DeleteBlueprintApiArg
+      >({
+        // invalidatesTags: [{ type: 'Blueprints' }],
+        queryFn: async ({ id: filename }) => {
+          try {
+            const path = await getBlueprintsPath();
+            const filepath = path + '/' + filename;
```
ochosi, line 102: Wouldn't it be safer to use `path.join()`?

src/store/cockpitApi.ts
```diff
@@ -89,8 +91,34 @@ export const cockpitApi = emptyCockpitApi.injectEndpoints({
...
+
+            await cockpit.spawn(['rm', '-r', filepath], {
```
ochosi, line 104: Does the `-r` help here?
If it's a file, it will be deleted without the argument, too. If it's a non-empty directory, the call will fail.

src/store/cockpitApi.ts
kingsleyzissou: Ya whoops, it would be a lot better

src/store/cockpitApi.ts
kingsleyzissou: Think I meant to put an -f flag rather than the -r. But maybe we should leave that off too

lucasgarfield:
Just started looking at this, deleting isn't working for me, I see this in the console:
image

lucasgarfield (approved):
Working nicely!

PR #2716, opened 2025-01-05
wizard: add user name validation (HMS-5285)

mgold1234:
this commit add validation to name field in Users step

JIRA: [HMS-5285](https://issues.redhat.com/browse/HMS-5285)

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -41,6 +41,7 @@ import {
   useRegistrationValidation,
   useHostnameValidation,
   useKernelValidation,
+  useUsersValidation,
 } from './utilities/useValidation';
 import {
   isAwsAccountIdValid,
@@ -230,6 +231,8 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   const firstBootValidation = useFirstBootValidation();
   // Details
   const detailsValidation = useDetailsValidation();
+  // Users
+  const usersValidation = useUsersValidation();
 
   let startIndex = 1; // default index
   if (isEdit) {
@@ -462,7 +465,10 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 key="wizard-users"
                 isHidden={!isUsersEnabled}
                 footer={
-                  <CustomWizardFooter disableNext={false} optional={true} />
+                  <CustomWizardFooter
+                    disableNext={usersValidation.disabledNext}
+                    optional={true}
+                  />
                 }
               >
                 <UsersStep />
--- a/src/Components/CreateImageWizard/validators.ts
+++ b/src/Components/CreateImageWizard/validators.ts
@@ -57,6 +57,17 @@ export const isFileSystemConfigValid = (partitions: Partition[]) => {
   return duplicates.length === 0;
 };
 
+export const isUserNameValid = (userName: string) => {
+  if (userName === undefined) return false;
+  const isLengthValid = userName.length <= 32;
+  const isNotNumericOnly = !/^\d+$/.test(userName);
+  const isPatternValid = /^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9_$]$/.test(
+    userName
+  );
+
+  return isLengthValid && isNotNumericOnly && isPatternValid;
+};
+
 export const getDuplicateMountPoints = (partitions: Partition[]): string[] => {
   const mountPointSet: Set<string> = new Set();
   const duplicates: string[] = [];
--- a/src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
@@ -72,6 +72,24 @@ const addValidUser = async () => {
   await waitFor(() => expect(nextButton).toBeEnabled());
 };
 
+const addInvalidUser = async () => {
+  const user = userEvent.setup();
+  const addUser = await screen.findByRole('button', { name: /add a user/i });
+  expect(addUser).toBeEnabled();
+  await waitFor(() => user.click(addUser));
+  const enterUserName = screen.getByRole('textbox', {
+    name: /blueprint user name/i,
+  });
+  const nextButton = await getNextButton();
+  await waitFor(() => user.type(enterUserName, '..'));
+  await waitFor(() => expect(enterUserName).toHaveValue('..'));
+  const enterSshKey = await screen.findByRole('textbox', {
+    name: /public SSH key/i,
+  });
+  await waitFor(() => user.type(enterSshKey, 'ssh-rsa d'));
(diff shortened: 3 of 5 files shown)


src/Components/CreateImageWizard/validators.ts
```diff
@@ -67,6 +67,19 @@ export const isFileSystemConfigValid = (partitions: Partition[]) => {
   return duplicates.length === 0;
 };
 
+export const isUserNameValid = (userName: string) => {
```
lucasgarfield, line 70: Let's use the rules from `man useradd` [as discussed back in November](https://github.com/osbuild/image-builder-frontend/pull/2551#discussion_r1849119104):

```

CAVEATS
       You may not add a user to a NIS or LDAP group. This must be performed on the corresponding server.

       Similarly, if the username already exists in an external user database such as NIS or LDAP, useradd will deny the user account
       creation request.

       Usernames may contain only lower and upper case letters, digits, underscores, or dashes. They can end with a dollar sign. Dashes
       are not allowed at the beginning of the username. Fully numeric usernames and usernames . or .. are also disallowed. It is not
       recommended to use usernames beginning with . character as their home directories will be hidden in the ls output.

       Usernames may only be up to 32 characters long.
```

lucasgarfield: You can/should play around with `useradd` yourself if something isn't completely clear.

For instance, look at this terminal output - `.` and `..` are both explicitly forbidden, but `...` is actually allowed (exactly as the man page says):
image

Remember, our goal with validation is always to introduce the least friction/work possible while simultaneously making sure that the image builds. We know that `useradd` will be called during the osbuild stages. Our goal right now is to make sure that the `useradd` command in that stage succeeds but also to be as permissive/unopinionated as possible about what the user actually does (if `useradd` allows it, so should we).

lucasgarfield (dismissed):

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -457,7 +460,9 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 key="wizard-users"
                 isHidden={!isUsersEnabled}
                 footer={
-                  <CustomWizardFooter disableNext={false} optional={true} />
+                  <CustomWizardFooter
+                    disableNext={usersValidation.disabledNext}
+                  />
```
regexowl, line 465: The `optional` prompt got lost in the process

```suggestion
                  <CustomWizardFooter
                    disableNext={usersValidation.disabledNext}
                    optional={true}
                  />
```

src/Components/CreateImageWizard/validators.ts
mgold1234: I think I cover all the options
```
  if (userName === undefined) return false;
  const isLengthValid = userName.length >= 1 && userName.length <= 32;
  const isNotDotOrDotDot = userName !== '.' && userName !== '..';
  const isNotNumericOnly = !/^\d+$/.test(userName);
  const isPatternValid = /^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9_$]$/.test(
    userName
  );
  const doesNotStartWithDash = !/^[-]/.test(userName);

  return (
    isLengthValid &&
    isNotDotOrDotDot &&
    isNotNumericOnly &&
    isPatternValid &&
    doesNotStartWithDash
  );
};
```

src/Components/CreateImageWizard/validators.ts
```diff
@@ -67,6 +67,25 @@ export const isFileSystemConfigValid = (partitions: Partition[]) => {
...
+  if (userName === undefined) return false;
+  const isLengthValid = userName.length >= 1 && userName.length <= 32;
+  const isNotDotOrDotDot = userName !== '.' && userName !== '..';
```
avitova, line 73: Is this not included in the regex isPatternValid? Deleting this condition did not allow adding the dots and double dots into the username. :thinking:

avitova (commented):
Amazing! This will be lots of work, so I think it's a good idea to split this into more pull requests. Do we maybe also want to start adding tests with each pull request? Or what's the plan for that?

src/Components/CreateImageWizard/validators.ts
mgold1234: yes, thank you for that!

src/Components/CreateImageWizard/validators.ts
```diff
@@ -67,6 +67,21 @@ export const isFileSystemConfigValid = (partitions: Partition[]) => {
...
 
+export const isUserNameValid = (userName: string) => {
+  if (userName === undefined) return false;
+  const isLengthValid = userName.length >= 1 && userName.length <= 32;
+
+  const isNotNumericOnly = !/^\d+$/.test(userName);
+  const isPatternValid = /^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9_$]$/.test(
+    userName
+  );
+  const doesNotStartWithDash = !/^[-]/.test(userName);
```
regexowl, line 78: This is already included in `isPatternValid`. The first character is limited to `[a-zA-Z0-9]`

src/Components/CreateImageWizard/validators.ts
```diff
@@ -67,6 +67,21 @@ export const isFileSystemConfigValid = (partitions: Partition[]) => {
   return duplicates.length === 0;
 };
 
+export const isUserNameValid = (userName: string) => {
+  if (userName === undefined) return false;
```
regexowl, line 71: Should we allow the user to proceed if no data was filled in on the Users step? I mean what if I click on "Add user" because I'm curious about the functionality, but then I see that I'd need to add a password/ssh key and I change my mind. I'd be kinda stuck, right?

src/Components/CreateImageWizard/validators.ts
regexowl: The `doesNotStartWithDash` I mean

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
avitova, line 1: Nice, thanks for including tests. :+1: I think I am good here now, I agree with what Klara commented on, other than that, I have no further comments.

src/Components/CreateImageWizard/validators.ts
regexowl: Also the min length based on `isPatternValid` is already two characters. The `userName.length >= 1` in `isLengthValid` is not needed.

src/Components/CreateImageWizard/validators.ts
regexowl: The button would also work :+1: agree with whatever is the easiest right now.

croissanne (approved):
lgtm!

croissanne (dismissed):

mgold1234:
@croissanne removed the other commit to different pr, becuase it seems a different work, sorry for that

croissanne (approved):

regexowl (approved):
Thank you! :heart:

PR #2717, opened 2025-01-06
wizard: add Administrator checkbox to users step (HMS-4903)

mgold1234:
this commit add Administrator checkbox to users step


src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -17,6 +19,8 @@ const UserInfo = () => {
   const userName = useAppSelector(userNameSelector);
   const userPasswordSelector = selectUserPasswordByIndex(index);
   const userPassword = useAppSelector(userPasswordSelector);
+  const userIsAdministratorSelector = selectUserAdministrator(index);
+  const userAdministrator = useAppSelector(userIsAdministratorSelector);
```
regexowl, line 23: For consistency

```suggestion
  const userIsAdministrator = useAppSelector(userIsAdministratorSelector);
```

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -32,6 +36,15 @@ const UserInfo = () => {
     dispatch(setUserPasswordByIndex({ index: index, password: value }));
   };
 
+  const handleCheckboxChange = (
+    _event: React.FormEvent<HTMLInputElement>,
+    value: boolean
+  ) => {
+    dispatch(
+      setUserAdministratorByIndex({ index: index, isAdministrator: value })
```
regexowl, line 44:
```suggestion
      setUserIsAdministratorByIndex({ index: index, isAdministrator: value })
```

src/store/wizardSlice.ts
```diff
@@ -368,6 +375,11 @@ export const selectUserPasswordByIndex =
     return state.wizard.users[userIndex]?.password;
   };
 
+export const selectUserAdministrator =
```
regexowl, line 378:
```suggestion
export const selectUserIsAdministrator =
```

regexowl:
An empty `groups` array hangs in the request if there are no groups added, causing the error in `Users.test.tsx` tests. Might need something like "if there's at least one group then add the groups array to the request".

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -587,6 +587,9 @@ const getUsers = (state: RootState): User[] | undefined => {
     if (user.ssh_key !== '') {
       result.ssh_key = user.ssh_key;
     }
+    if (user.groups !== undefined) {
```
lucasgarfield, line 590: I don't think user.groups can be undefined, its type is `string[]`. You can (and should) always use your IDE to inspect types.

image

lucasgarfield (changes requested):
Looking good, works nicely. 

Please add tests for both `edit` and `import` mode where `isAdmin === true`.

src/store/wizardSlice.ts
```diff
@@ -838,6 +850,20 @@ export const wizardSlice = createSlice({
...
       state.users[action.payload.index].ssh_key = action.payload.sshKey;
     },
+    setUserAdministratorByIndex: (
+      state,
+      action: PayloadAction<UserAdministratorPayload>
+    ) => {
+      const { index, isAdministrator } = action.payload;
+      state.users[index].isAdministrator = isAdministrator;
+      if (isAdministrator) {
+        state.users[index].groups?.push('wheel');
```
lucasgarfield, line 860: Do we need optional chaining here?

src/store/wizardSlice.ts
```diff
@@ -378,6 +385,11 @@ export const selectUserSshKeyByIndex =
     return state.wizard.users[userIndex]?.ssh_key;
   };
 
+export const selectUserAdministrator =
+  (userIndex: number) => (state: RootState) => {
+    return state.wizard.users[userIndex]?.isAdministrator;
```
lucasgarfield, line 390: Do we need optional chaining here?

src/store/wizardSlice.ts
```diff
@@ -838,6 +850,20 @@ export const wizardSlice = createSlice({
...
+    setUserAdministratorByIndex: (
+      state,
+      action: PayloadAction<UserAdministratorPayload>
+    ) => {
+      const { index, isAdministrator } = action.payload;
+      state.users[index].isAdministrator = isAdministrator;
+      if (isAdministrator) {
+        state.users[index].groups?.push('wheel');
+      } else {
+        state.users[index].groups = state.users[index].groups?.filter(
```
lucasgarfield, line 862: Do we need optional chaining here?

lucasgarfield (changes requested):

src/Components/CreateImageWizard/utilities/requestMapper.ts
mgold1234: Screenshot 2025-01-14 at 14 42 36
that's very intersting, because I see that groups can be undefined

mgold1234: without this condition, when I am editing the image, the Administrator checkbox always False

src/store/wizardSlice.ts
mgold1234: you correct, we define the user without undifined value

src/Components/CreateImageWizard/utilities/requestMapper.ts
avitova: IDEs are not always reliable in these kinds of things. I am just curious, which IDEs do you guys use, and do you have any extensions? :thinking: 
I think that your IDE might be misinterpreting the type, Michal. The `user` here comes from iterating through the array of `users`. It is of type `UserWithAdditionalInfo`, which is defined as:
`export type UserWithAdditionalInfo = {
  [K in keyof User]-?: NonNullable<User[K]>;
} & {
  isAdministrator: boolean;
};
`
[In the docs](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) they explain the usage of `-?` as removing optional attributes from type's properties. So it should not matter. Can you check again? We can try to debug this.
Damn, I did not know about `-?`, I love that I learned something today.

avitova (commented):
Thanks, I genuinely love that we have this split into more pull requests!

src/Components/CreateImageWizard/utilities/requestMapper.ts
mgold1234: yes, my IDE is Goland, and it is misinterpreting the type :)
Lucas advice me to start working with VS

mgold1234: solve it with 
# if (user.groups.length > 0) {
  result.groups = user.groups;
}

regexowl: Just a heads up - working on adding a rule that should help us with finding problems like this in the future: https://github.com/osbuild/image-builder-frontend/pull/2755

src/store/wizardSlice.ts
```diff
@@ -409,7 +421,7 @@ export const selectBlueprintDescription = (state: RootState) => {
 };
 
 export const selectFirstBootScript = (state: RootState) => {
-  return state.wizard.firstBoot?.script;
+  return state.wizard.firstBoot.script;
```
regexowl, line 424: This looks like an unrelated change, can you please remove it from this PR?

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -141,6 +141,40 @@ describe('Step Users', () => {
...
+      await goToUsersStep();
+      await addValidUser();
+      const isAdmin = screen.getByRole('checkbox', {
+        name: /administrator/i,
+      });
+      user.click(isAdmin);
+      await goToReviewStep();
+      // informational modal pops up in the first test only as it's tied
+      // to a 'imageBuilder.saveAndBuildModalSeen' variable in localStorage
+      //    await openAndDismissSaveAndBuildModal();
```
regexowl, line 162: This can be removed completely, the modal opens only on first try, which is covered in previous test.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -141,6 +141,40 @@ describe('Step Users', () => {
         },
       };
 
+      await waitFor(() => {
+        expect(receivedRequest).toEqual(expectedRequest);
+      });
+    });
+    test('with valid name and password and check the Administrator checkbox', async () => {
```
regexowl, line 148:
```suggestion
    test('with valid name, ssh key and checked Administrator checkbox', async () => {
```

regexowl (changes requested):
Added a few nitpicks. otherwise looks good :)

regexowl (approved):
I think this looks good.

`import` tests will need to be added after adding user customization to on-prem mapper. Adding `edit` tests might be safer when the step functionality is fully implemented as well. That way we don't forget to update them with new fields. 

I've created a [checklist task in JIRA](https://issues.redhat.com/browse/HMS-5365) to track what was already done and what still needs to be addressed.

PR #2727, opened 2025-01-08
wizard: create new component with eyeIcon button inside textInput (HMS-5289)

mgold1234:
this commit create new component with eyeIcon button inside textInput,
for password field
Screenshot 2025-01-09 at 13 21 59

JIRA: [HMS-5289](https://issues.redhat.com/browse/HMS-5289)

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -7,7 +7,11 @@ import {
   TextAreaProps,
   TextInput,
   TextInputProps,
+  Button,
+  InputGroup,
+  InputGroupItem,
 } from '@patternfly/react-core';
+import { EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
 
 import type { StepValidation } from './utilities/useValidation';
 
@@ -33,6 +37,57 @@ type HookValidatedInputPropTypes = TextInputProps &
     warning?: string;
     inputType?: 'textInput' | 'textArea';
   };
+
+export const HookPasswordValidatedInput = ({
+  ariaLabel,
+  placeholder,
+  dataTestId,
+  value,
+  ouiaId,
+  stepValidation,
+  fieldName,
+  onChange,
+  warning = undefined,
+  inputType,
+  isDisabled,
+}: HookValidatedInputPropTypes) => {
+  const [isPasswordVisible, setIsPasswordVisible] = useState(false);
+  const togglePasswordVisibility = () => {
+    setIsPasswordVisible(!isPasswordVisible);
+  };
+
+  return (
+    <>
+      <InputGroup>
+        <InputGroupItem isFill>
+          <HookValidatedInput
+            type={isPasswordVisible ? 'text' : 'password'}
+            ouiaId={ouiaId || ''}
+            data-testid={dataTestId}
+            value={value}
+            onChange={onChange!}
+            stepValidation={stepValidation}
+            ariaLabel={ariaLabel || ''}
+            fieldName={fieldName}
+            placeholder={placeholder || ''}
+            inputType={inputType || 'textInput'}
+            warning={warning || ''}
+            isDisabled={isDisabled || false}
+          />
+        </InputGroupItem>
+        <InputGroupItem>
+          <Button
+            variant="control"
+            onClick={togglePasswordVisibility}
+            aria-label={isPasswordVisible ? 'Show password' : 'Hide password'}
+          >
+            {isPasswordVisible ? <EyeSlashIcon /> : <EyeIcon />}
+          </Button>
+        </InputGroupItem>
+      </InputGroup>
+    </>
+  );
+};
 
 export const HookValidatedInput = ({
   dataTestId,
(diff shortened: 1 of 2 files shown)


lucasgarfield:
Some inspiration for you, we should try to follow PF as closely as possible:
https://v5-archive.patternfly.org/components/password-generator

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,65 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+  placeholder,
+  dataTestId,
+  value,
+  ouiaId,
+  stepValidation,
+  fieldName,
+  onChange,
+  type = 'text',
+  isPasswordVisible,
+  togglePasswordVisibility,
```
lucasgarfield, line 54: See my comment below - I think we can remove the `isPasswordVisible` and `togglePasswordVisibility` props and just keep track of their state inside of this component.

lucasgarfield (changes requested):

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,65 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
   warning?: string;
 }
 
+interface HookPasswordValidatedTextInputPropTypes
```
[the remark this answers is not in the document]
mgold1234: I remove this interface

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,61 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+  const togglePasswordVisibility = () => {
+    setIsPasswordVisible(!isPasswordVisible);
+  };
+
+  return (
+    <>
+      <InputGroup>
+        <InputGroupItem isFill>
+          <HookValidatedInput
+            isRequired
```
regexowl, line 58: Hmm, should this be marked as a required field? The user can either fill in the password or ssh key, right? So it's kinda required, but not really. I guess we can leave it now, but could maybe check how to indicate that password or ssh key need to be filled in.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,61 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+            type={isPasswordVisible ? 'text' : 'password'}
+            ouiaId={ouiaId || ''}
+            data-testid={dataTestId}
+            aria-label={ariaLabel || ''}
```
regexowl, line 62: There are duplicate aria labels in the props - `aria-label` and `ariaLabel`. `isDisabled` prop is missing. `warning` is missing also.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,61 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
...
+          </Button>
+        </InputGroupItem>
+      </InputGroup>
+      {warning !== undefined && (
+        <HelperText>
+          <HelperTextItem variant="warning" hasIcon>
+            {warning}
+          </HelperTextItem>
+        </HelperText>
+      )}
```
regexowl, line 87: Wouldn't this render duplicate warning? Warning should already get rendered in `HookValidatedInput`.

regexowl (commented):
Added a few comments, but maybe we're overlooking something since there's no password validation yet. Can we add that (could be in a separate PR) and then revisit this one to see how the validated input works with the added button?

src/Components/CreateImageWizard/ValidatedTextInput.tsx
mgold1234: I will add them as part of rest., what do you think?

mgold1234: I will pass some of the field as rest...

src/Components/CreateImageWizard/ValidatedTextInput.tsx
mgold1234: I will add the check if password or ssh key filled in the validation function

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -30,6 +34,45 @@ interface HookValidatedTextInputPropTypes extends TextInputProps {
   warning?: string;
 }
 
+export const HookPasswordValidatedInput = ({
+  ariaLabel,
+  value,
+  stepValidation,
+  fieldName,
+  ...rest
```
lucasgarfield, line 42: Generally in the code base we try to avoid prop spreading. It makes things less explicit. The official React docs even caution to "use spread syntax with restraint": https://react.dev/learn/passing-props-to-a-component#forwarding-props-with-the-jsx-spread-syntax

Would you mind eliminating the prop spreading here?

lucasgarfield (changes requested):
This looks great, exactly like the Patternfly example, and is working nicely when switching between visible/not visible. Just one requested change and we should be able to get it merged. :smile:

lucasgarfield (approved):
Nice work! :smile:

PR #2734, opened 2025-01-10
Wizard: Add kernel append input (HMS-5299)

regexowl:
This adds the kernel append input, removes kernel append from the OpenSCAP step and adds basic tests to cover the functionality.

TO DO:
- [x] parse "Required by OpenSCAP" correctly on edit
- [x] check for duplicate arguments

JIRA: [HMS-5299](https://issues.redhat.com/browse/HMS-5299)

### DIFF
--- a/src/Components/CreateImageWizard/ChippingInput.tsx
+++ b/src/Components/CreateImageWizard/ChippingInput.tsx
@@ -50,19 +50,19 @@ const ChippingInput = ({
   };
 
   const addItem = (value: string) => {
-    if (validator(value) && !list?.includes(value)) {
-      dispatch(addAction(value));
-      setInputValue('');
-      setErrorText('');
-    }
-
-    if (list?.includes(value)) {
+    if (list?.includes(value) || requiredList?.includes(value)) {
       setErrorText(`${item} already exists.`);
+      return;
     }
 
     if (!validator(value)) {
       setErrorText('Invalid format.');
+      return;
     }
+
+    dispatch(addAction(value));
+    setInputValue('');
+    setErrorText('');
   };
 
   const handleKeyDown = (e: React.KeyboardEvent, value: string) => {
--- a/src/Components/CreateImageWizard/ChippingInput.tsx
+++ b/src/Components/CreateImageWizard/ChippingInput.tsx
@@ -19,6 +19,7 @@ type ChippingInputProps = {
   ariaLabel: string;
   placeholder: string;
   validator: (value: string) => boolean;
+  requiredList?: string[] | undefined;
   list: string[] | undefined;
   item: string;
   addAction: (value: string) => UnknownAction;
@@ -30,6 +31,7 @@ const ChippingInput = ({
   placeholder,
   validator,
   list,
+  requiredList,
   item,
   addAction,
   removeAction,
@@ -64,7 +66,7 @@ const ChippingInput = ({
   };
 
   const handleKeyDown = (e: React.KeyboardEvent, value: string) => {
-    if (e.key === ' ' || e.key === 'Enter' || e.key === ',') {
+    if (e.key === ' ' || e.key === 'Enter') {
       e.preventDefault();
       addItem(value);
     }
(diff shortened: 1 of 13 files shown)


src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
```diff
@@ -1,9 +1,133 @@
...
+      if (
+        isKernelArgumentValid(value) &&
+        !kernelAppend.some((arg) => arg.name === value)
+      ) {
+        dispatch(addKernelArg({ name: value, isRequiredByOpenSCAP: false }));
+        setInputValue('');
+        setErrorText('');
+      }
+
+      if (kernelAppend.some((arg) => arg.name === value)) {
```
mgold1234, line 51: should I see this error in case I am adding the same kernel name?
now I dont see any error when adding the same name

mgold1234:
I think there is a bug when creating an image with OScap -> required kernel added to the kernel step,
and if you open the edit image, the required kernel change to false
https://github.com/user-attachments/assets/13668299-9b04-4b45-8a1f-c185646147a9

regexowl:
> I think there is a bug when creating an image with OScap -> required kernel added to the kernel step, and if you open the edit image, the required kernel change to false https://github.com/user-attachments/assets/13668299-9b04-4b45-8a1f-c185646147a9 Screenshot 2025-01-13 at 12 54 07

Ooh, I left the default parsing there, didn't I? Great catch! Will revert this to draft

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
regexowl: Tried to add a duplicate and got the correct error
image

What arguments did you try to add?

mgold1234: I tried 'true' value

mgold1234: ok, I got you, I think I check rhat I am adding 2 time the same value 'test',
if I tried to add same value that comes from the oscap profile the validation is working. 
but, maybe we also want to check if user try to add same value ?

regexowl:
The chipping input might be extractable from this one as well, moving to draft for the time being. https://github.com/osbuild/image-builder-frontend/pull/2747 will need to get merged first.

src/Components/CreateImageWizard/ChippingInput.tsx
```diff
@@ -64,7 +66,7 @@ const ChippingInput = ({
   };
 
   const handleKeyDown = (e: React.KeyboardEvent, value: string) => {
-    if (e.key === ' ' || e.key === 'Enter' || e.key === ',') {
+    if (e.key === ' ' || e.key === 'Enter') {
```
regexowl, line 69: Comma should be a valid character for kernel argument, I've removed it from the keyDown for the consistency sake.

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
```diff
@@ -2,8 +2,57 @@ import React from 'react';
...
+    {
+      skip: !complianceProfileID,
+    }
+  );
+
+  const requiredByOpenSCAP = kernelAppend.filter((arg) =>
+    oscapProfileInfo?.kernel?.append?.split(' ').includes(arg)
+  );
+
+  const notRequiredByOpenSCAP = kernelAppend.filter(
```
kingsleyzissou, line 38: Nitpick: could you not just have `!requiredByOpenSCAP`?

kingsleyzissou (commented):
Looks good, just a small nitpick that I added inline, but should block the PR.

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
regexowl: Ah right, the code was a bit redundant :sweat_smile: thank you! Added a new commit that should fix that. Also found one small bug so added a fix as well.

kingsleyzissou (approved):

PR #2737, opened 2025-01-12
wizard: add support of TextArea for ssh_key field (HMS-5304)

mgold1234:
this commit add support of TextArea for ssh_key field

JIRA: [HMS-5304](https://issues.redhat.com/browse/HMS-5304)

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -3,6 +3,8 @@ import React, { useState } from 'react';
 import {
   HelperText,
   HelperTextItem,
+  TextArea,
+  TextAreaProps,
   TextInput,
   TextInputProps,
 } from '@patternfly/react-core';
@@ -19,16 +21,18 @@ interface ValidatedTextInputPropTypes extends TextInputProps {
   placeholder?: string;
 }
 
-interface HookValidatedTextInputPropTypes extends TextInputProps {
-  dataTestId?: string | undefined;
-  ouiaId?: string;
-  ariaLabel: string | undefined;
-  value: string;
-  placeholder?: string;
-  stepValidation: StepValidation;
-  fieldName: string;
-  warning?: string;
-}
+type HookValidatedInputPropTypes = TextInputProps &
+  TextAreaProps & {
+    dataTestId?: string | undefined;
+    ouiaId?: string;
+    ariaLabel: string | undefined;
+    value: string;
+    placeholder?: string;
+    stepValidation: StepValidation;
+    fieldName: string;
+    warning?: string;
+    inputType?: 'textInput' | 'textArea';
+  };
 
 export const HookValidatedInput = ({
   dataTestId,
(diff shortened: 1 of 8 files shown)


src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -19,16 +21,18 @@ interface ValidatedTextInputPropTypes extends TextInputProps {
...
+  TextAreaProps & {
+    dataTestId?: string | undefined;
+    ouiaId?: string;
+    ariaLabel: string | undefined;
+    value: string;
+    placeholder?: string;
+    stepValidation: StepValidation;
+    fieldName: string;
+    warning?: string;
+    typeInput?: string;
```
lucasgarfield, line 34: Instead of making the type for this 'string', let's make the API more user friendly and explicit and set the type to either `'textInput' | 'textArea'`.

Then we could also rename this to 'type', which would be a bit more idiomatic for Patternfly-esque components.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -19,16 +21,18 @@ interface ValidatedTextInputPropTypes extends TextInputProps {
...
-  dataTestId?: string | undefined;
-  ouiaId?: string;
-  ariaLabel: string | undefined;
-  value: string;
-  placeholder?: string;
-  stepValidation: StepValidation;
-  fieldName: string;
-  warning?: string;
-}
+type HookValidatedTextInputPropTypes = TextInputProps &
```
lucasgarfield, line 24: We probably need to rename this component now that it is no longer a text input but supports both `input` and `area`. Maybe just `HookValidatedInput`.

src/Components/CreateImageWizard/ValidatedTextInput.tsx
mgold1234: we already have type, so I am changing it to inputType

src/Components/CreateImageWizard/ValidatedTextInput.tsx
```diff
@@ -19,16 +21,18 @@ interface ValidatedTextInputPropTypes extends TextInputProps {
...
-  ouiaId?: string;
-  ariaLabel: string | undefined;
-  value: string;
-  placeholder?: string;
-  stepValidation: StepValidation;
-  fieldName: string;
-  warning?: string;
-}
+type HookValidatedInputPropTypes = TextInputProps &
+  TextAreaProps & {
```
avitova, line 25: I was wondering why are we doing an intersection type instead of creating ValidatedTextInput and ValidatedTextArea components separately. I guess most of the logic is the same, but still, I am not sure if this will not be a complication in the future.
Btw if we really want to do this, we should also rename the name of the file, right?

mgold1234: regarding component separately, I tihnk it makes sense to avoid it if most of the logic is the same, and regarding the name file I think you right

lucasgarfield: Hm, you might be right @avitova. I'm the one who originally suggested trying to DRY out the code (it really seemed like a lot of duplication) but maybe it was just a "false" duplication... now that you point out the type intersection I think I'm inclined to agree with you.

I think all these validated inputs need some refactoring anyway. I'd say let's go ahead and merge this for now (it's working nicely) and I'll try to see what we could refactor it to look like.

lucasgarfield (approved):
Working nicely!

PR #2739, opened 2025-01-13
on-prem add blueprint pagination and fix filtering (HMS-5306)

kingsleyzissou:
This PR adds pagination support for on-prem blueprints and fixes the filtering of blueprints
when trying to search for blueprints.

/jira-epic COMPOSER-2411

JIRA: [HMS-5306](https://issues.redhat.com/browse/HMS-5306)


lucasgarfield (approved):
I've tested, working nicely.

We do have a new filtering utility function we could use for enhancing the search a bit (maybe) but that can be follow-up since other PRs depend on this.

Issue #2718, opened 2025-01-06
Wizard: Expandable should render only if there's at least one value


regexowl:
Expandables on the Review step are currently rendered even with no customizations for the relevant step. We should render only expandables that include customized values.

![image](https://github.com/user-attachments/assets/313d49a1-773b-4250-9d4b-469aacfa955e)

mgold1234:
fixed

PR #2740, opened 2025-01-13
wizard: expand steps only if there are values in the step

mgold1234:
add condition that check if there is value in Hostname, Locale, Timezone steps
and if there is no value, we dont see the extendable part in review step 

FIX ISSUE: (#2718)

JIRA: [HMS-5337](https://issues.redhat.com/browse/HMS-5337)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
@@ -46,6 +46,11 @@ import {
   selectDistribution,
   selectImageTypes,
   selectRegistrationType,
+  selectHostname,
+  selectLanguages,
+  selectKeyboard,
+  selectTimezone,
+  selectNtpServers,
 } from '../../../../store/wizardSlice';
 import { useFlag } from '../../../../Utilities/useGetEnvironment';
 
@@ -60,6 +65,11 @@ const Review = ({ snapshottingEnabled }: { snapshottingEnabled: boolean }) => {
   const complianceProfile = useAppSelector(selectComplianceProfileID);
   const compliancePolicy = useAppSelector(selectCompliancePolicyID);
   const registrationType = useAppSelector(selectRegistrationType);
+  const hostname = useAppSelector(selectHostname);
+  const languages = useAppSelector(selectLanguages);
+  const keyboard = useAppSelector(selectKeyboard);
+  const timezone = useAppSelector(selectTimezone);
+  const ntpServers = useAppSelector(selectNtpServers);
 
   const [isExpandedImageOutput, setIsExpandedImageOutput] = useState(true);
   const [isExpandedTargetEnvs, setIsExpandedTargetEnvs] = useState(true);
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
```diff
@@ -334,7 +346,8 @@ const Review = ({ snapshottingEnabled }: { snapshottingEnabled: boolean }) => {
           <UsersList />
         </ExpandableSection>
       )}
-      {isTimezoneEnabled && (
+      {((isTimezoneEnabled && timezone) ||
+        (isTimezoneEnabled && ntpServers && ntpServers.length > 0)) && (
```
regexowl, line 350: This could be simplified  to checking for the flag as one condition and for any existing values as a second ANDed one. So like this:

```suggestion
      {isTimezoneEnabled &&
        (timezone || (ntpServers && ntpServers.length > 0)) && (
```

src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
```diff
@@ -351,7 +364,8 @@ const Review = ({ snapshottingEnabled }: { snapshottingEnabled: boolean }) => {
           <TimezoneList />
         </ExpandableSection>
       )}
-      {isLocaleEnabled && (
+      {((isLocaleEnabled && languages && languages.length > 0) ||
+        (isLocaleEnabled && keyboard && keyboard.length > 0)) && (
```
regexowl, line 368: Similarly here, first check for the flag which doesn't combine with anything and then check whether there are any values:
```suggestion
      {isLocaleEnabled &&
        ((languages && languages.length > 0) ||
          (keyboard && keyboard.length > 0)) && (
```

src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
```diff
@@ -319,7 +331,7 @@ const Review = ({ snapshottingEnabled }: { snapshottingEnabled: boolean }) => {
         {/* Intentional prop drilling for simplicity - To be removed */}
         <ContentList snapshottingEnabled={snapshottingEnabled} />
       </ExpandableSection>
-      {isUsersEnabled && (
+      {isUsersEnabled && users.length > 0 && (
```
regexowl, line 334: The expandable renders empty when the "Add user" button gets clicked. This ties to https://github.com/osbuild/image-builder-frontend/pull/2716#discussion_r1916314663, option to remove the user or to ignore the user when there are no values.

Let's maybe remove the user check for now and return to it when the step is fully implemented, what do you think?

regexowl (commented):
Added a few nitpicks.

src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
mgold1234: agree thanks

regexowl (approved):
Works nicely, thank you!

PR #2754, opened 2025-01-16
validators: Remove undefined where only string value expected (HMS-5354)

regexowl:
This removes checks for an `undefined` value in validators that take only `string`.

JIRA: [HMS-5354](https://issues.redhat.com/browse/HMS-5354)


mgold1234 (approved):
/lgtm

PR #2755, opened 2025-01-16
ESLint: Set "no-unnecessary-condition" rule to error

regexowl:
This adds "no-unnecessary-condition" and sets the output to error. This should help us eliminate superfluous conditional chaining and uncover potentially faulty conditions.

Docs: https://typescript-eslint.io/rules/no-unnecessary-condition/

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -13,6 +13,8 @@ globals:
   render: 'readonly'
   mount: 'readonly'
 parser: "@typescript-eslint/parser"
+parserOptions:
+  project: ["tsconfig.json"]
 plugins:
   - import
 rules:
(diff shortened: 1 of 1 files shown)


regexowl:
My IDE is clever, I'm usually not (especially when I copy over a part of the code). Let's add a new rule to make our linting tighter.

regexowl:
Originally added all autofix changes, but those break the service. We're combining optional chaining and type assertion in some cases. Meaning we ignore the possibility of `undefined` values, but when we get one everything crashes. Fun. :upside_down_face:

lucasgarfield (approved):
I love it! This is going to really help the review process - one less thing to look for!

lucasgarfield:
> My IDE is clever, I'm usually not (especially when I copy over a part of the code). Let's add a new rule to make our linting tighter.

Hard disagree on the second part :wink:

PR #2758, opened 2025-01-16
POC: cockpit architectures

kingsleyzissou:
This is a proof of concept to see if we can use the cloudapi endpoint to get the arch, 
distrubtion and image type information that we need for creating the image through
the `CreateImageWizard` through the on-prem frontend.


kingsleyzissou:
This is just a POC for now and is still very much a work-in-progress. I will add the jira epic at a later stage.

I am testing this against the new endpoint that @bcl is working on from this [PR](https://github.com/osbuild/osbuild-composer/pull/4336).

kingsleyzissou:
This is not urgent for the MVP. We can revisit this later

PR #2761, opened 2025-01-17
Wizard: Fix repo status formatting

regexowl:
This fixes the formatting of repositories status.


regexowl:
Also this should align out status formatting with Repositories:

image

Not sure if we want to replace "Pending" with "In progress" also...

lucasgarfield (approved):
lgtm

PR #2764, opened 2025-01-19
wizrad: add validation to ssh key field (HMS-5349)

mgold1234:
this commit add validation to ssh key field

JIRA: [HMS-5349](https://issues.redhat.com/browse/HMS-5349)

### DIFF
--- a/src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
@@ -40,6 +40,7 @@ const goToReviewStep = async () => {
   await clickNext(); // Hostname
   await clickNext(); // Kernel
   await clickNext(); // Firewall
+  await clickNext(); // Services
   await clickNext(); // First boot
   await clickNext(); // Details
   await enterBlueprintName();
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -27,6 +27,7 @@ import RegistrationStep from './steps/Registration';
 import RepositoriesStep from './steps/Repositories';
 import ReviewStep from './steps/Review';
 import ReviewWizardFooter from './steps/Review/Footer/Footer';
+import ServicesStep from './steps/Services';
 import SnapshotStep from './steps/Snapshot';
 import Aws from './steps/TargetEnvironment/Aws';
 import Azure from './steps/TargetEnvironment/Azure';
@@ -146,6 +147,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   const isHostnameEnabled = useFlag('image-builder.hostname.enabled');
   const isKernelEnabled = useFlag('image-builder.kernel.enabled');
   const isFirewallEnabled = useFlag('image-builder.firewall.enabled');
+  const isServicesStepEnabled = useFlag('image-builder.services.enabled');
 
   // Remove this and all fallthrough logic when snapshotting is enabled in Prod-stable
   // =========================TO REMOVE=======================
@@ -236,7 +238,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
 
   let startIndex = 1; // default index
   if (isEdit) {
-    startIndex = 21;
+    startIndex = 22;
   }
 
   // Duplicating some of the logic from the Wizard component to allow for custom nav items status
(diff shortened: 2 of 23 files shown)


src/Components/CreateImageWizard/validators.ts
```diff
@@ -57,6 +57,24 @@ export const isFileSystemConfigValid = (partitions: Partition[]) => {
...
+    userName
+  );
+
+  return isLengthValid && isNotNumericOnly && isPatternValid;
+};
+
+export const isSshKeyValid = (sshKey: string) => {
+  const isLengthValid = sshKey !== undefined && sshKey.length >= 2;
+  const isPatternValid =
+    /^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp(256|384|521)) \S+/.test(sshKey);
```
regexowl, line 74: Do you have any source for the regex? If so, could you please add it to a comment so we can reference it if needed?

I wonder if we should try do to a full match instead of a partial one, should be safer that way. We would need to add a few more things to allow for the label part of the key. So maybe something like:
```
^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp(256|384|521))\s+[A-Za-z0-9+\/=]+\s+\S+$
```

What do you think?

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -72,6 +72,42 @@ const addValidUser = async () => {
...
+  await waitFor(() => user.type(enterUserName, '..'));
+  await waitFor(() => expect(enterUserName).toHaveValue('..'));
+  const enterSshKey = await screen.findByRole('textbox', {
+    name: /public SSH key/i,
+  });
+  await waitFor(() => user.type(enterSshKey, 'ssh-rsa d'));
+  await waitFor(() => expect(nextButton).toBeDisabled());
+};
+
+const addInvalidSSHKeyUser = async () => {
```
regexowl, line 93: We're already in the User test suite.

```suggestion
const addInvalidSSHKey = async () => {
```

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -72,6 +72,42 @@ const addValidUser = async () => {
   await waitFor(() => expect(nextButton).toBeEnabled());
 };
 
+const addInvalidUser = async () => {
```
regexowl, line 75: The helper functions could be cleaned up a bit.

I think we should have separate functions for clicking the "Add user" button, a separate function for filling in the username and a function for filling in the ssh key. That way we can chain them and easily use in the tests themselves so they'd be a bit more transparent.

So in the test, I'd expect to see something like:
```
goToUserStep()
clickAddUser()
addUsername(username)
addSshKey(sshKey)
```

That way you can also add a string as an argument and reuse the function for both valid and invalid values where needed.

Currently the `addInvalidUser` and `addInvalidSSHKeyUser` overlap quite a bit. Separating the steps would reduce the duplicity.

regexowl (changes requested):
Added some comments

src/Components/CreateImageWizard/validators.ts
mgold1234: I tried this  ```
 /^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp(256|384|521))\s+[A-Za-z0-9+=]+\s+\S+/.test(
      sshKey
    );
```
and the value 'ssh-rsa d' isnt valid
I took this pattern from 'edge' , they also check the ssh key validation

Comment (regexowl) [reply src/Components/CreateImageWizard/validators.ts]: Hmm, comment might be optional isn't it :thinking: this should work then
```
^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp(256|384|521))\s+[A-Za-z0-9+\/=]+(\s+\S+)?$
```

File: src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx (hunk 3 above)
Comment (mgold1234) [reply src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx]: like this approach thank you!

File: src/Components/CreateImageWizard/validators.ts (hunk 1 above)
Comment (mgold1234) [reply src/Components/CreateImageWizard/validators.ts]: do you think about this comment?
  // 1. Key types: ssh-rsa, ssh-dss, ssh-ed25519, or ecdsa-sha2-nistp(256|384|521).
  // 2. Base64-encoded key material.
  // 3. Optional comment at the end.

File: src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx [hunk 4]
```diff
@@ -201,10 +174,24 @@ describe('Step Users', () => {
...
       await clickRegisterLater();
       await goToUsersStep();
-      await addInvalidUser();
+      await clickAddUser();
+      await addUserName('ss.');
+      await addSshKey('ssh');
       const invalidUserMessage = screen.getByText(/invalid user name/i);
       await waitFor(() => expect(invalidUserMessage));
     });
+
```
regexowl, line 183: 'with invalid name' and 'with invalid ssh key' should be moved to the 'Step Users' describe as they're not testing generated blueprint request.

regexowl (commented):
Looks good! One more nitpick

regexowl:
Did you close the PR on purpose?

mgold1234:
I meant to delete the brach and close it by mistake

PR #2769, opened 2025-01-20
Wizard: Add firewall to Review step (HMS-5357)

regexowl:
This adds a firewall expandable to the Review step.

JIRA: [HMS-5357](https://issues.redhat.com/browse/HMS-5357)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
@@ -76,6 +76,7 @@ import {
   selectUserSshKeyByIndex,
   selectKernel,
   selectUserAdministrator,
+  selectFirewall,
 } from '../../../../store/wizardSlice';
 import { toMonthAndYear, yyyyMMddFormat } from '../../../../Utilities/time';
 import {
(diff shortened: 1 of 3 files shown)


regexowl:
Rebased on https://github.com/osbuild/image-builder-frontend/pull/2767

src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -913,6 +914,60 @@ export const KernelList = () => {
...
+        <TextListItem
+          component={TextListItemVariants.dt}
+          className="pf-v5-u-min-width"
+        >
+          Ports
+        </TextListItem>
+        <TextListItem component={TextListItemVariants.dd}>
+          <CodeBlock>
+            <CodeBlockCode>
+              {firewall.ports ? firewall.ports.join(' ') : 'None'}
```
avitova, line 932: I feel like this was the intention. :thinking: 
```suggestion
              {firewall.ports.length ? firewall.ports.join(' ') : 'None'}
```

avitova (commented):
Also, as I am thinking about it, it might be nice to be able to just add a list of services, divided by some kind of separator. :thinking: But I assume this is MVP, and we still need to see if people are using this?:)

src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
regexowl: Indeed :sweat_smile: updated!

regexowl:
> Also, as I am thinking about it, it might be nice to be able to just add a list of services, divided by some kind of separator. 🤔 But I assume this is MVP, and we still need to see if people are using this?:)

Right, the steps are currently going through the UX review as well. cockpit-composer originally took just a string, I wanted to incorporate some kind of validation as well. But might try to address that in a follow up as I see how that might be really nice to have :thinking: maybe we can just split the entire input string and then just work with the items the same way we currently do?

regexowl:
Oh wait, rebase is needed first :flushed:

PR #2770, opened 2025-01-20
Wizard: Add firewall to on-prem mapper (HMS-5358)

regexowl:
This adds firewall to on-prem mapper.

JIRA: [HMS-5358](https://issues.redhat.com/browse/HMS-5358)


regexowl:
Rebased on https://github.com/osbuild/image-builder-frontend/pull/2767, so that will need to get merged first.

avitova (approved):
This works as expected. Thanks for adding all new things to the on prem mapper. I had an old toml file that I created even with users and timezones - at times we did not have that in wizard. I just uploaded it, and it went through without any problems. :100:

regexowl:
Rebased! The tests that should pass are passing and the review persisted through the rebase. Merging (manually, because pr_check isn't happy, but this PR doesn't affect it in any way)

PR #2784, opened 2025-01-22
devDeps: Manually bump vitest

regexowl:
This bumps vitest and @vitest/coverage-v8 from 2.0.5 to 3.0.3.


regexowl:
Type needed to be added to `UNIQUE_VALIDATION_DELAY` so the vitest could properly mock it. It's a bit weird, but what vitest desires it shall get :shrug: (without the type the build ended in failure)

Also there is some warning mess in the test output now (`Sourcemap for...`). I've tried to find more info and it seems that it might be a regression. `@vitejs/plugin-react` most probably didn't catch up to the update yet.

And the output format changed a bit, listing all included checks, but that's just a cosmetic change.

src/constants.ts
```diff
@@ -237,7 +237,7 @@ export const EPEL_9_REPO_DEFINITION = {
 };
 
 export const DEBOUNCED_SEARCH_WAIT_TIME = 500;
-export const UNIQUE_VALIDATION_DELAY = 300;
+export const UNIQUE_VALIDATION_DELAY: number = 300;
```
mgold1234, line 240: is it relevant to that pr?

regexowl: > Type needed to be added to `UNIQUE_VALIDATION_DELAY` so the vitest could properly mock it. It's a bit weird, but what vitest desires it shall get 🤷 (without the type the build ended in failure)

Yup

regexowl:
Let's first bump to a higher 2.x version, then we can think about migrating to 3.x.

regexowl:
> Let's first bump to a higher 2.x version, then we can think about migrating to 3.x.

Or not, looks like the `Sourcemap for...`  mess in the output is already present in 2.1.9

PR #2787, opened 2025-01-22
Build images with cockpit (HMS-5417)

croissanne:
https://github.com/user-attachments/assets/e2d116fb-7250-44ad-8b4b-f231a51f7c3f

JIRA: [HMS-5417](https://issues.redhat.com/browse/HMS-5417)

### DIFF
--- a/src/store/cockpitApi.ts
+++ b/src/store/cockpitApi.ts
@@ -8,7 +8,6 @@ import path from 'path';
 // We also needed to create an alias in vitest to make this work.
 import cockpit from 'cockpit';
 import { fsinfo } from 'cockpit/fsinfo';
-import toml from 'toml';
 
 import {
   ListSnapshotsByDateApiArg,
@@ -16,6 +15,9 @@ import {
 } from './contentSourcesApi';
 import { emptyCockpitApi } from './emptyCockpitApi';
 import {
+  ComposeBlueprintApiResponse,
+  ComposeBlueprintApiArg,
+  CreateBlueprintRequest,
   GetArchitecturesApiResponse,
   GetArchitecturesApiArg,
   GetBlueprintsApiArg,
@@ -29,9 +31,10 @@ import {
   GetBlueprintApiArg,
   CreateBlueprintApiResponse,
   CreateBlueprintApiArg,
+  ComposeResponse,
 } from './imageBuilderApi';
 
-import { mapOnPremToHosted } from '../Components/Blueprints/helpers/onPremToHostedBlueprintMapper';
+import { mapHostedToOnPrem } from '../Components/Blueprints/helpers/onPremToHostedBlueprintMapper';
 import { BLUEPRINTS_DIR } from '../constants';
 
 const getBlueprintsPath = async () => {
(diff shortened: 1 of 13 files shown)


src/store/cockpitApi.ts
```diff
@@ -8,10 +8,12 @@ import path from 'path';
 // We also needed to create an alias in vitest to make this work.
 import cockpit from 'cockpit';
 import { fsinfo } from 'cockpit/fsinfo';
-import toml from 'toml';
+import { read_os_release } from 'os-release';
```
kingsleyzissou, line 11: Oh neat, where did this come from? I was trying to do something similar for the dropdown in the wizard

src/store/cockpitApi.ts
```diff
@@ -145,6 +146,79 @@ export const cockpitApi = emptyCockpitApi.injectEndpoints({
...
+        queryFn: async ({ id: filename }) => {
+          const blueprintsDir = await getBlueprintsPath();
+          const file = cockpit.file(
+            path.join(blueprintsDir, filename, `${filename}.json`)
+          );
+          const contents = await file.read();
+          file.close();
+          const parsed = JSON.parse(contents);
+
+          const cloudapi = cockpit.http('/run/cloudapi/api.socket', {
```
kingsleyzissou, line 196: I have a PR to extract this to the basequery so we can get rid of this :)
And just use the basequery function.

src/store/cockpitApi.ts
kingsleyzissou: Nevermind, just realised it's part of cockpit

src/store/cockpitApi.ts
croissanne: neat!

src/store/cockpitApi.ts
croissanne: indeed, it's part of cockpit. Though i think i'll make distribution not required in the cloudapi, that way we can drop it here at least.

src/store/cockpitApi.ts
```diff
@@ -110,6 +145,7 @@ export const cockpitApi = emptyCockpitApi.injectEndpoints({
         GetBlueprintsApiResponse,
         GetBlueprintsApiArg
       >({
+        providesTags: ['Blueprints'],
```
kingsleyzissou, line 148: I think we need to add the following to `enhancedCockpitApi`. 

https://github.com/osbuild/image-builder-frontend/blob/3c27f68b975bb52053f0c82da0cae65cf289ffb1/src/store/enhancedImageBuilderApi.ts#L143C1-L195C7

I think we can move the `providesTags` there too

src/store/cockpitApi.ts
```diff
@@ -295,6 +295,7 @@ export const cockpitApi = emptyCockpitApi.injectEndpoints({
         ComposeBlueprintApiResponse,
         ComposeBlueprintApiArg
       >({
+        invalidatesTags: ['Composes'],
```
kingsleyzissou, line 297: Think we can add this to `enhancedCockpitApi`

src/store/cockpitApi.ts
```diff
@@ -366,6 +367,7 @@ export const cockpitApi = emptyCockpitApi.injectEndpoints({
         },
       }),
       getComposes: builder.query<GetComposesApiResponse, GetComposesApiArg>({
+        providesTags: ['Composes'],
```
kingsleyzissou, line 369: As above, could be added to `enhancedCockpitAPi`

kingsleyzissou (commented):
Great stuff. Just had some comments about the tags stuff that could be added to the `enhancedCockpitApi` instead but pretty minor.

kingsleyzissou (approved):
LGTM, think the PR just needs a description :)

kingsleyzissou (approved):

PR #2788, opened 2025-01-22
Wizard: add support of upload custom repositories

dominikvagner:
### Summary 
This PR adds support for upload custom repositories. The upload repos are displayed and can be used in the same way as other custom repos, except that they have a badge next to them explaining they are of the upload type an thus don't have an URL.

_Kudos to @croissanne for help and making the needed backend changes for this._
Related ticket: [[🔗 HMS-5314](https://issues.redhat.com/browse/HMS-5314)]

image

### Testing Steps
1. Create 2 BPs that are using a an upload repo and have a set snapshot date, with and without having a selected package from the repo, then build and image with it. 
2. Create 2 BPs that are using a an upload repo and the latest content, with and without having a package selected from the repo, then build and image with it.


src/Components/CreateImageWizard/steps/Repositories/components/UploadRepositoryLabel.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import { Label, Tooltip } from '@patternfly/react-core';
+import { UploadIcon } from '@patternfly/react-icons';
+
+const UploadRepositoryLabel = () => {
+  return (
+    <Tooltip content="Upload repository: Doesn't have an URL, snapshots are taken when content is uploaded to the custom repo.">
```
xbhouse, line 8: i'm not nearly as familiar with this code as the IB team but this overall looks great to me 😄 

one thing i'm unsure about is the copy here. i'm not sure if the part about not having a URL is important or confusing for users (since there is still the snapshot URL). 

maybe we could have it be similar to what we have in the content sources UI: `Upload repository: Snapshots will only be taken when new content is uploaded.` maybe @regexowl might have some thoughts here too?

dominikvagner: yeah, copy is not my strong suit so I am unsure about that as well, any suggestions are welcome 😄 _(your suggestion already sounds better)_ 😅

xbhouse: awesome, thank you!! this looks great to me 🎉

xbhouse (approved):
ack from me! @regexowl is the pr check failure related? or is there anything we might need in iqe for this?

src/Components/CreateImageWizard/steps/Repositories/components/UploadRepositoryLabel.tsx
regexowl: Excellent, thank you! :heart:

regexowl:
> ack from me! @regexowl is the pr check failure related? or is there anything we might need in iqe for this?

I presume that one was a flake. Triggered retest and will keep my eyes on it :eyes:

PR #2789, opened 2025-01-22
Wizard: Add remove user button in users step (HMS-5378)

mgold1234:
this commit add remove user button in users step
Screenshot 2025-01-24 at 17 38 21

JIRA: [HMS-5378](https://issues.redhat.com/browse/HMS-5378)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
+++ b/src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
-import { Button, FormGroup, Checkbox } from '@patternfly/react-core';
-import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+import { Button, FormGroup, Checkbox, Tooltip } from '@patternfly/react-core';
+import { ExternalLinkAltIcon, TrashIcon } from '@patternfly/react-icons';
 
 import { GENERATING_SSH_KEY_PAIRS_URL } from '../../../../../constants';
 import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
@@ -14,6 +14,7 @@ import {
   setUserPasswordByIndex,
   setUserSshKeyByIndex,
   setUserAdministratorByIndex,
+  removeUser,
 } from '../../../../../store/wizardSlice';
 import { useUsersValidation } from '../../../utilities/useValidation';
 import {
@@ -62,6 +63,10 @@ const UserInfo = () => {
     dispatch(
       setUserAdministratorByIndex({ index: index, isAdministrator: value })
     );
+  };
+
+  const onRemoveUserClick = () => {
+    dispatch(removeUser(index));
   };
 
   return (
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -115,6 +130,34 @@ const UserInfo = () => {
...
+              className="pf-v5-u-pl-sm pf-v5-u-pt-0 pf-v5-u-pb-0 pf-v5-u-pr-0"
+            >
+              <HelpIcon />
+            </Button>
+          </Popover>
+          <Button variant="secondary" onClick={onRemoveUserClick}>
+            Remove a user
+          </Button>
+        </EmptyStateFooter>
+      </EmptyState>
```
regexowl, line 160: Nice! I think a remove button will be a good addition.

Let's make it simpler - we already have an empty user state and there's not really a reason to add another one. I'd also drop the popover, "Remove user" as a button text is self-explanatory enough.

regexowl: How about something like one of these?

image

The first one is a "plain" button. The second is tertiary with an icon and a tooltip.

regexowl (changes requested):
Can you also add a test to cover the button please? Maybe checking if the empty state with add button reappears when the remove button gets clicked or something like that?

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
mgold1234: I implement the button with the tooltip. if I remove the emptyState, the rectangle of the button looks wired

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -62,6 +62,15 @@ const clickAddUser = async () => {
   expect(addUser).toBeEnabled();
   await waitFor(() => user.click(addUser));
 };
+
+
```
lucasgarfield, line 66: The extra line here is causing the dev-check to fail.

lucasgarfield (changes requested):
Looking good, I like it! Couple of small change requests. Also, would you mind updating your commit message so it has proper spelling and capitalization?

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -115,6 +120,16 @@ const UserInfo = () => {
           name="user Administrator"
         />
       </FormGroup>
+      <Tooltip position="top-start" content={<div>Remove user</div>}>
```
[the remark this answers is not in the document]
lucasgarfield: Ah, I see that the Patternfly example uses them. I'm still not convinced we need them.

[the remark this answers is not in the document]
mgold1234: right, intersting why they add it in Patternfly

lucasgarfield (approved):
Working nicely, thank you!

PR #2807, opened 2025-01-28
Wizard: Add services to on-prem mapper (HMS-5405)

regexowl:
This adds services to on-prem mapper.

JIRA: [HMS-5405](https://issues.redhat.com/browse/HMS-5405)


avitova (approved):
Amazing, works as expected. Just a question - is masked going to be in the scope of you work?

regexowl:
> Amazing, works as expected. Just a question - is masked going to be in the scope of you work?

Good question. The services are currently split the way they were in both cockpit-composer and OpenSCAP step (masked grouped in with disabled). I'll open a thread to double check it's really what we want. Thank you!

PR #2809, opened 2025-01-28
scripts: add start:federated script

amirfefer:
This PR adds `start:federated` script for running image-builder-frontend as a static federated module.


regexowl:
Just to clear up the scope - this is just about the script, the page itself shouldn't load when ran as a static module as of yet, right?

amirfefer:
@regexowl, indeed, the script just serves IBF as a static module, not for running the app independently .
The context here is to enable running IBF  with a local chrome / different env, like a local  fedora services env :)

regexowl:
Got it! Thanks!

regexowl (approved):

regexowl:
pr_check failed on eph setup, this PR should be fine though. Merging manually.

PR #2823, opened 2025-01-31
.github/workflows/release: work around git clean

croissanne:
The osbuild release action cleans the repository, removing the tarball
    that gets uploaded to the release. This tarball needs to be built before
    schutzbot commits the version bump, so move it so it doesn't get
    cleaned.


kingsleyzissou (approved):
Trial & error haha

PR #2828, opened 2025-02-03
store/cockpit: don't use top-level constants that use cockpit

croissanne:
Only invoke the cockpit library when actually contacting the cloudapi.

The cockpit library isn't defined in the service, so any invocation of the cockpit library when the application loads will make the frontend crash.


regexowl (approved):
:heart:

regexowl:
Thank you! pr_check and bonfire pipeline will most likely fail, let's merge manually when everything else is green.

kingsleyzissou (approved):
Whoops. Thanks for fixing this

PR #2832, opened 2025-02-03
Wizard: Disable registration by default (HMS-5437)

### LINKED ISSUE #2826: Activation keys API is slow/unstable
Image

This is happening on my personal developers.redhat.com account. It's also been quite slow to load recently when it does work. Let's disable selecting an activation key automatically for now until they can get these issues resolved. We probably also need to report this to them, they may be unaware.

regexowl:
Fixes #2826

This sets default registration value to `register-later` and adjusts the tests.

JIRA: [HMS-5437](https://issues.redhat.com/browse/HMS-5437)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Registration/Registration.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/Registration.tsx
@@ -103,7 +103,7 @@ const Registration = () => {
   const registrationType = useAppSelector(selectRegistrationType);
 
   const [showOptions, setShowOptions] = useState(
-    registrationType === 'register-later'
+    registrationType === 'register-now-rhc'
   );
 
   // TO DO: Remove when rhc starts working for RHEL 10 Beta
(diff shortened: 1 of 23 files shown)


regexowl:
@lucasgarfield rhsm issues were resolved meaning activation keys are available again. Response time is also quite fast (1.19s), we might not need to even merge this

regexowl:
Closing this one, activation keys are working again and the queries are reasonably fast (around one second).

lucasgarfield:
I think we can close this again. I discussed the issue with Teamnado (the folks responsible for the activation keys API) and it seems like the issue is caused by the huge number of activation keys (~1050) on the account most of us Red Hatters share in production. Typical users would have at most 20 keys, and the API seems fast enough. In my own experimentation the activation keys API is slower than I would like (often requests take from 500 to 1000 ms) but I don't think that's enough to justify disabling the feature.

PR #2841, opened 2025-02-04
Wizard: remove confirmPassword from Mapper

mgold1234:
this commit remove confirmPassword from Mapper because we decide that this field is not relevant anymore to Users step, and it should remove from the mapper as well


regexowl (approved):
Thanks!

PR #2844, opened 2025-02-04
Wizard: Add validation to user's password in User step (HMS-5466)

mgold1234:
This commit adds password validation to the User step:
- Moved password validation logic to a separate `checkPasswordValidity` function, returning detailed results (strength and validation state).
- Simplified validation in `PasswordValidatedInput` by using `checkPasswordValidity` results directly.
- Added dynamic password strength indicator showing success or error based on requirements.
- Integrated environment-specific validation messages (e.g., for Azure).
- Improved code separation between presentation and validation for better maintainability.
- Unit tests: Adds tests for invalid passwords, covering both default and Azure cases.

use this Password strength - https://www.patternfly.org/components/password-strength/ as insperation 

JIRA: [HMS-5466](https://issues.redhat.com/browse/HMS-5466)

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -7,11 +7,7 @@ import {
   TextAreaProps,
   TextInput,
   TextInputProps,
-  Button,
-  InputGroup,
-  InputGroupItem,
 } from '@patternfly/react-core';
-import { EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
 
 import type { StepValidation } from './utilities/useValidation';
 
@@ -59,57 +55,6 @@ type ErrorMessageProps = {
 };
 
 type ValidationResult = 'default' | 'success' | 'error';
-
-export const HookPasswordValidatedInput = ({
-  ariaLabel,
-  placeholder,
-  dataTestId,
-  value,
-  ouiaId,
-  stepValidation,
-  fieldName,
-  onChange,
-  warning = undefined,
-  inputType,
-  isDisabled,
-}: HookValidatedInputPropTypes) => {
-  const [isPasswordVisible, setIsPasswordVisible] = useState(false);
-  const togglePasswordVisibility = () => {
-    setIsPasswordVisible(!isPasswordVisible);
-  };
-
-  return (
-    <>
-      <InputGroup>
-        <InputGroupItem isFill>
-          <HookValidatedInput
-            type={isPasswordVisible ? 'text' : 'password'}
-            ouiaId={ouiaId || ''}
-            data-testid={dataTestId}
-            value={value}
-            onChange={onChange!}
-            stepValidation={stepValidation}
-            ariaLabel={ariaLabel || ''}
-            fieldName={fieldName}
-            placeholder={placeholder || ''}
-            inputType={inputType || 'textInput'}
-            warning={warning || ''}
-            isDisabled={isDisabled || false}
-          />
-        </InputGroupItem>
-        <InputGroupItem>
-          <Button
-            variant="control"
-            onClick={togglePasswordVisibility}
-            aria-label={isPasswordVisible ? 'Show password' : 'Hide password'}
-          >
-            {isPasswordVisible ? <EyeSlashIcon /> : <EyeIcon />}
-          </Button>
-        </InputGroupItem>
-      </InputGroup>
-    </>
-  );
-};
 
 export const ValidatedInputAndTextArea = ({
   value,
--- a/src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
+++ b/src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
@@ -16,11 +16,9 @@ import {
   setUserAdministratorByIndex,
   removeUser,
 } from '../../../../../store/wizardSlice';
+import { PasswordValidatedInput } from '../../../utilities/PasswordValidatedInput';
 import { useUsersValidation } from '../../../utilities/useValidation';
-import {
-  HookPasswordValidatedInput,
-  ValidatedInputAndTextArea,
-} from '../../../ValidatedInput';
+import { ValidatedInputAndTextArea } from '../../../ValidatedInput';
 const UserInfo = () => {
   const dispatch = useAppDispatch();
   const index = 0;
@@ -41,7 +39,7 @@ const UserInfo = () => {
   };
 
   const handlePasswordChange = (
-    _event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+    _event: React.FormEvent<HTMLInputElement>,
     value: string
   ) => {
     dispatch(setUserPasswordByIndex({ index: index, password: value }));
@@ -81,16 +79,12 @@ const UserInfo = () => {
           fieldName="userName"
         />
       </FormGroup>
-      <FormGroup isRequired label="Password">
-        <HookPasswordValidatedInput
-          ariaLabel="blueprint user password"
-          value={userPassword || ''}
-          onChange={(_e, value) => handlePasswordChange(_e, value)}
-          placeholder="Enter password"
-          stepValidation={stepValidation}
-          fieldName="userPassword"
-        />
-      </FormGroup>
+      <PasswordValidatedInput
+        value={userPassword || ''}
+        ariaLabel="blueprint user password"
+        placeholder="Enter password"
+        onChange={(_e, value) => handlePasswordChange(_e, value)}
+      />
       <FormGroup isRequired label="SSH key">
         <ValidatedInputAndTextArea
           inputType={'textArea'}
--- a/src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
@@ -1,0 +1,95 @@
+import React, { useState } from 'react';
+
+import {
+  FormGroup,
+  FormHelperText,
+  HelperText,
+  HelperTextItem,
+  InputGroup,
+  InputGroupItem,
+  TextInput,
+  Button,
+  TextInputProps,
+} from '@patternfly/react-core';
+import { EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
+
+import { checkPasswordValidity } from './useValidation';
+
+import { useAppSelector } from '../../../store/hooks';
+import { selectImageTypes } from '../../../store/wizardSlice';
+
+type ValidatedPasswordInput = TextInputProps & {
+  value: string;
+  placeholder: string;
+  ariaLabel: string;
+  onChange: (event: React.FormEvent<HTMLInputElement>, value: string) => void;
+};
+
+export const PasswordValidatedInput = ({
+  value,
+  placeholder,
+  ariaLabel,
+  onChange,
+}: ValidatedPasswordInput) => {
+  const environments = useAppSelector(selectImageTypes);
+  const [isPasswordVisible, setIsPasswordVisible] = useState(false);
+
+  const { validationState, strength } = checkPasswordValidity(
+    value,
+    environments.includes('azure')
+  );
+  const { ruleLength, ruleCharacters } = validationState;
+
+  const togglePasswordVisibility = () => {
+    setIsPasswordVisible(!isPasswordVisible);
+  };
+
+  const passStrLabel = (
+    <HelperText>
+      <HelperTextItem variant={strength.variant} icon={strength.icon}>
+        {strength.text}
+      </HelperTextItem>
+    </HelperText>
+  );
+
+  return (
+    <FormGroup label="Password" isRequired labelInfo={passStrLabel}>
+      <>
+        <InputGroup>
+          <InputGroupItem isFill>
+            <TextInput
+              isRequired
+              type={isPasswordVisible ? 'text' : 'password'}
+              value={value}
+              onChange={onChange}
+              aria-label={ariaLabel}
+              placeholder={placeholder}
+            />
+          </InputGroupItem>
+          <InputGroupItem>
+            <Button
+              variant="control"
+              onClick={togglePasswordVisibility}
+              aria-label={isPasswordVisible ? 'Hide password' : 'Show password'}
+            >
+              {isPasswordVisible ? <EyeSlashIcon /> : <EyeIcon />}
+            </Button>
+          </InputGroupItem>
+        </InputGroup>
+      </>
+      <FormHelperText>
+        <HelperText component="ul">
+          <HelperTextItem variant={ruleLength} component="li" hasIcon>
+            Password must be at least 6 characters long.
+          </HelperTextItem>
+          {environments.includes('azure') && (
+            <HelperTextItem variant={ruleCharacters} component="li" hasIcon>
+              Must include at least 3 of the following: lowercase letters,
+              uppercase letters, numbers, symbols.
+            </HelperTextItem>
+          )}
+        </HelperText>
+      </FormHelperText>
+    </FormGroup>
+  );
+};
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -1,4 +1,6 @@
-import { useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
+
+import { CheckCircleIcon } from '@patternfly/react-icons';
 
 import { UNIQUE_VALIDATION_DELAY } from '../../../constants';
 import { useLazyGetBlueprintsQuery } from '../../../store/backendApi';
@@ -31,9 +33,11 @@ import {
   selectLanguages,
   selectKeyboard,
   selectTimezone,
+  selectImageTypes,
 } from '../../../store/wizardSlice';
 import { keyboardsList } from '../steps/Locale/keyboardsList';
 import { languagesList } from '../steps/Locale/languagesList';
+import { HelperTextVariant } from '../steps/Packages/components/CustomHelperText';
 import { timezones } from '../steps/Timezone/timezonesList';
 import {
   getDuplicateMountPoints,
@@ -70,6 +74,7 @@ export function useIsBlueprintValid(): boolean {
   const services = useServicesValidation();
   const firstBoot = useFirstBootValidation();
   const details = useDetailsValidation();
+  const users = useUsersValidation();
   return (
     !registration.disabledNext &&
     !filesystem.disabledNext &&
@@ -81,9 +86,31 @@ export function useIsBlueprintValid(): boolean {
     !firewall.disabledNext &&
     !services.disabledNext &&
     !firstBoot.disabledNext &&
-    !details.disabledNext
+    !details.disabledNext &&
+    !users.disabledNext
   );
 }
+
+type ValidationFlags = {
+  isUserNameValidValue: boolean;
+  isSshKeyValidValue: boolean;
+  isPasswordValidValue: boolean;
+};
+
+type PasswordValidationResult = {
+  isValid: boolean;
+  strength: {
+    variant: HelperTextVariant;
+    icon: JSX.Element | null;
+    text: string;
+  };
+  validationState: ValidationState;
+};
+
+type ValidationState = {
+  ruleLength: HelperTextVariant;
+  ruleCharacters: HelperTextVariant;
+};
 
 export function useRegistrationValidation(): StepValidation {
   const registrationType = useAppSelector(selectRegistrationType);
@@ -380,8 +407,23 @@ export function useServicesValidation(): StepValidation {
   };
 }
 
+const getUserNameErrorMsg = (userName: string): string => {
+  if (userName && !isUserNameValid(userName)) {
+    return 'Invalid user name';
+  }
+  return '';
+};
+
+const getSshKeyErrorMsg = (userSshKey: string): string => {
+  if (userSshKey && !isSshKeyValid(userSshKey)) {
+    return 'Invalid SSH key';
+  }
+  return '';
+};
+
 export function useUsersValidation(): StepValidation {
   const index = 0;
+  const environments = useAppSelector(selectImageTypes);
   const userNameSelector = selectUserNameByIndex(index);
   const userName = useAppSelector(userNameSelector);
   const userPasswordSelector = selectUserPasswordByIndex(index);
@@ -389,27 +431,118 @@ export function useUsersValidation(): StepValidation {
   const userSshKeySelector = selectUserSshKeyByIndex(index);
   const userSshKey = useAppSelector(userSshKeySelector);
   const users = useAppSelector(selectUsers);
+
+  const userNameError = getUserNameErrorMsg(userName);
+  const sshKeyError = getSshKeyErrorMsg(userSshKey);
+
+  const { isUserNameValidValue, isSshKeyValidValue, isPasswordValidValue } =
+    calculateValidationFlags(
+      userName,
+      userPassword,
+      userSshKey,
+      environments.includes('azure')
+    );
+
   const canProceed =
     // Case 1: there is no users
     users.length === 0 ||
-    // Case 2: All fields are empty
-    (userName === '' && userPassword === '' && userSshKey === '') ||
-    // Case 3: userName is valid and SshKey is valid
-    (userName &&
-      isUserNameValid(userName) &&
-      userSshKey &&
-      isSshKeyValid(userSshKey));
+    // Case 2: userName is valid and SshKey or Password is valid
+    (isUserNameValidValue && (isSshKeyValidValue || isPasswordValidValue));
 
   return {
     errors: {
-      userName:
-        userName && !isUserNameValid(userName) ? 'Invalid user name' : '',
-      userSshKey:
-        userSshKey && !isSshKeyValid(userSshKey) ? 'Invalid SSH key' : '',
+      userName: userNameError,
+      userSshKey: sshKeyError,
     },
     disabledNext: !canProceed,
   };
 }
+
+const calculateValidationFlags = (
+  userName: string,
+  userPassword: string,
+  userSshKey: string,
+  isAzure: boolean
+): ValidationFlags => {
+  const isUserNameValidValue = !!userName && isUserNameValid(userName);
+  const isSshKeyValidValue = !!userSshKey && isSshKeyValid(userSshKey);
+  const isPasswordValidValue =
+    !!userPassword && checkPasswordValidity(userPassword, isAzure).isValid;
+  return { isUserNameValidValue, isSshKeyValidValue, isPasswordValidValue };
+};
+
+export const checkPasswordValidity = (
+  password: string,
+  isAzure: boolean
+): PasswordValidationResult => {
+  if (!password) {
+    return {
+      isValid: false,
+      strength: getStrength(0, 0, false),
+      validationState: {
+        ruleLength: 'indeterminate',
+        ruleCharacters: 'indeterminate',
+      },
+    };
+  }
+  const isEncrypted = /^\$([^$]+)\$/.test(password);
+  if (isEncrypted) {
+    return {
+      isValid: true,
+      strength: getStrength(0, 0, false),
+      validationState: {
+        ruleLength: 'success',
+        ruleCharacters: 'success',
+      },
+    };
+  }
+
+  const trimmedValue = password.trim();
+  const isLengthValid = trimmedValue.length >= 6 && trimmedValue.length <= 128;
+  const { rulesCount, strCount } = countCharacterTypes(password);
+
+  const validationState: ValidationState = {
+    ruleLength: isLengthValid ? 'success' : 'error',
+    ruleCharacters: rulesCount >= 3 ? 'success' : 'error',
+  };
+
+  return {
+    isValid: isLengthValid,
+    strength: getStrength(strCount, rulesCount, isAzure),
+    validationState: validationState,
+  };
+};
+
+const getStrength = (
+  strCount: number,
+  rulesCount: number,
+  isAzure: boolean
+): PasswordValidationResult['strength'] => {
+  const isValidStrength = isAzure
+    ? strCount >= 6 && rulesCount >= 3
+    : strCount >= 6;
+
+  return isValidStrength
+    ? { variant: 'success', icon: <CheckCircleIcon />, text: 'Strong' }
+    : { variant: 'default', icon: null, text: '' };
+};
+
+const countCharacterTypes = (value: string) => {
+  const lowercaseCount = (value.match(/[a-z]/g) || []).length;
+  const uppercaseCount = (value.match(/[A-Z]/g) || []).length;
+  const digitsCount = (value.match(/\d/g) || []).length;
+  const specialCount = (value.match(/\W/g) || []).length;
+
+  const rulesCount = [
+    lowercaseCount,
+    uppercaseCount,
+    digitsCount,
+    specialCount,
+  ].filter((count) => count > 0).length;
+  const strCount = lowercaseCount + uppercaseCount + digitsCount + specialCount;
+
+  return { rulesCount, strCount };
+};
 
 export function useDetailsValidation(): StepValidation {
   const name = useAppSelector(selectBlueprintName);
--- a/src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
@@ -5,6 +5,7 @@ import { userEvent } from '@testing-library/user-event';
 import { CREATE_BLUEPRINT, EDIT_BLUEPRINT } from '../../../../../constants';
 import { mockBlueprintIds } from '../../../../fixtures/blueprints';
 import { usersCreateBlueprintRequest } from '../../../../fixtures/editMode';
+import { addTargetEnvAzure } from '../../CreateImageWizard.test';
 import {
   blueprintRequest,
   clickBack,
@@ -13,7 +14,6 @@ import {
   getNextButton,
   interceptBlueprintRequest,
   interceptEditBlueprintRequest,
-  openAndDismissSaveAndBuildModal,
   renderEditMode,
   verifyCancelButton,
 } from '../../wizardTestUtils';
@@ -26,6 +26,8 @@ import {
 let router: RemixRouter | undefined = undefined;
 const validUserName = 'best';
 const validSshKey = 'ssh-rsa d';
+const validPassword = 'validPassword';
+const invalidPassword = 'inval';
 
 const goToUsersStep = async () => {
   await clickNext();
@@ -86,6 +88,13 @@ const addUserName = async (userName: string) => {
   });
   await waitFor(() => user.type(enterUserName, userName));
   await waitFor(() => expect(enterUserName).toHaveValue(userName));
+};
+
+const addPassword = async (mockPassword: string) => {
+  const user = userEvent.setup();
+  const enterPassword = screen.getByPlaceholderText(/enter password/i);
+  await waitFor(() => user.type(enterPassword, mockPassword));
+  await waitFor(() => expect(enterPassword).toHaveValue(mockPassword));
 };
 
 describe('Step Users', () => {
@@ -159,10 +168,51 @@ describe('Step Users', () => {
     const invalidUserMessage = screen.getByText(/invalid ssh key/i);
     await waitFor(() => expect(invalidUserMessage));
   });
+
+  test('try to create Azure image with invalid password', async () => {
+    const user = userEvent.setup();
+    await renderCreateMode();
+    await waitFor(() => user.click(screen.getByTestId('upload-azure')));
+    await clickNext();
+    await addTargetEnvAzure();
+    await clickRegisterLater();
+    await goToUsersStep();
+    await clickAddUser();
+    await addUserName(validUserName);
+    await addPassword(invalidPassword);
+
+    const invalidUserMessage = screen.getByText(
+      /Password must be at least 6 characters long./i
+    );
+    const warningUserMessage = screen.getByText(
+      /Must include at least 3 of the following: lowercase letters, uppercase letters, numbers, symbols./i
+    );
+    await waitFor(() => expect(invalidUserMessage));
+    await waitFor(() => expect(warningUserMessage));
+    const nextButton = await getNextButton();
+    await waitFor(() => expect(nextButton).toBeDisabled());
+  });
+
+  test('with invalid password', async () => {
+    await renderCreateMode();
+    await goToRegistrationStep();
+    await clickRegisterLater();
+    await goToUsersStep();
+    await clickAddUser();
+    await addUserName(validUserName);
+    await addPassword(invalidPassword);
+
+    const invalidUserMessage = screen.getByText(
+      /Password must be at least 6 characters long./i
+    );
+    await waitFor(() => expect(invalidUserMessage));
+    const nextButton = await getNextButton();
+    await waitFor(() => expect(nextButton).toBeDisabled());
+  });
 });
 
 describe('User request generated correctly', () => {
-  test('with valid name, ssh key and checked Administrator checkbox', async () => {
+  test('create image with valid name, password, ssh key and checked Administrator checkbox', async () => {
     const user = userEvent.setup();
     await renderCreateMode();
     await goToRegistrationStep();
@@ -171,16 +221,14 @@ describe('User request generated correctly', () => {
     await clickAddUser();
     await addUserName(validUserName);
     await addSshKey(validSshKey);
+    await addPassword(validPassword);
     const nextButton = await getNextButton();
     await waitFor(() => expect(nextButton).toBeEnabled());
     const isAdmin = screen.getByRole('checkbox', {
       name: /administrator/i,
     });
-    user.click(isAdmin);
+    await waitFor(() => user.click(isAdmin));
     await goToReviewStep();
-    // informational modal pops up in the first test only as it's tied
-    // to a 'imageBuilder.saveAndBuildModalSeen' variable in localStorage
-    await openAndDismissSaveAndBuildModal();
     const receivedRequest = await interceptBlueprintRequest(CREATE_BLUEPRINT);
     const expectedRequest = {
       ...blueprintRequest,
@@ -189,6 +237,7 @@ describe('User request generated correctly', () => {
           {
             name: 'best',
             ssh_key: 'ssh-rsa d',
+            password: validPassword,
             groups: ['wheel'],
           },
         ],
@@ -234,7 +283,6 @@ describe('Users edit mode', () => {
     const receivedRequest = await interceptEditBlueprintRequest(
       `${EDIT_BLUEPRINT}/${id}`
     );
-    const expectedRequest = usersCreateBlueprintRequest;
-    expect(receivedRequest).toEqual(expectedRequest);
+    expect(receivedRequest).toEqual(usersCreateBlueprintRequest);
   });
 });
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -492,6 +492,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 key="wizard-users"
                 navItem={customStatusNavItem}
                 isHidden={!isUsersEnabled}
+                status={usersValidation.disabledNext ? 'error' : 'default'}
                 footer={
                   <CustomWizardFooter
                     disableNext={usersValidation.disabledNext}
--- a/src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
+++ b/src/test/Components/CreateImageWizard/CreateImageWizard.test.tsx
@@ -19,6 +19,30 @@ const getAzureSourceDropdown = async () => {
   return sourceDropdown;
 };
 
+export const addTargetEnvAzure = async () => {
+  const user = userEvent.setup();
+  expect(
+    await screen.findByRole('radio', {
+      name: /use an account configured from sources\./i,
+    })
+  ).toHaveFocus();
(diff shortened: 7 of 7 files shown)


src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -85,6 +92,20 @@ export const HookPasswordValidatedInput = ({
...
+          </HelperTextItem>
+        </HelperText>
+      )}
+      {warning !== undefined && warning !== '' && (
+        <HelperText>
+          <HelperTextItem variant="warning" hasIcon>
+            {warning}
+          </HelperTextItem>
+        </HelperText>
+      )}
```
regexowl, line 108: Umm, there is already validation implemented in the `HookValidatedInput`. This just seems to be duplicating a bunch of code. Couldn't you just add the password validation to the `useUsersValidation`?

Try to take a look at `useValidation` changes here: https://github.com/osbuild/image-builder-frontend/pull/2837/files#diff-e8f80e37ebfb45fc960d754061f3cd5877b1110ca14915bdcc1408571a04dad2

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -102,11 +123,14 @@ export const HookValidatedInput = ({
   type = 'text',
   inputType,
   warning = undefined,
+  isEmpty,
```
regexowl, line 126: I don't think this should be needed.

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -50,28 +51,34 @@ export const HookPasswordValidatedInput = ({
...
   isDisabled,
+  isEmpty,
 }: HookValidatedInputPropTypes) => {
+  const validated = isEmpty
+    ? 'default'
+    : stepValidation.errors[fieldName] === 'default'
+    ? 'default'
+    : stepValidation.errors[fieldName]
+    ? 'error'
+    : 'success';
```
regexowl, line 62: This is already handled in `HookValidatedInput`, why the duplicity?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -50,28 +51,34 @@ export const HookPasswordValidatedInput = ({
...
     setIsPasswordVisible(!isPasswordVisible);
   };
 
   return (
     <>
+      {inputType === 'textInput'}
       <InputGroup>
         <InputGroupItem isFill>
-          <HookValidatedInput
+          <TextInput
```
regexowl, line 73: I think this should stay as `HookValidatedInput` so we can use the validation features we already have.

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: why not?
without it, if there is no value we see the sucess status or error message, if user decide to delete the value 
Screenshot 2025-02-09 at 14 53 00

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: I tried to implemented it in the `HookValidatedInput` and the error appears at the right side intead of under, I will push again this implement in different commit
Screenshot 2025-02-09 at 15 14 17

mgold1234: I implemened it diffrent, thank you

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -132,18 +91,37 @@ export const HookValidatedInput = ({
...
+                passwordField ? (isPasswordVisible ? 'text' : 'password') : type
+              }
+              onChange={onChange!}
+              validated={validated}
+              aria-label={ariaLabel || ''}
+              onBlur={handleBlur}
+              placeholder={placeholder || ''}
+              isDisabled={isDisabled || false}
+            />
+          </InputGroupItem>
```
regexowl, line 110: This is a great rewrite, love it! :heart:

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -171,12 +194,13 @@ describe('User request generated correctly', () => {
     await clickAddUser();
     await addUserName(validUserName);
     await addSshKey(validSshKey);
+    await addPassword('thisIsValidPass@@')
```
regexowl, line 197:
```suggestion
    await addPassword('thisIsValidPass@@');
```

src/Components/CreateImageWizard/validators.ts
```diff
@@ -79,6 +79,19 @@ export const isSshKeyValid = (sshKey: string) => {
   return isPatternValid;
 };
 
+export const isPasswordValid = (password: string) => {
+  if (password === undefined) return false;
```
regexowl, line 83: `password` is a string, shouldn't the initial value be `''`?

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -79,16 +76,19 @@ const UserInfo = () => {
           onChange={(_e, value) => handleNameChange(_e, value)}
           stepValidation={stepValidation}
           fieldName="userName"
+          isEmpty={userName === ''}
```
regexowl, line 79: When this gets removed, the input field should still behave the same as the `validated` value is affected by `isPristine`:
```typescript
const [isPristine, setIsPristine] = useState(!value ? true : false);
```

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -79,16 +76,19 @@ const UserInfo = () => {
...
-        <HookPasswordValidatedInput
+        <HookValidatedInput
+          passwordField={true}
           ariaLabel="blueprint user password"
           value={userPassword || ''}
           onChange={(_e, value) => handlePasswordChange(_e, value)}
           placeholder="Enter password"
           stepValidation={stepValidation}
           fieldName="userPassword"
+          isEmpty={userPassword === ''}
```
regexowl, line 91: I believe this could get removed as well.

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
```diff
@@ -101,6 +101,7 @@ const UserInfo = () => {
           placeholder="Paste your public SSH key"
           stepValidation={stepValidation}
           fieldName="userSshKey"
+          isEmpty={userSshKey === ''}
```
regexowl, line 104: And this one also can get removed.

src/Components/CreateImageWizard/ValidatedInput.tsx
regexowl: The "emptiness" of the input is already handled by the `isPristine` state. I've tried to remove the `isEmpty` and the behaviour of the field didn't change in any way.

regexowl (commented):
Added a few smaller nit picks, but this is looking really good. Thank you!

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
mgold1234: I record it with and without empty

https://github.com/user-attachments/assets/d483a8d8-0cf8-4ba3-9cce-d9d3309a5373


https://github.com/user-attachments/assets/a7a69d49-94c2-40a1-ac0d-74b36ae11ffd

src/Components/CreateImageWizard/validators.ts
mgold1234: thank you, I think we dont need to check it at all, so I remove it, becuase user can choose to fill ssh- key or password

src/Components/CreateImageWizard/steps/Users/component/UserInfo.tsx
mgold1234: @regexowl wdyt?

regexowl: Ah, yeah, I see. But that's an error in validation. `useUsersValidation` needs to return an empty error string if `userName` is undefined. So similarly to:
```typescript
      userPassword: !userPassword
        ? ''
        : !isPasswordValid(userPassword)
        ? 'Password must be between 8 and 128 characters, contain at least one uppercase letter, one lowercase letter, and one special character.'
        : '',
```
or the sshKey part. Otherwise you're trying to pass `undefined` to a function that should take only string as an argument.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -341,17 +344,26 @@ export function useUsersValidation(): StepValidation {
...
+        ? 'Invalid user name'
+        : '',
+      userPassword: !userPassword
+        ? 'default'
+        : !isPasswordValid(userPassword)
+        ? 'Password must be between 8 and 128 characters, contain at least one uppercase letter, one lowercase letter, and one special character.'
+        : '',
       userSshKey: !userSshKey
-        ? ''
+        ? 'default'
```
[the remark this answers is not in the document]
regexowl: Oh, I see, what happened. Will add a comment in a bit!

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -341,17 +344,26 @@ export function useUsersValidation(): StepValidation {
...
+      userName: !userName
+        ? 'default'
+        : !isUserNameValid(userName)
+        ? 'Invalid user name'
+        : '',
+      userPassword: !userPassword
+        ? 'default'
+        : !isPasswordValid(userPassword)
+        ? 'Password must be between 8 and 128 characters, contain at least one uppercase letter, one lowercase letter, and one special character.'
+        : '',
```
regexowl, line 364: The default values shouldn't be added here, but to the `ValidationInput` component. Will add a comment there

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -102,6 +52,7 @@ export const HookValidatedInput = ({
   type = 'text',
   inputType,
   warning = undefined,
+  passwordField = false,
```
regexowl, line 55: Oh, I can't add a comment to the specific line - on line 60 there is the `const validate`, you should add the `'default'` there instead of in the `useValidation`.

The `validated` is switching between the validation states - meaning the 'default', 'error' or 'success' apply directly to the text input and render the appropriate version of the input.

In `useValidation` the `StepValidation` schema is following:
```typescript
export type StepValidation = {
  errors: {
    [key: string]: string;
  };
  disabledNext: boolean;
};
```
meaning the `error` should contain an error text that will get rendered for an error state. It doesn't affect the formatting of the input and should be empty when there is no problem with validation.

mgold1234: ok, so I think I got your point, added another condition in the `validate` function instead of in the `useValidation `

src/Components/CreateImageWizard/validators.ts
```diff
@@ -79,6 +84,27 @@ export const isSshKeyValid = (sshKey: string) => {
...
+
+  if (isEncrypted) return true;
+  return (
+    isLengthValid &&
+    hasUpperCase &&
+    hasNoSpaces &&
+    hasLowerCase &&
+    hasSpecialChar
+  );
+};
```
lucasgarfield, line 106: I might be misunderstanding something but as far as I can tell in the image definitions only Azure images seem to have any restrictions.

https://github.com/search?q=repo%3Aosbuild/images%20PwqualityConfConfig&type=code

If you haven't done so, you can always test this by building and booting an image with a simple password.

Remember, the goal of our validation is to prevent users from accidentally building images that fail to build or boot but to otherwise be as unopinionated / not-annoying as possible. And password rules are definitely annoying. :wink:

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -36,59 +36,9 @@ type HookValidatedInputPropTypes = TextInputProps &
     fieldName: string;
     warning?: string;
     inputType?: 'textInput' | 'textArea';
+    passwordField?: boolean;
   };
 
-export const HookPasswordValidatedInput = ({
```
lucasgarfield, line 41: What's the reason for getting rid of this wrapper component and moving everything into `<HookValidatedInput>`? We discussed this at length in previous pull requests - why change the approach now?

regexowl: I might know what the main motivation was in this case. Previously we had `<HookValidatedInput>` nested in the `<HookValidatedPasswordInput>` to get the validated input with the 'hide password' button without major duplication.

The helper error text was a part of `<HookValidatedInput>` meaning it was looking like this when the validation failed:

image

One option to solve this would be to add the button conditionally to the `<HookValidatedInput>` component as it is now, which is a solution that I quite like as it doesn't duplicate anything unnecessarily and with not too many changes seems quite clean.

mgold1234: @lucasgarfield , I changed the approach, becasue I figure out that we can prevent all those duplicated code with adding the button conditionally -
```
 {passwordField && (
            <InputGroupItem>
              <Button
                variant="control"
                onClick={togglePasswordVisibility}
                aria-label={
                  isPasswordVisible ? 'Hide password' : 'Show password'
                }
              >
                {isPasswordVisible ? <EyeSlashIcon /> : <EyeIcon />}
              </Button>
            </InputGroupItem>
          )}
        </InputGroup>
```
and I solve the porblem that Klara mentioned with the inputGroupItem, now the code looks more clean and without duplication rows, and I preffer this solution

lucasgarfield: Ah, that makes sense. I understand the reasoning now.

However, I’d still like to challenge this a bit more. This seems to be becoming a “mega” component.  

Taking a step back, the real problem here is just that we don’t have control over the order that the components are rendered in. There is a tight coupling between the input and the validation. Would it be a better design if we could loosen that coupling?

At the end of the day, all we really need to do is render an <InputGroup> and then render a <ValidationMessage> component below it.

The <InputGroup> could be anything. It might be a plain input or a text area, and it might have all sorts of different “extra” components (buttons, dropdowns, etc…). There are lots of different possible input groups in Patternfly. Take a look at the examples here:
https://www.patternfly.org/components/input-group

So my idea would be to simply create a <PasswordInput> component and a separate <PasswordValidationMessage> component, and then just render them one after the other. Could we give that a try?

That would also allow us to make a really nice custom validation component for the password, I would use the Patternfly example as inspiration:
https://www.patternfly.org/components/password-strength

src/Components/CreateImageWizard/validators.ts
mgold1234: ok, I did some tests in rhel env, and the only limit for password is that the minimum length is 6.
so I changed it to 6.
now we need to decide if we want to add warning to the user if he fill weak password - 
like only small letters or only digit?
and then we can put a link how to create strong password 
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/4/html/security_guide/s1-wstation-pass#s2-wstation-pass-create
wdyt?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -38,11 +38,12 @@ type HookValidatedInputPropTypes = TextInputProps &
     inputType?: 'textInput' | 'textArea';
   };
 
-type passwordPropTypes = TextInputProps & {
+interface PasswordInputProps extends TextInputProps {
```
lucasgarfield: Why `interface` instead of `type`?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -106,26 +139,21 @@ export const PasswordInput = ({
...
+}: PasswordInputProps) => {
+  const errorMessage = stepValidation.errors[fieldName];
+  const showErrorMessage = !isPristine && errorMessage;
   if (!errorMessage) {
     return null;
   }
 
   return (
     <>
+      {value === value}
```
lucasgarfield: ?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -83,6 +115,7 @@ export const PasswordInput = ({
     <>
       <InputGroup>
         <InputGroupItem isFill>
+          {isPristine === isPristine}
```
lucasgarfield: ?

lucasgarfield (changes requested):

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: it looks like you didnt update the pr with a lot of changes https://github.com/osbuild/image-builder-frontend/commit/a67760ce695d9443b50103897f007b1152cdf1e2

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: same

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: same

src/Components/CreateImageWizard/validators.ts
```diff
@@ -79,6 +85,41 @@ export const isSshKeyValid = (sshKey: string) => {
   return isPatternValid;
 };
 
+interface PasswordValidationResult {
```
lucasgarfield, line 88: Didn't we decide to use `type` instead of `interface` just for the sake of consistency?

src/Components/CreateImageWizard/validators.ts
```diff
@@ -79,6 +85,41 @@ export const isSshKeyValid = (sshKey: string) => {
...
+  isValid: boolean;
+  strength: HelperTextVariant;
+}
+
+const isPasswordLengthValid = (password: string): boolean => {
+  if (!password) return true;
```
lucasgarfield, line 94: I think this is confusing because a password with length 0 violates the minimum length requirement of 6. I would expect `isPasswordLengthValid('')` to return `false`. 

I think we should handle the special case where there is no password (`password === ''`) separately.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -392,6 +396,49 @@ export function useServicesValidation(): StepValidation {
   };
 }
 
+function validateUserName(userName: string): string {
```
lucasgarfield, line 399: Can you please use arrow function expressions to be consistent with the rest of the code base?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -392,6 +396,49 @@ export function useServicesValidation(): StepValidation {
...
+  if (!isUserNameValid(userName)) {
+    return 'Invalid user name';
+  }
+  return '';
+}
+
+function validatePassword(
```
lucasgarfield, line 406: Can you please use arrow function expressions to be consistent with the rest of the code base?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -392,6 +396,49 @@ export function useServicesValidation(): StepValidation {
...
+  }
+
+  if (!passwordValidationResult.isValid) {
+    return DEFAULT_PASSWORD_ERROR_MESSAGE;
+  }
+
+  return '';
+}
+
+function validateSshKey(userSshKey: string): string {
```
lucasgarfield, line 435: This function doesn't validate the SSH key, it just returns the error message for an invalid key. Naming is very hard, but a good starting point is to name functions after what they do.

src/Components/CreateImageWizard/utilities/useValidation.tsx
lucasgarfield: This function doesn't validate the password, it returns the error message for an invalid password. Naming is very hard, but a good starting point is to name functions after what they do.

src/Components/CreateImageWizard/utilities/useValidation.tsx
lucasgarfield: This function doesn't validate the username, it returns the error message for an invalid username. Naming is very hard, but a good starting point is to name functions after what they do.

lucasgarfield: Maybe `getUserNameErrorMsg` or something?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +139,25 @@ export const HookPasswordValidatedInput = ({
   );
 };
 
+const getValidationState = ({
```
lucasgarfield, line 142: I mentioned this in another PR but this is just a normal function, so it should just have normal arguments (not a Props object). A good rule of thumb is to minimize the number of arguments a function takes, but 2 is probably fine, I don't think we need to pack them in an object.

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -38,48 +47,89 @@ type HookValidatedInputPropTypes = TextInputProps &
...
+  const errorMessage = stepValidation.errors[fieldName];
+  const hasError = errorMessage !== '';
+
+  return (
+    <>
+      <PasswordInput
+        value={value}
+        placeholder={placeholder || ''}
+        errorMessage={errorMessage}
+        onChange={onChange!}
```
lucasgarfield, line 82: Do we need the non-null assertion operator here? :thinking:

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -38,48 +47,89 @@ type HookValidatedInputPropTypes = TextInputProps &
...
+        onBlur={onBlur!}
```
lucasgarfield, line 83: Do we need the non-null assertion operator here? :thinking:

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -38,48 +47,89 @@ type HookValidatedInputPropTypes = TextInputProps &
...
+  placeholder,
+}: ValidatedPasswordInput) => {
+  const errorMessage = stepValidation.errors[fieldName];
+  const hasError = errorMessage !== '';
+
+  return (
+    <>
+      <PasswordInput
+        value={value}
+        placeholder={placeholder || ''}
```
lucasgarfield, line 80: Why the `|| ''` here? Can we not just pass `undefined` down to the `<TextInput>`?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -15,15 +15,24 @@ import { EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
...
   ouiaId?: string;
   ariaLabel: string | undefined;
   helperText: string | undefined;
   validator: (value: string | undefined) => boolean;
   value: string;
   placeholder?: string;
-}
+};
+
+type ValidationResult = {
```
lucasgarfield, line 28: I don't think this needs to be an object since it's just three strings. You can just do something like
```javascript
type ValidationResult = 'default' | 'success' | 'error'
```

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -38,48 +47,89 @@ type HookValidatedInputPropTypes = TextInputProps &
...
-            stepValidation={stepValidation}
-            ariaLabel={ariaLabel || ''}
-            fieldName={fieldName}
+            validated={validated}
             placeholder={placeholder || ''}
-            inputType={inputType || 'textInput'}
-            warning={warning || ''}
-            isDisabled={isDisabled || false}
+            type={isPasswordVisible ? 'text' : 'password'}
+            onChange={onChange!}
```
lucasgarfield, line 124: Do we need the non-null assertion operator here? 🤔

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -392,6 +396,49 @@ export function useServicesValidation(): StepValidation {
   };
 }
 
+function validateUserName(userName: string): string {
+  if (!isUserNameValid(userName)) {
+    return 'Invalid user name';
+  }
+  return '';
```
lucasgarfield, line 403: It seems like the purpose of these functions is to generate the error message. If there is no error message, would it make more sense to return `undefined`?

lucasgarfield (changes requested):

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: because it detects that the properties onChange and onBlur might be undefined at runtime, even though the code seems to assume they are always defined. The reason is that they are defined as Optional props.

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: same

src/Components/CreateImageWizard/utilities/useValidation.tsx
mgold1234: no, because if I will return undefined, then I need to change `
export type StepValidation = {
  errors: {
    [key: string]: string| undefined;
  };
  disabledNext: boolean;
};`
and I think it is not make more sense

src/Components/CreateImageWizard/validators.ts
```diff
@@ -78,6 +79,41 @@ export const isSshKeyValid = (sshKey: string) => {
...
+  let rulesCount = 0;
+  if (/[a-z]/.test(password)) rulesCount++; // Lowercase
+  if (/[A-Z]/.test(password)) rulesCount++; // Uppercase
+  if (/\d/.test(password)) rulesCount++; // Digit
+  if (/\W/.test(password)) rulesCount++; // Symbol
+
+  return rulesCount >= 3 ? 'success' : 'error';
+};
+
+export const isPasswordValid = (password: string): PasswordValidationResult => {
```
lucasgarfield, line 106: This function is called `isPasswordValid` and so I would expect it to return a boolean, not an object.

lucasgarfield (changes requested):

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: yes of course I took a look on that, and it seems to me totally different from other fields, and I thought it is just advice, I will revise this pull request

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: the thing is that if I am using the new refactor component <ValidatedInputAndTextArea>
I have a problem that I mentioned before with the place of the error message, and I solved that with adding a condition to HookValidatedInput  - https://github.com/osbuild/image-builder-frontend/commit/93c677848d7b9fa81fa489b8d1d29ed180777ae2
but you dont like this solution 
Screenshot 2025-03-10 at 19 41 36

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: implemented it like this example, thank you

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -52,47 +54,61 @@ type ValidationInputProp = TextInputProps &
...
+}: ValidatedPasswordInput) => {
+  const errorMessage = stepValidation.errors[fieldName];
+  const hasError = errorMessage !== '';
+
+  const [isPristine, setIsPristine] = useState(!value);
   const [isPasswordVisible, setIsPasswordVisible] = useState(false);
   const togglePasswordVisibility = () => {
     setIsPasswordVisible(!isPasswordVisible);
   };
+  const validated = getValidationState(isPristine, errorMessage);
```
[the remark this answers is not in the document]
mgold1234: refactor the passsword validation to new component

src/Components/CreateImageWizard/validators.ts
mgold1234: changed it thank you

src/Components/CreateImageWizard/validators.ts
mgold1234: changed that to be with minimum 6 length

regexowl:
The "needs IQE update" label should be used when an IQE update is needed for the pr_check to pass. This step will need an update eventually when it's released to stage-stable, but should be fine right now since pr_check doesn't run on stage-preview. Removing the label.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -155,14 +162,64 @@ describe('Step Users', () => {
...
+    );
+    await waitFor(() => expect(invalidUserMessage));
+    const nextButton = await getNextButton();
+    await waitFor(() => expect(nextButton).toBeDisabled());
+  });
 });
 
 describe('User request generated correctly', () => {
-  test('with valid name, ssh key and checked Administrator checkbox', async () => {
+  test('create Guest-Image with valid name, password, ssh key and checked Administrator checkbox', async () => {
```
regexowl, line 222: This is very nitpicky - do we have to specify the guest-image here? I mean this should work for all targets, but Azure. So I would just mention Azure where it's applicable as it makes more sense to stress an exception than specify the "default".

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -171,12 +228,13 @@ describe('User request generated correctly', () => {
...
     await addUserName(validUserName);
     await addSshKey(validSshKey);
+    await addPassword('mockPassword123');
     const nextButton = await getNextButton();
     await waitFor(() => expect(nextButton).toBeEnabled());
     const isAdmin = screen.getByRole('checkbox', {
       name: /administrator/i,
     });
-    user.click(isAdmin);
+    await user.click(isAdmin);
```
regexowl, line 237: Using `await` with user actions in the main part of the tests causes `missing act()` warning in the test output.

So we either need to:
- remove the await (which can cause sync problems)
- replace it with `await waitFor(() => user.click(imageOutput));`
- or better - move the longer clicking and typing parts into a helper function where we can await everything, but the test output stays clean of the `missing act()` errors

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { useState } from 'react';
```
regexowl, line 1: There's a following warning in the test output:
```
Text input: Text input requires either an id or aria-label to be specified
```

seems like there's aria-label correctly specified in the PasswordValidatedInput, so this might be because of a different field, but we should take a look at it.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -389,28 +433,108 @@ export function useUsersValidation(): StepValidation {
...
+  if (strCount < 3)
+    return { variant: 'error', icon: <ExclamationCircleIcon />, text: 'Weak' };
+  if (strCount < 5)
+    return {
+      variant: 'warning',
+      icon: <ExclamationTriangleIcon />,
+      text: 'Medium',
+    };
+  return { variant: 'success', icon: <CheckCircleIcon />, text: 'Strong' };
+};
```
regexowl, line 523: How would I get to see any of these apart from the "Strong"?

image

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -389,28 +433,108 @@ export function useUsersValidation(): StepValidation {
...
+
+const countCharacterTypes = (value: string) => {
+  const counts = {
+    lowercase: (value.match(/[a-z]/g) || []).length,
+    uppercase: (value.match(/[A-Z]/g) || []).length,
+    digits: (value.match(/\d/g) || []).length,
+    special: (value.match(/\W/g) || []).length,
+  };
+  return {
+    rulesCount: Object.values(counts).filter((count) => count > 0).length,
```
regexowl, line 533: Couldn't we just return a number instead of an object? Are we using the structure of object anywhere? I can see this: `rulesCount >= 3`, but no distinction between `lowercase`, `uppercase`, `digits` or `special`.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -389,28 +433,108 @@ export function useUsersValidation(): StepValidation {
...
+    strCount: Object.values(counts).reduce((a, b) => a + b, 0),
```
regexowl, line 534: Is this just a length of the string? If I understand correctly, it just adds all the counts together, right?

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -389,28 +433,108 @@ export function useUsersValidation(): StepValidation {
...
+): ValidationFlags => {
+  const isUserNameValidValue = !!userName && isUserNameValid(userName);
+  const isSshKeyValidValue = !!userSshKey && isSshKeyValid(userSshKey);
+  const isPasswordValidValue =
+    !!userPassword && checkPasswordValidity(userPassword).isValid;
+  return { isUserNameValidValue, isSshKeyValidValue, isPasswordValidValue };
+};
+
+export const checkPasswordValidity = (
+  value: string
```
regexowl, line 471: Very nitpicky, but could we rename `value` to `password`?

regexowl (commented):
Went through the changes - it feels like a lot was discussed outside of the PR so if I commented on something that was already agreed on in a different thread/channel, let me know.

Overall I like the new validation pattern, we could use this as a template for reworking other validation fields :wrench:

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
regexowl: Checked again - it's caused by removing the `id` from `<TextInput>`

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const passStrLabel = (
+    <HelperText>
+      <HelperTextItem variant={strength.variant} icon={strength.icon}>
+        {strength.text}
+      </HelperTextItem>
+    </HelperText>
+  );
+
+  return (
+    <Form>
```
regexowl, line 54: There is already `<Form>` in the `User.tsx` index file, so this triggers:
```
Warning: validateDOMNesting(...): <form> cannot appear as a descendant of <form>.
```
in the test output.

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+    setIsPasswordVisible(!isPasswordVisible);
+  };
+
+  const passStrLabel = (
+    <HelperText>
+      <HelperTextItem variant={strength.variant} icon={strength.icon}>
+        {strength.text}
+      </HelperTextItem>
+    </HelperText>
+  );
```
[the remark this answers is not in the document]
regexowl: Following the password strength component more closely - the helper text shouldn't switch so strong label will stay and the second helper will be updated to not duplicate info.

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+            </InputGroupItem>
+          </InputGroup>
+        </>
+        <FormHelperText>
+          <HelperText component="ul">
+            {environments.includes('azure') ? (
+              <>
+                <HelperTextItem variant={ruleLength} component="li" hasIcon>
+                  A password for the target environment Azure must be at least 6
+                  characters long. Please enter a longer password.
```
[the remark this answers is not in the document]
regexowl: This is just about the suggested text change, sorry about not being more clear :sweat_smile:

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+                </HelperTextItem>
+                <HelperTextItem variant={ruleCharacters} component="li" hasIcon>
+                  WARNING: This password seems weak, please use with caution or
+                  include at least 3 of the following: lowercase letter,
+                  uppercase letters, numbers, symbols
+                </HelperTextItem>
```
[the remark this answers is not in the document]
regexowl: Also just about the suggested text change, maybe we could even update it to:
```suggestion
                <HelperTextItem variant={ruleCharacters} component="li" hasIcon>
                  For strong password include at least 3 of the following: lowercase letter,
                  uppercase letters, numbers, symbols
                </HelperTextItem>
```

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+      >
+        <>
+          <InputGroup>
+            <InputGroupItem isFill>
+              <TextInput
+                isRequired
+                type={isPasswordVisible ? 'text' : 'password'}
+                value={value}
+                onChange={onChange}
+                arial-label={ariaLabel}
```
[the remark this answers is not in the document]
regexowl: I wouldn't notice that :flushed:

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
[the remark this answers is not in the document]
mgold1234: I will keep the Strong label as we see here https://www.patternfly.org/components/password-strength/
and change the Warning to strong password

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+                <HelperTextItem variant={ruleCharacters} component="li" hasIcon>
+                  WARNING: This password seems weak, please use with caution or
+                  include at least 3 of the following: lowercase letter,
+                  uppercase letters, numbers, symbols
+                </HelperTextItem>
+              </>
+            ) : (
+              <HelperTextItem variant={ruleLength} component="li" hasIcon>
+                Password helps protect your account, we recommend a password of
+                at least 6 characters.
```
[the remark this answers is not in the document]
mgold1234: yes we can

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -155,14 +162,64 @@ describe('Step Users', () => {
...
+    const user = userEvent.setup();
+    await renderCreateMode();
+    await goToRegistrationStep();
+    await clickRegisterLater();
+    await goToUsersStep();
+    await clickAddUser();
+    const imageOutput = screen.getByRole('button', {
+      name: /image output/i,
+    });
+    await user.click(imageOutput);
```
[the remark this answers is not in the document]
mgold1234: when I put screen.logTestingPlaygroundURL() after row 174, I dont see the Microsoft Azure option in the image output
[1:08](https://redhat-internal.slack.com/archives/D05E4AXU10D/p1742296112676939)
and that is the reason I go to user and then come back to image output step
and if I print it after row 181, I see the Azure option
Screenshot 2025-03-18 at 13 03 54
not sure why we cant see it immediately

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+                placeholder={placeholder}
+              />
+            </InputGroupItem>
+            <InputGroupItem>
+              <Button
+                variant="control"
+                onClick={togglePasswordVisibility}
+                aria-label={
+                  isPasswordVisible ? 'Show password' : 'Hide password'
+                }
```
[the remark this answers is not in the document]
mgold1234: that's really help :)

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
mgold1234: the aria-label also will solve it :)

src/Components/CreateImageWizard/utilities/useValidation.tsx
mgold1234: as discussed in the slack channel, we will see only the Strong label as it shows in https://www.patternfly.org/components/password-strength/
remove other conditions

src/Components/CreateImageWizard/utilities/useValidation.tsx
mgold1234: right :)

src/Components/CreateImageWizard/utilities/useValidation.tsx
mgold1234: good point

src/Components/CreateImageWizard/utilities/useValidation.tsx
mgold1234: right

src/Components/CreateImageWizard/utilities/useValidation.tsx
mgold1234: correct

src/.env
```diff
@@ -0,0 +1,3 @@
+USER_INVALID_PASSWORD='inval';
+USER_VALID_PASSWORD='validPassword';
```
regexowl, line 2: Why are you setting testing values as environment variables? Those should be just constants in the test file or a fixture.

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -159,10 +167,53 @@ describe('Step Users', () => {
     const invalidUserMessage = screen.getByText(/invalid ssh key/i);
     await waitFor(() => expect(invalidUserMessage));
   });
+
+  test('try to create Azure image with invalid password', async () => {
+    const userPassword = process.env.USER_INVALID_PASSWORD || 'inval';
```
regexowl, line 172: Just constants, not env variables

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -159,10 +167,53 @@ describe('Step Users', () => {
...
+      /for a strong password, include at least 3 of the following: lowercase letters, uppercase letters, numbers, symbols./i
+    );
+    await waitFor(() => expect(invalidUserMessage));
+    await waitFor(() => expect(warningUserMessage));
+    const nextButton = await getNextButton();
+    await waitFor(() => expect(nextButton).toBeDisabled());
+  });
+
+  test('with invalid password', async () => {
+    const userPassword = process.env.USER_INVALID_PASSWORD || 'inval';
```
regexowl, line 197: Not env variables

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const passStrLabel = (
+    <HelperText>
+      <HelperTextItem variant={strength.variant} icon={strength.icon}>
+        {strength.text}
+      </HelperTextItem>
+    </HelperText>
+  );
+
+  return (
+    <FormGroup label="Password" isRequired labelInfo={passStrLabel}>
```
regexowl, line 56: I wouldn't say this is a strong password :sweat_smile: could we show the label for Azure targets only? What do you think?

image

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <FormHelperText>
+        <HelperText component="ul">
+          <HelperTextItem variant={ruleLength} component="li" hasIcon>
+            Password must be at least 6 characters long.
+          </HelperTextItem>
+          {environments.includes('azure') && (
+            <HelperTextItem variant={ruleCharacters} component="li" hasIcon>
+              For a strong password, include at least 3 of the following:
+              lowercase letters, uppercase letters, numbers, symbols.
+            </HelperTextItem>
```
regexowl, line 89: Since the char categories are required, let's shorten the helper text a bit more

image

```suggestion
            <HelperTextItem variant={ruleCharacters} component="li" hasIcon>
              Must include at least 3 of the following:
              lowercase letters, uppercase letters, numbers, symbols.
            </HelperTextItem>
```

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
mgold1234: As I understand from this conversation -https://redhat-internal.slack.com/archives/C05LNKHBCDA/p1742306355820419?thread_ts=1742303474.431899&cid=C05LNKHBCDA
if we have one condition (all the validation rule are passing-> so the passwrod strengh should show

regexowl: I understand it the same way. Let's keep it in this PR, open an issue and discuss how to handle this in a follow up. Showing "strong password" for weak passwords could be both confusing and misleading for the user, but since we don't have any other criteria for "other targets" passwords, we will need to explore when to show the label.

mgold1234: I dont have strong opinion here, but I agree that it is not make sense to see the strong label with just few letters, 
it is very simple fix, to keep it only for Azure targets,
so I am ok to keep this discussion here, and see what @lucasgarfield will say

regexowl (approved):
Thank you! Let's discuss the strong label and update in a follow up if needed, this is already a big PR as is.

PR #2849, opened 2025-02-05
vitests: Set global retry (HMS-5458)

regexowl:
This sets retry globally to `3`, so we don't have to add it to the tests individually.

For the tests that are reliable this shouldn't have any effect, but it will trigger on tests that proved to be a bit flakey (like revisit tests).

JIRA: [HMS-5458](https://issues.redhat.com/browse/HMS-5458)

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,5 +1,6 @@
-import react from '@vitejs/plugin-react';
 import path from 'path';
+
+import react from '@vitejs/plugin-react';
 
 const config = {
   plugins: [react()],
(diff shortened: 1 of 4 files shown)


vitest.config.ts
```diff
@@ -1,6 +1,7 @@
-import react from '@vitejs/plugin-react';
 import path from 'path';
 
+import react from '@vitejs/plugin-react';
+
```
regexowl, line 4: This was highlighted as an ESLint problem so thought I might as well fix it in one PR.

regexowl:
There is no review, disabling the auto-merge

mgold1234 (approved):
/lgtm

regexowl:
pr_check is failing, merging manually, this is just an update to test settings and shouldn't break anything

PR #2854, opened 2025-02-05
validators: Remove unneeded condition for service validator (HMS-5464)

regexowl:
After the regex was fixed this condition should be removed.

JIRA: [HMS-5464](https://issues.redhat.com/browse/HMS-5464)


croissanne (approved):

croissanne (approved):
Indeed, ty!

regexowl:
> Indeed, ty!

Thank you for fixing the validation pattern :sweat_smile:

PR #2858, opened 2025-02-06
Wizard: change password to **** if there is value in password field

mgold1234:
this changes password to **** if there is value in password field

Screenshot 2025-02-06 at 12 56 54

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
@@ -818,7 +818,7 @@ export const UsersList = () => {
             Password
           </TextListItem>
           <TextListItem component={TextListItemVariants.dd}>
-            {userPassword ? userPassword : 'None'}
+            {userPassword ? '●'.repeat(8) : 'None'}
           </TextListItem>
           <TextListItem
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -817,7 +817,7 @@ export const UsersList = () => {
             Password
           </TextListItem>
           <TextListItem component={TextListItemVariants.dd}>
-            {userPassword ? userPassword : 'None'}
+            {userPassword ? '●'.repeat(8) : 'None'}
```
regexowl, line 820: I wonder if we should allow to show the password on the review step with a similar toggle we use on the Users step (the eye icon).

Or maybe not render the password at all? I guess that would be a bit weird. It's just that rendering a hidden password does not seem super useful. It gives the information that _a_ password was selected, but when the user want's to check what password that was they have to return to the user step.

regexowl: Ah, gotcha. Alright, let's get this in and if we need a "show password" toggle we can add it in a follow up.

regexowl (approved):

PR #2863, opened 2025-02-07
cockpit/spec: add obsoletes & fix release workflow

croissanne:
With #2862  moving the artefact out of the checkout dir got reverted.

Adds an obsoletes for cockpit-composer.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,6 +25,8 @@ jobs:
       - name: Make dist
         run: |
           make dist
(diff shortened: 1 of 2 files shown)


thozza (approved):
LGTM. Looking at the diff made me wonder if the minimum required osbuild-composer version is correct? Since you mentioned that you've been waiting for the package search functionality which landed just recently.

croissanne:
> LGTM. Looking at the diff made me wonder if the minimum required osbuild-composer version is correct? Since you mentioned that you've been waiting for the package search functionality which landed just recently.

oh good point, let me fix that

PR #2867, opened 2025-02-10
dependabot: Increase limit of opened pull requests

regexowl:
This will temporarily increase limit of opened pull requests, until the frontend-components bumps get resolved.


regexowl:
This might be nice so we don't let other deps "unmaintained" while frontend-components bumps are blocked.

regexowl:
pr_check failed, but this will have no effect on IQE, merging manually

PR #2871, opened 2025-02-10
Wizard: Remove first boot script from services when no script (HMS-5481)

regexowl:
There is a bug that makes `custom-first-boot` service stay in services when the first boot script is removed.

How to reproduce:
1. create a blueprint with a first boot script
2. download the blueprint and confirm `custom-first-boot` was added to enabled services
3. click on "Edit blueprint"
4. go to First boot step and remove the script
5. save edited blueprint

Current behaviour:
- `custom-first-boot` service should be still enabled even with the removed first boot script

Updated behaviour:
- `custom-first-boot` is no longer in the blueprint after first boot script got removed

JIRA: [HMS-5481](https://issues.redhat.com/browse/HMS-5481)

### DIFF
--- a/src/Components/CreateImageWizard/utilities/requestMapper.ts
+++ b/src/Components/CreateImageWizard/utilities/requestMapper.ts
@@ -566,6 +566,13 @@ const getServices = (state: RootState): Services | undefined => {
     enabledSvcs = [...enabledSvcs, FIRST_BOOT_SERVICE];
   }
   if (
+    (!selectFirstBootScript(state) ||
+      selectFirstBootScript(state).length === 0) &&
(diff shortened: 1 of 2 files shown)


lucasgarfield (approved):
Nice catch! I wonder if this type of thing would be worth checking for in our tests? :thinking:

regexowl:
> Nice catch! I wonder if this type of thing would be worth checking for in our tests? 🤔

It would be for sure. Making the code block work properly with the tests is a bit of pain, but tried to add one which uses the upload function with an empty script (body of the code block does not seem reachable via testing-library methods).

lucasgarfield (approved):
Thanks for adding the test! The fix works great!

PR #2872, opened 2025-02-10
Wizard: Refactor hostname and blueprint name inputs (HMS-5616)

mgold1234:
This commit Implement refactor of HookValidatedInput for hostname and blueprint name
fields, addressing the following bugs:

1) Fixes a bug where the validation symbol persisted after a user deleted the value in the hostname field.
2) Fixes a bug where the validation symbol persisted after a user deleted the value in the blueprint name field.

These changes improve code maintainability and provide a more consistent user experience.
FIX ISSUE: [2917](https://github.com/osbuild/image-builder-frontend/issues/2917), https://github.com/osbuild/image-builder-frontend/issues/2971
JIRA: [HMS-5616](https://issues.redhat.com/browse/HMS-5616)

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -50,6 +50,7 @@ type ValidationInputProp = TextInputProps &
       event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
       value: string
     ) => void;
+    isRequired?: boolean;
   };
 
 type ErrorMessageProps = {
@@ -117,12 +118,13 @@ export const ValidatedInputAndTextArea = ({
   onChange,
   ariaLabel,
   inputType = 'textInput',
+  isRequired,
 }: ValidationInputProp) => {
   const errorMessage = stepValidation.errors[fieldName];
   const hasError = errorMessage !== '';
 
   const [isPristine, setIsPristine] = useState(!value);
-  const validated = getValidationState(isPristine, errorMessage);
+  const validated = getValidationState(isPristine, errorMessage, isRequired);
 
   const handleBlur = () => {
     if (value) {
@@ -164,9 +166,14 @@ export const ValidatedInputAndTextArea = ({
 
 const getValidationState = (
   isPristine: boolean,
-  errorMessage: string
+  errorMessage: string,
+  isRequired: boolean | undefined
 ): ValidationResult => {
-  const validated = isPristine ? 'default' : errorMessage ? 'error' : 'success';
+  const validated = isPristine
+    ? 'default'
+    : (isRequired && errorMessage) || errorMessage
+    ? 'error'
+    : 'success';
 
   return validated;
 };
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -250,15 +250,14 @@ export function useHostnameValidation(): StepValidation {
   const errorMessage =
     'Invalid hostname. The hostname should be composed of up to 64 7-bit ASCII lower-case alphanumeric characters or hyphens forming a valid DNS domain name. It is recommended that this name contains only a single label, i.e. without any dots.';
 
-  if (!isHostnameValid(hostname)) {
-    return {
-      errors: {
-        hostname: errorMessage,
-      },
-      disabledNext: true,
-    };
-  }
-  return { errors: {}, disabledNext: false };
+  const hostnameError = !isHostnameValid(hostname) ? errorMessage : '';
+
+  return {
+    errors: {
+      hostname: hostnameError,
+    },
+    disabledNext: !!hostnameError,
+  };
 }
 
 export function useKernelValidation(): StepValidation {
@@ -479,6 +478,9 @@ export function useDetailsValidation(): StepValidation {
   }, [blueprintId, name, setUniqueName, trigger, nameValid]);
 
   let nameError = '';
+  if (!name) {
+    nameError = 'Blueprint name is required';
+  }
   if (name && !nameValid) {
     nameError = 'Invalid blueprint name';
   } else if (uniqueName === false) {
--- a/src/Components/CreateImageWizard/steps/Details/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Details/index.tsx
@@ -19,7 +19,10 @@ import {
 } from '../../../../store/wizardSlice';
 import { useGenerateDefaultName } from '../../utilities/useGenerateDefaultName';
 import { useDetailsValidation } from '../../utilities/useValidation';
-import { HookValidatedInput } from '../../ValidatedInput';
+import {
+  HookValidatedInput,
+  ValidatedInputAndTextArea,
+} from '../../ValidatedInput';
 
 const DetailsStep = () => {
   const dispatch = useAppDispatch();
@@ -55,15 +58,15 @@ const DetailsStep = () => {
         blueprint.
       </Text>
       <FormGroup isRequired label="Blueprint name" fieldId="blueprint-name">
-        <HookValidatedInput
+        <ValidatedInputAndTextArea
           ariaLabel="blueprint name"
-          dataTestId="blueprint"
           value={blueprintName}
           isDisabled={false}
           onChange={handleNameChange}
           placeholder="Add blueprint name"
           stepValidation={stepValidation}
           fieldName="name"
+          isRequired={true}
         />
         <FormHelperText>
           <HelperText>
--- a/src/Components/CreateImageWizard/steps/Hostname/components/HostnameInput.tsx
+++ b/src/Components/CreateImageWizard/steps/Hostname/components/HostnameInput.tsx
@@ -8,7 +8,7 @@ import {
   selectHostname,
 } from '../../../../../store/wizardSlice';
 import { useHostnameValidation } from '../../../utilities/useValidation';
-import { HookValidatedInput } from '../../../ValidatedInput';
+import { ValidatedInputAndTextArea } from '../../../ValidatedInput';
 
 const HostnameInput = () => {
   const dispatch = useAppDispatch();
@@ -22,7 +22,7 @@ const HostnameInput = () => {
 
   return (
     <FormGroup label="Hostname">
-      <HookValidatedInput
+      <ValidatedInputAndTextArea
         ariaLabel="hostname input"
         value={hostname}
         onChange={handleChange}


lucasgarfield:
Nice catch, the validation should not appear when there is no hostname.

I’m not sure about the approach, though. In this PR we’re hiding the validation by returning an error object with a fieldname of ‘default’. That technically works and is supported by the HookValidatedInput but it just feels kind of surprising/counter-intuitive.

I think it feels counter-intuitive because `<HookValidatedInput>` does not surface validation on pristine state components. So there is already an existing pattern for not displaying validation – that’s what I intuitively would expect to use. In other words, set isPristine to true when the field’s value returns to the default state (in this case, ‘’).

This might be a good opportunity to refactor <HookValidatedInput> a little bit. The double nested ternary operator is hard to understand. It definitely does support this ‘feature’ where if the fieldName is ‘default’ then but let’s take a step back and ask… should it? Maybe we can get rid of the nested ternary and the concept of a ‘default’ state. Having to manage both ‘default’ and ‘isPristine’ feels duplicitous.

In all the validation hooks, it only seems to be used in one other place, and there is a comment specifically calling it out as a hacky solution…
image

Therefore, I am a little hesitant to introduce this new pattern to the code base. I’d prefer refactoring a bit. What do you think?

lucasgarfield (commented):
Nice catch, the validation should not appear when there is no hostname.

I’m not sure about the approach, though. In this PR we’re hiding the validation by returning an error object with a fieldname of ‘default’. That technically works and is supported by the HookValidatedInput but it just feels kind of surprising/counter-intuitive.

I think it feels counter-intuitive because `<HookValidatedInput>` does not surface validation on pristine state components. So there is already an existing pattern for not displaying validation – that’s what I intuitively would expect to use. In other words, set isPristine to true when the field’s value returns to the default state (in this case, ‘’).

This might be a good opportunity to refactor <HookValidatedInput> a little bit. The double nested ternary operator is hard to understand. It definitely does support this ‘feature’ where if the fieldName is ‘default’ then but let’s take a step back and ask… should it? Maybe we can get rid of the nested ternary and the concept of a ‘default’ state. Having to manage both ‘default’ and ‘isPristine’ feels duplicitous.

In all the validation hooks, it only seems to be used in one other place, and there is a comment specifically calling it out as a hacky solution…
image

Therefore, I am a little hesitant to introduce this new pattern to the code base. I’d prefer refactoring a bit. What do you think?

mgold1234:
after some investigation, the solution that I did here https://github.com/osbuild/image-builder-frontend/pull/2844/files#diff-5206f54a601e8c56290e1a412fbf3f35172f16b892c7674852c91b4ea846bcd9R62
will fix that problem, so I can close this pr

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -188,15 +188,22 @@ export function useHostnameValidation(): StepValidation {
...
-      errors: {
-        hostname: errorMessage,
-      },
-      disabledNext: true,
-    };
+    hostnameError = errorMessage;
+    disabledNext = true;
+  } else if (!hostname) {
+    // Hack to keep the error message from flickering in create mode
+    hostnameError = 'default';
```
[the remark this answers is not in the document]
mgold1234: do you think we can fix that in a different way?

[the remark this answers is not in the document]
regexowl: I've originally added the comment as a quick reaction to the code comment, let me expand a bit :sweat_smile: 

I think hostname validation might not be the right place to solve this issue - the problem lies in the way `ValidatedInput` handles the `validated` value. Meaning if we update `hostnameValidation` we will later need to update other inputs that use `HookValidatedInput` in the same way, causing duplicate and redundant code.

Also as the original comment in the code said, it was a bit of a hack, that solves the error text flashing when the value is being populated by hooks, which is not happening in the case of hostname as that one is populated by user input only.

So what I think we might need to do in this case - going back to what @lucasgarfield mentioned:
> I think it feels counter-intuitive because <HookValidatedInput> does not surface validation on pristine state components. So there is already an existing pattern for not displaying validation – that’s what I intuitively would expect to use. In other words, set isPristine to true when the field’s value returns to the default state (in this case, ‘’).

We could probably return to the "pristine" state when the input is cleared, which would set the field validation to "default"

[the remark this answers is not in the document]
mgold1234: thank you @regexowl I got your point, and it is make sense to fix it in the `HookValidatedInput`
to prevent duplicate code

regexowl:
One small bug in FSC - the input doesn't get validated without a value
image

this is how the error for invalid value looks like:
image

mgold1234:
nice catch @klara, I am looking at stage, if user delete the value in Blueprint name we should see an error be cecause it's mandatory field, am I wrong?
I fixed the FSC respectivally

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -606,7 +606,6 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
             name="Details"
             id="step-details"
             navItem={customStatusNavItem}
-            status={detailsValidation.disabledNext ? 'error' : 'default'}
```
regexowl, line 609: If we remove this we'll lose the status indication of Details step, so this needs to stay.

src/Components/CreateImageWizard/validators.ts
```diff
@@ -33,13 +33,23 @@ export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
 };
 
 export const isMountpointMinSizeValid = (minSize: string) => {
+  if (!minSize) {
+    return false;
+  }
```
regexowl, line 38: This might be fine here as 'Must be larger than 0' is kinda fitting for both invalid and missing value, but ... (continuing in a comment lower)

src/Components/CreateImageWizard/validators.ts
```diff
@@ -33,13 +33,23 @@ export const isGcpEmailValid = (gcpShareWithAccount: string | undefined) => {
...
 };
 
-export const isBlueprintNameValid = (blueprintName: string) =>
-  blueprintName.length >= 2 &&
-  blueprintName.length <= 100 &&
-  /\w+/.test(blueprintName);
+export const isBlueprintNameValid = (blueprintName: string) => {
+  if (!blueprintName) {
+    return false;
+  }
```
regexowl, line 45: This might be better in `useDetailsValidation`, because then we could add a custom error message that could be something like: 'Blueprint name is required'

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -454,7 +454,7 @@ export function useDetailsValidation(): StepValidation {
   }, [blueprintId, name, setUniqueName, trigger, nameValid]);
 
   let nameError = '';
-  if (name && !nameValid) {
+  if (!nameValid) {
```
regexowl, line 457: We might want to add an empty value condition here so we're able to render a custom error message for a missing value.

src/Components/CreateImageWizard/CreateImageWizard.tsx
mgold1234: thanks for that, I know but we have kind of wired bug, 
when we open the create image screen for the first time, we see that this step appears with error, maybe it takes few second until we get the value in blueprint name field?
or maybe it is related to https://github.com/osbuild/image-builder-frontend/issues/2929?

regexowl: This is connected to fact that we populate the name of blueprint after getting to the step. Meaning there is an empty value for blueprint name in all the previous steps, thus the error icon in navigation.

I can think of two possible ways of how to resolve that - either we could update details step validation or `customStatusNavItem` in CreateImageWizard.tsx to reflect that there's a bit more unique pattern in place.

Or we could trigger the blueprint name generation earlier so the field gets filled in and the validation doesn't trip on the empty field.

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -115,8 +119,17 @@ export const HookValidatedInput = ({
...
-    setIsPristine(false);
+    if (!value && !isRequired) {
+      setIsPristine(true);
+    } else {
+      setIsPristine(false);
+    }
   };
+  useEffect(() => {
+    if (!value && !isRequired) {
+      setIsPristine(true);
```
lucasgarfield, line 130: Lines 129 and 130 seem to duplicate 122 and 123.

lucasgarfield (changes requested):
Should this be rebased on top of https://github.com/osbuild/image-builder-frontend/pull/2922?

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -479,6 +478,9 @@ export function useDetailsValidation(): StepValidation {
   }, [blueprintId, name, setUniqueName, trigger, nameValid]);
 
   let nameError = '';
+  if (!name) {
```
lucasgarfield, line 481: :+1: This is a great catch, nice job! Did we have an issue tracking this or did you just find it while working on the pull request? It seems like this bug has been here for a long time!

lucasgarfield (changes requested):

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -50,6 +50,7 @@ type ValidationInputProp = TextInputProps &
       event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
       value: string
     ) => void;
+    isRequired?: boolean;
```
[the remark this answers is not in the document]
mgold1234: right, if I define it as false I can remove that, thank you

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -117,21 +118,18 @@ export const ValidatedInputAndTextArea = ({
   onChange,
   ariaLabel,
   inputType = 'textInput',
+  isRequired = false,
```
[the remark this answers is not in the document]
mgold1234: remove the default, thank you

src/Components/CreateImageWizard/utilities/useValidation.tsx
mgold1234: I will open one - https://github.com/osbuild/image-builder-frontend/issues/2971

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -117,21 +118,18 @@ export const ValidatedInputAndTextArea = ({
...
 
   const [isPristine, setIsPristine] = useState(!value);
   const validated = getValidationState(isPristine, errorMessage);
 
-  const handleBlur = () => {
-    if (value) {
-      setIsPristine(false);
-    }
-  };
+  const handleBlur = () => setIsPristine(!value && !isRequired);
```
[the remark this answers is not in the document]
mgold1234: no it is not make sense, it makes sense to check that in the validation

regexowl (approved):
LGTM, thank you! :sparkles: And the blueprint name bug is a really nice catch! :bug:

Issue #2864, opened 2025-02-07
Searching for a package should switch view to "Available"


lucasgarfield:
If a user is on the "Selected" tab and begins searching for a package (simply by typing into the search field), the view should switch back to the "Available" tab to show packages being searched for.

![Image](https://github.com/user-attachments/assets/bb8bcb9d-c065-490b-a058-d10f5a19e936)

regexowl:
Resolved in https://github.com/osbuild/image-builder-frontend/pull/2873

PR #2873, opened 2025-02-10
Wizard: Switch view to "Available" when searching for package (HMS-5480)

mgold1234:
this fix an issue when user search for a package, the viewv stay "Selected" instead of switching to "Available

FIX ISSUE: (https://github.com/osbuild/image-builder-frontend/issues/2864)

JIRA: [HMS-5480](https://issues.redhat.com/browse/HMS-5480)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -784,6 +784,7 @@ const Packages = () => {
   ) => {
     setSearchTerm(selection);
     setToggleSourceRepos(RepoToggle.INCLUDED);
+    setToggleSelected('toggle-available');
   };
 
   const handleClear = async () => {
--- a/src/test/Components/CreateImageWizard/steps/Packages/Packages.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Packages/Packages.test.tsx
@@ -327,9 +327,10 @@ describe('Step Packages', () => {
     await clearSearchInput();
     await typeIntoSearchBox('test');
 
-    const availablePackages = await getRows();
-    expect(availablePackages[0]).toHaveTextContent('test');
-    expect(availablePackages[1]).toHaveTextContent('test-sources');
+    await toggleSelected();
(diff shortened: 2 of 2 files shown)


lucasgarfield:
Looks good. Thanks for taking this one! 

Would you mind getting the tests passing?

mgold1234:
after this small change - setToggleSelected('toggle-available');
one was failing-
`should display relevant results in selected first`
I investigate it and figure out that when we press on  

    await toggleSelected();
    await clearSearchInput();
    await typeIntoSearchBox('test');
    we now see all the list - available and selected packages ( this is the change that I did)
    so we should update the fixture with all the pakcages -
    https://github.com/osbuild/image-builder-frontend/pull/2873/files#diff-0e6259b7bae68a7c74bc6a5d690681a6399caa397d3ba5abe457064c699e8c9bR31
    
    
and now there are 3 tests that failing, and it is very wired,

you can see in the recording that the packages that were added to the fixutre now appear twice.

https://github.com/user-attachments/assets/a06cc288-14ea-43bb-95bf-766ebac11c6e

I think this is a bug in the mock data, I couldnt reproduce that manully


@lucasgarfield need your advice here

regexowl:
Ah nice! The problem with the tests suddenly makes sense after noticing it should sort selected packages :sweat_smile: but that brings a question - aren't we losing the option to sort through those?

Imagine a situation where you have one hundred selected packages. And you want to know if `tmux` is one of them. You'd have to click through pages, scroll through and check manually, right? Previously you could just search the package in "Selected packages" and the selected packages would get filtered. But now it would automatically switch you to available and you would have to switch back to Selected to see the results.

mgold1234:
While I understand the concern about losing the ability to filter within the 'Selected' tab, my belief is that user primarily initiates a search when he intend to add new packages. User familiar with his current selections would likely navigate directly to the 'Selected' tab for review, rather than using search to locate specific items within that list.
@lucasgarfield wdyt?

regexowl (approved):
I mean sure, let's give it a try and see what users think

PR #2878, opened 2025-02-11
Wizard: Fix nav status icon (HMS-5489)

### LINKED ISSUE #2700: Wizard: Invalid values are not obvious on import
The error icon doesn't appear next to the step name if the step was skipped on import.

How to reproduce:
1. import toml blueprint with invalid hostname like "-base-image-"
2. select a target and click next
3. click "Review and finish" button

Outcome:
- hostname step doesn't have error icon next to its name in the left side navigation, the "Create blueprint" button is disabled without any indication why

Expected outcome:
- hostname step has the error icon next to its name

There might a problem with the pristine status of the navigation item, if clicked after skipping to review the validation gets applied properly and clicking away with invalid value adds the error icon immediately.

regexowl:
Fixes #2700

Status icon rendered only when the step was previously visited, this caused problems with imported blueprints.

How to reproduce a bug:
1. import a blueprint with invalid hostname
2. skip to Review

Current behaviour:
- the Create blueprint button is disabled, but there is no indication of error next to the Hostname step in navigation

After fix:
- the error status should be rendered next to the Hostname nav item

JIRA: [HMS-5489](https://issues.redhat.com/browse/HMS-5489)

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -282,16 +282,14 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
     );
 
     // Only this code is different from the original
-    const status =
-      (step.isVisited && step.id !== activeStep?.id && step.status) ||
-      'default';
+    const status = (step.id !== activeStep.id && step.status) || 'default';
 
     return (
       <WizardNavItem
         key={step.id}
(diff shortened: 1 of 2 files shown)


regexowl:
Registration and Details which get populated upon visit shouldn't have an error status next to their name.

ezr-ondrej (approved):
Oh, right, yeah looks like a reasonable fix :+1:

PR #2888, opened 2025-02-12
test: add gating tests (tmt)

croissanne:
Adds gating tests, and runs them on PRs.

Also let's release into fedora 42!


mgold1234 (dismissed):
/lgtm

croissanne:
Actually, let's revisit adding gating tests until we have the playwright tests up and running. Then we could just point those at the installed rpm. I'll open a separate PR for releasing into f42.

PR #2894, opened 2025-02-13
Wizard: Add step validation to Locale (HMS-5505)

regexowl:
This adds step validation to Locale, allowing to properly validate imported values.

JIRA: [HMS-5505](https://issues.redhat.com/browse/HMS-5505)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
+++ b/src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
@@ -3,6 +3,8 @@ import React, { useEffect, useState } from 'react';
 import {
   Button,
   FormGroup,
+  HelperText,
+  HelperTextItem,
   MenuToggle,
   MenuToggleElement,
   Select,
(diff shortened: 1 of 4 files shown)


mgold1234:
if I am choosing value in 'Languages' that I already choose, there is no error message show, but the value is not added, myabe we should disable choosen value?

regexowl:
> if I am choosing value in 'Languages' that I already choose, there is no error message show, but the value is not added, myabe we should disable choosen value?

Ooo, nice catch! Let me take a look at that :eyes:

regexowl:
Added a new commit to handle the already added options: 
image

Was also thinking about filtering the selected values away from the list, but feel like this communicates things better than just removing the option.

src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
```diff
@@ -20,12 +22,17 @@ import {
...
 import { keyboardsList } from '../keyboardsList';
 
 const KeyboardDropDown = () => {
   const keyboard = useAppSelector(selectKeyboard);
   const dispatch = useAppDispatch();
 
+  const stepValidation = useLocaleValidation();
+
+  const [errorText, setErrorText] = useState(stepValidation.errors['keyboard']);
+
```
mgold1234, line 35: why we need empty line here?

regexowl: We don't, removed

mgold1234:
@regexowl looks great, thank you!

mgold1234 (approved):
/lgtm

PR #2897, opened 2025-02-13
src: add certificates into files

avitova:
Missing:

- [ ] adding a mechanism (update-ca-trust) to set certificates as trusted

This does not add any UI components, but this is a PR to prepare for when we have mocks, and can add the Certificates step. :)


avitova:
Sorry, closing this - let's try a different approach.

PR #2903, opened 2025-02-17
Wizard: Remove deprecated selects

regexowl:
This swaps deprecated PF4 selects for non-deprecated PF5 ones in few places.


avitova:
lgtm, let me know once this is ready from the iqe side.

lucasgarfield (approved):
@avitova has already reviewed and approved. This PR was pending updates to IQE but we have made the decision to abandon IQE tests. Our current unit test coverage is considered to be adequate as a stop gap until Playwright tests are ready, the only real difference being mocked vs. real APIs.

PR #2909, opened 2025-02-18
Playwright tests (HMS-5597)

croissanne:
Initial test using the playwright container, as playwright isn't officially supported on fedora or el.

JIRA: [HMS-5597](https://issues.redhat.com/browse/HMS-5597)


package.json
```diff
@@ -34,6 +34,7 @@
     "@babel/preset-react": "7.26.3",
     "@babel/preset-typescript": "7.26.0",
     "@patternfly/react-icons": "5.4.2",
+    "@playwright/test": "^1.50.1",
```
regexowl, line 37: Can you please pin the versions of newly added deps?

croissanne: oh woops, yes of course

playwright/test.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    await frame.getByRole('heading', { name: 'Images About image builder' });
+    await frame.getByRole('heading', { name: 'Blueprints' });
+    await frame.getByRole('heading', { name: 'No blueprints yet' });
+    await frame.getByTestId('create-blueprint-action-emptystate').click();
+    await frame.getByRole('heading', { name: 'Image output' });
+    await frame.getByTestId('checkbox-guest-image').click();
+    await frame.getByRole('button', { name: 'Next', exact: true }).click();
+
+    // fs
+    await frame.getByRole('heading', { name: 'Additional packages' });
```
regexowl, line 24: Should this be FS heading?

croissanne: oh wait, why does that work then O.o  if it's looking for packages twice, maybe the nav in the sidebar is playing tricks.

kingsleyzissou (approved):

croissanne:
hmm, seems like the latest additions I made to the edit test don't work, will fix those first.

regexowl:
Looks like some tests were marked as flaky in the Schutzbot output. Vitest has default number of retries set to 3, wanna set up the same thing for playwright?

Also wondering why is the Schutzbot run marked as success when there were flakes <_<

croissanne:
> Looks like some tests were marked as flaky in the Schutzbot output. Vitest has default number of retries set to 3, wanna set up the same thing for playwright?
> 
> Also wondering why is the Schutzbot run marked as success when there were flakes <_<

oh wow, i needs more than 90 seconds to generate the metadata and do the package search ;_; I'll just make the test EXTRA slow

regexowl:
> > Looks like some tests were marked as flaky in the Schutzbot output. Vitest has default number of retries set to 3, wanna set up the same thing for playwright?
> > Also wondering why is the Schutzbot run marked as success when there were flakes <_<
> 
> oh wow, i needs more than 90 seconds to generate the metadata and do the package search ;_; I'll just make the test EXTRA slow

:snail:

regexowl:
Extra extra slow might be needed?

image

croissanne:
hmm failed in a different place now, still looking at it

regexowl (dismissed):

kingsleyzissou (dismissed):

regexowl:
Looks like pr_check might fail on eph setup, `chrome-service` is not doing great. This shouldn't affect IQE in any way thought, should it?

regexowl (approved):
★_★

PR #2913, opened 2025-02-18
Wizard: Remove obsolete css

regexowl:
This removes css styles for `DualListSelector` we no longer use and a temporary pagination fix that was already addressed directly in PF.


mgold1234 (approved):
/lgtm

PR #2922, opened 2025-02-20
Wizard: refactor HookValidatedInput component

mgold1234:
This commit splits the HookValidatedInput component into three separate functions to improve modularity and readability:

- `getValidationState`: Calculates the validation state ('default', 'success', 'error') based on whether the input is pristine and if there is an error message.
- `ValidatedInputAndTextArea`: Renders the TextInput or TextArea component, utilizing the `getValidationState` output.
- `ErrorMessage`: Displays validation error messages.

This refactoring enhances code maintainability and testability, and the updated structure is now implemented for the username field.

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import {
   HelperText,
@@ -15,7 +15,7 @@ import { EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
 
 import type { StepValidation } from './utilities/useValidation';
 
-interface ValidatedTextInputPropTypes extends TextInputProps {
+type ValidatedTextInputPropTypes = TextInputProps & {
   dataTestId?: string | undefined;
   ouiaId?: string;
   ariaLabel: string | undefined;
@@ -23,7 +23,7 @@ interface ValidatedTextInputPropTypes extends TextInputProps {
   validator: (value: string | undefined) => boolean;
   value: string;
   placeholder?: string;
-}
+};
 
 type HookValidatedInputPropTypes = TextInputProps &
   TextAreaProps & {
@@ -37,6 +37,26 @@ type HookValidatedInputPropTypes = TextInputProps &
     warning?: string;
     inputType?: 'textInput' | 'textArea';
   };
+
+type ValidationInputProp = TextInputProps &
+  TextAreaProps & {
+    value: string;
+    placeholder: string;
+    stepValidation: StepValidation;
+    fieldName: string;
+    inputType?: 'textInput' | 'textArea';
+    ariaLabel: string;
+    onChange: (
+      event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+      value: string
+    ) => void;
+  };
+
+type ErrorMessageProps = {
+  errorMessage: string;
+};
+
+type ValidationResult = 'default' | 'success' | 'error';
 
 export const HookPasswordValidatedInput = ({
   ariaLabel,
@@ -86,6 +106,78 @@ export const HookPasswordValidatedInput = ({
         </InputGroupItem>
       </InputGroup>
     </>
+  );
+};
+
+export const ValidatedInputAndTextArea = ({
+  value,
+  stepValidation,
+  fieldName,
+  placeholder,
+  onChange,
+  ariaLabel,
+  inputType = 'textInput',
+}: ValidationInputProp) => {
+  const errorMessage = stepValidation.errors[fieldName];
+  const hasError = errorMessage !== '';
+
+  const [isPristine, setIsPristine] = useState(!value);
+  const validated = getValidationState(isPristine, errorMessage);
+
+  const handleBlur = () => {
+    if (value) {
+      setIsPristine(false);
+    }
+  };
+
+  useEffect(() => {
+    if (!value) {
+      setIsPristine(true);
+    }
+  }, [value, setIsPristine]);
+
+  return (
+    <>
+      {inputType === 'textArea' ? (
+        <TextArea
+          value={value}
+          onChange={onChange}
+          validated={validated}
+          onBlur={handleBlur}
+          placeholder={placeholder}
+          aria-label={ariaLabel}
+        />
+      ) : (
+        <TextInput
+          value={value}
+          onChange={onChange}
+          validated={validated}
+          onBlur={handleBlur}
+          placeholder={placeholder}
+          aria-label={ariaLabel}
+        />
+      )}
+      {hasError && <ErrorMessage errorMessage={errorMessage} />}
+    </>
+  );
+};
+
+const getValidationState = (
+  isPristine: boolean,
+  errorMessage: string
+): ValidationResult => {
+  const validated = isPristine ? 'default' : errorMessage ? 'error' : 'success';
+
+  return validated;
+};
+
+export const ErrorMessage = ({ errorMessage }: ErrorMessageProps) => {
+  return (
+    <HelperText>
+      <HelperTextItem variant="error" hasIcon>
+        {errorMessage}
+      </HelperTextItem>
+    </HelperText>
   );
 };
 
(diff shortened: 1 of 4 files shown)


src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,120 @@ export const HookPasswordValidatedInput = ({
   );
 };
 
+const useValidationState = ({
```
lucasgarfield, line 117: The `use` prefix is reserved for hooks, which have some special characteristics. You might check the docs here to refresh your memory:
https://react.dev/learn/reusing-logic-with-custom-hooks

This is just a helper function, so you could name it something like `getValidationState`.

Honestly, I think you could probably inline this function.

If you do choose to keep it as a separate function, it might be worth making it have a bit looser coupling by just passing in an `errorMessage` prop (it would be a string) instead of `stepValidation` and `fieldName`. Right now the component has to "know" about the `stepValidation` object and its `fieldName` field which makes it tightly coupled to that particular paradigm. We might want to get rid of the `stepValidation` object in the future, after all.

lucasgarfield (changes requested):

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,120 @@ export const HookPasswordValidatedInput = ({
...
+    </>
+  );
+};
+
+export const ErrorMessage = ({
+  stepValidation,
+  fieldName,
+}: ErrorMessageProps) => {
+  const errorMessage = stepValidation.errors[fieldName];
+  if (!errorMessage) {
```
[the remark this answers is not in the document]
mgold1234: like it!

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,120 @@ export const HookPasswordValidatedInput = ({
...
+        />
+      )}
+    </>
+  );
+};
+
+export const ErrorMessage = ({
+  stepValidation,
+  fieldName,
+}: ErrorMessageProps) => {
```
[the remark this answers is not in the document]
mgold1234: Screenshot 2025-02-25 at 15 06 21
without it we have the error symbol all the time

src/Components/CreateImageWizard/ValidatedInput.tsx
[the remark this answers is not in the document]
mgold1234: without it we have the error symbol all the time
Screenshot 2025-02-25 at 15 06 21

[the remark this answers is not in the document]
mgold1234: becase I define 'hasError' wrong,
this do the trick 
` const hasError = errorMessage !== '';`

src/Components/CreateImageWizard/validators.ts
```diff
@@ -58,7 +58,9 @@ export const isFileSystemConfigValid = (partitions: Partition[]) => {
 };
 
 export const isUserNameValid = (userName: string) => {
-  if (userName === undefined) return false;
+  if (!userName) {
```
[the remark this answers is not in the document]
lucasgarfield: And, of course, `isUserNameValid` should evaluate to `false` for `''` because it is not a valid user name.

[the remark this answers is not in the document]
lucasgarfield: Finally, won't we need to apply whatever pattern we use (either my suggestion or this) to the other validation functions? For instance, `isSshKeyValid()` will also need to handle the `''` case, etc...

lucasgarfield (changes requested):

src/Components/CreateImageWizard/validators.ts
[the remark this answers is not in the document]
mgold1234: I changed the sshKey condition here as well,  maybe I should do it in a different pr?
`userSshKey:
        userSshKey && !isSshKeyValid(userSshKey) ? 'Invalid SSH key' : '',`

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,102 @@ export const HookPasswordValidatedInput = ({
...
+
+  return (
+    <>
+      <TextAndTextAreaInput
+        inputType={inputType}
+        ariaLabel={ariaLabel || ''}
+        value={value}
+        placeholder={placeholder || ''}
+        errorMessage={errorMessage}
+        onChange={onChange!}
```
lucasgarfield, line 138: Why do we need a non-null assertion operator here?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,102 @@ export const HookPasswordValidatedInput = ({
...
+        onBlur={onBlur!}
```
lucasgarfield, line 139: Why do we need a non-null assertion operator here?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,102 @@ export const HookPasswordValidatedInput = ({
...
+  inputType = 'textInput',
+}: ValidationInputProp) => {
+  const errorMessage = stepValidation.errors[fieldName];
+  const hasError = errorMessage !== '';
+
+  return (
+    <>
+      <TextAndTextAreaInput
+        inputType={inputType}
+        ariaLabel={ariaLabel || ''}
```
lucasgarfield, line 134: Why `|| ''`?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,102 @@ export const HookPasswordValidatedInput = ({
...
+        value={value}
+        placeholder={placeholder || ''}
```
lucasgarfield, line 136: Why || ''?

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -89,6 +114,102 @@ export const HookPasswordValidatedInput = ({
...
+        errorMessage={errorMessage}
+        onChange={onChange!}
+        onBlur={onBlur!}
+      />
+      {hasError && <ErrorMessage errorMessage={errorMessage} />}
+    </>
+  );
+};
+
+export const TextAndTextAreaInput = ({
```
lucasgarfield, line 146: I'm looking at this component and wondering if we really need it. It's just a wrapper around the text input and the text area components with a blur handler and state management of `isPristine`. Could we just move `isPristine` up into ValidatedInputAndTextArea along with the blur handler and the useEffect and eliminate this component?

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: because it detects that the properties onChange and onBlur might be undefined at runtime, even though the code seems to assume they are always defined. The reason is that they are defined as Optional props.
and we know that it is not optional in our case, so I will define it as mandatory 
```
   onChange: (
      event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
      value: string
    ) => void;
```

thank you for that

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: same

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: same

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: same

src/Components/CreateImageWizard/ValidatedInput.tsx
mgold1234: yes, thanks

lucasgarfield (approved):
This came out a very nice refactor! Thank you, @mgold1234.

lucasgarfield:
pr check failing: 21:45:16 ERROR: deploy failed: Could not do 'wait for all deployed resources to be ready' at /var/lib/jenkins/jenkins-osbuild-image-builder-frontend-pr-check-6219-bcgp7G/.bonfire_venv/lib64/python3.8/site-packages/bonfire/openshift.py:85 in time

admin merging

PR #2931, opened 2025-02-24
Blueprints: invalid message variant on upload (HMS-5555)

### LINKED ISSUE #2919: Blueprints: `isInvalidFormat` error should have `error` variant
Image

The "Not compatible with the blueprints format." text should be formatted as error, same as this:

Image

ksiekl:
Fixed #2919

Error message has error variant when uploading invalid blueprint.

JIRA: [HMS-5555](https://issues.redhat.com/browse/HMS-5555)

### DIFF
--- a/src/Components/Blueprints/ImportBlueprintModal.tsx
+++ b/src/Components/Blueprints/ImportBlueprintModal.tsx
@@ -273,7 +273,7 @@ export const ImportBlueprintModal: React.FunctionComponent<
             <HelperText>
               <HelperTextItem
                 variant={
-                  isRejected ? 'error' : isOnPrem ? 'warning' : 'default'
+                  isRejected ? 'error' : isInvalidFormat ? 'error' : isOnPrem ? 'warning' : 'default'
                 }
               >
                 {isRejected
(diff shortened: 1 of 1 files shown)


src/Components/Blueprints/ImportBlueprintModal.tsx
```diff
@@ -273,7 +273,7 @@ export const ImportBlueprintModal: React.FunctionComponent<
             <HelperText>
               <HelperTextItem
                 variant={
-                  isRejected ? 'error' : isOnPrem ? 'warning' : 'default'
+                  isRejected ? 'error' : isInvalidFormat ? 'error' : isOnPrem ? 'warning' : 'default'
```
regexowl, line 276: I'll just add that if you decide to go with the ternary operator, we could remove one level of nesting by chaining the `isRejected` and `isInvalidFormat` since they both lead to `'error'` variant.

```suggestion
                  isRejected || isInvalidFormat ? 'error' : isOnPrem ? 'warning' : 'default'
```

ksiekl:
Thanks! I changed it to the switch statement although I never got quite used to seeing "switch (true)" but I definitely thought the nested operator was also funky looking. Thanks for the autofix recommendation, I got it set up :)

regexowl (approved):
Works great, thanks! :sparkles: 

In case you'd like to rebase here in UI (reviews usually persist through those), we use the "Update with rebase" option:

image

I'll rebase and then we can merge this.

PR #2937, opened 2025-02-25
Wizard: Add masked services input (HMS-5561)

regexowl:
This exposes masked services in the UI by adding a masked services input.

image

JIRA: [HMS-5561](https://issues.redhat.com/browse/HMS-5561)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
@@ -458,7 +458,9 @@ const Review = () => {
           </ExpandableSection>
         )}
       {isServicesStepEnabled &&
-        (services.enabled.length > 0 || services.disabled.length > 0) && (
+        (services.enabled.length > 0 ||
+          services.disabled.length > 0 ||
+          services.masked.length > 0) && (
           <ExpandableSection
             toggleContent={composeExpandable(
               'Systemd services',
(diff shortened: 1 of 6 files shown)


regexowl:
@lucasgarfield I believe we discussed that newly added disabled services should get added to the masked. Is that correct? Currently the disabled services are just in disabled so this should fix the issue.

lucasgarfield:
@regexowl, I think what we may actually want to do is add a third input to allow users to specifically choose between disabling or masking a service... there is a fairly big difference between the two and the on-prem blueprints expose both options: https://osbuild.org/docs/user-guide/blueprint-reference/#systemd-services

Then we would continue to mask the OpenSCAP services (and they would appear in the masked services input). 

Does that sound reasonable? :thinking:

regexowl:
> @regexowl, I think what we may actually want to do is add a third input to allow users to specifically choose between disabling or masking a service... there is a fairly big difference between the two and the on-prem blueprints expose both options: https://osbuild.org/docs/user-guide/blueprint-reference/#systemd-services
> 
> Then we would continue to mask the OpenSCAP services (and they would appear in the masked services input).
> 
> Does that sound reasonable? 🤔

I see. I was admittedly confused by masked services not being exposed in on-prem UI
image
 
Let me rework this :eyes: Thank you!

lucasgarfield:
Ah, right! Cockpit composer strikes again! :grin: 

For what it's worth, you can check `man systemctl` to find the difference between the two. I think the biggest difference is probably this:
`Using enable on masked units is not supported and results in an error.`

I can definitely imagine that users might want to be able to choose between those two options!

ondrejbudai:
I randomly stumbled upon this PR. Yes, it makes sense to support both. Somewhat simplified explanation of the difference:

- disabled units don't start on boot. They can still be started manually, or via a dependency.
- masked units don't start on boot, nor they can be started manually, nor via a dependency.

regexowl:
Alright, this should do the trick. Thanks!

lucasgarfield (approved):
Great work! Thank you.

PR #2941, opened 2025-02-26
Wizard: Remove "Too many results" state

regexowl:
This removes the "Too many results" state that was rendered when a search term returned more than 100 results.


regexowl:
I believe this should be currently blocked by https://issues.redhat.com/browse/HMS-5551 as the current pattern allows for the exact match to be buried within the results (even though the increased limit should cover this in most cases)

regexowl:
There are two changes that had to be addressed in the mocked fixtures:
1. sorting of the packages is now handled directly in the API so there's no need to re-sort things on our end
2. search was changed to prefix search only, so instead of `*foo*` results for "foo" search we're getting only `foo*` now.

mgold1234 (approved):
/lgtm

PR #2943, opened 2025-02-26
Wizard: fix indentation of manual file system config (HMS-5564)

### LINKED ISSUE #2910: Wizard/Review: Fix FSC expandable indentation
Indentation of the list descriptions is bigger than in other lists:

Image

ksiekl:
Fixes #2910

Indentation of manual file system configuration in the review step of the wizard was aligned.

JIRA: [HMS-5564](https://issues.redhat.com/browse/HMS-5564)

### DIFF
--- a/src/Components/CreateImageWizard/steps/FileSystem/FileSystemTable.tsx
+++ b/src/Components/CreateImageWizard/steps/FileSystem/FileSystemTable.tsx
@@ -54,7 +54,7 @@ export const MinimumSizePopover = () => {
         variant="plain"
         aria-label="File system configuration info"
         aria-describedby="file-system-configuration-info"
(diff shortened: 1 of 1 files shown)


regexowl (approved):
This did the trick! I wonder if we could just set all the padding to 0 so we have just one className modifier. There seems to be to a space between the "Minimum size" text and the popover button already which might be sufficient.

But otherwise this looks perfectly fine so approved and let me know if you'd like to merge this as is or want to look into consolidating the className into one :)

ksiekl:
yea sure I'll checkout the className first :)

regexowl (approved):
LGTM! Thank you for cleaning up the classNames :sparkles::broom:

PR #2953, opened 2025-03-03
Wizard: Fix kernel validation and add import validation test (HMS-5582)

regexowl:
This fixes problems with kernel step validation and adds a new test that checks validation of steps on import.

JIRA: [HMS-5582](https://issues.redhat.com/browse/HMS-5582)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Kernel/components/KernelName.tsx
+++ b/src/Components/CreateImageWizard/steps/Kernel/components/KernelName.tsx
@@ -23,6 +23,7 @@ import {
   selectKernel,
 } from '../../../../../store/wizardSlice';
 import { useKernelValidation } from '../../../utilities/useValidation';
+import { isKernelNameValid } from '../../../validators';
 
 const initialOptions = ['kernel', 'kernel-debug'];
 let kernelOptions = initialOptions;
@@ -48,7 +49,9 @@ const KernelName = () => {
       if (!filteredKernelPkgs.some((kernel) => kernel === filterValue)) {
         filteredKernelPkgs = [
           ...filteredKernelPkgs,
-          `Custom kernel package "${filterValue}"`,
+          isKernelNameValid(filterValue)
+            ? `Custom kernel package "${filterValue}"`
+            : `"${filterValue}" is not a valid kernel package name`,
         ];
       }
       if (!isOpen) {
@@ -165,7 +168,11 @@ const KernelName = () => {
         >
           <SelectList>
             {selectOptions.map((option) => (
-              <SelectOption key={option} value={option}>
+              <SelectOption
+                key={option}
+                value={option}
+                isDisabled={/not a valid kernel package name/i.test(option)}
+              >
                 {option}
               </SelectOption>
             ))}
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -167,33 +167,24 @@ export function useSnapshotValidation(): StepValidation {
 export function useTimezoneValidation(): StepValidation {
   const timezone = useAppSelector(selectTimezone);
   const ntpServers = useAppSelector(selectNtpServers);
-  const errors = {};
-
-  if (timezone) {
-    if (!timezones.includes(timezone)) {
-      Object.assign(errors, { timezone: 'Unknown timezone' });
-    }
-  }
+  const invalidServers = [];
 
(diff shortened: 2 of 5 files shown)


src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -251,14 +251,10 @@ export function useHostnameValidation(): StepValidation {
...
+  const errors = {};
 
   if (!isKernelNameValid(kernel.name)) {
-    return {
-      errors: {
-        kernel: 'Invalid format.',
-      },
-      disabledNext: true,
-    };
+    Object.assign(errors, { kernel: 'Invalid format.' });
```
lucasgarfield, line 257: I'm not opposed to this but it's a somewhat new pattern... what's the advantage of using `Object.assign()`?

regexowl: I've started using `Object.assign()` on steps that have multiple validated fields. The main reason is that we need to return an object with more than one possible error values.

In this case we were previously returning either
```typescript
errors: { kernel: ... }
```

OR 
``` typescript
errors: { kernelAppend: ... }
```

but there was no pattern for cases when both imported values are invalid at the same time AFAIK. This way we can just keep adding possible problems into the `errors` object and then return the entire thing at once.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -262,14 +262,10 @@ export function useHostnameValidation(): StepValidation {
 
 export function useKernelValidation(): StepValidation {
   const kernel = useAppSelector(selectKernel);
+  const errors = {};
 
   if (!isKernelNameValid(kernel.name)) {
```
mgold1234, line 267: personally, I prefer this approach 
```
  const kerneler = !isKernelNameValid(kernel.name) ? 'Invalid format.' : '';

  return {
    errors: {
      kernel: kerneler,
    },
    disabledNext: !!kerneler,
  };
```
it is more readable for me :) WDYT?
I checked that and it is working the same

regexowl: Sure, I've updated the code. :+1:  Will open a follow up for the other validators that use `Object.assign()`

regexowl: I lied, I've added the changes to this PR after all :zipper_mouth_face:

mgold1234:
@regexowl your changes looks good, I am not sure if that this is a bug or correct beavior, 
if I am trying to add invalid name, its adding it to the list 
Screenshot 2025-03-13 at 16 05 20

mgold1234:
one small thing, should we check if user try to add only numbers?

regexowl:
> @regexowl your changes looks good, I am not sure if that this is a bug or correct beavior, if I am trying to add invalid name, its adding it to the list Screenshot 2025-03-13 at 16 05 20

Yeah, I don't think we should allow to add invalid value even if we validate it after.

regexowl:
This should do the trick :magic_wand:

src/Components/CreateImageWizard/steps/Kernel/components/KernelName.tsx
```diff
@@ -165,7 +168,13 @@ const KernelName = () => {
         >
           <SelectList>
             {selectOptions.map((option) => (
-              <SelectOption key={option} value={option}>
+              <SelectOption
+                key={option}
+                value={option}
+                isDisabled={
+                  /not a valid kernel package name/i.test(option) ? true : false
```
avitova, line 175: I think this is redundant - `.test` already returns a value.
```suggestion
                  /not a valid kernel package name/i.test(option)
```

avitova (commented):
Looks great!:)

src/Components/CreateImageWizard/steps/Kernel/components/KernelName.tsx
regexowl: Nice catch! Updated

avitova (approved):
Btw what kind of problem does this fix? I am overall good here, I tested this, works fine. :) I will leave up to you if you still want to wait for others to approve.

regexowl:
> Btw what kind of problem does this fix? I am overall good here, I tested this, works fine. :) I will leave up to you if you still want to wait for others to approve.

The kernel validation previously returned either `kernelName` error or `kernelAppend` error, but not both. So one field (append) would always seem to be valid.

PR #2954, opened 2025-03-03
Wizard: Deprecate `Chip` and `ChipGroup` components

regexowl:
`Chip` and `ChipGroup` components will be deprecated in PF6. This replaces them with recommended `Label` and `LabelGroup`.

### DIFF
--- a/src/Components/CreateImageWizard/ChippingInput.tsx
+++ b/src/Components/CreateImageWizard/ChippingInput.tsx
@@ -2,10 +2,10 @@ import React, { useState } from 'react';
 
 import {
   Button,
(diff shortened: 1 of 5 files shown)


lucasgarfield (approved):
Looks good to me! Would we want to rename ChippingInput to LabelInput?

regexowl:
> Looks good to me! Would we want to rename ChippingInput to LabelInput?

Yes, let's. I'll open a follow up with the name change.

regexowl:
Follow up here: https://github.com/osbuild/image-builder-frontend/pull/2958

PR #2963, opened 2025-03-05
Wizard: add satellite registration

avitova:
This adds an option to paste the Satellite registration command + CA. Saving this information into the blueprint is still not done; I will create a follow-up; this is big PR as is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "@sentry/webpack-plugin": "3.2.2",
     "@unleash/proxy-client-react": "5.0.0",
     "classnames": "2.5.1",
+    "jwt-decode": "4.0.0",
     "lodash": "4.17.21",
     "react": "18.3.1",
     "react-dom": "18.3.1",
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -104,7 +104,9 @@ export const ValidatedInputAndTextArea = ({
           </HelperTextItem>
         </HelperText>
       )}
-      {hasError && <ErrorMessage errorMessage={errorMessage} />}
+      {validated === 'error' && hasError && (
+        <ErrorMessage errorMessage={errorMessage} />
+      )}
     </>
   );
 };
--- a/src/Components/CreateImageWizard/steps/Registration/components/SatelliteRegistrationCommand.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/components/SatelliteRegistrationCommand.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+
+import {
+  FormGroup,
+  FormHelperText,
+  HelperText,
+  HelperTextItem,
+} from '@patternfly/react-core';
+
+import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
+import {
+  changeSatelliteRegistrationCommand,
+  selectSatelliteRegistrationCommand,
+} from '../../../../../store/wizardSlice';
+import { useRegistrationValidation } from '../../../utilities/useValidation';
+import { ValidatedInputAndTextArea } from '../../../ValidatedInput';
+
+const SatelliteRegistrationCommand = () => {
+  const dispatch = useAppDispatch();
+  const registrationCommand = useAppSelector(
+    selectSatelliteRegistrationCommand
+  );
+  const stepValidation = useRegistrationValidation();
+  const registrationDocs =
+    'https://docs.redhat.com/en/documentation/red_hat_satellite/6.16/html-single/managing_hosts/index#Customizing_the_Registration_Templates_managing-hosts';
+
+  const handleChange = (e: React.FormEvent, value: string) => {
+    dispatch(changeSatelliteRegistrationCommand(value));
+  };
+
+  return (
+    <FormGroup label="Registration command from Satellite" isRequired>
+      <ValidatedInputAndTextArea
+        inputType={'textArea'}
+        ariaLabel="registration command"
+        value={registrationCommand || ''}
+        onChange={handleChange}
+        placeholder="Registration command"
+        stepValidation={stepValidation}
+        fieldName="command"
+      />
+      <FormHelperText>
+        <HelperText>
+          <HelperTextItem>
+            To generate command from Satellite, follow the{' '}
+            <a href={registrationDocs} target="_blank" rel="noreferrer">
+              documentation
+            </a>
+            .
+          </HelperTextItem>
+        </HelperText>
+      </FormHelperText>
+    </FormGroup>
+  );
+};
+
+export default SatelliteRegistrationCommand;
--- a/src/Components/CreateImageWizard/utilities/certificates.ts
+++ b/src/Components/CreateImageWizard/utilities/certificates.ts
@@ -1,0 +1,5 @@
+export const isValidPEM = (cert: string): boolean => {
+  return /-----BEGIN CERTIFICATE-----[\s\S]+-----END CERTIFICATE-----/.test(
+    cert.trim()
+  );
+};
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -1,6 +1,9 @@
 import React, { useEffect, useState } from 'react';
 
 import { CheckCircleIcon } from '@patternfly/react-icons';
+import { jwtDecode } from 'jwt-decode';
+
+import { isValidPEM } from './certificates';
 
 import { UNIQUE_VALIDATION_DELAY } from '../../../constants';
 import { useLazyGetBlueprintsQuery } from '../../../store/backendApi';
@@ -34,6 +37,8 @@ import {
   selectKeyboard,
   selectTimezone,
   selectImageTypes,
+  selectSatelliteCaCertificate,
+  selectSatelliteRegistrationCommand,
 } from '../../../store/wizardSlice';
 import { keyboardsList } from '../steps/Locale/keyboardsList';
 import { languagesList } from '../steps/Locale/languagesList';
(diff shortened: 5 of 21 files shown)


lucasgarfield:
Love the validation guardrails!

package.json
```diff
@@ -19,6 +19,7 @@
     "@sentry/webpack-plugin": "3.2.1",
     "@unleash/proxy-client-react": "4.5.2",
     "classnames": "2.5.1",
+    "jwt-decode": "^4.0.0",
```
lucasgarfield, line 22: New dependency! Can you please remove the `^`? We rely on Dependabot to handle everything.

Also, I'd be curious to know about the decision to use jwt-decode. What was your selection criteria?

lucasgarfield: I'm re-reading my comment and I realized it is a bit curt. :sweat_smile: 

`jwt-decode` is probably fine but it would be nice to know why we chose it, especially in the commit message for the future - is it the most popular or well maintained library for this? Did we look into any alternatives, etc...? I don't have any specific problem with `jwt-decode` itself I'm just totally ignorant so some extra context would be nice. :smile:

regexowl:
+1 on pining the dep version, otherwise looks great! :heart: One question - could `isValidPEM` go into `validators.ts` or are there more plans for `certificates.ts`?

Also can't see the flag in stage unleash, could you please add it there as well? We might even want to gate it behind a preview constraint as this will probably affect IQE :thinking:

src/Components/CreateImageWizard/steps/Registration/components/SatelliteRegistrationCommand.tsx
```diff
@@ -0,0 +1,52 @@
...
+        ariaLabel="registration command"
+        value={registrationCommand || ''}
+        onChange={handleChange}
+        placeholder="Registration command"
+        stepValidation={stepValidation}
+        fieldName="command"
+      />
+      <FormHelperText>
+        <HelperText>
+          {/* TODO: Add link to our docs */}
```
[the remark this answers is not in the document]
avitova: Nope, I assume we would need the feature first before documentation, as they would probably need to create some screenshots. This way we can just not expose the feature because of the flag, create docs, add the link to docs, and only then potentially expose the feature.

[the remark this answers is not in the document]
regexowl: Fair enough!

src/Components/CreateImageWizard/utilities/certificates.ts
```diff
@@ -0,0 +1,5 @@
+export const isValidPEM = (cert: string): boolean => {
+  return /-----BEGIN CERTIFICATE-----[\s\S]+-----END CERTIFICATE-----/.test(
```
ezr-ondrej, line 2: Nit: isn't `[\s\S]+`  equal to `.+`?

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -122,6 +132,51 @@ export function useRegistrationValidation(): StepValidation {
...
+      Object.assign(errors, {
+        command: 'No registration command for Satellite registration',
+      });
+    }
+    try {
+      const match = registrationCommand?.match(
+        /Bearer\s+([\w-]+\.[\w-]+\.[\w-]+)/
+      );
+      if (!match) {
+        Object.assign(errors, { command: 'Invalid token' });
```
ezr-ondrej, line 153: This can also be cause by the token not being there right?
```suggestion
        Object.assign(errors, { command: 'Invalid or missing token' });
```

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -122,6 +132,51 @@ export function useRegistrationValidation(): StepValidation {
...
+            return {
+              errors: errors,
+              disabledNext:
+                caCertificate === undefined || !isValidPEM(caCertificate),
+            };
+          }
+        }
+      }
+    } catch {
+      Object.assign(errors, { command: 'Invalid token' });
```
ezr-ondrej, line 175: ditto
```suggestion
      Object.assign(errors, { command: 'Invalid or missing token' });
```

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -122,6 +132,51 @@ export function useRegistrationValidation(): StepValidation {
...
+        /Bearer\s+([\w-]+\.[\w-]+\.[\w-]+)/
+      );
+      if (!match) {
+        Object.assign(errors, { command: 'Invalid token' });
+      } else {
+        const token = match[1];
+        const decoded = jwtDecode(token);
+        if (decoded.exp) {
+          const currentTime = Date.now() / 1000;
+          if (decoded.exp < currentTime) {
```
ezr-ondrej, line 159: Improvement idea, **not a blocker**: this could potentially have a buffer.
I mean the image build takes about 15 minutes, so token expiring in half an hour doesn't give you much room to use the image :stuck_out_tongue:
We might want to figure out a warning intead in a follow-up :)

ezr-ondrej (commented):
Great work! :orange_heart:

package.json
avitova: Alright, that makes sense. I removed the `^`, and added information into the commit, hopefully that's what you meant. :)

ezr-ondrej:
Ugh there is a conflict already, that was fast :thinking:

avitova:
> +1 on pining the dep version, otherwise looks great! ❤️ One question - could isValidPEM go into validators.ts or are there more plans for certificates.ts?
Uhm, I'd like to add also isDer/isCer, but I need to look into that one. On top of that, the plan might be that users could paste more certificates into one textfield, and we'd parse them all. But I think we should do this as a followup. But yeah, it will likely be a bit more complicated.

> Also can't see the flag in stage unleash, could you please add it there as well? We might even want to gate it behind a preview constraint as this will probably affect IQE 🤔
Done, that's a good point, thanks!

src/Components/CreateImageWizard/utilities/certificates.ts
avitova: Hmmm, I tried to figure this out as well, we'd need to add "s" at the end of the regex, something like this:
`/-----BEGIN CERTIFICATE-----.+-----END CERTIFICATE-----/s`

Otherwise it would not accept certificates with more lines. I can do that if it makes the regex a bit more clear.

src/Components/CreateImageWizard/utilities/useValidation.tsx
avitova: I like this idea, and I can do that as a followup. I still need to add the Satellite registration to the review step, add the command + certificate to the blueprint, so that it is saved. This could fit in too. :pencil2:  :notebook:

regexowl:
> > +1 on pining the dep version, otherwise looks great! ❤️ One question - could isValidPEM go into validators.ts or are there more plans for certificates.ts?
> 
> Uhm, I'd like to add also isDer/isCer, but I need to look into that one. On top of that, the plan might be that users could paste more certificates into one textfield, and we'd parse them all. But I think we should do this as a followup. But yeah, it will likely be a bit more complicated.

Got it! I don't have anything against splitting validators into separate files. Just curious :+1:  Thinking about it we should probably do a bit of a cleanup around utilities and unsorted files in the Wizard folder, but that's completely outside of this PR's scope :sweat_smile:

src/Components/CreateImageWizard/steps/Registration/components/SatelliteRegistrationCommand.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <HelperTextItem>
+            To generate command from Satellite, follow the documentation.
```
ezr-ondrej, line 44: Should we link the current Satellite registration for now? :)
https://docs.redhat.com/en/documentation/red_hat_satellite/6.16/html-single/managing_hosts/index#Customizing_the_Registration_Templates_managing-hosts

regexowl: We could, both no link and using a temporary relevant link work for me. We just need to remember to open a follow up when the new documentation is ready. :+1:

ezr-ondrej (dismissed):
All right then, I think we can merge, the only change with the feature flag turned off should be change to radios, which should be fine in the interim :)

regexowl:
Switch from a single checkbox to two radios might impact IQE, @tkoscieln could you please check if the tests will need an adjustment?

avitova:
Oopsie, did not see the approval. I just put the docs link and fixed a bug, hope y'all do not mind. :sparkles:

ezr-ondrej (approved):
I'm still fine, we need to fix the tests tho.
@tkoscieln would you do that and then merge this PR? or what's the process now?

tkoscieln:
I'll deploy the PR locally and see if it breaks something. If not, we can merge, if yes, I will do the necessary adjustments and then we can merge

ezr-ondrej:
the failures in the CI seems to be related :)
```
>               raise WidgetOperationFailed("Failed to set the checkbox to requested value.")
E               widgetastic.exceptions.WidgetOperationFailed: Failed to set the checkbox to requested value.
```

tkoscieln:
Well, I'll do the adjustment then :smiley:

regexowl:
@avitova Heya, could you please resolve the conflict and rebase? We're unblocked on the IQE issue :sparkles:

avitova:
Hopefully this is it, there were some more changes after I managed to resolve conflicts, hopefully we have everything now.

regexowl:
I see, Playwright tests will need an update. In `playwright/test.spec.ts` the line 
```typescript
await page.getByTestId('automatically-register-checkbox').uncheck();
```
might need to be swapped for something like this I think?
```typescript
await page.getByTestId('register-later-radio').click();
```

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -104,7 +104,9 @@ export const ValidatedInputAndTextArea = ({
           </HelperTextItem>
         </HelperText>
       )}
-      {hasError && <ErrorMessage errorMessage={errorMessage} />}
+      {validated === 'error' && hasError && (
+        <ErrorMessage errorMessage={errorMessage} />
+      )}
```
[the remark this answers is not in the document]
regexowl: Thank you! Agree that the previous logic made more sense :+1:

regexowl (approved):
Thank you, this is great :heart:

PR #2964, opened 2025-03-05
Wizard: rename repo snapshot step to repeatable builds HMS-5599

jlsherrill:
This renames the repository snapshot wizard step to "Reproducible builds" to emphasize the benefit to the user.
It also makes the snapshot date be auto set to the current date, and changes the 'reset' button  to do the same.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Snapshot/Snapshot.tsx
+++ b/src/Components/CreateImageWizard/steps/Snapshot/Snapshot.tsx
@@ -18,11 +18,13 @@ import {
   changeUseLatest,
   changeSnapshotDate,
 } from '../../../../store/wizardSlice';
+import { yyyyMMddFormat } from '../../../../Utilities/time';
 import { isSnapshotDateValid } from '../../validators';
 
 export default function Snapshot() {
   const dispatch = useAppDispatch();
   const snapshotDate = useAppSelector(selectSnapshotDate);
+
   const useLatest = useAppSelector(selectUseLatest);
   return (
     <>
@@ -31,8 +33,8 @@ export default function Snapshot() {
           id="use latest snapshot radio"
           ouiaId="use-latest-snapshot-radio"
           name="use-latest-snapshot"
-          label="Use latest content"
-          description="Use the newest repository state available when building this image."
+          label="Disable repeatable build."
+          description="Use the newest repository content available when building this image."
           isChecked={useLatest}
           onChange={() => !useLatest && dispatch(changeUseLatest(true))}
         />
@@ -40,8 +42,8 @@ export default function Snapshot() {
           id="use snapshot date radio"
           ouiaId="use-snapshot-date-radio"
           name="use-snapshot-date"
-          label="Use a snapshot"
-          description="Target a date and build images with repository information from this date."
+          label="Enable repeatable build"
+          description="Build this image with the repository content of a selected date."
           isChecked={!useLatest}
           onChange={() => useLatest && dispatch(changeUseLatest(false))}
         />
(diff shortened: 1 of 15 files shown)


jlsherrill:
image

src/store/wizardSlice.ts
```diff
@@ -245,6 +245,14 @@ export const initialState: wizardState = {
   users: [],
 };
 
+const currentDateString = () => {
```
lucasgarfield, line 248: We've got a bunch of time related helper functions in `/src/Utilities/time.ts`... you might find what you need there or if not can put this function there.

lucasgarfield:
We typically use a rebase Git workflow instead of a merge workflow, would you mind getting rid of the merge commit and just rebasing on main?

src/store/wizardSlice.ts
jlsherrill: perfect!

lucasgarfield:
We should probably add tests for this... and looking into that actually made me realize that we don't test the requests we generate for the Snapshot step at all in "Create" mode. We do test the request in edit mode for a blueprint that contains a snapshot.

jlsherrill:
@lucasgarfield i'm happy to give it a try, but is this not doing that? https://github.com/osbuild/image-builder-frontend/blob/main/src/test/Components/CreateImageWizard/steps/Snapshot/Snapshot.test.tsx#L141-L176

or is there another type of test you're expecting.

regexowl:
Those should be the correct tests. Going through the file it seems that `Snapshot.test.tsx` doesn't follow the usual structure we use in other test files - I'm happy to clean that up in a follow up.

Since the title of the step will be changing this might have effect on IQE tests. Let's see what pr_check thinks, should be just a small adjustment though.

Otherwise this looks really good I think, thank you for working on this! :sparkles:

regexowl:
Hmm, tests seem to be unhappy - `selectUseSnapshot` helper function will need to be updated with the new radio name. It also looks like there's a difference between expected and received snapshot dates in the test runs. Have to run for today, but can check it out tomorrow.

image
(from the `Snapshot.test.tsx` run after radio name update)

src/Components/CreateImageWizard/steps/Snapshot/Snapshot.tsx
```diff
@@ -85,9 +87,17 @@ export default function Snapshot() {
...
               <Button
                 variant="link"
-                onClick={() => dispatch(changeSnapshotDate(''))}
+                onClick={async () => {
+                  //Patternfly DatePicker seems to only clear error text if value is reset to '',
+                  // if you have an invalid date (2000-01-010000) and try to reset it, it must be set to '' first
+                  dispatch(changeSnapshotDate(''));
+                  setTimeout(() => {
+                    dispatch(changeSnapshotDate(yyyyMMddFormat(new Date())));
+                  }, 1);
```
jlsherrill, line 99: I wanted to highlight this.  i would thought that just 
```
                    dispatch(changeSnapshotDate(yyyyMMddFormat(new Date())));
```
would work, but if you type in something like  '2024-01-010000'  and then click reset, it won't actually reset to the current date.  By resetting to '' and then resetting to the date, it works fine.

jlsherrill:
UXD suggested 'build' instead of 'builds', so i have updated to that.

regexowl:
Alright, looks like IQE will indeed need an adjustment to the new naming. @tkoscieln could you please take a look at this when you have some time? The step title for repo snapshot was changed and so were the radio button labels and descriptions.

regexowl:
Hey @jlsherrill! Could you please resolve the conflict and rebase? We're unblocked on the IQE issue

jlsherrill:
after some discussion, due to the potential expectation with reproducible build, we've decided to go with repeatable build

regexowl (approved):
Thank you :heart:

Issue #2924, opened 2025-02-20
Wizard: When unselecting a package, it shouldn't immediately disappear


regexowl:
When removing a previously added package from the "Selected" it shouldn't immediately disappear. Rathe the checkbox should get unchecked, but the package should remain in the list until another packages table part is toggled or the step changes.

This will allow the user to select the package again in a case it was removed by accident.

regexowl:
Fixed by https://github.com/osbuild/image-builder-frontend/pull/2965

PR #2965, opened 2025-03-05
Wizard: Current unselected package fix

adamkankovsky:
Modify the logic of removing a package so that it remains in the selected table for possible re-addition until another package is removed or the wizard is switched to another screen.

FIX ISSUE: https://github.com/osbuild/image-builder-frontend/issues/2924

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -138,6 +138,9 @@ const Packages = () => {
       origin: ContentOrigin.EXTERNAL,
     });
 
+  const [currentlyRemovedPackages, setCurrentlyRemovedPackages] = useState<
+    IBPackageWithRepositoryInfo[]
+  >([]);
   const [isRepoModalOpen, setIsRepoModalOpen] = useState(false);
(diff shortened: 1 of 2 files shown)


lucasgarfield (dismissed):
Thank you Adam!

This PR absolutely meets the spec requested in the issue, but playing with it and removing more than one package still feels a bit disorienting due to the disappearing rows, at least to me.

What would you think if instead of just keeping track of the last removed package, we kept track of all removed packages as long as the user is on the “Selected Packages” tab?

Couple of other tips since you’re a first time contributor:

You’ve got some linting errors causing the unit tests to fail. We run those in the “Development checks / dev-check” – those have to pass for us to merge the PR. 

You can run the tests locally with `npm run test`. You can also just check the linting problems with `npm run lint`.

adamkankovsky:
@lucasgarfield 
Honestly I did both options, but in the end I decided for this one because it matched the description better. However, it is not difficult to convert it to an array. 
I apologize for the linter error, I did it quite late so I didn't let all the tests finish in order to save time.

regexowl (approved):
This works great, thank you! :sparkles:

regexowl:
One timeout in eph setup, giving pr_check another try

/retest

PR #2975, opened 2025-03-10
Wizard: Fix default key creation (HMS-5623)

regexowl:
Default activation key creation was broken, resulting in `requested serviceLevel is not supported for the organization`.

This PR removes serviceLevel from the create request body, defaulting to "Not defined", same as for "Role" and "Usage".

JIRA: [HMS-5623](https://issues.redhat.com/browse/HMS-5623)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
@@ -164,7 +164,6 @@ const ActivationKeysList = () => {
         await createActivationKey({
           body: {
             name: defaultActivationKeyName,
-            serviceLevel: 'Self-Support',
           },
         });
         window.localStorage.setItem(


regexowl:
How to test:
1. have no activation keys
2. open Wizard
3. go to Registration step

Current behaviour:
- a key is "created" (there is a default key name pre-selected in the Activation key dropdown), but information about the key cannot be loaded and an error is rendered

After the fix:
- a key is created, default name is selected in the dropdown and information about the key is fetched including "Not defined" for Role, SLA and Usage.

ksiekl:
I didn't manage to reproduce the error in the main:
activation_key
Could someone else give it a try as well please? :sparkles:

regexowl:
@ksiekl Thank you! Yeah, not sure why that would be though. This is all I see in the Activation Keys app, wonder if there's something wrong with my stage account maybe 
image

ksiekl:
Yeah, that's weird! No idea tho
image

regexowl:
My SKUs got borked during import x_x all's good. Thanks a lot @ksiekl for double and even triple checking for me! :sweat_smile: I'm closing this one.

PR #2976, opened 2025-03-10
on-prem: add OpenSCAP customizations (HMS-5824)

kingsleyzissou:
This PR adds the following:
- enables the OpenSCAP step for on-prem
- adds a warning if the `openscap-scanner` and `scap-security-guide` packages aren't installed on the host
- this implementation runs the `oscap` tool on the host to fetch the available profiles and eventual customizations

JIRA: [HMS-5824](https://issues.redhat.com/browse/HMS-5824)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Oscap/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/index.tsx
@@ -10,6 +10,8 @@ import {
 } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 
+import OscapOnPremSpinner from './OnPremSpinner';
+import OscapOnPremWarning from './OnPremWarning';
 import { Oscap, removeBetaFromRelease } from './Oscap';
 
 import {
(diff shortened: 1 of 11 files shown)


kingsleyzissou:
Screen recording below:

https://github.com/user-attachments/assets/87a70d70-5485-40ce-bb35-29c9a6dc1623

Loading the profiles is a bit slow :/

src/Components/CreateImageWizard/steps/Oscap/index.tsx
```diff
@@ -88,6 +94,41 @@ const OscapStep = () => {
     dispatch(clearKernelAppend());
   };
 
+  if (process.env.IS_ON_PREMISE && !onPremOpenSCAPAvailable) {
```
kingsleyzissou, line 97: Maybe this should be taken up one level for two reasons:
- it takes a bit of time to execute the check for the packages so the warning shows for a moment even if the packages are available
- some of the queries to get the profile information are attempted and fail with errors if OpenSCAP is not available

lucasgarfield:
This is working really nicely! What do you think we need to do to get it out of draft?

It could use a bit of polish here and there... should we do follow-ups or try to get everything in this PR? :thinking:

kingsleyzissou:
> This is working really nicely! What do you think we need to do to get it out of draft?
> 
> It could use a bit of polish here and there... should we do follow-ups or try to get everything in this PR? 🤔

Thanks for the review, yeah maybe we could do the polish ups and try get everything done here

kingsleyzissou:
@lucasgarfield the only thing that is outstanding now is the copy & the customizations that aren't disabled.
For the customizations maybe we can do that in a follow up, or would you prefer to get it all done here?

croissanne (approved):

croissanne (approved):

lucasgarfield:
> @lucasgarfield the only thing that is outstanding now is the copy & the customizations that aren't disabled. For the customizations maybe we can do that in a follow up, or would you prefer to get it all done here?

Follow up is fine!

Sanne has approved, so I'm merging.

PR #2978, opened 2025-03-11
Blueprints: hiding unused activation key in when editing Blueprint (HMS-5631)

### LINKED ISSUE #2960: False activation key displayed in edit mode
In order to boost registration of machines to Red Hat Insights, Image Builder automatically selects an activation key when the user is creating a blueprint. Either the user’s last used key is selected (kept track of in the browser storage) or, if they do not have an activation key, an activation key is created for them.

There is a bug when editing a blueprint related to this behavior.

If a blueprint is created without an activation key, when the user goes to edit the blueprint, an activation key is visible in the disabled select.

Steps to reproduce:

1. Create a blueprint. Do not add an activation key.
2. Edit the blueprint. Visit the activation keys step. 

Desired solution:

The disabled select should not display an activation key in edit mode if the blueprint does not have one.

ksiekl:
Fixed #2960

The Select dropdown in Registration wizard step no longer shows activation key when user decides to register later.

JIRA: [HMS-5631](https://issues.redhat.com/browse/HMS-5631)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
@@ -241,7 +241,9 @@ const ActivationKeysList = () => {
           variant={SelectVariant.typeahead}
           onToggle={handleToggle}
           onSelect={setActivationKey}
-          selections={activationKey}
+          selections={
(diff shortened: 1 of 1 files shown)


regexowl:
pr_check fails in eph setup, but passed previously. Merging manually

PR #2983, opened 2025-03-12
Wizard: add/rm firstboot service based on state

avitova:
I noticed there is a bug that does not show a relevant list of services that we create on user's behalf.

How to reproduce:
1. start creating a blueprint with a first boot script
2. go to the review step - the custom-first-boot service is not displayed there, but we are adding it on user's behalf
3. click on "Edit blueprint"
4. go to First boot step and remove the script
5. go to the review step - the custom-first-boot service is displayed there, but we will remove it after clicking "edit"

Overall, I thought that the code in requestMapper was a bit unclear, and I needed to do something similar for the Satellite follow-up, so I was digging deeper, and found a small bug.
Let me know what you think. I tried to prevent the code from removing/adding service in case we just edit an existing script.

### DIFF
--- a/src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
+++ b/src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
@@ -12,8 +12,11 @@ import {
   HelperTextItem,
 } from '@patternfly/react-core';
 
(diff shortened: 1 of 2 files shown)


regexowl (approved):
This works beautifully. Thank you for cleaning up the code :heart:

PR #2984, opened 2025-03-12
Fedora-services: add support for fedora env

amirfefer:
This pull request introduces enhancements to the Image Builder Frontend service to support Fedora services. 
For more details, and a full dev setup docs, please refer to the [README file](https://github.com/FedoraConsole/community-chrome).

The key updates include:

- Support for Fedora Services:
Adds compatibility for Fedora services, by introducing `isFedoraEnv` in the existing `useGetEnvironment` hook. This new env is set by the ChromeAPI.
```js
  const { isFedoraEnv } = useGetEnvironment();
```

- Keeping the hosted (consoleDot) and on premise environment isolated.

- Aligned Blueprint Wizard:
Hides irrelevant steps in the blueprint wizard when configuring Fedora and CentOS distributions, streamlining the user experience.

- Expanded Distribution Support:
Adds support for Fedora and CentOS distributions (Fedora 40, 41 & CentOS 9)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/Footer/Footer.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/Footer/Footer.tsx
@@ -20,6 +20,7 @@ import {
   useUpdateBlueprintMutation,
 } from '../../../../../store/backendApi';
 import { resolveRelPath } from '../../../../../Utilities/path';
+import { useGetEnvironment } from '../../../../../Utilities/useGetEnvironment';
 import { mapRequestFromState } from '../../../utilities/requestMapper';
 import { useIsBlueprintValid } from '../../../utilities/useValidation';
 
@@ -33,6 +34,7 @@ const ReviewWizardFooter = () => {
     useUpdateBlueprintMutation({ fixedCacheKey: 'updateBlueprintKey' });
   const { auth } = useChrome();
   const { composeId } = useParams();
+  const { isFedoraEnv } = useGetEnvironment();
   const [isOpen, setIsOpen] = useState(false);
   const store = useStore();
   const onToggleClick = () => {
@@ -50,7 +52,7 @@ const ReviewWizardFooter = () => {
   }, [isUpdateSuccess, isCreateSuccess, resetCreate, resetUpdate, navigate]);
 
   const getBlueprintPayload = async () => {
-    if (!process.env.IS_ON_PREMISE) {
+    if (!process.env.IS_ON_PREMISE && !isFedoraEnv) {
       const userData = await auth?.getUser();
       const orgId = userData?.identity?.internal?.org_id;
       const requestBody = orgId && mapRequestFromState(store, orgId);
--- a/src/Utilities/useGetEnvironment.ts
+++ b/src/Utilities/useGetEnvironment.ts
@@ -2,14 +2,19 @@ import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome'
 import { useFlag as useUnleashFlag } from '@unleash/proxy-client-react';
 
 export const useGetEnvironment = process.env.IS_ON_PREMISE
-  ? () => ({ isBeta: () => false, isProd: () => true })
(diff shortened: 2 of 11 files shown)


src/Utilities/useGetEnvironment.ts
```diff
@@ -47,4 +52,15 @@ const onPremFlag = (flag: string): boolean => {
 
 // Since some of these flags are only relevant to
 // the service, we need a way of bypassing this for on-prem
-export const useFlag = !process.env.IS_ON_PREMISE ? useUnleashFlag : onPremFlag;
+export const useFlag = (flag: string) => {
+  const { isFedoraEnv } = useGetEnvironment();
+
+  const onPremFlagValue = onPremFlag(flag);
+  const unleashFlagValue = useUnleashFlag(flag);
```
ezr-ondrej, line 59: Should this be skipped for fedora and on-prem? I'm assuming this will fail in those environments, or it wont? :)

src/Components/CreateImageWizard/steps/Review/Footer/Footer.tsx
```diff
@@ -52,7 +54,9 @@ const ReviewWizardFooter = () => {
   const getBlueprintPayload = async () => {
     if (!process.env.IS_ON_PREMISE) {
       const userData = await auth?.getUser();
-      const orgId = userData?.identity?.internal?.org_id;
+      const orgId = isFedoraEnv
+        ? userData?.identity?.user?.username
+        : userData?.identity?.internal?.org_id;
```
ezr-ondrej, line 59: would be nice to abstract it away a little 🤔 I'm not sure how, but this feels slightly dirty :(

ezr-ondrej (approved):
It is not so lean I'd hope, but it does the job done, I'm all for just merging as is, but I'd love @lucasgarfield or @regexowl opinion on that as they will be for the most part on hook to maintain it :)
I guess we should have the smoke test PR at least open before we merge this.

src/Components/CreateImageWizard/steps/Review/Footer/Footer.tsx
lucasgarfield: Maybe just a helper function `getOrgId` for now? That way at least the dirty part is isolated. :thinking:

lucasgarfield (commented):
I haven't had a chance to test this yet but looking through it I have one initial thought. I wasn't even sure if I should bring it up because it seems like a nitpick but I decided it is important to me...

I’m a little bit worried that we will forget why all of these things are potentially now undefined. I know it will be a bit more verbose but instead of using optional chaining (`.?`) could we just short circuit (`!isFedoraEnv &&`) instead? That way things are explicit and future us knows exactly what is going on.

src/Components/CreateImageWizard/steps/Review/Footer/Footer.tsx
ezr-ondrej: Agreed! :)

amirfefer: well, we don't really need the orgId :)  similar to the onPrem implementation.

src/Utilities/useGetEnvironment.ts
amirfefer: We can't call a hook conditionally, but no worries, it just returns undefined in those env.

amirfefer:
Thanks, @ezr-ondrej  and @lucasgarfield!
I updated this PR and also added some tests :)

src/Utilities/useGetEnvironment.ts
ezr-ondrej: Ah right! I always forget :)

regexowl:
Wow, this works really nicely, thank you! It seems the issue with orgId is resolved, so if there's nothing else I'd be happy to merge.

regexowl (approved):

PR #2987, opened 2025-03-14
Wizard: Disable "No results found" options (HMS-5641)

regexowl:
The "No results found for <searchTerm>" options were previously enabled, but ignored upon click. This disables them properly and adds tests to check if they're disabled.

Before:
image

After:
image

JIRA: [HMS-5641](https://issues.redhat.com/browse/HMS-5641)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Timezone/components/TimezoneDropDown.tsx
+++ b/src/Components/CreateImageWizard/steps/Timezone/components/TimezoneDropDown.tsx
@@ -43,9 +43,6 @@ const TimezoneDropDown = () => {
       filteredTimezones = timezones.filter((timezone: string) =>
         String(timezone).toLowerCase().includes(filterValue.toLowerCase())
       );
-      if (!filteredTimezones.length) {
-        filteredTimezones = [`No results found for "${filterValue}"`];
-      }
       if (!isOpen) {
         setIsOpen(true);
       }
@@ -70,7 +67,7 @@ const TimezoneDropDown = () => {
   };
 
   const onSelect = (_event: React.MouseEvent, value: string) => {
-    if (value && !value.includes('No results')) {
+    if (value) {
       setInputValue(value);
       setFilterValue('');
       setErrorText('');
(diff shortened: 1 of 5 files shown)


src/Components/CreateImageWizard/steps/Timezone/components/TimezoneDropDown.tsx
```diff
@@ -145,7 +145,11 @@ const TimezoneDropDown = () => {
       >
         <SelectList>
```
ksiekl, line 146: It works great! I just have one question - right now we are relying on text matching to see if the "No results found" is a selectable option, which is a bit unintuitive to me. Wouldn't it be better to check if there's 0 options to select from and then render the "No results found" text? :) Feel free to correct me, it's just not the way I expected this to be implemented.

regexowl: Hmm, I see your point. I must admit I was simply following the [PF docs for typeahead selects](https://www.patternfly.org/components/menus/select#typeahead) during the implementation. Meaning since they push "No results found" to the `selectOptions` array in the filtering `useEffect`, so did I. :thinking:

regexowl: That being said we could also refactor a bit :sweat_smile: The code is not always strictly following the docs

regexowl: Added a wip commit https://github.com/osbuild/image-builder-frontend/pull/2987/commits/74af2233f2187f4f953adf86cd00a9677b17a807 that would render the options based on `selectOptions` length, what do you think?

regexowl: Ugh, I'm great at numbers obviously... another try

ksiekl: Yeah! I like it! thanks for the explanation :smile:

ksiekl (approved):

regexowl:
The review persisted through the commit squashing, very nice :open_mouth:

ksiekl:
Yeah that's neat!

PR #2990, opened 2025-03-14
Wizard: Fix activation keys URL (HMS-5642)

regexowl:
This updates activation keys URL to behave the same way the content one does.

JIRA: [HMS-5642](https://issues.redhat.com/browse/HMS-5642)


regexowl:
Switching from full to appended path will allow us to get rid of `isProd` distinction and will also stop taking us out of local dev space directly to consoleDot.

mgold1234 (approved):
/lgtm

PR #2996, opened 2025-03-17
dependabot: ignore playwright dependency

### LINKED ISSUE #2905: Import modal: Error text doesn't render on "Upload"
The validation doesn't appear when uploading a blueprint via the "Upload" button. It appears only on drag and drop

Image

### LINKED ISSUE #2919: Blueprints: `isInvalidFormat` error should have `error` variant

croissanne:
Since the container embedded in our CI images is pinned, any update here
would result in the container having to be pulled again in CI
runs. Let's reduce the frequency that playwright gets updated, making it
easier to keep the container and the dependencies here somewhat in sync.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,9 @@ updates:
       time: "04:00"
     open-pull-requests-limit: 3
     rebase-strategy: "auto"
+    ignore:
+      - dependency-name: "@playwright/test"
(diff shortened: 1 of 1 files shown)


regexowl (approved):
Thank you! :heart:

PR #2999, opened 2025-03-17
Race condition activation key (HMS-5651)

ksiekl:
If a user skips too quickly to the Review step in Wizard, the step rerenders with the newly created activation key, and the "Create blueprint" button stays disabled until the activation key loads. 

activation_key_rerender

JIRA: [HMS-5651](https://issues.redhat.com/browse/HMS-5651)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/ActivationKeysList.tsx
@@ -164,12 +164,12 @@ const ActivationKeysList = () => {
             name: defaultActivationKeyName,
             serviceLevel: 'Self-Support',
           },
-        });
+        }).unwrap();
+
         window.localStorage.setItem(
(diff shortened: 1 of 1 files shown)


regexowl:
The unwrapping of the request works nicely! There's one more detail though - when you open a "register-later" blueprint in Edit mode, it doesn't have an activation key, making the "Save changes" button disabled and failing the edit tests.

ksiekl:
Thanks! I was trying to figure out what made the pipeline fail :thinking:

lucasgarfield (commented):
This works nicely. It seems to fix a subtle bug where the "Next" button should have been disabled but wasn't... 

I think in the future it might be a little hard for us to remember why we needed to unwrap the promise and not refetch - could you update the commit message to explain why this works? 

Then I think it will be ready to merge! :+1:

regexowl (approved):
Works nicely, thank you!

PR #3000, opened 2025-03-17
devDeps: Manually bump msw from 2.7.0 to 2.7.3 (HMS-5652)

regexowl:
This bumps msw from 2.7.0 to 2.7.3

JIRA: [HMS-5652](https://issues.redhat.com/browse/HMS-5652)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,7 +77,7 @@
         "madge": "8.0.0",
         "mini-css-extract-plugin": "2.9.2",
         "moment": "2.30.1",
-        "msw": "2.7.0",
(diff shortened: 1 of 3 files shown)


regexowl:
While updating the version of msw and running `npm ci` file `src/mockServiceWorker.js` gets automagically updated as well. Since dependabot doesn't want to touch anything outside of `package.json` and `package-lock.json` this creates a conflict between the original and `npm ci` updated file.

Which is the reason why we need to bump msw manually, at least for now.

ksiekl (approved):
Looks good

PR #3004, opened 2025-03-18
Wizard: Replace deprecated selects on Azure step (HMS-5655)

regexowl:
This replaces deprecated selects on the Azure step for non-deprecated ones.

JIRA: [HMS-5655](https://issues.redhat.com/browse/HMS-5655)


avitova:
Not sure if this is the desired behaviour, I'd consider this a bug, mainly if the second select component has a different behaviour:

https://github.com/user-attachments/assets/35f97b69-672b-4cb1-9d36-322f412e95cc

^ the "hms" string stays. Other than that, I think this looks good and works well.

regexowl:
> Not sure if this is the desired behaviour, I'd consider this a bug, mainly if the second select component has a different behaviour:
> Screencast.From.2025-03-19.10-28-24.mp4
> 
> ^ the "hms" string stays. Other than that, I think this looks good and works well.

Great catch! I've updated the `onClear` logic, should do the trick.

avitova (approved):
LGTM!:)

PR #3012, opened 2025-03-19
Add currents reporter for Playwright tests (HMS-5660)

marusak:
This adds currents reporter for Playwright tests

JIRA: [HMS-5660](https://issues.redhat.com/browse/HMS-5660)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@babel/preset-typescript": "7.26.0",
     "@patternfly/react-icons": "5.4.2",
     "@playwright/test": "1.50.1",
+    "@currents/playwright": "1.9.4",
     "@redhat-cloud-services/eslint-config-redhat-cloud-services": "2.0.12",
     "@redhat-cloud-services/frontend-components-config": "6.3.8",
     "@redhat-cloud-services/tsc-transform-imports": "1.0.23",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "@babel/preset-typescript": "7.26.0",
     "@patternfly/react-icons": "5.4.2",
     "@playwright/test": "1.50.1",
+    "@currents/playwright": "^1.9.4",
```
regexowl, line 38: Can you pin the version, dependabot likes pinned version more

```suggestion
    "@currents/playwright": "1.9.4",
```

regexowl:
You'll probably need to run `npm i` after pinning the version and adding re-generated `package-lock.json` to the PR as well. Otherwise looks good, thank you!

regexowl:
Added a simple description to make pr-best-practices happy, feel free to change it any way you like

package.json
marusak: Fixed!

regexowl:
Hmm, Schutzbot seems to be struggling still

marusak:
> The reporter was added for all playwright invocations

I think it does not hurt, but also locally "list" should do? Or you also want "html" for local runs?

regexowl:
Agree, I like having an option of the 'html' reporter for local runs.

regexowl (approved):

PR #3020, opened 2025-03-21
codecov: Ignore API, PW and store folders and update config (HMS-5792)

regexowl:
API, PW and  store folders contain mostly files that are copied, auto-generated or test files. We don't need to collect coverage for those.

There were previously errors in expected values, this should fix them.

JIRA: [HMS-5792](https://issues.redhat.com/browse/HMS-5792)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,15 +1,15 @@
 coverage:
   status:
-    patch: no
+    patch: off
     project:
       default:
         threshold: 5%
 codecov:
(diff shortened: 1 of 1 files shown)


regexowl:
Feel free to suggest more filed/folders to add to ignore. I just quickly ran through those.

croissanne (approved):

croissanne (approved):

PR #3021, opened 2025-03-21
Wizard: Show "Strong" label only for Azure when password meets criteria (HMS-5674)

mgold1234:
Currently, we only want to display the "Strong" label when the password meets the required strength conditions and only if it is Azure target. This change ensures that the label remains hidden for non-Azure cases.

Let’s just disable it for now until we can decide on further adjustments.

JIRA: [HMS-5674](https://issues.redhat.com/browse/HMS-5674)


regexowl (approved):

PR #3026, opened 2025-03-23
Wizard: Refactor Minimum Size input in File system step

mgold1234:
This commit refactors the Minimun size field by replacing HookValidatedInput with the new ValidatedInputAndTextArea component. It fixes a bug where the error icon remained visible after the user cleared the value in the field.

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -167,7 +167,7 @@ export function useFilesystemValidation(): StepValidation {
   const duplicates = getDuplicateMountPoints(partitions);
   for (const partition of partitions) {
     if (partition.min_size === '') {
-      errors[`min-size-${partition.id}`] = 'partition size is required';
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -166,7 +166,10 @@ export function useFilesystemValidation(): StepValidation {
 
   const duplicates = getDuplicateMountPoints(partitions);
   for (const partition of partitions) {
-    if (!isMountpointMinSizeValid(partition.min_size)) {
+    if (partition.min_size === '') {
+      errors[`min-size-${partition.id}`] = 'partition size is required';
```
regexowl, line 170: One tiny nitpick, otherwise looks great. Thank you!

```suggestion
      errors[`min-size-${partition.id}`] = 'Partition size is required';
```

regexowl (approved):
LGTM! Thank you! :heart:

PR #3031, opened 2025-03-25
Wizard: Remove `if (!value)` check from validators (HMS-5681)

regexowl:
This removes check for whether the value is defined from validators and moves it to the place the validator is called where needed.

JIRA: [HMS-5681](https://issues.redhat.com/browse/HMS-5681)


lucasgarfield (approved):
Thank you!

PR #3042, opened 2025-03-26
Segment tracking

ksiekl:
Initial attempt at tracking analytics when user creates a BP - very drafty, very not official

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
@@ -33,7 +33,7 @@ export const CreateSaveAndBuildBtn = ({
   setIsOpen,
   isDisabled,
 }: CreateDropdownProps) => {
-  const { analytics, isBeta } = useChrome();
+  const { analytics, auth, isBeta } = useChrome();
   const packages = useAppSelector(selectPackages);
 
   const [buildBlueprint] = useComposeBlueprintMutation();
@@ -43,10 +43,12 @@ export const CreateSaveAndBuildBtn = ({
   const dispatch = useAppDispatch();
   const onSaveAndBuild = async () => {
     const requestBody = await getBlueprintPayload();
+    const user = await auth.getUser();
     setIsOpen(false);
 
     if (!process.env.IS_ON_PREMISE && requestBody) {
       const analyticsData = {
+        user: user,
         image_name: requestBody.name,
         description: requestBody.description,
         distribution: requestBody.distribution,
@@ -70,6 +72,8 @@ export const CreateSaveAndBuildBtn = ({
         `${AMPLITUDE_MODULE_NAME}-blueprintCreated`,
         analyticsData
       );
+      // eslint-disable-next-line no-console
+      console.log('ANALYTICS: ', analyticsData);
     }
     const blueprint =
       requestBody &&
--- a/src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
@@ -45,15 +45,32 @@ export const CreateSaveAndBuildBtn = ({
     const requestBody = await getBlueprintPayload();
     setIsOpen(false);
 
-    if (!process.env.IS_ON_PREMISE) {
-      analytics.track(`${AMPLITUDE_MODULE_NAME}-blueprintCreated`, {
-        module: AMPLITUDE_MODULE_NAME,
-        isPreview: isBeta(),
-        type: 'createBlueprintAndBuildImages',
+    if (!process.env.IS_ON_PREMISE && requestBody) {
+      const analyticsData = {
+        image_name: requestBody.name,
+        description: requestBody.description,
+        distribution: requestBody.distribution,
+        openscap: requestBody.customizations.openscap,
+        image_request_types: requestBody.image_requests.map(
+          (req) => req.image_type
+        ),
+        image_request_architectures: requestBody.image_requests.map(
+          (req) => req.architecture
+        ),
+        // imageRequests: requestBody.image_requests,
+        // customizations: requestBody.customizations,
+        metadata: requestBody.metadata,
         packages: packages.map((pkg) => pkg.name),
-      });
+        file_system_configuration: requestBody.customizations.filesystem,
+        module: AMPLITUDE_MODULE_NAME,
+        is_preview: isBeta(),
+        type: 'createBlueprintAndBuildImages',
+      };
+      analytics.track(
+        `${AMPLITUDE_MODULE_NAME}-blueprintCreated`,
+        analyticsData
+      );
     }
-
     const blueprint =
       requestBody &&
       (await createBlueprint({


src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -43,17 +43,38 @@ export const CreateSaveAndBuildBtn = ({
   const dispatch = useAppDispatch();
   const onSaveAndBuild = async () => {
     const requestBody = await getBlueprintPayload();
+    const user = await auth.getUser();
     setIsOpen(false);
 
-    if (!process.env.IS_ON_PREMISE) {
-      analytics.track(`${AMPLITUDE_MODULE_NAME}-blueprintCreated`, {
+    if (!process.env.IS_ON_PREMISE && requestBody) {
+      const analyticsData = {
```
ondrejbudai, line 50: @lucasgarfield Cannot we just send the whole request body? Would it be too "expensive"? Or is it an issue to send a whole object with nested values? Just curious, I don't know anything about this. 😅

ondrejbudai: Ah, I've just realized that the request body can contain secrets (user password, activation key, etc.), so it's probably best to handle new fields in the request on an opt-in basis.

ksiekl: I'm not sure if it contains user password, but there is definitely the activation key, so yea, I can keep just selecting specific parts of the request body.

ksiekl:
I didn’t add the optional pieces of code yet (the ones from the doc), since I’m not sure if this approach is final. Let me know what you think @lucasgarfield and I can add those in :crossed_fingers:

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
lucasgarfield: It would in fact contain the user's password... so I guess that answers this question! :sweat_smile:

lucasgarfield:
@ksiekl have you verified this is working in Segment for your stage user ID?

ksiekl:
Whoop, yes, it shows my account id, I somehow managed to delete it when cleaning up my commits but it's back again :woman_technologist:

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -138,13 +171,12 @@ export const CreateSaveButton = ({
...
-        module: AMPLITUDE_MODULE_NAME,
-        isPreview: isBeta(),
-        type: 'createBlueprint',
-        packages: packages.map((pkg) => pkg.name),
-      });
+    if (!process.env.IS_ON_PREMISE && requestBody) {
+      const analyticsData = createAnalytics(requestBody, packages, isBeta);
+      analytics.track(
+        `${AMPLITUDE_MODULE_NAME}-blueprintCreated`,
+        analyticsData
```
regexowl, line 178: Type will be different here, right? Original was `'createBlueprint'`, but the one in analytics data is `'createBlueprintAndBuildImages'`. The original tracking is used by recommendations team, so this could probably make some mess in their analytics.

src/Components/CreateImageWizard/steps/Review/Footer/EditDropdown.tsx
```diff
@@ -22,19 +27,64 @@ type EditDropdownProps = {
...
+      (req) => req.architecture
+    ),
+    imageRequests: requestBody.image_requests,
+    customizations: requestBody.customizations,
+    metadata: requestBody.metadata,
+    packages: packages.map((pkg) => pkg.name),
+    file_system_configuration: requestBody.customizations.filesystem,
+    module: AMPLITUDE_MODULE_NAME,
+    is_preview: isBeta(),
+    type: 'createBlueprintAndBuildImages',
```
regexowl, line 53: Do we want to use the same type as for `CreateDropdown`, or do we want to differentiate the actions? Maybe just something like:
```suggestion
    type: 'editBlueprintAndBuildImages',
```
Also is the addition of packages info we're sending to recommendations team in create action intentional here? Just curious, I bet they wouldn't mind getting more info about how are packages used :)

regexowl:
Could you also please update PR description when this is not drafty and becomes official?

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
ksiekl: Ah, good catch! :see_no_evil:

src/Components/CreateImageWizard/steps/Review/Footer/EditDropdown.tsx
ksiekl: The plan is not very specific, but I'd agree that it makes sense to differentiate the types, so I changed it :) As for the packages, they are intentional, they were mentioned in the doc connected to this task in Jira :memo:

lucasgarfield:
I think you accidentally made a merge commit @ksiekl! I sure wish we could disable those in the Github UI, but alas...

You can get it fixed by doing:

```
git checkout main
git pull
git checkout segmentTracking
git rebase -i
```

Then drop the merge commit. Then...

```
git rebase main
git push -f $your_remote_name
```

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -22,13 +22,41 @@ import {
...
+    description: requestBody.description,
+    distribution: requestBody.distribution,
+    openscap: requestBody.customizations.openscap,
+    image_request_types: requestBody.image_requests.map(
+      (req) => req.image_type
+    ),
+    image_request_architectures: requestBody.image_requests.map(
+      (req) => req.architecture
+    ),
+    imageRequests: requestBody.image_requests,
```
ondrejbudai, line 49: This key is in camelCase, but other fields are in snake_case, is there a reason for it?

src/Components/CreateImageWizard/steps/Review/Footer/EditDropdown.tsx
```diff
@@ -22,19 +27,63 @@ type EditDropdownProps = {
...
+    description: requestBody.description,
+    distribution: requestBody.distribution,
+    openscap: requestBody.customizations.openscap,
+    image_request_types: requestBody.image_requests.map(
+      (req) => req.image_type
+    ),
+    image_request_architectures: requestBody.image_requests.map(
+      (req) => req.architecture
+    ),
+    imageRequests: requestBody.image_requests,
```
ondrejbudai, line 46: The same inconsistency is here.

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
ksiekl: Ah yes, the reason being I didn't notice :see_no_evil:

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -22,13 +22,41 @@ import {
...
+    customizations: requestBody.customizations,
```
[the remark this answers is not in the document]
ksiekl: Right, yea. I only took the account number and organization from it, if there is any other data we could need from customizations, lmk :v:

src/Components/CreateImageWizard/utilities/requestMapper.ts
lucasgarfield, line 1: Now that we're using the useChrome hook to get the accountID, should we revert everything in requestMapper.ts to how it was before? I also assume it would mean we could revert the changes to the tests.

lucasgarfield (changes requested):

src/Components/CreateImageWizard/utilities/requestMapper.ts
ksiekl: Oh right, yeah, we can revert that. Done!

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -141,12 +191,12 @@ export const CreateSaveButton = ({
     const requestBody = await getBlueprintPayload();
     setIsOpen(false);
 
-    if (!process.env.IS_ON_PREMISE && !isFedoraEnv) {
-      analytics.track(`${AMPLITUDE_MODULE_NAME}-blueprintCreated`, {
-        module: AMPLITUDE_MODULE_NAME,
-        isPreview: isBeta(),
+    if (!process.env.IS_ON_PREMISE && !isFedoraEnv && requestBody) {
+      const analyticsData = createAnalytics(requestBody, packages, isBeta);
+      analytics.track(`${AMPLITUDE_MODULE_NAME} - Blueprint Created`, {
```
regexowl, line 196: Same here, changing `${AMPLITUDE_MODULE_NAME}-blueprintCreated` to `${AMPLITUDE_MODULE_NAME} - Blueprint Created` can potentially break already set up Amplitude tracking.

src/test/fixtures/editMode.ts
```diff
@@ -33,6 +33,7 @@ export const expectedSubscription = {
   insights: true,
   rhc: true,
   organization: 5,
+  account: 5,
```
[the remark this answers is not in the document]
ksiekl: yup, thanks for checking that!

src/Components/CreateImageWizard/steps/Review/Footer/CreateDropdown.tsx
```diff
@@ -47,15 +84,20 @@ export const CreateSaveAndBuildBtn = ({
...
```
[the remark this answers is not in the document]
ksiekl: Also, the same happened rn for "Blueprint Updated", so I changed that too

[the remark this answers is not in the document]
regexowl: Gotcha, what I've meant is that the recommendations team is already tracking data via Amplitude for several months now and they're using the original names of events.

src/Components/CreateImageWizard/utilities/requestMapper.ts
lucasgarfield: Hmm, something might have gone wrong @ksiekl, I'm still seeing all of the changers to requestMapper.ts and the associated tests in this PR. :thinking: Could you check your local branch again and push?

lucasgarfield: Oh I see, you removed them in the second commit. Please disregard my previous comment. :sweat_smile:

lucasgarfield (changes requested):

src/Components/CreateImageWizard/steps/Review/Footer/EditDropdown.tsx
```diff
@@ -22,19 +28,72 @@ type EditDropdownProps = {
   isDisabled: boolean;
 };
 
+const createAnalytics = (
```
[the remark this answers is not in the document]
ksiekl: Right, I expected them to collect different data at first and then got the dev blindness :see_no_evil: Thanks for noticing!

regexowl (approved):
Ok, I think this looks great. Thank you!

PR #3043, opened 2025-03-27
store: Safeguard splicing (HMS-5793)

regexowl:
This adds a check for existing/non-existing item before splicing an array.

Based on https://github.com/osbuild/image-builder-frontend/pull/3029#discussion_r2011489983

> There's a potential issue with the group removal logic. When findIndex() doesn't find the group, it returns -1, which would cause splice(-1, 1) to remove the last element in the array rather than doing nothing. 

JIRA: [HMS-5793](https://issues.redhat.com/browse/HMS-5793)


ksiekl (approved):
Nice check!

PR #3045, opened 2025-03-27
tsconfig: Exclude `/pkg` folder during build

regexowl:
After running `make cockpit/devel`, the generated files do not pass `npm run build`, outputing failures and making dev console and output of `npm run build` cluttered with errors.

This excludes `/pkg` folder during build time, removing the errors.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -24,5 +24,6 @@
         "./src/test/mocks/*"
       ]
     }
-  }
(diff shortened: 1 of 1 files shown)


regexowl:
Not sure if this is the most elegant solution, open to any suggestions

croissanne (approved):

PR #3055, opened 2025-03-31
[wip] on-prem: enable building & uploading AWS images (HMS-5823)

kingsleyzissou:
- read `/etc/osbuild-worker/osbuild-worker.toml` for aws configs
- if set, enable aws building & uploading

JIRA: [HMS-5823](https://issues.redhat.com/browse/HMS-5823)

### DIFF
--- a/src/Components/CreateImageWizard/steps/TargetEnvironment/Aws/index.tsx
+++ b/src/Components/CreateImageWizard/steps/TargetEnvironment/Aws/index.tsx
@@ -68,19 +68,21 @@ const Aws = () => {
         by Red Hat, to your own AWS account.
       </Text>
       <FormGroup label="Share method:">
-        <Radio
-          id="radio-with-description"
-          label="Use an account configured from Sources."
-          name="radio-7"
-          description="Use a configured source to launch environments directly from the console."
-          isChecked={shareMethod === 'sources'}
-          onChange={() => {
-            dispatch(changeAwsSourceId(undefined));
-            dispatch(changeAwsAccountId(''));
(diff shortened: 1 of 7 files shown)


kingsleyzissou:
TODO:
- handle the logo for the target environment (we just get alt text for now)
- we need to know the region for the upload request
  - either configure the image wizard to give us this info
  - or read it from the `osbuild-worker.toml` config

src/Components/CreateImageWizard/steps/TargetEnvironment/Aws/index.tsx
```diff
@@ -12,6 +12,10 @@ import {
   HelperText,
   HelperTextItem,
   Button,
+  Select,
```
kingsleyzissou, line 15: For this to work we would need to update to include the `region` which we obviously can't do since we'd then need workers in each region.
https://github.com/osbuild/image-builder-crc/blob/48d1c3b953b106fd53bf98ff32fbfc7719206456/internal/v1/api.yaml#L1316-L1330

Seems like a hard-blocker towards using a dropdown for the upload region

kingsleyzissou:
I've hit a blocker for this in `image-builder-crc` added a comment inline. We might have to just read the aws region from `/etc/osbuild-worker/osbuild-worker.toml` and maybe show an error if the field is missing.

kingsleyzissou:
I will mark this PR as ready once #3103 lands

croissanne:
Hmm technically I guess asking for the account id to share with isn't really necessary on prem, aswe just upload straight to their account.

PR #3060, opened 2025-04-01
Wizard: show a specific error message when description validation fails

### LINKED ISSUE #3036: Blueprint description says "invalid" but not why
I pasted 5 paragraphs of lorem ipsum, it says this is invalid (fair), but it doesn't say why:

Image

ondrejbudai:
Prior to this commit, the wizard showed "Invalid description" when the description was too long. This was rather unhelpful for users.

The actual size check happens in validators.ts, but the functions there just return a boolean value, even if they check for multiple criteria.

I'm not entirely sure, why the validation functions live in a separate file, so I decided to move this specific condition to useValidation.tsx so the condition, and the relevant user-facing error message can be defined next to each other.

Fixes #3036

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -42,7 +42,6 @@ import { timezones } from '../steps/Timezone/timezonesList';
 import {
   getDuplicateMountPoints,
   isBlueprintNameValid,
-  isBlueprintDescriptionValid,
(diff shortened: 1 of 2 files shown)


regexowl (approved):
The error message is much more helpful now, thanks! :sparkles: 

(I believe generally the validators live in a separate file as it makes the `useValdation` a bit more concise and the full pattern for validation of a specific value can be found quickly in one place, but since we're just checking length of the description it makes perfect sense to do the validation directly here)

PR #3071, opened 2025-04-03
Wizard: Add support for multiple users (HMS-5911)

regexowl:
This adds support for multiple users in the Wizard. The users are rendered in tabs and can be both added and removed. When clicking on remove button, modal pops up to inform user that the changes cannot be undone.

JIRA: [HMS-5911](https://issues.redhat.com/browse/HMS-5911)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Users/components/RemoveUserModal.tsx
+++ b/src/Components/CreateImageWizard/steps/Users/components/RemoveUserModal.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 
 import { Button, Modal } from '@patternfly/react-core';
 
+import calculateNewIndex from './calculateNewIndex';
+
 import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
 import { removeUser, selectUsers } from '../../../../../store/wizardSlice';
 
@@ -12,6 +14,7 @@ type RemoveUserModalProps = {
   tabIndex: number;
   setIndex: React.Dispatch<React.SetStateAction<number>>;
   isOpen: boolean;
+  userName: string;
 };
 
 const RemoveUserModal = ({
@@ -21,6 +24,7 @@ const RemoveUserModal = ({
   tabIndex,
   setIndex,
   isOpen,
+  userName,
 }: RemoveUserModalProps) => {
   const dispatch = useAppDispatch();
   const users = useAppSelector(selectUsers);
@@ -30,16 +34,11 @@ const RemoveUserModal = ({
   };
 
   const onConfirm = () => {
-    const tabIndexNum = tabIndex;
-    let nextTabIndex = activeTabKey;
-
-    if (tabIndexNum < activeTabKey) {
-      // if a preceding tab is closing, keep focus on the new index of the current tab
-      nextTabIndex = activeTabKey - 1 > 0 ? activeTabKey - 1 : 0;
-    } else if (activeTabKey === users.length - 1) {
-      // if the closing tab is the last tab, focus the preceding tab
-      nextTabIndex = users.length - 2 > 0 ? users.length - 2 : 0;
-    }
+    const nextTabIndex = calculateNewIndex(
+      tabIndex,
+      activeTabKey,
+      users.length
+    );
 
     setActiveTabKey(nextTabIndex);
     setIndex(nextTabIndex);
@@ -49,7 +48,7 @@ const RemoveUserModal = ({
 
   return (
     <Modal
-      title="Remove user?"
+      title={`Remove user${userName ? ` ${userName}` : ''}?`}
       isOpen={isOpen}
       onClose={onClose}
       width="50%"
--- a/src/Components/CreateImageWizard/steps/Users/components/RemoveUserModal.tsx
+++ b/src/Components/CreateImageWizard/steps/Users/components/RemoveUserModal.tsx
@@ -1,0 +1,72 @@
+import React from 'react';
+
+import { Button, Modal } from '@patternfly/react-core';
+
+import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
+import { removeUser, selectUsers } from '../../../../../store/wizardSlice';
+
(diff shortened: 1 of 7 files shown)


src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
```diff
@@ -71,88 +86,111 @@ const UserInfo = () => {
...
+                icon={<ExternalLinkAltIcon />}
+                iconPosition="right"
+                href={GENERATING_SSH_KEY_PAIRS_URL}
+                className="pf-v5-u-pl-0"
+              >
+                Learn more about SSH keys
+              </Button>
+            </FormGroup>
+            <FormGroup className="pf-v5-u-pb-md">
+              <Checkbox
```
ksiekl, line 163: When I create multiple users, give them names, I can click on "Administrator" label (_not_ the checkbox), and the wheels just keep on rolling: 

wheel

I think this might be connected to the bug with the random creation of new tabs tho, so maybe if that is fixed, this bug will resolve "itself" too.

regexowl:
It's me, I'm "this guy" in this scenario, should be fixed now. Really nice catches, thank you!

src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
```diff
@@ -71,88 +86,111 @@ const UserInfo = () => {
...
+        onSelect={onSelect}
+        onAdd={onAdd}
+        onClose={onClose}
+      >
+        {users.map((user, index) => (
+          <Tab
+            aria-label={`User ${user.name} tab`}
+            key={user.name}
+            eventKey={index}
+            title={<TabTitleText>{user.name || 'New user'}</TabTitleText>}
```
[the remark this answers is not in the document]
regexowl: I can't index :flushed:  thanks!

src/Components/CreateImageWizard/steps/Users/components/RemoveUserModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    setActiveTabKey(nextTabIndex);
+    setIndex(nextTabIndex);
+    dispatch(removeUser(tabIndex));
+    setShowRemoveUserModal(!isOpen);
+  };
+
+  return (
+    <Modal
+      title="Remove user?"
```
avitova, line 52: I think we might want to add name of the user to be deleted here.

src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
```diff
@@ -27,17 +32,29 @@ import { isUserGroupValid } from '../../../validators';
...
+  };
+
+  const onAdd = () => {
+    setActiveTabKey(users.length);
+    setIndex(index + 1);
+    dispatch(addUser());
+  };
+
+  const onClose = (_event: React.MouseEvent, tabIndex: number) => {
+    setShowRemoveUserModal(true);
```
avitova, line 55: I was wondering. In case no values are on input - name, password, ssh key, and groups are empty - do we want to remove them without the dialog? Just a thought for improvement.

avitova (dismissed):
1) If I create a user, there is immediately  "Invalid password" in errors structure. I do not know if we mind, since it can be empty, and there is different behaviour for ssh keys. I probably would fix it.

2) If you create a user, add ssh key that is invalid, remove a different user, the ssh key error disappears.

I ran into weird bugs, but could not reproduce it, so I guess that's it for now. :eyes:

src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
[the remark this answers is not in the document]
regexowl: Yeah, that would be super useful, opened an issue to address this in a follow up https://github.com/osbuild/image-builder-frontend/issues/3097

regexowl:
Added two commits with changes. The only not resolved issue is the ssh key validation one - it seems to be connected to handling `isPristine` in validated input, same as in https://github.com/osbuild/image-builder-frontend/pull/3072#issuecomment-2794130788

The error stays in the `errors` object and  if you click away from the field after the error disappeared, the helper text reappears again.

src/Components/CreateImageWizard/steps/Users/components/RemoveUserModal.tsx
regexowl: Good point, added!

src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
regexowl: Makes sense, updated :+1:

src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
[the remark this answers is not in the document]
ksiekl: 🧮 it's always a off-by-one problem :D

avitova:
Ah, sorry, some more nitpicks I did not notice before.
1) Should it be possible to have more users with the same name?
2) When adding new tab when the navigation is full, I would expect that the navigation shifts so that we can see the new tab:
https://github.com/user-attachments/assets/53c64e6b-72ae-477b-8003-890f9f11b89a

regexowl:
Wasn't that hard actually, focusing on newly added user should be working now :sweat_smile:

regexowl:
> Looks good, I am good here. Suggestion for future: I would probably disable the "+" button to add new users as long as there is an empty user. 🤔 If you like that idea, we can do that in a followup, I do not want to block this.

Agree this would be nice, might require some digging in Patternfly though as disabling the add button doesn't seem to be something we have access to by default. Opened an issue so we can investigate later https://github.com/osbuild/image-builder-frontend/issues/3114

lucasgarfield (approved):
Approving and merging since @avitova acked it earlier.

PR #3092, opened 2025-04-10
Promote WSL to production stable (HMS-5950)

ochosi:
This simple pull request promotes WSL images to production stable.

It also adds a Popover to reflect that this image type is currently not supported by Red Hat. The Popover and text reflect the UX mockup.

image

/jira-epic HMS-5949

JIRA: [HMS-5950](https://issues.redhat.com/browse/HMS-5950)

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
@@ -29,17 +29,13 @@ import {
   selectImageTypes,
 } from '../../../../store/wizardSlice';
 import isRhel from '../../../../Utilities/isRhel';
-import {
-  useFlag,
-  useGetEnvironment,
-} from '../../../../Utilities/useGetEnvironment';
+import { useGetEnvironment } from '../../../../Utilities/useGetEnvironment';
 
 const TargetEnvironment = () => {
   const arch = useAppSelector(selectArchitecture);
   const environments = useAppSelector(selectImageTypes);
   const distribution = useAppSelector(selectDistribution);
   const { isFedoraEnv } = useGetEnvironment();
-  const wslFlag = useFlag('image-builder.wsl.enabled');
 
   const { data } = useGetArchitecturesQuery(
     {
@@ -353,7 +349,7 @@ const TargetEnvironment = () => {
             data-testid="checkbox-image-installer"
           />
         )}
-        {supportedEnvironments?.includes('wsl') && wslFlag && (
+        {supportedEnvironments?.includes('wsl') && (
           <Checkbox
             label={
               <>
--- a/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
@@ -359,9 +359,7 @@ const TargetEnvironment = () => {
                   position="right"
                   headerContent={
                     <TextContent>
-                      <Text>
-                        WSL is not officially supported by Red Hat.
-                      </Text>
+                      <Text>WSL is not officially supported by Red Hat</Text>
                     </TextContent>
                   }
                   bodyContent={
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
```diff
@@ -353,9 +349,48 @@ const TargetEnvironment = () => {
...
+                <Popover
+                  maxWidth="30rem"
+                  position="right"
+                  bodyContent={
+                    <TextContent>
+                      <Text component={TextVariants.h5}>
+                        WSL is not officially supported by Red Hat.
+                      </Text>
+                      <Text>
+                        Unfortunately Windows Subsystem for Linux is currently not supported by Red Hat.
```
regexowl, line 366: Looks great! Tiny nitpick, I'd personally remove "unfortunately" to avoid implying a stance towards Red Hat not supporting WSL and to keep things neutral.

But if the text already went through someone from copy or UX then I'm happy to merge as is.

ochosi: This was Shayna's suggestion (see the Figma link in the Epic), but I agree, let's get some more eyes on the actual text.

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
```diff
@@ -353,9 +349,49 @@ const TargetEnvironment = () => {
...
           />
         )}
-        {supportedEnvironments?.includes('wsl') && wslFlag && (
+        {supportedEnvironments?.includes('wsl') && (
           <Checkbox
-            label="WSL - Windows Subsystem for Linux (.tar.gz)"
+            label={
+              <>
+                WSL - Windows Subsystem for Linux (.tar.gz)
+                <Popover
```
marusak, line 357: Popover has header, content and footer. If you move title to header you don't need to specify text variant and don't need to do `<br />` if you move `Learn more` into footer. https://www.patternfly.org/components/popover/

But if you use Popovers like this everywhere then I guess let's keep it consistent :)

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
```diff
@@ -353,9 +349,49 @@ const TargetEnvironment = () => {
...
+                          href="https://access.redhat.com/solutions/6338661"
+                        >
+                          Learn more
+                        </Button>
+                      </Text>
+                    </TextContent>
+                  }
+                >
+                  <Button
+                    className="pf-v5-u-pl-sm pf-v5-u-pt-0 pf-v5-u-pb-0"
```
marusak, line 385: Is this necessary? These are from the utilities lib?

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
ochosi: And this is with `header` and `footer`
image

It looks slightly different with the footer, but the code is cleaner and it also looks nicely structured. I'd be in favor of this version.

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
marusak: fair enough! Then I guess they are needed - you use them in other places as well, thanks for checking it out!

src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
regexowl: I believe you can even do 
```typescript
    headerContent="WSL is not officially supported by Red Hat."
    bodyContent="
          Unfortunately Windows Subsystem for Linux is currently
          not supported by Red Hat."
```
should work the same

ochosi: Yeah I saw that in PatternFly's docs but was unsure if the `<Text>` elements were a convention we were following.

regexowl: Gotcha, I don't think we have a specific convention set, but it indeed looks like we're currently using the text component quite often. It allows to split the content into paragraphs without `<br />`, so I guess that's why.

ochosi: Ok, I left the text component in but amended the rest. Let me know if you'd like more fixups!

regexowl (approved):
Thank you, looks great! :heart:

PR #3094, opened 2025-04-10
ImagesTable: Fix AWS regions indicator (off by one)

ochosi:
The AWS regions indicator that is currently baked into each row containing an AWS image is currently off by one due to incorrect parentheses in the code.
This is now fixed, so it displays '(1)' when there is one region (which is the default, us-east-1) instead of '(0)' which was both wrong and confusing.

Before:
image

After:
image


regexowl (approved):
Very nice catch, thank you

PR #3098, opened 2025-04-11
.github: new stage user on each playwright run

croissanne:
SSIA

---

Did a run which retains all videos, as you can see a new user is used: https://app.currents.dev/i/wGpgxSmts5YdAWYi/test/b72ba50710186a3d173d-736ee36bae07ab26c41c?testExplorerTab=attempts


tkoscieln (approved):
I'm a bit worried depending on Ethel API could introduce flakiness into the tests, but let's see how stable it is, we can always try different approach later.

PR #3103, opened 2025-04-14
on-prem: cloud provider config modal (COMPOSER-2488)

kingsleyzissou:
Add a modal/page to configure the `osbuild-worker` with the required configs for the various cloud providers

JIRA: COMPOSER-2488

### DIFF
--- a/playwright/test.spec.ts
+++ b/playwright/test.spec.ts
@@ -1,3 +1,6 @@
+import { readFileSync } from 'node:fs';
+
+import TOML from '@ltd/j-toml';
 import { expect, test } from '@playwright/test';
 import { v4 as uuidv4 } from 'uuid';
 
@@ -164,5 +167,75 @@ test.describe.serial('test', () => {
     await frame.getByTestId('blueprint-action-menu-toggle').click();
     await frame.getByRole('menuitem', { name: 'Delete blueprint' }).click();
     await frame.getByRole('button', { name: 'Delete' }).click();
+  });
+
+  test('cockpit worker config', async ({ page }) => {
+    if (isHosted()) {
+      return;
+    }
+
+    await ensureAuthenticated(page);
+    await closePopupsIfExist(page);
+    // Navigate to IB landing page and get the frame
+    await navigateToLandingPage(page);
+    await page.goto('/cockpit-image-builder');
+    const frame = ibFrame(page);
+
+    const header = frame.getByText('Configure AWS Uploads');
+    if (!(await header.isVisible())) {
+      await frame
+        .getByRole('button', { name: 'Configure Cloud Providers' })
+        .click();
+      await expect(header).toBeVisible();
+    }
+
+    const bucket = 'cockpit-ib-playwright-bucket';
+    const credentials = '/test/credentials';
+    const switchInput = frame.locator('#aws-config-switch');
+    await expect(switchInput).toBeVisible();
+
+    // the test is a little bit flaky, if it fails the first time
+    // while setting the configs, the second time the config should
+    // already be loaded and visible, if it is go back to the landing page
+    // https://github.com/osbuild/image-builder-frontend/issues/3429
+    if (await switchInput.isChecked()) {
+      await frame.getByRole('button', { name: 'Cancel' }).click();
+      await expect(
+        frame.getByRole('heading', { name: 'All images' })
+      ).toBeVisible();
+    } else {
+      const switchToggle = frame.locator('.pf-v6-c-switch');
+      await switchToggle.click();
+
+      await frame
+        .getByPlaceholder('AWS bucket')
+        // this doesn't need to exist, we're just testing that
+        // the form works as expected
+        .fill(bucket);
+      await frame.getByPlaceholder('Path to AWS credentials').fill(credentials);
+      await frame.getByRole('button', { name: 'Submit' }).click();
+      await expect(
+        frame.getByRole('heading', { name: 'All images' })
+      ).toBeVisible();
+    }
+
+    const config = readFileSync('/etc/osbuild-worker/osbuild-worker.toml');
+    // this is for testing, the field `aws` should exist
+    // eslint-disable-next-line
+    const parsed = TOML.parse(config) as any;
+    expect(parsed.aws?.bucket).toBe(bucket);
+    expect(parsed.aws?.credentials).toBe(credentials);
+
+    await frame
+      .getByRole('button', { name: 'Configure Cloud Providers' })
+      .click();
+    await expect(header).toBeVisible();
+    await expect(frame.locator('#aws-config-switch')).toBeChecked();
+
+    await expect(frame.getByPlaceholder('AWS bucket')).toHaveValue(bucket);
+    await expect(frame.getByPlaceholder('Path to AWS credentials')).toHaveValue(
+      credentials
+    );
+    await frame.getByRole('button', { name: 'Cancel' }).click();
   });
 });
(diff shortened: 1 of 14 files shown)


kingsleyzissou:
Last step is to restart the worker.

package.json
```diff
@@ -7,6 +7,7 @@
     "npm": ">=7.0.0"
   },
   "dependencies": {
+    "@ltd/j-toml": "^1.38.0",
```
kingsleyzissou, line 10: I've just added this one in addition to the other package that we have. The existing `toml` package only parses TOML, it isn't bi-directional.

I could replace the TOML package completely, but maybe that would be better for a follow up PR

src/store/cockpit/types.ts
```diff
@@ -18,3 +18,20 @@ export type Package = {
   version: string;
   release: string;
 };
+
+export type AWSWorkerConfig =
+  | {
+      bucket?: string | undefined;
+      region?: string | undefined;
```
kingsleyzissou, line 25: The region isn't actually used at all by the osbuild-worker, but it makes sense to save this info here for the on-prem version

src/Components/CloudProviderConfig/CloudProviderConfig.tsx
```diff
@@ -12,6 +25,32 @@ import { useAppDispatch } from '../../store/hooks';
...
+  onClose: MouseEventHandler<HTMLButtonElement>;
+}) => {
+  return (
+    <EmptyState variant={EmptyStateVariant.xl}>
+      <EmptyStateIcon icon={ExclamationIcon} color="#C9190B" />
+      <Title headingLevel="h4" size="lg">
+        Error
+      </Title>
+      <EmptyStateBody>
+        There was an error reading the `/etc/osbuild-worker/osbuild-worker.toml`
```
kingsleyzissou, line 46: Maybe the copy here could be improved, any suggestions?

kingsleyzissou:
I added a note to the the guides about configuring the worker globally for AWS uploads, since this PR essentially follows that pattern.

https://github.com/osbuild/osbuild.github.io/pull/159

src/store/cockpit/cockpitApi.ts
```diff
@@ -584,6 +585,92 @@ export const cockpitApi = contentSourcesApi.injectEndpoints({
...
+              [
+                'systemctl',
+                'start',
+                'osbuild-composer.socket',
+                'osbuild-composer',
+              ],
+              {
+                superuser: 'require',
+              }
+            );
```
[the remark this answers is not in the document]
kingsleyzissou: I'm on the fence about this. We do sort of need to restart the `osbuild-local-worker.socket` but this is taken care of by restarting the service and the `osbuild-composer.socket`. @croissanne do you have any opinions on this?

kingsleyzissou:
https://github.com/user-attachments/assets/09605004-b52a-4eeb-8948-1b2e626a3f4e

avitova:
Makes sense. Thanks!

kingsleyzissou:
Converting to draft since this might change if we use `image-builder-cli` as the backend.

kingsleyzissou:
Should be ready now, I fixed the bug where the input wasn't getting updated.
I've also removed the region field from the config, I'll do that in a separate PR.

If we do end up using `ibcli` we can just drop this at a later point.

src/store/cockpit/cockpitApi.ts
```diff
@@ -652,6 +652,23 @@ export const cockpitApi = contentSourcesApi.injectEndpoints({
...
 
+            // TODO: maybe give a warning that we need admin access?
+            await cockpit.spawn(['systemctl', 'stop', 'osbuild-*'], {
+              superuser: 'require',
+            });
+
+            await cockpit.spawn(
+              [
+                'systemctl',
+                'start',
```
croissanne, line 663: `restart`

src/store/cockpit/cockpitApi.ts
```diff
@@ -652,6 +652,23 @@ export const cockpitApi = contentSourcesApi.injectEndpoints({
...
+                'osbuild-composer.socket',
+                'osbuild-composer',
```
croissanne, line 665: I'm not even sure if we need to restart composer, just the worker(s) should be fine.

`osbuild-worker@*.service`

croissanne (changes requested):
Mostly looks good to me, the restarting is broken though I think.

src/store/cockpit/cockpitApi.ts
kingsleyzissou: Okay cool, let me try that out

croissanne (changes requested):
LGTM, works fine. 

One thing that's missing is maybe a wee playwright test, where it goes through the config wizard, and just compares the contents of /etc/osbuild-worker/osbuild-worker.toml? And maybe if the worker restarted (though can't immediately think of a way to do that).

src/store/cockpit/cockpitApi.ts
croissanne: ah ok, ye let's just restart all of them. lgtm thanks!

kingsleyzissou:
> LGTM, works fine.
> 
> One thing that's missing is maybe a wee playwright test, where it goes through the config wizard, and just compares the contents of /etc/osbuild-worker/osbuild-worker.toml? And maybe if the worker restarted (though can't immediately think of a way to do that).

Whoops, somehow missed this. Yeah that seems like a good idea!

kingsleyzissou:
> and just compares the contents of /etc/osbuild-worker/osbuild-worker.toml

Turns out this was _very_ difficult to do. I couldn't figure out how to access `cockpit` from the playwright tests so it made checking the `worker.toml` hard

croissanne:
> > and just compares the contents of /etc/osbuild-worker/osbuild-worker.toml
> 
> Turns out this was _very_ difficult to do. I couldn't figure out how to access `cockpit` from the playwright tests so it made checking the `worker.toml` hard

ooh ok, I thought just using `fs` and bindmounting the /etc directory into the playwright container would be ok. But fair enough.

playwright/test.spec.ts
```diff
@@ -165,4 +165,66 @@ test.describe.serial('test', () => {
...
+
+    const bucket = 'cockpit-ib-playwright-bucket';
+    const credentials = '/test/credentials';
+    const switchInput = frame.locator('#aws-config-switch');
+    await expect(switchInput).toBeVisible();
+
+    // the test is a little bit flaky, if it fails the first time
+    // while setting the configs, the second time the config should
+    // already be loaded and visible, if it is go back to the landing page
+    if (await switchInput.isChecked()) {
```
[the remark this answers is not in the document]
croissanne: Mind adding it to the comment itself? Otherwise lgtm!

kingsleyzissou:
> > ooh ok, I thought just using fs and bindmounting the /etc directory into the playwright container would work. But fair enough.
> 
> Uhh right that didn't even cross my mind. I can look into that in a follow up

Turns out it was as straightforward as that. I've added the `fs` check to this pr

playwright/test.spec.ts
```diff
@@ -165,4 +165,66 @@ test.describe.serial('test', () => {
...
+      .getByRole('button', { name: 'Configure Cloud Providers' })
+      .click();
+    await expect(header).toBeVisible();
+    await expect(frame.locator('#aws-config-switch')).toBeChecked();
+
+    await expect(frame.getByPlaceholder('AWS bucket')).toHaveValue(bucket);
+    await expect(frame.getByPlaceholder('Path to AWS credentials')).toHaveValue(
+      credentials
+    );
+    await frame.getByRole('button', { name: 'Cancel' }).click();
```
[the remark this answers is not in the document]
kingsleyzissou: I just assumed that node fs wasn't available in playwright :/

PR #3108, opened 2025-04-15
Wizard: fix a rerender loop in AzureResourceGroups

ondrejbudai:
resourceGroup is used in useEffect, so if assigned a new array to it every time this function is called, it would cause the useEffect to run every time, leading to an infinite loop. This is because useEffect uses Object.is to determine if the value has changed, and [] creates a new array every time. Thus, we use a constant empty array to avoid this issue.

Alternatively JSON.stringify(resourceGroups), or a deep comparison could be used as a useEffect dependency, but that feels like just a waste of resources.

Without this patch, I get this warning when running the basic test for upload to azure in `AzureTarget.test.tsx`:

```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```


regexowl (approved):
Great patch, thank you :heart:

PR #3110, opened 2025-04-15
ImagesTable/Toolbar: add alert with blueprint linter errors

croissanne:
Includes a fix button which should resolve the errors automatically.


src/test/mocks/handlers.js
```diff
@@ -211,6 +211,14 @@ export const handlers = [
     const id = params['id'];
     return HttpResponse.json({ id: id });
   }),
+  http.post(
+    `${IMAGE_BUILDER_API}/experimental/blueprints/:id/fixup`,
+    ({ params }) => {
+      const id = params['id'];
+      getMockBlueprintResponse(id).lint.errors = [];
```
croissanne, line 218: This is a bit of a hack. I'm not sure if this is allowed, to write to the mocked data. I'm not sure if there's an alternative either.

regexowl:
Code looks good, how could I test this? I've tried to create a blueprint with OpenSCAP profile, removed some mount points and packages, but can't see the alert.

croissanne:
> Code looks good, how could I test this? I've tried to create a blueprint with OpenSCAP profile, removed some mount points and packages, but can't see the alert.

oh it only works with compliance policies currently, sorry ._.

regexowl (approved):
Works amazingly! :wrench:

PR #3117, opened 2025-04-16
    ImagesTable/BlueprintsSideBar: Add Segment tracking (HMS-5989)

lucasgarfield:
Needed so the Intercom team can decide whether or not to display an onboarding video (first time user with no images or blueprints).

Verified to be working in Segment:
image


regexowl (approved):
Looks good, thanks!

PR #3118, opened 2025-04-16
Wizard/Oscap: Add analytics (HMS-5990)

lucasgarfield:
The Intercom team has some ideas about helping new users determine which security profile is right for them. Tracking differentiates between "vanilla" OpenSCAP and Compliance so that any messages from Intercom can be targeted.


regexowl:
Looks good, @lucasgarfield can you please resolve the conflict?

regexowl:
Looks like some files sneaked their way to the PR @lucasgarfield

regexowl (approved):

PR #3121, opened 2025-04-17
Wizard: Update periods at the end of lines

regexowl:
Removing/adding some periods for the sake of consistency.

I believe the description of the step should end with a period, helped texts and options should not (at least that's the prevailing pattern in the Wizard as of now).


croissanne (approved):

lucasgarfield (approved):
Consistency is key!

Eventually it might be nice to take the Patternfly guidelines for copy and put them as instructions to our AI reviewer... :thinking:

PR #3122, opened 2025-04-17
Wizard: support tailored customizations

croissanne:
This splits the policy and profile selectors as they're drifting apart.

Instead of querying the customizations attached to the profile, query
the customizations attached to the policy, as these take into account
tailoring. As a result unnecessary customizations won't be added.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
@@ -59,7 +59,7 @@ const ComplianceSelectOption = ({ policy }: ComplianceSelectOptionPropType) => {
   ): ComplianceSelectOptionValueType => ({
     policyID,
     profileID,
-    toString: () => title,
+    toString: () => title || 'None',
   });
 
   const descr = (
@@ -89,7 +89,6 @@ const PolicySelector = () => {
   const hasWslTargetOnly = useHasSpecificTargetOnly('wsl');
   const dispatch = useAppDispatch();
   const [isOpen, setIsOpen] = useState(false);
-  const [selected, setSelected] = useState<string>('None');
 
   const {
     data: policies,
@@ -231,13 +230,11 @@ const PolicySelector = () => {
           handlePackages(oldOscapPackages, newOscapPackages);
           handleServices(response.services);
           handleKernelAppend(response.kernel?.append);
-          const compl = selection as ComplianceSelectOptionValueType;
-          setSelected(compl.toString());
           dispatch(
             changeCompliance({
-              profileID: compl.profileID,
-              policyID: compl.policyID,
-              policyTitle: compl.toString(),
+              profileID: selection.profileID,
+              policyID: selection.policyID,
+              policyTitle: selection.toString(),
             })
           );
         });
@@ -290,7 +287,7 @@ const PolicySelector = () => {
         } as React.CSSProperties
       }
     >
-      {selected}
+      {policyTitle || 'Select a policy'}
     </MenuToggle>
   );
 
@@ -301,7 +298,7 @@ const PolicySelector = () => {
         isOpen={isOpen}
         onSelect={handleSelect}
         onOpenChange={handleToggle}
-        selected={selected}
+        selected={policyID}
         toggle={toggleCompliance}
         shouldFocusFirstItemOnOpen={false}
       >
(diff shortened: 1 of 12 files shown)


croissanne:
https://github.com/osbuild/image-builder-frontend/pull/3134 will make this easier

src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
```diff
@@ -0,0 +1,314 @@
...
+          width: '200px',
+        } as React.CSSProperties
+      }
+    >
+      {selected}
+    </MenuToggle>
+  );
+
+  return (
+    <FormGroup data-testid="profiles-form-group" label={<>Policy</>}>
```
regexowl, line 298:
```suggestion
    <FormGroup data-testid="profiles-form-group" label="Policy">
```

src/Components/CreateImageWizard/steps/Oscap/components/ProfileSelector.tsx
```diff
@@ -481,86 +350,52 @@ const ProfileSelector = () => {
...
-        </Select>
-      )}
+          {isSuccess && selectOptions.length === 0 && (
+            <SelectOption isDisabled>
+              {`No results found for "${filterValue}"`}
+            </SelectOption>
+          )}
+        </SelectList>
+      </Select>
+      )
```
regexowl, line 398: image

The closing bracket will need to be removed so it doesn't render after the select.

regexowl (commented):
Super happy about splitting policy and profile selectors tbh. Works great, I've tried to edit a policy and the changes showed in the Wizard correctly.

Two tiny nitpicks, but otherwise good on my side, thank you!

src/Components/CreateImageWizard/steps/Oscap/components/ProfileSelector.tsx
croissanne: I wondered what that was!

src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+  const complianceOptions = () => {
+    if (!policies || policies.data === undefined) {
+      return [];
+    }
+
+    const res = [
+      <SelectOption
+        key="compliance-none-option"
+        value={{ toString: () => 'None', compareTo: () => false }}
```
avitova, line 265: Why do we have this here? Is it not enough to not select any profile?

src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
```diff
@@ -0,0 +1,314 @@
...
+      const partition: Partition = {
+        mountpoint: filesystem.mountpoint,
+        min_size: size.toString(),
+        unit: unit as Units,
+        id: uuidv4(),
+      };
+      return partition;
+    });
+
+    if (newPartitions) {
```
avitova, line 188: What is the purpose of this? Even if newPartitions is empty, this will go through, I think.
And I do not think that map can return anything else than an array, or is that possible?

avitova (commented):
Not much to comment here, looks good, mostly. :)

src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
croissanne: The empty option is there to have a way to deselect a policy in case you already chose one. Similar to the ProfileSelector.

src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
croissanne: ah you're right, this might need fixing in both the profile and policy selector.

src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
avitova: But there is a way to do that now, using the close button. :thinking: I was just a bit confused, because we do not use this in different steps (like Timezone, Locale (Keyboard), etc.). But this is not a topic of this PR, so I am happy to just leave this as is, and open this conversation elsewhere. :)

https://github.com/user-attachments/assets/4deb4e18-cdc5-4c7e-b551-139660552574

regexowl: There is an inconsistency between policy and profile selectors - default value shown, rendering clear button - I'd leave it for a follow up as it's not entirely in the scope of this PR. But we should still address it at some point.

regexowl (approved):
LGTM!

PR #3127, opened 2025-04-22
Wizard: edit compliance according to updated mockups (HMS-5963)

ksiekl:
Changes the compliance step based on most recent mockups

JIRA: [HMS-6000](https://issues.redhat.com/browse/HMS-6000)

compliance
image
image
image
image


src/test/Components/CreateImageWizard/steps/Oscap/Compliance.test.tsx
```diff
@@ -39,7 +39,10 @@ const goToComplianceStep = async () => {
   await clickRegisterLater();
   await clickNext(); // Compliance
   await screen.findByRole('heading', { name: /Compliance/ });
-  const button = await screen.findByLabelText('Insights compliance');
+  const button = await screen.findByRole('button', {
+    name: /Compliance policies/,
+  });
+  // const button = await screen.findByLabelText('Insights compliance');
```
avitova, line 45: Let's delete this commented line completely.

src/Components/CreateImageWizard/steps/Oscap/index.tsx
```diff
@@ -101,77 +100,56 @@ const OscapContent = () => {
...
-            variant="link"
-            icon={<ExternalLinkAltIcon />}
-            iconPosition="right"
-            isInline
-            href={COMPLIANCE_AND_VULN_SCANNING_URL}
-          >
-            Documentation
-          </Button>
-        </Text>
+      {policies?.data?.length === 0 && complianceType === 'compliance' && (
```
avitova, line 140: Just a thought, but the alert will not be shown in case policies or policies.data is undefined. This will probably happen only in case something is incorrect, and I think we want this, but I just wanted to mention it, in case it's not the intention.

src/Components/CreateImageWizard/steps/Oscap/components/ProfileSelector.tsx
```diff
@@ -490,7 +495,7 @@ const ProfileSelector = () => {
       isDisabled={isFetchingPolicies}
       style={
         {
-          width: '200px',
+          width: '100%',
```
avitova, line 498: THANKS GOD, I was getting soooooo triggered by this. Thank you for fixing this. xd

avitova (changes requested):
Let's delete the commented line in tests, otherwise I think this looks good. :)

src/test/Components/CreateImageWizard/steps/Oscap/Compliance.test.tsx
ksiekl: Oh right, forgot abt it! thanks! :heart_decoration: 
image

src/Components/CreateImageWizard/steps/Oscap/components/ProfileSelector.tsx
ksiekl: She got a glowup :nail_care:

src/Components/CreateImageWizard/steps/Oscap/components/OpenSCAPFGLabel.tsx
```diff
@@ -1,33 +1,7 @@
...
-          variant="plain"
-          aria-label="About OpenSCAP"
-          isInline
-          className="pf-v5-u-pl-sm pf-v5-u-pt-0 pf-v5-u-pb-0 pf-v5-u-pr-0"
-        >
-          <HelpIcon />
-        </Button>
-      </Popover>
-    </>
-  );
```
regexowl, line 30: I wonder if we should keep the popover even though it's not visible in the mocks. If not we can completely remove this component and just add the "Profile" as a label to the `<FormGroup>` component.

src/Components/CreateImageWizard/steps/Oscap/components/OscapProfileInformation.tsx
```diff
@@ -115,10 +99,48 @@ export const OscapProfileInformation = ({
                 component={TextListItemVariants.dt}
                 className="pf-v5-u-min-width"
               >
-                Operating system:
+                Reference ID:
+              </TextListItem>
+              <TextListItem
+                data-testid="oscap-profile-info-ref-id"
```
regexowl, line 105: There's a plan to stop using `data-testid` and `ouiaId` in the code base and testing. It's an anti-pattern that allows us to lean on arbitrary locators instead of "what's really there". Could you please remove it?

src/Components/CreateImageWizard/steps/Oscap/index.tsx
```diff
@@ -101,78 +100,59 @@ const OscapContent = () => {
...
-            component="a"
-            target="_blank"
-            variant="link"
-            icon={<ExternalLinkAltIcon />}
-            iconPosition="right"
-            isInline
-            href={COMPLIANCE_AND_VULN_SCANNING_URL}
-          >
-            Documentation
-          </Button>
```
regexowl, line 173: We're losing the compliance and documentation links in the process. I got that they're not in the mock then, but still wonder if we shouldn't keep them or give the users another option to quickly navigate to compliance (there's a link in the alert, but that renders only when there are no policies defined if I understand correctly)

@lucasgarfield any ideas?

regexowl (commented):
Looks great, left some small nitpicks. One note on merging, this will have a decent overlap with @croissanne's https://github.com/osbuild/image-builder-frontend/pull/3122, so we should think about the order we're merging in and take into account that there will be most probably conflicts.

src/Components/CreateImageWizard/steps/Oscap/components/OpenSCAPFGLabel.tsx
ksiekl: Yeah, good point

src/Components/CreateImageWizard/steps/Oscap/components/OscapProfileInformation.tsx
ksiekl: Oh well, the tests are bothered now...

regexowl: Oh, sorry overlooked that it's not a new `data-testid`, but it's already used in the test. Feel free to add it back. Sorry about that :flushed: (we should remove the testids, but that's out of scope for this PR)

avitova (approved):
LGTM:)

src/Components/CreateImageWizard/steps/Oscap/components/OscapProfileInformation.tsx
ksiekl: No problem! I put it back :)

regexowl (approved):
Looks great, thank you! :rocket:

PR #3130, opened 2025-04-23
test: Remove WSL flag from unleash test mock (HMS-6002)

regexowl:
The WSL flag was removed from the codebase, this removes it from the test mock as well.

JIRA: [HMS-6002](https://issues.redhat.com/browse/HMS-6002)


ochosi (approved):
Thanks for correcting my mistake!

PR #3131, opened 2025-04-23
src: Remove `pkgRecs` flag

regexowl:
Package recommendations are fully deployed to both stage and production. We can remove the flag from the codebase.


ochosi (approved):
Verified by toggling 'preview' and checking that package recommendations don't dis/appear.

PR #3132, opened 2025-04-23
src: Remove `firstboot` flag

regexowl:
Firstboot also seems to be released fully to both stage and prod. The flag is not needed anymore.


ochosi (approved):
Verified by toggling 'preview' and checking that firstboot doesn't dis/appear.

regexowl:
Looking at the failing test - should first boot be available for Fedora service?

src/test/FedoraServices.test.tsx
```diff
@@ -27,6 +27,7 @@ const goToDetailsStep = async () => {
   await clickNext(); // Kernel
   await clickNext(); // Firewall
   await clickNext(); // Services
+  await clickNext(); // First boot
```
regexowl, line 30: Assuming the First boot step should be available in the Fedora service as well. If it's an incorrect assumption let me know and I'll switch this for hiding the step for Fedora env instead.

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -666,7 +665,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 key="wizard-first-boot"
                 navItem={customStatusNavItem}
                 status={firstBootValidation.disabledNext ? 'error' : 'default'}
-                isHidden={!isFirstBootEnabled}
+                isHidden={!!process.env.IS_ON_PREMISE || isFedoraEnv}
```
regexowl, line 668: Based this on the tests in the end, assuming the first boot step shouldn't be visible in on-prem or Fedora service.

avitova (commented):
Adding my 2 cents, I think we can release this feature.

ochosi (approved):
LGTM though we should assess with @kingsleyzissou if there are any blockers not to enable first boot on prem, too.

kingsleyzissou:
> LGTM though we should assess with @kingsleyzissou if there are any blockers not to enable first boot on prem, too.

I think theoretically it should work on-prem. But maybe we could keep it disabled not to block this PR and we can look at it properly for on-prem

kingsleyzissou (approved):

PR #3136, opened 2025-04-23
Wizard: Use SearchInput component for packages (HMS-6011)

### LINKED ISSUE #3133: Inconsistent search box UIs
We currently have slightly inconsistent search boxes in the wizard. Compare the repos and packages steps:

Image

Image

One has the search box embedded in the input field, the other one has a separator/box for the search glass icon.

ochosi:
As a small simplification and also to make the UX more consistent, let's use a SearchInput widget over the custom InputGroup structure. This means the search boxes will also be consistent between the repository and package search pages.
Fixes #3133

JIRA: [HMS-6011](https://issues.redhat.com/browse/HMS-6011)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -15,19 +15,16 @@ import {
   EmptyStateIcon,
   EmptyStateVariant,
   Icon,
-  InputGroup,
-  InputGroupItem,
-  InputGroupText,
   Pagination,
   PaginationVariant,
   Popover,
+  SearchInput,
   Spinner,
   Stack,
   Tab,
   Tabs,
   TabTitleText,
   Text,
-  TextInput,
   ToggleGroup,
   ToggleGroupItem,
   Toolbar,
@@ -40,7 +37,6 @@ import {
   HelpIcon,
   OptimizeIcon,
   SearchIcon,
-  TimesIcon,
 } from '@patternfly/react-icons';
 import {
   ExpandableRowContent,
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizard/steps/Packages/Packages.tsx
```diff
@@ -1251,39 +1247,22 @@ const Packages = () => {
...
-              </InputGroup>
+              <SearchInput
+                data-ouia-component-id="packages-search-input"
+                type="text"
+                validated={
+                  debouncedSearchTermLengthOf1 &&
+                  !debouncedSearchTermIsGroup
+                    ? 'error'
+                    : 'default'
+                }
```
regexowl, line 1258: This is a great catch, completely agree with switching the components here. :heart:  The only prop that doesn't seem to be supported is `validated` (which might have been the reason why was the component written in a different way before). 

To be perfectly honest, I wasn't 100% sold on screaming at the user if the searchTerm has just one character anyway. We could just throw the entire `debouncedSearchTermLengthOf1` maybe? :shrug: Then we would have no need for the `validated` anymore as we don't check length for groups and there's not other validation in place.

ochosi: I completely agree. Also the state machine of the debounce and all is completely broken. If you first search for something and then gradually remove characters from the search term until you're down to 1 it shows an error message **and** the results :woozy_face: 

Also, the validator still seems to work, despite the error. Here's a screenshot of the code on this branch
image

ochosi:
If we want to retain the validation and validation styles we could also pivot to using [TextInputGroup](https://www.patternfly.org/components/text-input-group), which would allow for a very consistent visual style to SearchInput.

src/Components/CreateImageWizard/steps/Packages/Packages.tsx
regexowl: Oof, we should do something about this. 

image

Anyway - looks like the search doesn't start until there are at least two characters, which might be an intentional limitation on content-services side. Meaning we should inform user about it.

Originally we just had the "The search value is too short" state if I recall correctly. The input field validation was added later. So the options would probably be:
- replacing the previous components with `SearchInput`, dropping input validation and relying on the state only as "incorrect" input doesn't mean invalid values being added to the state or possible build failures
- or go with the `TextInputGroup`, leaving the validation for the input as currently is


I'm slightly leaning towards the first option, but both are good solutions I'd say.

regexowl: Opened an issue for the incorrect state rendering: https://github.com/osbuild/image-builder-frontend/issues/3139

ochosi: I have updated the PR to reflect option 1, but am happy to rework it with `TextInputGroup` if you or @lucasgarfield feel this is a better path.

src/Components/CreateImageWizard/steps/Packages/Packages.tsx
```diff
@@ -1248,39 +1244,16 @@ const Packages = () => {
...
-                      aria-label="clear-package-search"
-                      onClick={handleClear}
-                      icon={<TimesIcon />}
-                      ouiaId="clear-package-search-button"
-                    />
-                  </InputGroupItem>
-                )}
-              </InputGroup>
+              <SearchInput
+                data-ouia-component-id="packages-search-input"
```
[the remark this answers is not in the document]
ochosi: I'm happy to drop that property in this PR if you prefer that. Otherwise, should we do this in a separate PR as a more concerted effort across the whole codebase?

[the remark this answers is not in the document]
regexowl: Good point, we shouldn't need those no more. Opened a PR: https://github.com/osbuild/image-builder-frontend/pull/3144

We could similarly start getting rid of `data-testid`s as well, since those are also an anti-pattern. But that will be a bit more involved as tests will need to be updated.

regexowl (approved):
I'd say looks great! Thank you :heart:

PR #3140, opened 2025-04-24
Wizard: Fix 'too short' state in Packages step

### LINKED ISSUE #3139: Wizard/Packages: State for "Too short" seems unreliable
How to reproduce:
1. go to Packages step
2. enter "a" in the search input -> you should see "The search value is too short" state
3. enter one more "a" -> you should see search results
4. remove search input
5. enter "s" only

Current behaviour:
The "The search value is too short" state is not rendered, instead the previously found results are there.

Image

Expected behaviour:
The "The search value is too short" state is rendered.

ochosi:
Currently, when a search term is entered and then reduced to a single character in the Wizard's packages step, the previous results are still shown.
What is expected is that the empty state hinting at the search term being too short would be shown.

The condition in bodyContent was wrong, because it assumed that the list of results would be 0 when in fact it wasn't being cleared.

Fixes #3139

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -1166,10 +1166,7 @@ const Packages = () => {
 
   const bodyContent = useMemo(() => {
     switch (true) {
-      case debouncedSearchTermLengthOf1 &&
-        !debouncedSearchTermIsGroup &&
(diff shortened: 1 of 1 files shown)


regexowl (approved):
Works nicely, thank you!

ochosi:
Packages-search-too-short

Dang, I had just created the gif - too slow! :laughing:

PR #3126, opened 2025-04-22
tests/playwright: Add Locale customization test (HMS-4590)

avitova:
Adds a Locale customization test that creates a BP, edits BP, exports and imports it back and verifies the BP content.

PR #3141, opened 2025-04-24
tests/playwright: Add Timezone customization test

avitova:
Adds a Timezone customization test that creates a BP, edits BP, exports and imports it back and verifies the BP content.

Let's see if there are the same error as we have #3126 3126

### DIFF
--- a/playwright/Customizations/Timezone.spec.ts
+++ b/playwright/Customizations/Timezone.spec.ts
@@ -1,0 +1,118 @@
+import { expect } from '@playwright/test';
+import { v4 as uuidv4 } from 'uuid';
+
+import { test } from '../fixtures/cleanup';
+import { isHosted } from '../helpers/helpers';
+import { login } from '../helpers/login';
+import { navigateToOptionalSteps, ibFrame } from '../helpers/navHelpers';
+import {
+  registerLater,
+  fillInDetails,
+  createBlueprint,
+  fillInImageOutputGuest,
+  deleteBlueprint,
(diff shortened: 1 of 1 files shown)


playwright/Customizations/Timezone.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  await test.step('Create BP', async () => {
+    await createBlueprint(frame, blueprintName);
+  });
+
+  await test.step('Edit BP', async () => {
+    await frame.getByRole('button', { name: 'Edit blueprint' }).click();
+    await frame.getByLabel('Revisit Timezone step').click();
+    await expect(frame.getByText('Canada/Saskatchewan')).not.toBeVisible();
```
regexowl, line 77:
```suggestion
    await expect(frame.getByText('Canada/Saskatchewan')).toBeHidden();
```

regexowl (commented):
Looks good! One tiny nitpick

playwright/Customizations/Timezone.spec.ts
avitova: Thanks! Good catch!:)

regexowl (approved):
Thank you!

regexowl:
@tkoscieln The increased timeouts didn't seem to help. Could there be a problem with file parallelism, overwhelmed workers or something like that? But also I've increased timeouts for `expect` only, maybe we need to do more stuff? :shrug:

PR #3144, opened 2025-04-25
src: Remove `ouia-id`s (HMS-6012)

regexowl:
Since we're moving away from IQE the `ouia-id` atributes should be no longer needed. This removes them from the entire code base.

JIRA: [HMS-6012](https://issues.redhat.com/browse/HMS-6012)

PR #3145, opened 2025-04-25
Wizard: Update search input placeholders

regexowl:
The placeholders we tend to use are usually quite specific. "Type to search" felt quite vague for some reason. Also repositories had no placeholder, I believe "filter" is clearer than "search" here as we're not fetching any results, just filtering through the already existing list of repositories.


regexowl:
This might be very subjective, opened to any suggestions.

ochosi (dismissed):
Those both look nice from where I stand!

regexowl:
> Those both look nice from where I stand!

Thank you! Will rebase and update the tests when https://github.com/osbuild/image-builder-frontend/pull/3144 is merged. There might be another conflict otherwise.

ochosi (approved):

PR #3161, opened 2025-04-29
playwright: Increase timeout for `expect`

regexowl:
This increases timeout for `expect` from 30s to 50s.


ksiekl:
ou yea, much needed! (I could not resist making it a minute instead of 50 seconds, but I respect your unwavering determination to not round it)

ksiekl (approved):

PR #3162, opened 2025-04-29
Wizard: add Firewall test (HMS-5974)

ksiekl:
Playwright test to the firewall step in Wizard :mage:

### DIFF
--- a/playwright/Customizations/Firewall.spec.ts
+++ b/playwright/Customizations/Firewall.spec.ts
@@ -1,0 +1,139 @@
+import { expect } from '@playwright/test';
+import { v4 as uuidv4 } from 'uuid';
+
+import { test } from '../fixtures/cleanup';
+import { isHosted } from '../helpers/helpers';
+import { login } from '../helpers/login';
+import { navigateToOptionalSteps, ibFrame } from '../helpers/navHelpers';
+import {
+  registerLater,
+  fillInDetails,
+  createBlueprint,
+  fillInImageOutputGuest,
+  deleteBlueprint,
+  exportBlueprint,
+  importBlueprint,
+} from '../helpers/wizardHelpers';
+
+test('Create a blueprint with Firewall customization', async ({
+  page,
+  cleanup,
+}) => {
+  const blueprintName = 'test-' + uuidv4();
+
+  // Delete the blueprint after the run fixture
+  await cleanup.add(() => deleteBlueprint(page, blueprintName));
+
+  // Login, navigate to IB and get the frame
+  await login(page);
+  const frame = await ibFrame(page);
+
+  await test.step('Navigate to optional steps in Wizard', async () => {
+    await navigateToOptionalSteps(frame);
+    await registerLater(frame);
+  });
+
+  await test.step('Select and correctly fill the ports in Firewall step', async () => {
+    await frame.getByRole('button', { name: 'Firewall' }).click();
+    await frame.getByPlaceholder('Add ports').fill('80:tcp');
+    await frame.getByRole('button', { name: 'Add ports' }).click();
+    await expect(frame.getByText('80:tcp')).toBeVisible();
+  });
+
+  await test.step('Select and correctly fill the disabled services in Firewall step', async () => {
+    await frame
+      .getByPlaceholder('Add disabled service')
+      .fill('disabled_service');
+    await frame.getByRole('button', { name: 'Add disabled service' }).click();
+    await expect(frame.getByText('disabled_service')).toBeVisible();
+  });
+
+  await test.step('Select and correctly fill the enabled services in Firewall step', async () => {
+    await frame.getByPlaceholder('Add enabled service').fill('enabled_service');
+    await frame.getByRole('button', { name: 'Add enabled service' }).click();
+    await expect(frame.getByText('enabled_service')).toBeVisible();
+  });
+
+  await test.step('Select and incorrectly fill the ports in Firewall step', async () => {
+    await frame.getByPlaceholder('Add ports').fill('x');
+    await frame.getByRole('button', { name: 'Add ports' }).click();
+    await expect(frame.getByText('Invalid format.').nth(0)).toBeVisible();
+  });
+
+  await test.step('Select and incorrectly fill the disabled services in Firewall step', async () => {
+    await frame.getByPlaceholder('Add disabled service').fill('1');
+    await frame.getByRole('button', { name: 'Add disabled service' }).click();
+    await expect(frame.getByText('Invalid format.').nth(1)).toBeVisible();
+  });
+
+  await test.step('Select and incorrectly fill the enabled services in Firewall step', async () => {
+    await frame.getByPlaceholder('Add enabled service').fill('ťčš');
+    await frame.getByRole('button', { name: 'Add enabled service' }).click();
+    await expect(frame.getByText('Invalid format.').nth(2)).toBeVisible();
+  });
+
+  await test.step('Fill the BP details', async () => {
+    await frame.getByRole('button', { name: 'Review and finish' }).click();
+    await fillInDetails(frame, blueprintName);
+  });
+
+  await test.step('Create BP', async () => {
+    await createBlueprint(frame, blueprintName);
+  });
+
+  await test.step('Edit BP', async () => {
+    await frame.getByRole('button', { name: 'Edit blueprint' }).click();
+    await frame.getByLabel('Revisit Firewall step').click();
+
+    await frame.getByPlaceholder('Add ports').fill('90:tcp');
+    await frame.getByRole('button', { name: 'Add ports' }).click();
+    await frame.getByPlaceholder('Add disabled service').fill('x');
+    await frame.getByRole('button', { name: 'Add disabled service' }).click();
+    await frame.getByPlaceholder('Add enabled service').fill('y');
+    await frame.getByRole('button', { name: 'Add enabled service' }).click();
+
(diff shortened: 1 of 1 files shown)


playwright/Customizations/Firewall.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  await test.step('Select and correctly fill the ports in Firewall step', async () => {
+    await frame.getByRole('button', { name: 'Firewall' }).click();
+    await frame.getByPlaceholder('Add ports').fill('80:tcp');
+    await frame.getByPlaceholder('Add ports').press('Enter');
+  });
+
+  await test.step('Select and correctly fill the disabled services in Firewall step', async () => {
+    await frame
+      .getByPlaceholder('Add disabled service')
+      .fill('enabled_service');
```
avitova, line 45: Adding enabled_service into disabled service...lol. xd

avitova (commented):
Nice job! More teeests! Can you maybe also add something that tries to remove some of ports, disabled services/enabled services. Dunno, maybe in the Edit step or somewhere? So that we check more of the functionality.

playwright/Customizations/Firewall.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  await test.step('Navigate to optional steps in Wizard', async () => {
+    await navigateToOptionalSteps(frame);
+    await registerLater(frame);
+  });
+
+  await test.step('Select and correctly fill the ports in Firewall step', async () => {
+    await frame.getByRole('button', { name: 'Firewall' }).click();
+    await frame.getByPlaceholder('Add ports').fill('80:tcp');
+    await frame.getByPlaceholder('Add ports').press('Enter');
```
regexowl, line 39: Could you please add some kind of "final" check to the "correctly fill in..." tests? You're adding a port and then we could check if it got rendered and is visible on the page.

playwright/Customizations/Firewall.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    await frame
+      .getByPlaceholder('Add disabled service')
+      .fill('enabled_service');
+    await frame.getByPlaceholder('Add disabled service').press('Enter');
+  });
+
+  await test.step('Select and correctly fill the enabled services in Firewall step', async () => {
+    await frame
+      .getByPlaceholder('Add enabled service')
+      .fill('disabled_service');
```
regexowl, line 52: ... and disabled to enabled :broom:

playwright/Customizations/Firewall.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    await frame.getByPlaceholder('Add disabled service').press('Enter');
+    await expect(frame.getByText('Invalid format.').nth(1)).toBeVisible();
+  });
+
+  await test.step('Select and incorrectly fill the enabled services in Firewall step', async () => {
+    await frame.getByPlaceholder('Add enabled service').fill('ťčš');
+    await frame.getByPlaceholder('Add enabled service').press('Enter');
+    await expect(frame.getByText('Invalid format.').nth(2)).toBeVisible();
+  });
+  await test.step('Fill the BP details', async () => {
```
regexowl, line 73: Minor nitpick time!

```suggestion
  });

  await test.step('Fill the BP details', async () => {
```

playwright/Customizations/Firewall.spec.ts
tkoscieln: I second this, it would be cool if you would check that the stuff you fill in was actually added.

playwright/Customizations/Firewall.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    await importBlueprint(page, blueprintName);
+  });
+
+  await test.step('Review imported BP', async (step) => {
+    step.skip(!isHosted(), 'Importing is not available in the plugin');
+    await fillInImageOutputGuest(page);
+    await page.getByRole('button', { name: 'Firewall' }).click();
+
+    await expect(frame.getByText('80:tcp')).toBeVisible();
+    await expect(frame.getByText('x').nth(0)).toBeVisible();
```
tkoscieln, line 119: The expects here don't cover all the values that are visible in review (missing disabled services and second port value). Is this on purpose? :thinking:

playwright/Customizations/Firewall.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  await test.step('Select and correctly fill the enabled services in Firewall step', async () => {
+    await frame
+      .getByPlaceholder('Add enabled service')
+      .fill('disabled_service');
+    await frame.getByPlaceholder('Add enabled service').press('Enter');
+  });
+
+  await test.step('Select and incorrectly fill the ports in Firewall step', async () => {
+    await frame.getByPlaceholder('Add ports').fill('x');
+    await frame.getByPlaceholder('Add ports').press('Enter');
```
tkoscieln, line 58: Not really a fan of using key shortcuts to add the values. It works, but not fan of it. @regexowl @avitova thoughts? :thinking:

playwright/Customizations/Firewall.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  // Login, navigate to IB and get the frame
+  await login(page);
+  const frame = await ibFrame(page);
+
+  await test.step('Navigate to optional steps in Wizard', async () => {
+    await navigateToOptionalSteps(frame);
+    await registerLater(frame);
+  });
+
+  await test.step('Select and correctly fill the ports in Firewall step', async () => {
```
tkoscieln, line 36: Any reason for making these steps so atomized? Not complaining but curious why

playwright/Customizations/Firewall.spec.ts
regexowl: Specifically for `<LabelInput>` component we allow both button clicking and key down. I'd say we probably want to test both (as we do in unit tests now), but doesn't have to be necessarily here as the functionality should work across all `<LabelInput>`s in the code base (NTP servers, ports, firewall services inputs, systemd services inputs).

Maybe some "components test file / utilities" could be the right place?

tkoscieln: Oooh right. I agree, I'd move testing this somewhere else and use the buttons here.

playwright/Customizations/Firewall.spec.ts
regexowl: I'm usually a fan of short tests (=== read the name of failing test and understand what' specific step failed). But that's a subjective thing so just my two cents :face_in_clouds:

playwright/Customizations/Firewall.spec.ts
avitova: Hmm, in that case we should add this to other tests too. I'd prefer button clicking too, as I think it should be the most intuitive option, and definitely should not be broken.

playwright/Customizations/Firewall.spec.ts
ksiekl: 5pm kinda brain :no_mouth:

playwright/Customizations/Firewall.spec.ts
ksiekl: I picked the nit!

playwright/Customizations/Firewall.spec.ts
ksiekl: No idea what my thoughts were, it was a week ago! I added the values

playwright/Customizations/Firewall.spec.ts
tkoscieln: Well it's more readable, but we may need to extract these steps (the ones that are already not) into helper functions as well because we may need to reuse them in the other tests, so that's why I wouldn't want to atomize it further and keep it in the same format as the template/other tests. But that's just my opinion and I might be in the wrong trying to enforce this :smile: @regexowl thoughts?

regexowl: I guess it depends. Both approaches seem equally functional to me.

As for extracting logic into helpers, I was thinking about how we're currently splitting unit tests into smaller reusable functions. An example could be this test:
```typescript
  test('revisit step button on Review works', async () => {
    await renderCreateMode();
    await goToLocaleStep();
    await searchForKeyboard('us');
    await selectKeyboard();
    await goToReviewStep();
    await clickRevisitButton();
    await screen.findByRole('heading', { name: /Locale/ });
  });
```
where all the awaited parts except for `.findByRole` are reusable helper functions. So we can have super concise and easily readable tests and then main logic split into helpers that are a bit "hidden" away, but can be shared between multiple components.

On the other hand I'm not super sure if this will be applicable to Playwright tests also as currently we're using more of a streamlined approach there, going from the very first start of the Wizard to import/export and just splitting the Wizard modes

ksiekl: Yeah, I don't think that this particular test can be split into reusable helper functions more. My humble opinion is to let it be atomized, but if @tkoscieln or you @regexowl have a very strong opinion about some change I could do, let me know please! But if not, this is the last review thread and we could merge it :pleading_face:

regexowl: As mentioned above - I'm happy with smashing things into small pieces, but have no strong opinion otherwise  :sweat_smile: Let's see what @tkoscieln thinks

tkoscieln: Let's leave it like it is. If we were to split it into helper functions or whatever we can easily adjust it later, so let's merge this :+1:

tkoscieln (approved):

PR #3171, opened 2025-04-30
playwright: Disable file parallelism

regexowl:
This disables file parallelism in Playwright tests.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -13,8 +13,8 @@ if (process.env.CURRENTS_PROJECT_ID && process.env.CURRENTS_RECORD_KEY) {
 
 export default defineConfig({
   testDir: 'playwright',
-  fullyParallel: true,
-  workers: 2,
+  fullyParallel: false,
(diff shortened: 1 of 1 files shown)


regexowl:
Doesn't seem to be needed, closing!

playwright.config.ts
```diff
@@ -13,7 +13,7 @@ if (process.env.CURRENTS_PROJECT_ID && process.env.CURRENTS_RECORD_KEY) {
 
 export default defineConfig({
   testDir: 'playwright',
-  fullyParallel: true,
+  fullyParallel: false,
   workers: 2,
```
tkoscieln, line 17: Not sure if this will do the trick? https://playwright.dev/docs/api/class-testconfig#test-config-fully-parallel
I think it would be safer to set the worker amount to 1 as well?
```suggestion
  workers: 1,
```

tkoscieln (changes requested):

PR #3183, opened 2025-05-02
package.json: replace toml package

kingsleyzissou:
We need to be able to stringify objects into a toml format for on-prem. This is needed to save the worker config, unfortunately the current toml package only parses toml and can't stringify back to toml.


package.json
```diff
@@ -7,6 +7,7 @@
     "npm": ">=7.0.0"
   },
   "dependencies": {
+    "@ltd/j-toml": "^1.38.0",
```
regexowl, line 10:
```suggestion
    "@ltd/j-toml": "1.38.0",
```

regexowl (commented):
Looks good, can you please pin the dep version? It'd be better for dependabot

package.json
kingsleyzissou: Oh whoops I thought I did save exact. Thanks

PR #3186, opened 2025-05-02
Wizard: Fix users bug - 'wheel' group removal

regexowl:
If 'wheel' group was removed the Administrator checkbox stayed check. This unchecks it.


lucasgarfield (approved):
lgtm thank you!

PR #3190, opened 2025-05-05
ImagesTable: remove else after return

avitova:
I noticed that we have this ongoing trend in the codebase of using "else" block after return in the `if`. Let's make sure we do not add more; I tried to fix a few in ImagesTable; more will likely come later. I wanted to prevent big PR and conflicts with other PRs. :)


regexowl (approved):
Yay, spring cleanup! :broom:

PR #3209, opened 2025-05-09
tsconfig: Specify `include`

regexowl:
Specify, what folders to include in the compilation process. This limits the scope, might reduce compilation time and will resolve errors outputted to the console during build.


avitova (approved):
Okay, let's try this.

PR #3210, opened 2025-05-09
LandingPage: Add Edge decommission alert (HMS-6156)

ochosi:
This alert will only be shown when the 'immutable' tab has focus.
Below are screenshots of both tabs.

image
image

This will likely require further work after UX mockups are complete/ready. E.g. do we want to use this alert type? Is the copy correct/complete?


regexowl (dismissed):
LGTM, thank you! :heart: Is this still a draft/blocked by something? Happy to merge when ready.

ochosi:
@regexowl TY :heart:
It's a draft mostly because the UX mockups didn't seem final and we're unsure when to publish this, I think

ochosi:
Update: This is scheduled to land on May 30, so I'll keep it a draft until then.

ochosi:
This is now ready for review. I have updated the branch to match the copy shared in [CPUX-5380](https://issues.redhat.com/browse/CPUX-5380).

There's only one minor difference. Instead of going for an inline hyperlink, I used a newline and a separate hyperlink/button element, as we do everywhere else. This makes the link stand out more and it's also more consistent.

image

regexowl (approved):
LGTM :heart:

PR #3216, opened 2025-05-12
Wizard: Fix password placeholder (HMS-6290)

regexowl:
User password gets redacted from a blueprint, meaning if there was any password set, it does not render on import.

Previously a placeholder was rendered as a value of password, making the value not editable. This renders the password indication as a placeholder text, solving the issue.

JIRA: [HMS-6290](https://issues.redhat.com/browse/HMS-6290)

### DIFF
--- a/src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
@@ -54,10 +54,10 @@ export const PasswordValidatedInput = ({
             <TextInput
               isRequired
               type={isPasswordVisible ? 'text' : 'password'}
-              value={hasPassword ? '•'.repeat(8) : value}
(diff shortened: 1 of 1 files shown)


regexowl:
Can't test this one as `hasPassword` is not in prod yet and stage doesn't work.

regexowl:
Tested, works!

ochosi (dismissed):
Awesome, what a simple fix!

avitova (commented):
If you create a blueprint, then you try to edit it, the option to make it visible again is disabled. I think we might try to fix this.

regexowl:
> If you create a blueprint, then you try to edit it, the option to make it visible again is disabled. I think we might try to fix this.

Great catch, thank you! Should be fixed now.

avitova (approved):
Niiice! LGTM now!:)

PR #3217, opened 2025-05-12
Wizard: Fix Wizard height

regexowl:
The height of Wizard seems to be randomly shrinking, this sets the height to a fixed value, ensuring the footer is in a fixed position.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.scss
+++ b/src/Components/CreateImageWizard/CreateImageWizard.scss
@@ -76,10 +76,7 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
 }
 
 .pf-v5-c-wizard__main {
-    flex: 1 1 auto;
-    display: flex;
-    flex-direction: column;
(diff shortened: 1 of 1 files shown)


regexowl:
This is not an elegant solution, I've been having no luck with `flex` though :thinking: Tried to revert changes done in previous PRs that touched the styling and ended up with the original problem - the Wizard usually rendered fine, but this happened when the window got resized:

image

Setting fixed height seems to alleviate that problem, but also causes double scroll bars in some cases.

ochosi:
Yeah, I've seen this problem today as well - sucks! I wonder if we could ask someone from the PatternFly team or someone from ConsoleDot if this is a known problem and if there's a common solution.

At the same time, we'll likely do away with the wizard during the UI refresh after Summit, so maybe it isn't worth it...

lucasgarfield:
I think I found a slightly more elegant solution (more elegant because it removes the need for CSS entirely):
https://github.com/osbuild/image-builder-frontend/pull/3234

Issue #3156, opened 2025-04-28
Wizard: Make "Invalid format" validation error more specific


regexowl:
For `LabelInput` component the default validation error is "Invalid format". We should specify, what format is expected to make it easier for the user to input correct data.

PR #3219, opened 2025-05-12
CreateImageWizard: Implement user-friendly error messages (HMS-8704)

schuellerf:
This is a suggestion for #3156 which I didn't know existed when I started implementing this draft…

I didn't test the implementation, nor do I know if the return value type `string | null` is a good or even valid approach.

@ksiekl please check if this branch helps and feel free to continue working on it (or ping me if I should continue)

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 
 import { CheckCircleIcon } from '@patternfly/react-icons';
-import { jwtDecode, JwtPayload } from 'jwt-decode';
+import { jwtDecode } from 'jwt-decode';
 
 import { getListOfDuplicates } from './getListOfDuplicates';
 
@@ -30,6 +30,7 @@ import {
   selectLanguages,
   selectKeyboard,
   selectTimezone,
+  selectSatelliteCaCertificate,
   selectSatelliteRegistrationCommand,
   selectImageTypes,
   UserWithAdditionalInfo,
@@ -45,11 +46,11 @@ import {
   isMountpointMinSizeValid,
   isSnapshotValid,
   isHostnameValid,
-  isKernelNameValid,
   isUserNameValid,
   isSshKeyValid,
   isNtpServerValid,
   isKernelArgumentValid,
+  isKernelNameValid,
   isPortValid,
   isServiceValid,
   isUserGroupValid,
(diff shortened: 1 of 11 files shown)


src/Components/CreateImageWizard/validators.ts
```diff
@@ -113,28 +122,46 @@ export const isHostnameValid = (hostname: string) => {
...
+  const check =
     kernelName.length < 65 &&
     /^([a-z0-9]|[a-z0-9][a-z0-9-_.+]*)[a-z0-9]$/.test(kernelName) &&
-    /[a-zA-Z]+/.test(kernelName) // contains at least one letter
-  );
+    /[a-zA-Z]+/.test(kernelName); // contains at least one letter
+  if (check) {
+    return null;
+  } else {
+    return 'Expected format: <kernel-name>. Example: kernel-5.14.0-284.11.1.el9_2.x86_64';
```
ksiekl, line 133: Hmm, this is causing some issues - the error message shows as an option in the dropdown and can be selected, I'm thinking if it was caused by changing the type of return value from boolean to string :thinking: 
image

ksiekl (commented):
Very nice! The approach to return `string | null` works nicely, it's simple and effective. However, I think in one dropdown it might be causing issues (when it was returning boolean before, it made the error message disabled, now the error message is clickable as if it was an option:
image
)
So maybe to make it a bit more robust the validators could return a tuple or an object like `{ valid: boolean, error?: string }`? It would be a more scalable solution, but honestly, I don't see a way we could scale the validators more in the future... 

So another option that could be easier to do is to just disable the error message in the dropdown, although I have not looked into that yet, so idk how easy that would be!

src/Components/CreateImageWizard/validators.ts
schuellerf: I'm pretty sure this can be fixed without changing the return type signature
Do you want to continue fixing this?
I still need to setup my PC to get the local version up & running 😇

schuellerf: @ksiekl 😍 thanks for testing this - there really is a text-search to disable the dropdown…
I just hard-patched the new string in
https://github.com/osbuild/image-builder-frontend/pull/3219/commits/8b1cb77e1e347d4fbb55337a7d1a447a9fa861d1
not sure if this should be the way to go in the future…

regexowl:
Hmm, but then we'd need to think about how to work with on import errors...

schuellerf:
@regexowl my idea was to keep the error message as close (code-wise) to the regex as possible
but feel free to change/rewrite. I'm perfectly fine with other solutions 👌. Also no need to keep my code/input!

ksiekl:
@regexowl Refactored like we talked about, could you check it out please? :pray:

src/Components/CreateImageWizard/validators.ts
```diff
@@ -68,8 +68,8 @@ export const isUserGroupValid = (group: string) => {
   return (
     group.length <= 32 &&
     /^[a-zA-Z0-9_][a-zA-Z0-9_-]*(\$)?$/.test(group) &&
-    /[a-zA-Z]+/.test(group) // contains at least one letter
-  );
+    /[a-zA-Z]+/.test(group)
+  ); // contains at least one letter
```
regexowl, line 72: These comments were originally line specific - relevant to `/[a-zA-Z]+/.test(group)` part of the regex only, do we need to move them?

regexowl (changes requested):
Ran through the rest of the changes - just the question about the comments above and there are some more tests that will need to be updated, otherwise looks great! Thank you both so much :heart:

src/Components/CreateImageWizard/validators.ts
ksiekl: Ah yes sorry, moved them :smile:

src/test/Components/Blueprints/ImportBlueprintModal.test.tsx
```diff
@@ -555,9 +555,6 @@ describe('Import modal', () => {
         await screen.findByPlaceholderText('Paste your public SSH key')
       )
     );
-    expect(
-      await screen.findByText(/Invalid user groups: 0000/)
```
ksiekl, line 559: Deleted this because the validation is not persistent, but that is a scope for a PR on its own, filed an issue [here](https://github.com/osbuild/image-builder-frontend/issues/3394).

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -526,7 +496,6 @@ export function useUsersValidation(): UsersStepValidation {
   }
 
   for (let index = 0; index < users.length; index++) {
-    const invalidGroups = [];
```
regexowl, line 529: I see the problem, but leaving the groups code in doesn't seem to cause any issues with the tests. Can we leave the code in to keep at least some groups field validation and follow up by fixing the issue you've opened?

src/test/Components/Blueprints/ImportBlueprintModal.test.tsx
```diff
@@ -555,9 +555,6 @@ describe('Import modal', () => {
         await screen.findByPlaceholderText('Paste your public SSH key')
       )
     );
-    expect(
```
regexowl, line 558: The same here, I've tried running the tests with this expect and the groups validation and they seem to be passing just fine, I'd keep this in to keep some validation even if not perfect :shrug:

regexowl (changes requested):
Looks great! Two questions about the user groups validation issue

regexowl (approved):
Thank you :heart:

schuellerf:
🥳 thank you @ksiekl for cleaning up my mess and finishing the PR 🤗

PR #3221, opened 2025-05-12
Wizard: encode satellite cmd properly

lzap:
I know nothing about the modern JS/TS, review with care. Problems:

* The file /usr/local/sbin/register-satellite is not executable.
* The file does not have a shebang.
* The file contents is for some reason all URL encoded.
* Finally, the file is missing an end newline.

/epic https://issues.redhat.com/browse/HMS-6296

### DIFF
--- a/src/Components/CreateImageWizard/utilities/requestMapper.ts
+++ b/src/Components/CreateImageWizard/utilities/requestMapper.ts
@@ -473,9 +473,7 @@ const getSatelliteCommand = (files?: File[]): string => {
   const satelliteCommandFile = files?.find(
     (file) => file.path === '/usr/local/sbin/register-satellite'
   );
-  return satelliteCommandFile?.data
-    ? decodeURIComponent(atob(satelliteCommandFile.data))
-    : '';
+  return satelliteCommandFile?.data ? atob(satelliteCommandFile.data) : '';
 };
 
 const uploadTypeByTargetEnv = (imageType: ImageTypes): UploadTypes => {
@@ -587,7 +585,8 @@ const getCustomizations = (state: RootState, orgID: string): Customizations => {
     });
     files.push({
       path: '/usr/local/sbin/register-satellite',
-      data: btoa(encodeURIComponent(satCmd)),
+      data: btoa(satCmd),
+      mode: '0774',
       data_encoding: 'base64',
       ensure_parents: true,
     });
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -263,23 +263,23 @@ WantedBy=graphical.target
 export const FIRST_BOOT_SERVICE = 'custom-first-boot';
 
 export const SATELLITE_SERVICE_DATA = btoa(`[Unit]
-  Description=Custom first boot script
-  ConditionFileIsExecutable=/usr/local/sbin/register-satellite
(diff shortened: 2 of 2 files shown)


src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -587,7 +587,8 @@ const getCustomizations = (state: RootState, orgID: string): Customizations => {
     });
     files.push({
       path: '/usr/local/sbin/register-satellite',
-      data: btoa(encodeURIComponent(satCmd)),
+      data: btoa("#!/bin/bash\n\n" + satCmd + "\n"),
```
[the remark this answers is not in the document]
lzap: I guess I need to remove it from `getSatelliteCommand` too but I do not know the reason why the value is URL decoded in the first place. @avitova ?

lzap:
@avitova had a good point that prefixing/suffixing the string will stack each edit, so removed that. Newline actually does not matter, looks like GNU bash has no problems with that. Also dropped the URI decoding which was not necessary.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -587,7 +587,8 @@ const getCustomizations = (state: RootState, orgID: string): Customizations => {
     });
     files.push({
       path: '/usr/local/sbin/register-satellite',
-      data: btoa(encodeURIComponent(satCmd)),
+      data: btoa(satCmd),
```
avitova, line 590: Are we not adding shebang then?

lzap:
Tested new/edit and it works as expected: 

image

src/Components/CreateImageWizard/utilities/requestMapper.ts
lzap: No, it is not necessary, I have changed the service unit it now calls `bash -x /the/script`

src/constants.ts
```diff
@@ -250,7 +250,7 @@ After=osbuild-first-boot.service
 
 [Service]
 Type=oneshot
-ExecStart=/usr/local/sbin/custom-first-boot
+ExecStart=/bin/bash -x /usr/local/sbin/custom-first-boot
```
lzap, line 253: This makes the shebang unnecessary. The newline is also not needed, it looks like bash does not care if a script does not have it and EOF will trigger the script as well.

avitova (approved):
This looks good, let's wait for this ^ and then merge.

lzap:
Rebased, moved the explicit `bash` exec to the correct place, thanks that was sneaky!

I am keeping the executable flag check tho, I think it is worth keeping the script executable to be able to re-run it later (e.g. when it fails).

lzap:
After chat with @ondrejbudai I was experimenting a bit with something unrelated but I am done, no more experiements let's not play around when in crunch.

```
  [Install]
  WantedBy=basic.target
  WantedBy=multi-user.target
  WantedBy=graphical.target
```

regexowl:
The changes look good to me :+1: I'll leave the final word to @avitova as I'm not super familiar with Satellite

avitova (approved):
Yeah, let's get this in finally. Thanks everyone!:)

PR #3220, opened 2025-05-12
Wizard: Fix activation keys list refresh (HMS-6294)

regexowl:
We've been refetching the data, but not re-rendering the list. This should fix the issue.

JIRA: [HMS-6294](https://issues.redhat.com/browse/HMS-6294)

PR #3222, opened 2025-05-12
Wizard: Update AWS sources filtering (HMS-6298)

regexowl:
Same issue as in #3220. The list of sources did refetch, but did not re-render.

JIRA: [HMS-6298](https://issues.redhat.com/browse/HMS-6298)


avitova (approved):
Yeah, this makes perfect sense, thanks!

PR #3224, opened 2025-05-12
Satellite: Add instructions for obtaining certificate + token expiration info

avitova:
Let me know what you think. To be honest, I kinda do not like what the code for handling of "will expire in approximately in" looks like, but if this is better for the user, I guess this is the best way to do it. 

Adding screencast so that everyone can check the changes. ;)

https://github.com/user-attachments/assets/e1b599ca-3b4d-49cd-945f-9a48136136e2


lzap:
I am unable to do code review, so dropping a comment right here. It is great, one detail is that minutes probably do not matter, I think the shortest relevant time might be one hour (typical build queue on stage) meaning: the image might be build after token expired. So in my eyes, the only relevant milestone is one hour. Just a personal preference.

I still see the CA cert marked as required, this should not be required at all.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -153,44 +202,13 @@ export function useRegistrationValidation(): StepValidation {
...
-      Object.assign(errors, { command: 'Invalid or missing token' });
-    }
+    const tokenErrors = validateSatelliteToken(registrationCommand);
+    Object.assign(errors, tokenErrors);
+
     return {
       errors: errors,
       disabledNext:
-        Object.keys(errors).length > 0 || caCertificate === undefined,
+        Object.keys(errors).filter((key) => key !== 'expired').length > 0,
```
ondrejbudai, line 216: nitpick: `Object.keys(errors).some(key => key !== 'expired')` is shorter and faster.

avitova: Thanks for the review. I think that the current validation is much more readable. ;)

avitova:
Thanks, everyone, for the reviews. I decided to refactor the Satellite validation function; I believe this is much better now. The certificate is no longer required. I also disabled the next button for when the token is expired after talking to @ondrejbudai.
I think we're there.

https://github.com/user-attachments/assets/8c317853-7a1a-4a56-a0e2-0434a97f73d5

regexowl:
@avitova Can you please check out the AI suggestions? Otherwise looks good to me :)

avitova:
Oh, finally! This took me a while to crack - Errors in input validated field do not show before you click outside the component. Should be okay now.

regexowl (approved):
LGTM, thanks! :sparkles:

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -112,52 +112,62 @@ type ValidationState = {
...
+    return errors;
+  }
+  const tokenRemainingS = tokenValidityRemaining(expiresInSeconds);
+  if (tokenRemainingS <= 0) {
+    errors.command = `The token is expired. Check out the Satellite documentation to extend the Token lifetime.`;
+    return errors;
+  }
+  const expirationString = getExpirationString(tokenRemainingS);
+  if (expirationString !== undefined) {
+    errors.expired = `The token expires in ${expirationString}. Check out the Satellite documentation to extend the Token lifetime.`;
```
ondrejbudai, line 167: I suggest dropping the capital T to be consistent with the beggining of the sentence.

regexowl (dismissed):

avitova:
@regexowl sorry, I had to unflag the secrets so that they are not flagged as security leak:D

regexowl (approved):

PR #3225, opened 2025-05-12
Wizard: fix satellite command encoding and permissions

avitova:
I suggest creating a helper function that replaces `registrationType !== 'register-later'` and `registrationType !== 'register-satellite'. But maybe after summit, idk. It was very hard to find every `registrationType !== 'register-later'` a in the codebase, and I hope I found every occurrence.


src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -166,6 +166,25 @@ const getLatestRelease = (distribution: Distributions) => {
     : distribution;
 };
 
+function removeShebangAndNewline(satelliteCommand: string) {
```
avitova, line 169: I created this function so that when we revisit the wizard to edit it, we should not show to user that there is a shebang and newline.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -587,8 +606,9 @@ const getCustomizations = (state: RootState, orgID: string): Customizations => {
     });
     files.push({
       path: '/usr/local/sbin/register-satellite',
-      data: btoa(encodeURIComponent(satCmd)),
+      data: btoa('#!/bin/bash\n' + satCmd + '\n'),
```
avitova, line 609: adding shebang - here we kinda do not care if there is any shebang or newline. But we have properly documented that users are just supposed to add the command, so I think we are good here.

avitova:
Closing,this is solved in #3221, I moved the rest - the check for activation keys into #3224 .

PR #3226, opened 2025-05-12
WIP: Ansible Automation Platform

lucasgarfield:
Coded with Aider + Claude Sonnet 3.7


lucasgarfield:
So far this cost $3.34 in API tokens.

Todo:
- [ ] Get review step working
- [ ] Get edit mode working (I think ib-crc changes must be made first)
- [ ] Add validation to forms

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -368,6 +369,9 @@ function commonRequestToState(
  */
 export const mapRequestToState = (request: BlueprintResponse): wizardState => {
   const wizardMode = 'edit';
+  // Check if any image request has AAP registration
+  const aapRegistration = request.image_requests.find(req => req.aap_registration);
```
lucasgarfield, line 373: This is wrong.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -377,20 +381,28 @@ export const mapRequestToState = (request: BlueprintResponse): wizardState => {
...
     registration: {
       registrationType:
-        request.customizations?.subscription && isRhel(request.distribution)
-          ? request.customizations.subscription.rhc
-            ? 'register-now-rhc'
-            : 'register-now-insights'
-          : getSatelliteCommand(request.customizations.files)
-          ? 'register-satellite'
-          : 'register-later',
+        aapRegistration?.aap_registration
```
lucasgarfield, line 384: This is wrong and should be rewritten with a helper function that uses a switch for readability.

src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
```diff
@@ -195,6 +195,9 @@ const Review = () => {
 
   const isUsersEnabled = useFlag('image-builder.users.enabled');
   const { isFedoraEnv } = useGetEnvironment();
+  // This is the review step. In the registration section,
+  // add relevant fields for AAP if it is being used for registration.
+  // AI!
```
lucasgarfield, line 200: Remove this.

src/constants.ts
```diff
@@ -288,6 +288,7 @@ WantedBy=graphical.target
 `);
 
 export const SATELLITE_SERVICE = 'register-satellite';
+export const AAP_SERVICE = 'ansible-controller-registration';
```
kingsleyzissou, line 291: We take care of this in `image-builder-crc`, so this breaks the image build.
If you really want to add this to the list of services, it should be (see here [line](https://github.com/kingsleyzissou/image-builder/blob/5204665cf9de6bf22ce4591fa2cc976b366bc4b4/internal/v1/handler_compose_image.go#L29) and [here](https://github.com/kingsleyzissou/image-builder/blob/5204665cf9de6bf22ce4591fa2cc976b366bc4b4/internal/v1/handler_compose_image.go#L1306)):
`aap-first-boot-reg.service`

lzap:
Small remark as I am trying this out. It reads "Ansible Controller URL" but then it shows `example.towerhost.net` string which is not a URL, that is what I call a hostname, so I am confused which one to enter.

image

src/constants.ts
lzap: Yep, I can confirm I just tested this PR and it is erroring out with: `Failed to enable unit, unit ansible-controller-registration.service does not exist`

We should move all the code logic into the backend, UI should not be aware of any details here, it must use the API from the CRC.

lucasgarfield:
> Small remark as I am trying this out. It reads "Ansible Controller URL" but then it shows `example.towerhost.net` string which is not a URL, that is what I call a hostname, so I am confused which one to enter.
> image

I removed the placeholders for now, let's just keep the UX as basic as possible until we've verified everything is working, then we can polish it up a bit.

src/Components/CreateImageWizard/steps/Registration/Registration.tsx
```diff
@@ -112,13 +112,14 @@ const Registration = () => {
     'image-builder.satellite.enabled'
   );
 
+  const isAAPRegistrationEnabled = useFlag('image-builder.aap.enabled') || true; // Default to true for now, remove when flag is implemented
```
regexowl, line 115: Making a note here for myself so we don't forget to remove the override.

src/Components/CreateImageWizard/steps/Registration/AAPRegistration.tsx
```diff
@@ -0,0 +1,161 @@
...
+        />
+      </FormGroup>
+      <FormGroup label="Host Config Key" isRequired>
+        <ValidatedInputAndTextArea
+          value={hostConfigKey || ''}
+          onChange={(_event, value) => handleHostConfigKeyChange(value)}
+          ariaLabel="host config key"
+          isRequired
+          stepValidation={stepValidation}
+          fieldName=""
```
regexowl, line 107: The validation wouldn't appear for this field.

```suggestion
          fieldName="hostConfigKey"
```

src/Components/CreateImageWizard/steps/Registration/AAPRegistration.tsx
```diff
@@ -0,0 +1,161 @@
...
+      </FormGroup>
+      <FormGroup label="Job Template ID" isRequired>
+        <ValidatedInputAndTextArea
+          value={jobTemplateId || ''}
+          onChange={(_event, value) => handleJobTemplateIdChange(value)}
+          ariaLabel="job template id"
+          isRequired
+          stepValidation={stepValidation}
+          fieldName="jobTemplateId"
+          placeholder=""
```
regexowl, line 97: I'd prefer to either fill the placeholders or mark them as optional in `ValidatedInputAndTextArea`. But also happy with just opening an issue and updating it in a follow up so we don't block on this.

src/Components/CreateImageWizard/ValidatedInput.tsx
```diff
@@ -168,6 +169,7 @@ export const ValidatedInput = ({
         aria-label={ariaLabel || ''}
         onBlur={handleBlur}
         placeholder={placeholder || ''}
+        {...props}
```
regexowl, line 172: Here also - it would be nicer to spread the props so we have more control. But it works and if we open an issue to look at this in a follow up I'll be happy.

regexowl (commented):
Most of the code looks fine, I've added some comments. The generated request seems to be correct, couldn't test further as I'm not familiar with AAP at all :flushed: 

I presume the on-prem tests are failing because there is no Registration step in on-prem currently.

Also validation doesn't appear when a required field is empty, but I think that's connected to how we handle the pristine state for validated inputs and isn't necessary in the scope of this PR. Since we indicate that the fields are required I'd say it makes sense that the Next button is disabled even without an error message.

kingsleyzissou:
> Small remark as I am trying this out. It reads "Ansible Controller URL" but then it shows example.towerhost.net string which is not a URL, that is what I call a hostname, so I am confused which one to enter.

This was my fault. Fix here: https://github.com/osbuild/image-builder-crc/pull/1606

kingsleyzissou:
> I presume the on-prem tests are failing because there is no Registration step in on-prem currently.

Yeah, we should probably fix the tests to account for this

api/schema/composerCloudApi.v2.yaml
regexowl, line 1: Just a heads up - the schemas were removed last week, replaced by direct pull https://github.com/osbuild/image-builder-frontend/pull/3454

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -200,6 +202,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
 
   // Feature flags
   const complianceEnabled = useFlag('image-builder.compliance.enabled');
+  // const isAAPRegistrationEnabled = useFlag('image-builder.aap.enabled') || true; // Default to true for now, remove when flag is implemented
```
regexowl, line 205: Just checked and the flag is currently disabled in both prod and stage unleash, how should we set it up? Enabled for stage preview only?

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -200,6 +211,60 @@ export function useRegistrationValidation(): StepValidation {
...
+        (!tlsCertificateAuthority || tlsCertificateAuthority.trim() === '')
+      ) {
+        errors.certificate =
+          'HTTPS URL requires either a custom TLS certificate or confirmation that no custom certificate is needed';
+      }
+    } else {
+      // If URL is just HTTP, always require a TLS certificate
+      if (!tlsCertificateAuthority || tlsCertificateAuthority.trim() === '') {
+        errors.certificate = 'HTTP URL requires a custom TLS certificate';
+      }
```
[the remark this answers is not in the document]
ksiekl: Hush, robot

src/Components/CreateImageWizard/CreateImageWizard.tsx
regexowl: We'll also need to set the flag for tests, leaving this here so we don't forget to take a look at that :sweat_smile:

regexowl: The flags are set to enable AAP in stage-preview only for now :+1:

regexowl (commented):
Thank you, this is looking really good! Added a few small comments and nitpicks, but overall works nicely :sparkles:

src/test/Components/CreateImageWizard/steps/Oscap/Compliance.test.tsx
```diff
@@ -61,6 +61,7 @@ const goToReviewStep = async () => {
   await clickNext(); // Kernel
   await clickNext(); // Firewall
   await clickNext(); // Systemd services
+  await clickNext(); // AAP
```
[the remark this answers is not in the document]
ksiekl: Thanks for this! I was bamboozled about what was going wrong with that test :heart_decoration:

kingsleyzissou (approved):

PR #3232, opened 2025-05-13
Wizard: Add RHEL10 to ReleaseLifecycle chart

ochosi:
This adds RHEL10 lifecycle dates to the chart we show when customers select RHEL8.

Open question: Should we also show this when RHEL9 is selected, now that we want to promote RHEL10 adoption?


lucasgarfield:
As a future improvement, I think the RHEL Lifecycle Api exposes this information now. I would continue to hard code it in as you have done here since GA is this week but just wanted to make this comment for awareness (cc @regexowl).

regexowl:
Looks great! As discussed in the frontend channel, we can start rendering the chart for RHEL 9 as soon as it's not a default release anymore (meaning follow up with switching the default + showing chart for 9). I'm happy to merge this when it's out of draft :rocket:

regexowl (approved):
Looks great, thank you!

Issue #3097, opened 2025-04-11
Wizard: Indicate which user tab contains error


regexowl:
Follow up to https://github.com/osbuild/image-builder-frontend/pull/3071

If there are any errors present in one of the users, the tabs should indicate, which one it is.

PR #3233, opened 2025-05-13
Wizard: indicate which user tab contains error (HMS-8640)

ksiekl:
Fixes: #3097

Added exclamation marks to the tabs where the user contains error.

JIRA: [HMS-8640](https://issues.redhat.com/browse/HMS-8640)

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -486,7 +486,8 @@ export function useServicesValidation(): StepValidation {
 
 const validateUserName = (
   users: UserWithAdditionalInfo[],
-  userName: string
+  userName: string,
+  currentIndex: number
 ): string => {
   if (!userName) {
     return 'Required value';
@@ -496,9 +497,10 @@ const validateUserName = (
   }
 
   // check for duplicate names
-  const duplicateName =
-    new Set(users.map((user) => user.name)).size !== users.length;
-  if (duplicateName) {
+  const count = users.filter(
+    (user, index) => user.name === userName && index !== currentIndex
+  ).length;
+  if (count > 0) {
     return 'Username already exists';
   }
   return '';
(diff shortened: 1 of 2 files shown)


regexowl (commented):
This will be super nice to have! Thanks for looking into it :heart:

This currently checks problems with username only, I think it might be good to indicate any error on the user tab:
image

Also found one bug - the error renders for all three users here:
image
(I've filled in only the usernames)

ksiekl:
I noticed the bug with multiple error icons - it specifically happens, when there are duplicate user names. It's a bug in the validator evaluation logic and I thought it would be better to have a separate issue and PR for that because it's a bit of a different scope. Is that ok?

ksiekl:
Well, nevermind, I edited the validation in this PR, cheerio! :cookie:

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -496,9 +497,10 @@ const validateUserName = (
   }
 
   // check for duplicate names
-  const duplicateName =
-    new Set(users.map((user) => user.name)).size !== users.length;
-  if (duplicateName) {
+  const count = users.filter(
+    (user, index) => user.name === userName && index !== currentIndex
+  ).length;
+  if (count > 0) {
```
avitova, line 503: Nice, I like this simplification very much! I was wondering, do we want to add the error even for the tab that had this username first? :thinking:

avitova (commented):
I have one comment for us to think about - let me know what you think. Other than that, I do not have anything.
Also, the error in tabs appears sooner than the validation error under the field. I think we should fix this later, but that's not in the scope of this PR, I guess.

regexowl:
>Also, the error in tabs appears sooner than the validation error under the field. I think we should fix this later, but that's not in the scope of this PR, I guess.

Agree that this might be outside of the scope. We can create an issue

avitova (approved):
Nice, let's see if the tests will pass, and we can merge! :partying_face:

avitova:
Bruuuuh, not the rebase again. :sob:

PR #3234, opened 2025-05-14
Wizard: Fix vertical height bug

lucasgarfield:
Recently the wizard has started rendering such that its vertical height matches the height of its body – and therefore only occupies some fraction of the page’s vertical space for many steps. The desired behavior is for its footer to be fixed at the bottom of the viewport.

Specifying the page section type prop as ‘wizard’ resolves the issue. This is the pattern used in the Patternfly documentation for in-page wizards. See here:
https://www.patternfly.org/components/wizard/react-demos#in-page

Using CSS to adjust Patternfly component rendering is an antipattern. More custom CSS means more time spent troubleshooting rendering issues. Custom CSS related to rendering of the wizard has been removed.

There is still an annoying issue. The body of the wizard is not scrollable and will overflow. This is counter to the Patternfly examples. There doesn’t seem to be any CSS in Image Builder that would cause this to occur. Perhaps the problem is somehow inherited from Insights Chrome? In any case, this will require further investigation.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.scss
+++ b/src/Components/CreateImageWizard/CreateImageWizard.scss
@@ -74,15 +74,3 @@ div.pf-v5-c-alert.pf-m-inline.pf-m-plain.pf-m-warning {
         margin-block-start: 0;
     }
 }
-
-.pf-v5-c-wizard__main {
-    flex: 1 1 auto;
-    display: flex;
-    flex-direction: column;
-    justify-content: space-between;
-}
-
-.pf-v5-c-wizard__footer {
-    position: sticky;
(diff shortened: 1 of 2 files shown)


ochosi (approved):
Tested this locally and it looks and works great! Such an improvement over what's in prod at the moment... We should really get this in asap!

ochosi:
Also, the fugly grey border around the wizard is gone with this patch. :100:

regexowl (approved):
I also like it, bye bye CSS :broom:

It's a shame the footer is not sticky even when we're not overriding anything, but since PF6 migration and UI revamp are ahead of us I don't think we need to block on that.

PR #3237, opened 2025-05-14
LandingPage: Add Users to `NewAlert`

regexowl:
This adds Users to the list of new customization options.


regexowl:
@ochosi @lucasgarfield I assume we want to flip the Users flag as soon as the two bug fixes are released to prod. So let's also add the Users to the alert in the same release.

ochosi (approved):

PR #3238, opened 2025-05-14
Add el10 releases

croissanne:
Hides the compliance step until we get the security guides in order.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -87,6 +87,7 @@ export const RELEASES = new Map([
   [RHEL_9_BETA, 'Red Hat Enterprise Linux (RHEL) 9 Beta'],
   [RHEL_10_BETA, 'Red Hat Enterprise Linux (RHEL) 10 Beta'],
   [CENTOS_9, 'CentOS Stream 9'],
+  [CENTOS_10, 'CentOS Stream 10'],
 ]);
 
 export const ON_PREM_RELEASES = new Map([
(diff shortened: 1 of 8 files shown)


src/constants.ts
```diff
@@ -100,10 +102,13 @@ export const FEDORA_RELEASES = new Map([
   [FEDORA_40, 'Fedora Linux 40'],
   [FEDORA_41, 'Fedora Linux 41'],
   [CENTOS_9, 'CentOS Stream 9'],
+  [CENTOS_10, 'CentOS Stream 10'],
 ]);
 
+export const RHEL_10_FULL_SUPPORT = ['2025-05-13', '2027-05-31'];
```
regexowl, line 108: Code looks good, does the full support end the same day the it will end for RHEL9?

croissanne: oh no
sorry that's meant to be 2030

ondrejbudai (dismissed):
🚢 it!

regexowl (dismissed):
Thank you! :heart:

croissanne:
oh we rebased ad the same time :D

regexowl:
Synced rebasing FTW!

croissanne:
Oh the test ids disappearing also broke the schutzbot tests :/

regexowl:
EC check seems to be stuck, trying to rerun konflux actions

/retest

lucasgarfield (changes requested):
Rather than hide the step, let's just disable the dropdown and surface a notification that tells the user that RHEL 10 profiles are coming soon.

PR #3243, opened 2025-05-15
Wizard: drop unwanted WantedBy and run after custom firstboot

lzap:
We need to fix the ordering of registrations vs custom first boot script. The PR also drops unused WantedBy statements.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -252,6 +252,8 @@ ConditionPathExists=!/var/local/.custom-first-boot-done
 Wants=network-online.target
 After=network-online.target
 After=osbuild-first-boot.service
+After=register-satellite.service
+After=aap-first-boot-reg.service
 
 [Service]
 Type=oneshot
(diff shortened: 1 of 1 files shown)


lzap:
I think we should merge it today if possible to ensure ordering for all the demos: @lucasgarfield @ondrejbudai @ochosi @avitova @regexowl can you review and help me to get this merged?

lzap:
We should leave a note about the ordering: https://issues.redhat.com/browse/HMS-8555

src/constants.ts
```diff
@@ -260,9 +262,7 @@ ExecStartPost=/usr/bin/touch /var/local/.custom-first-boot-done
 RemainAfterExit=yes
 
 [Install]
-WantedBy=basic.target
```
ochosi, line 263: Why did we have these and why do we not want them anymore?

ochosi:
> I think we should merge it today if possible to ensure ordering for all the demos: @lucasgarfield @ondrejbudai @ochosi @avitova @regexowl can you review and help me to get this merged?

@lzap I would suggest we update the PR description as well to reflect that this should be merged. I almost closed the tab again after reading it :)

src/constants.ts
lzap: @ondrejbudai noticed these are in fact incorrect or have no effect and I can confirm that these are useless

lzap:
Playwright timeout rebasing, anything else I need to do?

regexowl:
EC check seems to be stuck, re-trying

/retest

regexowl (approved):
Happy to merge as soon as konflux is back :rocket:

PR #3244, opened 2025-05-15
Wizard: fix token not having expiration fails

avitova:
The token does not need expiration date.

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 
 import { CheckCircleIcon } from '@patternfly/react-icons';
-import { jwtDecode } from 'jwt-decode';
+import { jwtDecode, JwtPayload } from 'jwt-decode';
 
 import { UNIQUE_VALIDATION_DELAY } from '../../../constants';
 import { useLazyGetBlueprintsQuery } from '../../../store/backendApi';
@@ -116,10 +116,10 @@ function tokenValidityRemaining(expireTimeInSeconds: number): number {
   return expireTimeInSeconds - currentTimeSeconds;
 }
 
-function getTokenExpirationTime(token: string): number | undefined {
+function decodeToken(token: string): JwtPayload | undefined {
   try {
     const decoded = jwtDecode(token) as { exp?: number };
-    return decoded.exp;
+    return decoded;
   } catch {
     return undefined;
   }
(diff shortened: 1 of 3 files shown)


ondrejbudai (approved):
Thanks!

avitova:
Unit tests failed, so I tried debugging locally, after a lot of attempts turning on performance mode on my laptop worked. So I think that tests should be green. :eyes:

regexowl (approved):
Looks like some async shenanigans were going on, the timeout did the trick. Thank you! :heart:

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -152,19 +152,22 @@ export function validateSatelliteToken(
...
-    errors.command = `The token is expired. Check out the Satellite documentation to extend the token lifetime.`;
-    return errors;
-  }
-  const expirationString = getExpirationString(tokenRemainingS);
-  if (expirationString !== undefined) {
-    errors.expired = `The token expires in ${expirationString}. Check out the Satellite documentation to extend the token lifetime.`;
+
+  if (satelliteToken.exp) {
+    const tokenRemainingS = tokenValidityRemaining(satelliteToken.exp);
+    if (tokenRemainingS <= 0) {
```
ochosi, line 163: We don't notify/warn customers about their token having no expiration time, right? I'm not sure what value is set in that case, but I assume 0?

I'm not sure I understand from a distance how this check is different in terms of blocking tokens that cannot expire without notifying the user about it.

regexowl: I assume we check (and when needed block on) how far the expiration date is only for tokens, that have an expiration date set:
```typescript
  if (satelliteToken.exp) {
    ...
  }
```

avitova: 1) If the token is set with no expiration time, there would be no "exp" in the token, afaik.
So for that situation, we would not add any errors or warnings, just let users go through.
Does that sound okay?

regexowl: Makes perfect sense to me :+1:

ochosi: @avitova We can later add a notice that not setting an expiration time is not considered best practice and secure, but that's fine for now!

regexowl: Good idea, opened an issue: https://github.com/osbuild/image-builder-frontend/issues/3248

PR #3253, opened 2025-05-19
src: Remove `image-builder.users.enabled` flag

regexowl:
The Users customization is available in production now, the flag is no longer used.


avitova:
Is this still needed?:)

kingsleyzissou (approved):

kingsleyzissou:
> Is this still needed?:)

Ya I'd say so, it's just an empty check now since this is in prod. So no harm removing it :)

PR #3256, opened 2025-05-20
Wizard: temporarily disable OCI (HMS-8569)

croissanne:
Until we get our accounts back, there's nothing we can do here.

---

screen oci


ondrejbudai (commented):
Looks sane to me.

ochosi (approved):
The tile resizes a little oddly, but that's not really a show-stopper and we should get this into production as quickly as possible.

PR #3257, opened 2025-05-20
build: add testing proxy to playwright CI

dominikvagner:
This PR switches the playwright tests in CI to use the consoledot testing
proxy instead of the development one in fec. This should result in
faster and more stable tests.


regexowl (approved):
Thank yooou :heart:

PR #3261, opened 2025-05-22
Revert "Wizard: temporarily disable OCI" (HMS-8592)

ochosi:
Once everything is back to 'normal' with Oracle Cloud, we should revert the temporary patch that allowed us to disable this specific image type in the frontend.

This reverts commit efed39dd9da8de903705e61faaa1bed6aad64be6.

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/TargetEnvironment.tsx
@@ -10,7 +10,6 @@ import {
   TextContent,
   TextVariants,
   Tile,
-  Tooltip,
 } from '@patternfly/react-core';
 import { HelpIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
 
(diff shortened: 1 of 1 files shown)


schuellerf:
Would it hurt to keep this, as it guarded by unleash/flags anyway?
…sure adds clutter to the code 🤔🤷

ochosi:
> Would it hurt to keep this, as it guarded by unleash/flags anyway?
> …sure adds clutter to the code 🤔🤷

I thought about this as well and was wondering if a more general way of toggling support for public clouds would be better than leaving this quick hack in.

Also, keep in mind that it resizes poorly (because showing the tooltip required an additional div).

regexowl:
I think reverting in this case makes sense. As far as I understand it the flags in unleash should have limited lifespan and should be generally used for gating pre-prod:stable functionality (also there's the resizing issue).

If we want to have the ability to dynamically enable/disable certain parts of wizard without releasing we should probably think about what parts would that need to be and how to achieve it without needing tens of flags.

I'd wait with merging this until confirmation that the incident is over and the flag can go for good though.

schuellerf:
once [this](https://issues.redhat.com/browse/HMS-8592) is confirmed the incident is over from my side. Followup issues should not affect the structure (cloud region etc.) and setup of the deployment…

regexowl:
Opened a manual revert: https://github.com/osbuild/image-builder-frontend/pull/3326

PR #3265, opened 2025-05-26
Wizard: Remove edge alert

regexowl:
This removes edge deprecation alert from Wizard screen.


ochosi (approved):
Oh dear, shame on me!

PR #3268, opened 2025-05-27
dependabot: Temporarily increase PR limit

regexowl:
This increases the limit for open dependabot pull requests from 3 to 5. We currently have two bumps that will require manual bumping/migrations, leaving us with little room for any other dependencies.


ksiekl (approved):
Sounds reasonable to me :)

avitova (approved):
Okay, I do not see a reason to block this.

PR #3287, opened 2025-06-02
Wizard: edit validation of port format in the Firewall step

### LINKED ISSUE #3269: Firewall ports should allow other correct formats
We should revisit the regex for ports and update it (for example "22/tcp" should be a valid port)

ksiekl:
Fixes #3269


avitova (approved):
LGTM:)

avitova:
konflux, plz

PR #3293, opened 2025-06-03
Wizard: conflicting packages

### LINKED ISSUE #3274: Modular packages vs. non-modular ones interact strangely
I created a new RHEL 9 blueprint, and I added the `nginx-all-modules` package from the 1.24 stream:

Image

This disabled all other modular nginx packages, but the non-modular `nginx` was still available:

Image

However, when I selected it, the 1.24 version of nginx also got selected:

Image

When I built and booted the image, just the nginx 1.24 version got installed:

```console
$ rpm -qa | grep nginx
nginx-filesystem-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.noarch
nginx-core-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.x86_64
nginx-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.x86_64
nginx-mod-http-image-filter-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.x86_64
nginx-mod-http-perl-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.x86_64
nginx-mod-http-xslt-filter-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.x86_64
nginx-mod-mail-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.x86_64
nginx-mod-stream-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.x86_64
nginx-all-modules-1.24.0-4.module+el9.5.0+22954+e2d70a1c.1.noarch
```

When I later revisited the blueprint, I found this state on the package screen that nicely summarizes what happened:

Image

---

I guess the solution is to disable also all non-modular packages that are also available in enabled module streams?

ksiekl:
Disable a package if it's a conflicting module stream or if it's a non-modular package whose base name is already covered by an enabled module stream.

Fixes #3274

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -1075,12 +1075,16 @@ const Packages = () => {
    * @returns Package (or group) is / is not selected
    */
   const isSelectDisabled = (pkg: IBPackageWithRepositoryInfo) => {
-    return (
-      (pkg.type === 'module' &&
-        modules.some((module) => module.name === pkg.module_name) &&
-        !modules.some((p) => p.stream === pkg.stream)) ||
-      false
-    );
+    const isModuleStreamConflict =
+      pkg.type === 'module' &&
+      modules.some((module) => module.name === pkg.module_name) &&
+      !modules.some((p) => p.stream === pkg.stream);
+
+    const isNonModuleDisabledByModule =
+      (!pkg.type || pkg.type === 'package') &&
+      modules.some((module) => module.name === pkg.name);
+
+    return isModuleStreamConflict || isNonModuleDisabledByModule;
   };
 
   const formatDate = (date: string | undefined) => {


regexowl:
Thanks for looking into this! I think we should also update the `isPackageSelected` so the checkbox doesn't autoselect and then gets disabled when a module is selected:

image

@ondrejbudai do you think the module checkboxes should get disabled when package is selected?

image

Otherwise we should probably fix a bug where you can select a package, a module with the same name and removing the module removes the package as well. It looks like the module info overwrites the package data.

ksiekl:
Hmm, I got a little lost in between the lines of all the modules, streams, and packages, but I believe this is what we want: The package's checkbox doesn't get enabled if a module with the same name is already chosen, module selection now ensures both name and stream match, and the logic of finding conflicts between modules stayed the same as in my commit before. I hope that's right, if it doesn't check out @regexowl, please tell me :smile:

regexowl (commented):
Looks great! The only remaining issue I've noticed is if you select a package first, the modules with the same name do not get disabled:

image

regexowl (approved):
Works great, thanks!

PR #3296, opened 2025-06-03
Remove the Edge Management federated module (HMS-8637)

ochosi:
This pull request removes the federated Edge Management module from the frontend.
With the service being retired on July 31, we want to drop all code related to it.

**Note that this pull request does not yet remove the EdgeAPI**, because I wasn't sure if we wanted to drop both at the same time. A separate pull request will be required to also remove that part.

JIRA: [HMS-8637](https://issues.redhat.com/browse/HMS-8637)


regexowl:
Merging is blocked until next week for now.

regexowl (approved):
We're unblocked, let's do this :rocket:

PR #3297, opened 2025-06-04
playwright: Cleanup fails gracefully if no BP is created (HMS-6035)

tkoscieln:
If no blueprint was created, the cleanup function fails gracefully without throwing error. Also includes a fix for a missing navigation to the IB landing page.


kingsleyzissou (approved):
Thanks

PR #3298, opened 2025-06-05
migrate on-prem to PF6 (HMS-8644)

kingsleyzissou:
The PF6 migration did not completely work for the on-prem version of the frontend. This PR
makes the necessary fixes for this.

PR #3307, opened 2025-06-11
Blueprints: disable analytics for on-prem delete

avitova:
Analytics broke on prem deletion of the blueprint.


croissanne:
I think this is captured in https://github.com/osbuild/image-builder-frontend/pull/3298 as well

croissanne:
oh no, it's not

croissanne:
thanks!

PR #3310, opened 2025-06-13
TargetEnvironment: Add OpenShift Virtualization

ochosi:
This is still a draft, I'm not sure if maybe some of the APIs are generated.


regexowl:
@ochosi Yes, the APIs are generated. What needs to be done is to update API files in `/api/schema` and then run `npm run api:generate`. This should re-generate relevant files in `/src/store`.

regexowl:
OpenShift Virtualization is now in API. I've updated this PR that should get the changes in https://github.com/osbuild/image-builder-frontend/pull/3331, or you can follow the steps above and it should get rid of the manual changes failure :)

regexowl:
Waiting for image-builder-crc changes.

PR #3317, opened 2025-06-16
Wizard: change order of the systemd services (HMS-8687)

ksiekl:
Changes the order of services in the systemd step we found confusing in Justin's demo

JIRA: [HMS-8687](https://issues.redhat.com/browse/HMS-8687)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Firewall/components/Services.tsx
+++ b/src/Components/CreateImageWizard/steps/Firewall/components/Services.tsx
@@ -22,19 +22,6 @@ const Services = () => {
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


regexowl:
Looks great! Should we switch the enabled and disabled on the Firewall step as well for consistency?

kingsleyzissou:
> Looks great! Should we switch the enabled and disabled on the Firewall step as well for consistency?

Yeah thanks, I already forgot about this 😅 
Agreed, we should probably update the firewall step too

regexowl (approved):
Looks great!

PR #3321, opened 2025-06-16
Wizard: Fix package search bug with RH repos in template

dominikvagner:
## Summary
This fixes a bug inside of the package search step where packages from additional (not the 2 default ones) RH repos were not coming up when searching. This only happens when using a content template, as that's currently the only way to add more RH repos.
Also fixes a small issue in the repositories step where multiple distributions weren't concatenated correctly together if a repo was for more than one distribution.

Issue Link: [[HMS-8559🔗](https://issues.redhat.com/browse/HMS-8559)]

## Testing steps
1. Create a template and include the `CodeReady` RH repo alongside the default ones.
2. Use that template in the `Repeatable Build` step.
3. Verify that you can search for packages in the `CodeReady` repo in the `Additional Packages` step.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -70,6 +70,7 @@ import {
   useSearchPackageGroupMutation,
   ApiSearchRpmResponse,
   ApiPackageSourcesResponse,
+  useGetTemplateQuery,
 } from '../../../../store/contentSourcesApi';
 import { useAppSelector } from '../../../../store/hooks';
 import { Package } from '../../../../store/imageBuilderApi';
@@ -89,6 +90,7 @@ import {
   addModule,
   removeModule,
   selectModules,
+  selectTemplate,
 } from '../../../../store/wizardSlice';
 import {
   getEpelDefinitionForDistribution,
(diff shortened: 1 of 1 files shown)


dominikvagner:
oh, didn't though of that 💭😅
I tried my best to fix it and the only thing that seems to do so is this, which I am not really happy with :/ 
nothing else that I tried did the job tho _(is on premise checks, some redux toolkit query setting and more)_ 😮‍💨 I am not really familiar with `redux/redux-toolkit` so that might be skill issue on my part 😆 
but if you think this can be done in a nicer way then please try or give me some tips and I will, if this isn't that big of an issue and is an acceptable solution then good! _(tested it and it worked in cockpit and in the service)_ 😄 

```diff
❯ gd
diff --git a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
index ad65c2d9..a7a8f0c8 100644
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -331,7 +331,6 @@ const Packages = () => {
     isSuccessDistroRepositories,
     distroRepositories,
     arch,
-    reposInTemplate,
     template,
     distribution,
     debouncedSearchTermIsGroup,
@@ -1466,7 +1465,6 @@ const Packages = () => {
     activeSortDirection,
     activeSortIndex,
     template,
-    reposInTemplate,
   ]);
```

regexowl:
> oh, didn't though of that 💭😅
> 
> I tried my best to fix it _(the infinite search in cockpit loop this change caused)_ and the only thing that seems to do so is this, which I am not really happy with :/ nothing else that I tried did the job tho _(is_on_premise checks, some redux toolkit query settings and more)_ 😮‍💨 I am not really familiar with `redux` or `redux-toolkit` so that might be skill issues on my part 😆 but if you think this can be done in a nicer way then please try or give me some tips and I will, if this isn't that big of an issue and is an acceptable solution then good! _(tested it and it worked in cockpit and in the service)_ 😄

Hmm, this looks like a tough one. My first instinct was to skip the queries when in on-prem, but that didn't really do the trick. I'd say I'm fine with this solution tbh. The repos don't get "stuck" in the Wizard when a template is added/removed and the search seems to be working just fine :shrug: Thanks a lot for looking into it! :heart:

regexowl (approved):

PR #3331, opened 2025-06-18
api: Update pull.sh, regenerate schemas and fix errors (HMS-8720)

regexowl:
This adds missing schema links to the `pull.sh` script, runs `npm run api` and updates code where needed.

JIRA: [HMS-8720](https://issues.redhat.com/browse/HMS-8720)


regexowl:
There were some API schemas missing from the `pull.sh` script. I'm not sure if there's a specific reason though. I'd say we should keep up to date with all the schemas to avoid problems.

kingsleyzissou (commented):
LGTM with one small nitpick, I would squash the 2nd and 3rd commit. The reason is then each individual commit compiles without an error. 

Edit: probably even all three together since the first commit might cause the manual api test to fail

PR #3334, opened 2025-06-18
GHA: enable the stale action to delete its saved state

thozza:
It turns out that the stale action is not able to delete its saved state due to missing permissions. As a result, it was not processing issues and PRs, that have been processed once, for almost a month.

The error in the job log was:
```
Warning: Error delete _state: [403] Resource not accessible by integration
```

The fix is to add `actions: write` to the action permissions


regexowl (approved):
Thank you!

PR #3336, opened 2025-06-18
Fix tests

kingsleyzissou:
With the switch from tiles to cards there were some missing fields in the 
Card components. This was causing some warnings to be spammed in the test output.
I fixed those up and took some time to clear up the other noisy warnings in the unit tests.


kingsleyzissou:
Fixed the tests but broke the playwright tests 😐

regexowl (approved):
Much better :relieved: thank you!

PR #3339, opened 2025-06-19
Wizard: Remove repository column from packages table (HMS-8701)

regexowl:
This removes the column with repository information from package table as it could be confusing and misleading.


avitova (approved):
I approve. Can you resolve the conflict?:)

avitova (approved):
I will approve again, let's see how we can merge with failures.

regexowl:
> I will approve again, let's see how we can merge with failures.

Thank you! The packit checks are not required, so should be fine. I'll keep an eye on flakes, there's been quite a few lately

PR #3342, opened 2025-06-19
PackageRecommendations: Add distribution to package recommendations

mgold1234:
- Add required distribution field to RecommendPackageRequest
- Add modelVersion field to RecommendationsResponse
- Update frontend to send RHEL major version to API
- Add analytics tracking for distribution and modelVersion

Enables version-specific package recommendations and model usage tracking.
JIRA: [HMS-8634](https://issues.redhat.com/browse/HMS-8634)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
@@ -82,6 +82,7 @@ const PackageRecommendations = () => {
           recommendPackageRequest: {
             packages: packages.map((pkg) => pkg.name),
             recommendedPackages: 5,
+            distribution: distribution.replace('-', ''),
           },
         });
 
@@ -98,6 +99,7 @@ const PackageRecommendations = () => {
               isPreview: isBeta(),
               shownRecommendations: response.data.packages,
               selectedPackages: packages.map((pkg) => pkg.name),
+              distribution: distribution.replace('-', ''),
             }
           );
         }
@@ -266,6 +268,7 @@ const PackageRecommendations = () => {
                                     (pkg) => pkg.name
                                   ),
                                   shownRecommendations: data.packages,
+                                  distribution: distribution.replace('-', ''),
                                 }
                               );
                               addRecommendedPackage(pkg);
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
```diff
@@ -45,6 +45,18 @@ const PackageRecommendations = () => {
...
+    const distroLower = distribution.toLowerCase();
+    if (distroLower.includes('rhel') && distroLower.includes('8'))
+      return 'rhel-8';
+    if (distroLower.includes('rhel') && distroLower.includes('9'))
+      return 'rhel-9';
+    if (distroLower.includes('rhel') && distroLower.includes('10'))
+      return 'rhel-10';
+    if (distroLower.includes('rhel')) return 'rhel-9';
+    return 'rhel-9';
+  };
```
regexowl, line 59: Distribution is already in the needed format, isn't it?

image

src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
```diff
@@ -82,6 +82,7 @@ const PackageRecommendations = () => {
           recommendPackageRequest: {
             packages: packages.map((pkg) => pkg.name),
             recommendedPackages: 5,
+            distribution: distribution,
```
regexowl, line 85: Looks great! :heart: Let's update this
```suggestion
            distribution: distribution.replace('-', ''),
```

to fix the `Recommendations couldn't be fetched` error and we're ready to merge :rocket:

You can test the recommendations for RHEL 9 with `blktrace` package, I can see some recommendations there :sparkles:

src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
```diff
@@ -98,6 +99,7 @@ const PackageRecommendations = () => {
               isPreview: isBeta(),
               shownRecommendations: response.data.packages,
               selectedPackages: packages.map((pkg) => pkg.name),
+              distribution: distribution,
```
regexowl, line 102: I think we could maybe leave the distro in analytics as is, but it also might be nice to have the format consistent with the data we are sending :thinking:  Let's maybe update this and the analytics bellow to also remove the dash? Then it might be good to put the replace in a constant so we can just pass that.

regexowl (approved):
Works great, thank you!

PR #3344, opened 2025-06-19
BlueprintCard: Make hand cursor appear on hover

lucasgarfield:
This commit fixes an issue where if both isClickable and isSelectable are passed to a card in Patternfly 6, the hand cursor no longer appears on hover and it is not obvious the card is clickable.


lucasgarfield:
This PR: 
image

avitova:
Thanks for this! However, the failure is related to changes somehow. Can you try to address the failure?

regexowl (approved):
Works great, thank you :sparkles:

PR #3351, opened 2025-06-20
Wizard: hide other repos for on-prem (HMS-6135)

kingsleyzissou:
There is no way of configuring 3rd party repos for on-prem as of yet, we should hide this tab from the users.

Fixes: HMS-6135


src/Components/sharedComponents/RenderComponent.tsx
regexowl, line 1: Should this be a part of the PR?

regexowl (commented):
One question, otherwise works well!

regexowl (approved):
Looks good. thanks!

PR #3353, opened 2025-06-23
Wizard: Update package recommendations description

regexowl:
This adds more information to the recommended packages expandable to highlight that:
- all needed dependencies will be included by default
- recommendations are optional and based on choices of other users

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
@@ -210,8 +210,10 @@ const PackageRecommendations = () => {
               <>
                 <Content>
                   <Content>
-                    Other users commonly add these packages with the ones you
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/steps/Packages/PackageRecommendations.tsx
```diff
@@ -210,8 +210,10 @@ const PackageRecommendations = () => {
               <>
                 <Content>
                   <Content>
-                    Other users commonly add these packages with the ones you
-                    selected.
+                    Dependencies for your selected packages will be
+                    automatically included in the blueprint. The recommended
+                    packages are optional suggestions based on what other users
+                    add in similar setups.
```
andreasn, line 216: I suggest turning these statements around. First explain what the feature does, then specify that they are not dependencies.

PR #3368, opened 2025-06-26
Wizard: Show error with duplicated values (HMS-6306)

regexowl:
This show an error when user imports duplicate values to a `<LabelInput>` field.

errors

### DIFF
--- a/src/Components/CreateImageWizard/LabelInput.tsx
+++ b/src/Components/CreateImageWizard/LabelInput.tsx
@@ -48,30 +48,36 @@ const LabelInput = ({
   const dispatch = useAppDispatch();
 
   const [inputValue, setInputValue] = useState('');
-  const [errorText, setErrorText] = useState(stepValidation.errors[fieldName]);
+  const [onStepInputErrorText, setOnStepInputErrorText] = useState('');
+  let [invalidImports, duplicateImports] = ['', ''];
+
+  if (stepValidation.errors[fieldName]) {
+    [invalidImports, duplicateImports] =
+      stepValidation.errors[fieldName].split('|');
+  }
 
   const onTextInputChange = (
     _event: React.FormEvent<HTMLInputElement>,
     value: string
   ) => {
     setInputValue(value);
-    setErrorText('');
+    setOnStepInputErrorText('');
   };
 
   const addItem = (value: string) => {
     if (list?.includes(value) || requiredList?.includes(value)) {
-      setErrorText(`${item} already exists.`);
+      setOnStepInputErrorText(`${item} already exists.`);
       return;
     }
 
     if (!validator(value)) {
-      setErrorText('Invalid format.');
+      setOnStepInputErrorText('Invalid format.');
       return;
     }
 
     dispatch(addAction(value));
     setInputValue('');
-    setErrorText('');
+    setOnStepInputErrorText('');
   };
 
   const handleKeyDown = (e: React.KeyboardEvent, value: string) => {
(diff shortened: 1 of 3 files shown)


ochosi:
How does a user end up in this situation? Copy-pasting values into the input box? Or really adding duplicates manually?

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -48,30 +48,36 @@ const LabelInput = ({
   const dispatch = useAppDispatch();
 
   const [inputValue, setInputValue] = useState('');
-  const [errorText, setErrorText] = useState(stepValidation.errors[fieldName]);
+  const [onStepInputErrorText, setOnStepInputErrorText] = useState('');
+  let [invalidImports, duplicateImports] = ['', ''];
+
+  if (stepValidation.errors[fieldName]) {
+    [invalidImports, duplicateImports] =
+      stepValidation.errors[fieldName].split('|');
```
ksiekl, line 56: This is a really nice update! I was thinking that sometimes we need to show multiple errors! :rocket:  The "|" concatenation is a cool idea and it works nicely, however I was thinking that maybe more scalable approach would be to have a better StepValidation type, like:
```
export type StepValidation = {
  errors: {
    [key: string]: string[];
  };
  disabledNext: boolean;
};

```
That way, you don't have to remember which set of errors goes first, and you don't have to split all the strings.

On the other hand, I'm not 100% sure it's needed that much - and it would take a lot of rewriting code.... So, I'm going to be annoying and leave that up to you :smile: this works for me.

src/Components/CreateImageWizard/utilities/getListOfDuplicates.ts
```diff
@@ -0,0 +1,6 @@
+export const getListOfDuplicates = (list: string[]) => {
+  const duplicates = list.filter((item, index) => list.indexOf(item) !== index);
+  const uniqueDuplicates = [...new Set(duplicates)];
+
+  return uniqueDuplicates;
```
ksiekl, line 5: nice oxymoron :)

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -125,9 +130,21 @@ const LabelInput = ({
...
+            <HelperTextItem variant={'error'}>
+              {onStepInputErrorText}
+            </HelperTextItem>
+          )}
+          {invalidImports && (
+            <HelperTextItem variant={'error'}>{invalidImports}</HelperTextItem>
+          )}
+          {duplicateImports && (
+            <HelperTextItem variant={'error'}>
+              {duplicateImports}
```
ksiekl, line 145: these messages all return the same HelperTextItem with the same variant

regexowl: Fair enough! I've removed some of the repetition :sweat_smile:

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -125,9 +135,13 @@ const LabelInput = ({
...
       </TextInputGroup>
-      {errorText && (
+      {errors.length > 0 && (
         <HelperText>
-          <HelperTextItem variant={'error'}>{errorText}</HelperTextItem>
+          {errors.map((error) => (
+            <HelperTextItem variant={'error'}>
+              {error}
+            </HelperTextItem>
+          ))}
```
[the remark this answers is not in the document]
regexowl: I know, I know, linter already screamed at me :smiling_face_with_tear:

src/Components/CreateImageWizard/LabelInput.tsx
ksiekl: Heh, sorry, in my mind I was still thinking about that StepValidation type differently, with it we could just have one HelperTextItem for all errors, but now you still had to push all the different errors into the `errors`, so some repetition was still needed :) Sorry for that confusion, but it looks gucci! :heart:

ksiekl (approved):

avitova:
Thanks, this helped a lot with getting the import further.

PR #3370, opened 2025-06-27
playwright: Add check to ensure test is authenticated

tkoscieln:
In order to fix the flakiness in cockpit plugin tests introduced by the single login, the tests need a check if they are authenticated at the start of their execution and relog in case the session expired.

### DIFF
--- a/playwright/Customizations/Filesystem.spec.ts
+++ b/playwright/Customizations/Filesystem.spec.ts
@@ -4,6 +4,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { FILE_SYSTEM_CUSTOMIZATION_URL } from '../../src/constants';
 import { test } from '../fixtures/cleanup';
 import { isHosted } from '../helpers/helpers';
+import { ensureAuthenticated } from '../helpers/login';
 import {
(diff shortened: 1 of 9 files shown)


playwright/Customizations/Filesystem.spec.ts
```diff
@@ -28,6 +29,9 @@ test('Create a blueprint with Filesystem customization', async ({
   // Delete the blueprint after the run fixture
   await cleanup.add(() => deleteBlueprint(page, blueprintName));
 
+  // Ensure authentication before starting
```
kingsleyzissou, line 32: Small nitpick:
The function has a good name and describe's well what it does. So the comment is a bit redundant.
I think a good principle for comments that the broader team uses is to favour comments that explain
why we're doing something rather than what it's doing. That way comments are only really necessary
the why isn't very clear or there's a specific logic for doing things that way.

:)

kingsleyzissou (commented):
I've added a small nitpick. Obviously since this is quite urgent we won't block on it

Will see how the tests do 🤞

regexowl (approved):
Thank you!

PR #3371, opened 2025-06-27
Disable f41 schutz test

kingsleyzissou:
The tests are really flaky at the moment, let's disable them until they have been fixed.
This should give us more time to debug the tests and figure out what's going on.

We can re-enable this with #3370


kingsleyzissou:
> The flakiness plagues also Fedora 42 and RHEL 9.6 runners. We could probably mark Schutzbot tests as not required for now. I believe the test will not run in merge queue then.

Maybe admin merging is a better option than marking those as not required

PR #3379, opened 2025-06-30
Wizard: Render labels for all FSC table columns

regexowl:
This adds column names for the Suffix and Unit columns that were previously not labeled.


ochosi (approved):
LGTM!

PR #3393, opened 2025-07-02
Wizard: Replace VMware radios with checkboxes (HMS-8778)

regexowl:
This replaces previously used VMware radios with checkboxes, allowing the user to select both VMware vSphere targets at the same time.

JIRA: [HMS-8778](https://issues.redhat.com/browse/HMS-8778)


src/test/Components/CreateImageWizard/steps/ImageOutput/ImageOutput.test.tsx
```diff
@@ -265,44 +259,6 @@ describe('Step Image output', () => {
     );
   });
 
-  test('VMware checkbox select and unselect works', async () => {
```
regexowl, line 268: This tested if switching between the radios and toggling parent category works correctly. Since we have simple checkboxes now it's no longer needed.

kingsleyzissou (approved):
Neat works nicely

PR #3395, opened 2025-07-03
Remove IQE scripts and configs (HMS-8779)

regexowl:
Since we're no longer testing via IQE, we can say bye bye to the `pr_check.sh` script and `iqe-trigger-integration.yml`.

JIRA: [HMS-8779](https://issues.redhat.com/browse/HMS-8779)


tkoscieln (approved):
I think this should be good to go, let's see what the test do, but LGTM

PR #3396, opened 2025-07-03
Remove devel dockerfile and update README (HMS-8780)

regexowl:
Since local fullstack deployment is being done via https://github.com/osbuild/osbuild-getting-started, we can get rid of the original dockerile.

README was also updated to no longer point to the `devel` folder as it just points you to the same repository. osbuild-getting-started is mentioned in the "Backend development" section.

JIRA: [HMS-8780](https://issues.redhat.com/browse/HMS-8780)


kingsleyzissou:
> I believe the dockerfile for `npm run devel` is no longer needed, would that be correct @croissanne? Not sure if it was used for anything else.

That was added in 5a34506eb6a4043bbe507514459c79f41c340ada. I think it was just a separate webpack config for development because of the proxy. So ya, not needed anymore :)

kingsleyzissou (approved):

PR #3397, opened 2025-07-03
Run actions with Node 22

regexowl:
The github actions were previously ran with Node 20, this bumps them to Node 22.

README was also updated to reflect currently used version of Node.


regexowl:
Not sure if the `build_deploy.sh` script is still necessary. It was added here: https://github.com/osbuild/image-builder-frontend/pull/828 and I'd assume is no longer in use since we migrated to konflux, but :shrug:

PR #3398, opened 2025-07-03
fix: landing page sticky header (HMS-8708)

cdcabrera:
Adds Patternfly sticky header styling to avoid end user confusion around hidden buttons.

Updates HMS-8708

Jul-03-2025 14-41-06
Jul-03-2025 14-55-40


ochosi:
Love it - ship it! 🚀

PR #3423, opened 2025-07-14
Wizard: Update Satellite registration based on mocks (HMS-9414)

regexowl:
This updates Satellite registration based on final revamp mocks.

JIRA: [HMS-9414](https://issues.redhat.com/browse/HMS-9414)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Registration/components/Registration.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/components/Registration.tsx
@@ -17,88 +17,96 @@ const Registration = () => {
 
   return (
     <FormGroup label='Registration method'>
-      <Radio
-        label='Automatically register and enable advanced capabilities.'
-        isChecked={
-          registrationType === 'register-now' ||
-          registrationType === 'register-now-insights' ||
-          registrationType === 'register-now-rhc'
-        }
-        onChange={(_event, checked) => {
-          if (checked) {
-            dispatch(changeRegistrationType('register-now-rhc'));
+      <Content className='pf-v6-u-pb-sm'>
+        <Radio
+          label='Automatically register and enable advanced capabilities.'
+          isChecked={
+            registrationType === 'register-now' ||
+            registrationType === 'register-now-insights' ||
+            registrationType === 'register-now-rhc'
           }
-        }}
-        id='register-system-now'
-        name='register-system-now'
-        autoFocus
-        body={
-          <>
-            <Content component='p'>
-              <Switch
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/steps/Registration/components/RegisterLaterAlert.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { INSIGHTS_URL, RHC_URL } from '../../../../../constants';
+
+const RegisterLaterAlert = () => {
+  return (
+    <Alert
+      variant="info"
+      title={
+        <>
+          You should register your image now if you may want to connect to Red
```
kingsleyzissou, line 14: Is this the copy we were given?

```
You should register your image now if you would like to connect to Red Hat insights...
```

Sounds slightly better to me.

regexowl: Yup, this is taken directly from the mocks: https://www.figma.com/design/9t3owYlJi4M23fUXWpgXCE/HMS-Image-Builder?node-id=3230-48342&t=r5LcLuq1Se2acEyk-0

regexowl: @kelseamann what do you think?

regexowl:
Waiting for copy update.

ksiekl (dismissed):
Thanks! It looks good to me, I just have some little nit picks but they are not blockers:
1. The "Manage Activation keys" button has lowercase "A" in the mocks, but that's such a small thing I don't think it matters
2. After clicking the "View details" button, among the info about the key there is also a "Description" field, that is not in the mocks, and if a key doesn't have description, it's value just stays empty and it looks a bit funny. I think the description field is nice, but maybe we shouldn't show it if it's empty, or put in the "Not defined" value like everywhere else?
image

src/Components/CreateImageWizard/steps/Registration/components/SatelliteDocumentationButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <Button
+      component='a'
+      target='_blank'
+      variant='link'
+      icon={<ExternalLinkAltIcon />}
+      iconPosition='right'
+      isInline
+      href={REGISTRATION_DOCS_URL}
+    >
```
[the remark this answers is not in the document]
regexowl: If that's correct we should probably update it in more places :face_with_peeking_eye:

PR #3439, opened 2025-07-17
Wizard: Fix stuck page for package search

### LINKED ISSUE #3381: No page reset after different search value
When going from the searchbar value that has lots of results to the value that has a few, if you end on some of the later pages of packages, there is this weird of `x out of y` where `x > y`, it goes to the last page, but the last page is empty for some reason, unless you click back and forward. Attaching the 

https://github.com/user-attachments/assets/6cbddb27-ff2a-4efe-9873-67e958c6fc76

regexowl:
Fixes #3381

This resets the paging to page 1 when search term is updated.


kingsleyzissou (approved):
Thx!

PR #3455, opened 2025-07-23
playwright: add timeout for cockpit tests

kingsleyzissou:
Playwright at times is making assertions before the osbuild-worker file is loaded. This is causing some flakiness in the tests, adding a timeout before some of the assertions should hopefully remove some of the flakiness.


regexowl (dismissed):
Nice! Let's give it a try! :rocket:

kingsleyzissou:
Still flaking :/ but also on the service tests. I'll have another look

regexowl (approved):

PR #3456, opened 2025-07-23
test: Remove typing delay for locale test (HMS-8890)

regexowl:
`Step Locale - unknown option is disabled` was quite flaky recently. Removing typing delay from the userEvent actions should solve the issue.

JIRA: [HMS-8890](https://issues.redhat.com/browse/HMS-8890)


regexowl:
We should probably default to no typing delay throughout our tests as we have no checks that would require the latency... I bet this will bring a ton of async issues though. Might take a look at it anyway :shrug:

PR #3459, opened 2025-07-24
test: Move `user` definition to wizard test utils (HMS-8912)

regexowl:
This moves `user` definition to `wizardTestUtils.tsx` so it can be shared with the test files.

The delay for user methods is also set to `null` by default as we're currently not testing with user typing latency. This should improve async behaviour and can be easily overriden for individual tests when needed.

Some early exits had to be safeguarded.

JIRA: [HMS-8912](https://issues.redhat.com/browse/HMS-8912)


regexowl:
Well, apparently this is not best practice as `userEvent` tracks several things which could lead to test bleeds and even more flakiness... so yeah, just closing this and won't touch it again.

PR #3463, opened 2025-07-25
ESLint: Set `no-unused-vars` rule to error (HMS-8923)

regexowl:
This rule should output error instead of just a warning.

JIRA: [HMS-8923](https://issues.redhat.com/browse/HMS-8923)


mgold1234 (approved):
/lgtm

PR #3479, opened 2025-07-30
Wizard: Remove alert from OpenSCAP step (HMS-8950)

regexowl:
Information about required packages, services and kernel arguments is back on the step. Meaning we can remove the alert as per UX recommendation.

JIRA: [HMS-8950](https://issues.redhat.com/browse/HMS-8950)


mgold1234 (approved):
/lgtm

PR #3486, opened 2025-07-31
Wizard: Add tooltip explaining password visibility in edit mode (HMS-8968)

### LINKED ISSUE #3303: Wizard: psswd for a user is not shown when editing a blueprint
Image
The eye icon should be clickable when editing a blueprint, I believe. Or maybe the option to see it should not be there at all.

mgold1234:
**Summary**
When editing a blueprint, password fields show an eye icon that appears clickable but is actually disabled for security reasons (passwords cannot be retrieved from the backend). This creates confusing UX where users expect the icon to work but it doesn't respond.

This change adds a tooltip that appears when hovering over the disabled eye button, explaining that "Passwords cannot be viewed when editing a blueprint for security reasons."

The fix wraps the disabled button in a span element to ensure the tooltip triggers properly, as disabled buttons don't receive mouse events.

**Jira**
[HMS-8968](https://issues.redhat.com/browse/HMS-8968)
Fixes #3303

🤖 Generated with AI

**Screen Captures**
| Before                                              | After                                   |
| --------------------------------------------------- | --------------------------------------- |
 | Screenshot 2025-07-31 at 17 22 22| Screenshot 2025-07-31 at 17 22 08 |

### DIFF
--- a/src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
@@ -10,6 +10,7 @@ import {
   InputGroupItem,
   TextInput,
   TextInputProps,
+  Tooltip,
 } from '@patternfly/react-core';
 import { EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
 
@@ -46,6 +47,21 @@ export const PasswordValidatedInput = ({
     setIsPasswordVisible(!isPasswordVisible);
   };
 
+  const isEditingWithoutValue = hasPassword && !value;
+
+  const PasswordToggleButton = () => {
+    return (
+        <Button
+            variant="control"
+            onClick={togglePasswordVisibility}
+            aria-label={isPasswordVisible ? 'Hide password' : 'Show password'}
+            isDisabled={isEditingWithoutValue}
+        >
+          {isPasswordVisible ? <EyeSlashIcon /> : <EyeIcon />}
+        </Button>
+    );
+  };
+
   return (
     <FormGroup label="Password" className="pf-v6-u-pb-md">
       <>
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -61,14 +62,29 @@ export const PasswordValidatedInput = ({
...
+                <Tooltip content="Passwords cannot be viewed when editing a blueprint for security reasons">
+                <span>
+                  <Button
+                      variant="control"
+                      onClick={togglePasswordVisibility}
+                      aria-label={isPasswordVisible ? 'Hide password' : 'Show password'}
+                      isDisabled={true}
+                  >
+                    {isPasswordVisible ? <EyeSlashIcon /> : <EyeIcon />}
+                  </Button>
```
regexowl, line 75: Can you please move the button to a reusable component to remove the duplicate code?

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -61,14 +62,29 @@ export const PasswordValidatedInput = ({
...
           <InputGroupItem>
-            <Button
-              variant="control"
-              onClick={togglePasswordVisibility}
-              aria-label={isPasswordVisible ? 'Hide password' : 'Show password'}
-              isDisabled={hasPassword && !value}
-            >
-              {isPasswordVisible ? <EyeSlashIcon /> : <EyeIcon />}
-            </Button>
+            {hasPassword && !value ? (
```
regexowl, line 65: `hasPassword && !value` could be also put into a constant, which could be reused for disabling the button

src/test/Components/CreateImageWizard/steps/Users/Users.test.tsx
```diff
@@ -506,4 +506,28 @@ describe('Users edit mode', () => {
...
+    expect(passwordButtons).toHaveLength(0);
+    
+    const allElements = await within(expandable).findAllByRole('button');
+    const passwordElements = allElements.filter(element => 
+      element.getAttribute('aria-label')?.toLowerCase().includes('password') ||
+      element.textContent?.toLowerCase().includes('password')
+    );
+  
+    expect(passwordElements).toHaveLength(0);
+  });
```
regexowl, line 532: Got lost in the test, it doesn't really do what it says in the name of it.

There are no buttons in the users expandable on the review step so I don't think we need to check if there are or their status.

What would make sense to me is:
- render edit mode
- go to the Users step
- hover on the password visibility button
- check for the tooltip
- unhover
- make sure the tooltip disappears

the test could also double check that the button is indeed disabled, I think that would be nice. What do you think? :)

regexowl (changes requested):
Thanks for looking into this! :heart: Added a few thoughts

regexowl (approved):
Works great, thanks! The test is great :heart:

PR #3487, opened 2025-07-31
remove as much edge mgmt code as possible

ondrejbudai:
We no longer include the edge mgmt federated module so let's remove all remaining traces of the integration.

Edit: happy to see that the CI is green. However, please do review this carefully. While this was done entirely by a human, the human doesn't have too much context, so there could be some stuff that I... ehm... I mean... the human misunderstood.


regexowl:
The failure looks too consistent for a flake, but required Oscap packages shouldn't be affected by these changes :eyes:

regexowl:
Oooooh, looks like there's just eight packages for the Oscap profile being returned now, at least from what I can see in the UI. `cronie` seems to be removed. So yeah, unrelated to changes in this PR, but we'll need to fix that before we're able to merge again :thinking:

PR #3488, opened 2025-08-01
playwright: Update required Oscap packages

regexowl:
Follow up to https://github.com/osbuild/image-builder-crc/pull/1659.

This updates test to check required Oscap packages to match updated set.


kingsleyzissou (approved):
Thanks for catching this

PR #3489, opened 2025-08-01
Revert "playwright: Add duration to test account creation request (HMS-8958)"

regexowl:
Reverts osbuild/image-builder-frontend#3485

Ethel API is fixed now, we don't need to specify "duration" anymore.


mgold1234 (approved):
/lgtm

PR #3502, opened 2025-08-06
Wizard: Resolve row reordering issue on selection and expansion (HMS-8924)

mgold1234:
**Summary**
- Fix issue when clicking the expandable arrow or selecting a package checkbox in the Packages step it caused unexpected row reordering.
- Updated sorting logic to ensure that selecting a package with a specific stream groups all related module streams together at the top.
- Ensured that rows expand in place and selection does not affect row position.

**Jira**
Fixes [HMS-8924](https://issues.redhat.com/browse/HMS-8924)

**How to Test**
1) Navigate to Image Builder wizard → Packages step
2) Search for packages to populate the table
3) Click on the expandable arrow (▶) next to any package row
OR Click on the checkbox to select a package

🤖 Generated with AI

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -50,6 +50,7 @@ import {
   Thead,
   Tr,
 } from '@patternfly/react-table';
+import { orderBy } from 'lodash';
 import { useDispatch } from 'react-redux';
 
 import CustomHelperText from './components/CustomHelperText';
@@ -66,7 +67,6 @@ import {
 } from '../../../../constants';
 import { useGetArchitecturesQuery } from '../../../../store/backendApi';
 import {
-  ApiPackageSourcesResponse,
   ApiRepositoryResponseRead,
   ApiSearchRpmResponse,
   useCreateRepositoryMutation,
@@ -700,7 +700,7 @@ const Packages = () => {
       );
     }
 
-    const unpackedData: IBPackageWithRepositoryInfo[] =
+    let unpackedData: IBPackageWithRepositoryInfo[] =
       combinedPackageData.flatMap((item) => {
         // Spread modules into separate rows by application stream
         if (item.sources) {
@@ -724,13 +724,16 @@ const Packages = () => {
       });
 
     // group by name, but sort by application stream in descending order
-    unpackedData.sort((a, b) => {
-      if (a.name === b.name) {
-        return (b.stream ?? '').localeCompare(a.stream ?? '');
-      } else {
-        return a.name.localeCompare(b.name);
-      }
-    });
+    unpackedData = orderBy(
+      unpackedData,
+      [
+        'name',
+        (pkg) => pkg.stream || '',
+        (pkg) => pkg.repository || '',
+        (pkg) => pkg.module_name || '',
+      ],
+      ['asc', 'desc', 'asc', 'asc'],
+    );
 
     if (toggleSelected === 'toggle-available') {
       if (activeTabKey === Repos.INCLUDED) {
@@ -866,8 +869,6 @@ const Packages = () => {
         dispatch(addPackage(pkg));
         if (pkg.type === 'module') {
           setActiveStream(pkg.stream || '');
-          setActiveSortIndex(2);
-          setPage(1);
           dispatch(
             addModule({
               name: pkg.module_name || '',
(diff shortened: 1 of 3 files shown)


src/store/wizardSlice.ts
```diff
@@ -824,7 +824,9 @@ export const wizardSlice = createSlice({
     },
     addPackage: (state, action: PayloadAction<IBPackageWithRepositoryInfo>) => {
       const existingPackageIndex = state.packages.findIndex(
-        (pkg) => pkg.name === action.payload.name,
+        (pkg) =>
```
regexowl, line 827: This would allow user to add duplicate packages (by name) if stream was different, wouldn't it? :eyes: 

I don't think we want that and I'd assume the UI is blocking this by disabling correct checkboxes.

src/store/wizardSlice.ts
```diff
@@ -835,10 +837,12 @@ export const wizardSlice = createSlice({
     },
     removePackage: (
       state,
-      action: PayloadAction<IBPackageWithRepositoryInfo['name']>,
+      action: PayloadAction<IBPackageWithRepositoryInfo>,
     ) => {
       const index = state.packages.findIndex(
-        (pkg) => pkg.name === action.payload,
+        (pkg) =>
```
regexowl, line 843: Similar here - there can't be two packages with the same in the state, so we don't need to differentiate between them by stream... all the names in the packages array should be unique, no? A side note - when you select a specific module, you should be able to select other package from the same module only, other module version and non-modular packages should have disabled checkboxes.

src/Components/CreateImageWizard/steps/Packages/Packages.tsx
```diff
@@ -876,13 +902,35 @@ const Packages = () => {
...
+        const remainingModulesWithSameStream = modules.filter(
+          (m) => m.name !== pkg.module_name && m.stream === pkg.stream,
+        );
+        // Check if there will be other selected packages with the same stream AFTER removal
+        const remainingPackagesWithSameStream = packages.filter(
+          (p) => p.name !== pkg.name && p.stream === pkg.stream,
+        );
+
+        if (
+          pkg.stream === activeStream &&
```
regexowl, line 924: I think this is the line @ksiekl referred to here: https://github.com/osbuild/image-builder-frontend/pull/3475#discussion_r2245140787 and I agree. When you unselect package/module the sorting shouldn't reset, but stay as is.

I can't find the mocks anymore, will try to search a bit later. The behavior should be as follow:
- when you select specific stream of a module, all the packages from the active stream should get sorted to the top, followed by all the "other" (packages, other app streams) that should be disabled.
- when you unselect a package, the sorting of the table shouldn't change in any way (so really the only change in the entire table would be that previously checked checkbox will be now unchecked)

src/Components/CreateImageWizard/steps/Packages/Packages.tsx
```diff
@@ -1100,14 +1195,15 @@ const Packages = () => {
         (module) => module.name === pkg.name,
       );
       isSelected =
-        packages.some((p) => p.name === pkg.name) && !isModuleWithSameName;
+        packages.some((p) => p.name === pkg.name && p.stream === pkg.stream) &&
```
regexowl, line 1198: Information about active stream is already being stored in `packages` and `enabled_modules`, maybe we can work with that?

image

Were there any bugs with incorrect packages being selected? I though both the jumping package and jumping expandable were a problem with unstable sorting.

regexowl (changes requested):
This is quite a bit rewrite for (what I assumed) was an issue with unstable `.sort` behaviour. :sweat_smile: 

To be honest I was expecting the changes to happen mostly in the two sort blocks, but this seems to reach quite far out of scope. Like changing the way required packages for openscap/compliance are handled...

I've left some comments, let me know, what you think.

src/Components/CreateImageWizard/steps/Oscap/components/useSelectorHandlers.tsx
```diff
@@ -13,16 +13,23 @@ import {
...
 import { Partition, Units } from '../../FileSystem/components/FileSystemTable';
 
 export const useSelectorHandlers = () => {
   const dispatch = useAppDispatch();
+  const currentPackages = useAppSelector(selectPackages);
 
   const clearCompliancePackages = (oscapPackages: string[]) => {
-    for (const pkg of oscapPackages) {
-      dispatch(removePackage(pkg));
+    for (const packageName of oscapPackages) {
```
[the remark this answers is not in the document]
mgold1234: few tests were failing becuase the re-sorting that I did when user unselected the last package, 
changed the re-sroting and remove those changes

src/test/Components/CreateImageWizard/steps/Packages/Packages.test.tsx
```diff
@@ -530,6 +530,39 @@ describe('Step Packages', () => {
       expect(secondAppStreamRow).toBeDisabled();
       expect(secondAppStreamRow).not.toBeChecked();
     });
+
+    test('module selection does not reset sort and page position', async () => {
+      const user = userEvent.setup();
+
+      await renderCreateMode();
+      await goToPackagesStep();
+      await selectCustomRepo();
```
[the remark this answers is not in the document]
mgold1234: removed it :) thank you

src/Components/CreateImageWizard/steps/Packages/Packages.tsx
mgold1234: while the stream matching wasn't essential for the jumping issue, it's a logical improvement that makes the selection logic more robust and precise. Both changes work together to create a better user experience.
Before: A package was considered selected if ANY package with the same name existed in the selection, regardless of stream
After: A package is only considered selected if both name AND stream match

This prevents edge cases where packages with the same name but different streams could show incorrect selection states.

mgold1234:
@kingsleyzissou great catch, thank you I fixed that

PR #3504, opened 2025-08-06
devDeps: fix npm vulnerabilities (HMS-9010)

kingsleyzissou:
npm was complaining about some high-severity vulnerabilities. Fix these by running `npm audit fix`

JIRA: [HMS-9010](https://issues.redhat.com/browse/HMS-9010)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10214,7 +10214,9 @@
       }
     },
     "node_modules/express": {
-      "version": "4.21.1",
+      "version": "4.21.2",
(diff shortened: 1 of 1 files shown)


regexowl (approved):
Nice catch!

I'm a bit wary when it comes to `npm audit fix` (or changing transitive dependencies without updating the direct ones to be more specific), but since all the tests are green and everything seems to run just fine I'd say we're good here. 

Thank you!

PR #3511, opened 2025-08-07
Wizard: Fix lint warnings and snapshot button behaviour

regexowl:
This fixes lint warnings for Review step and updates behaviour of the snapshot button.

The button for reviewing repeatable build was permanently disabled, now it should be disabled only when there are no snapshotable repositories selected. The content of snapshot popover has updated rendering of states.


mgold1234 (approved):
/lgtm

kingsleyzissou (approved):
Nice

PR #3519, opened 2025-08-11
Wizard: reintroduce registration step for cockpit (HMS-9195)

kingsleyzissou:
We can re-enable the registration step for cockpit by allowing users to pass in an activation and org id pair so that they can build an image that is automatically registered.

JIRA: [HMS-9030](https://issues.redhat.com/browse/HMS-9030)

### DIFF
--- a/playwright/helpers/wizardHelpers.ts
+++ b/playwright/helpers/wizardHelpers.ts
@@ -1,6 +1,6 @@
 import { expect, FrameLocator, type Page, test } from '@playwright/test';
 
-import { closePopupsIfExist, isHosted } from './helpers';
+import { closePopupsIfExist, isHosted, isRhel } from './helpers';
 import { ibFrame, navigateToLandingPage } from './navHelpers';
 
 /**
@@ -45,11 +45,12 @@ export const fillInDetails = async (
 
 /**
  * Select "Register later" option in the wizard
- * This function executes only on the hosted service
+ * This function executes only on if the registration
+ * step is visible (it won't be visible for Fedora)
  * @param page - the page object
  */
 export const registerLater = async (page: Page | FrameLocator) => {
-  if (isHosted()) {
+  if (await isRhel(page)) {
     await page.getByRole('button', { name: 'Register' }).click();
     await page.getByRole('radio', { name: 'Register later' }).click();
   }
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -318,7 +318,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
       'parentId' in step && step.parentId === 'step-optional-steps';
 
     useEffect(() => {
-      if (process.env.IS_ON_PREMISE) {
+      if (!isRhel(distribution)) {
         if (step.id === 'step-oscap' && step.isVisited) {
           setWasRegisterVisited(true);
         }
@@ -460,7 +460,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 name='Register'
                 id='step-register'
                 key='step-register'
-                isHidden={!!process.env.IS_ON_PREMISE || !isRhel(distribution)}
+                isHidden={!isRhel(distribution)}
                 navItem={CustomStatusNavItem}
                 status={
                   wasRegisterVisited
--- a/src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
@@ -1,0 +1,123 @@
+import React, { MutableRefObject, useEffect, useRef } from 'react';
+
+import {
+  Button,
+  Content,
+  ContentVariants,
+  FormGroup,
+  FormGroupLabelHelp,
+  Popover,
+} from '@patternfly/react-core';
+import { ExternalLinkAltIcon } from '@patternfly/react-icons';
+
+import { ACTIVATION_KEYS_URL, CDN_PROD_URL } from '../../../../../constants';
+import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
+import {
+  changeActivationKey,
+  changeBaseUrl,
+  changeOrgId,
+  changeServerUrl,
+  selectActivationKey,
+  selectOrgId,
+} from '../../../../../store/wizardSlice';
+import { ValidatedInput } from '../../../ValidatedInput';
+
+const ManualRegistrationPopover = ({
+  ref,
+}: {
+  ref: MutableRefObject<null>;
+}) => {
+  return (
+    <Popover
+      triggerRef={ref}
+      headerContent='About Activation Keys & Organization ID'
+      position='right'
+      minWidth='30rem'
+      bodyContent={
+        <Content>
+          <Content component={ContentVariants.p}>
+            Activation keys assist you in registering and configuring systems.
+            Metadata such as role, system purpose, and usage can be
+            automatically attached to systems via an activation key, and
+            monitored with Subscription Watch.
+          </Content>
+          <Content component={ContentVariants.p}>
+            The Organization ID is the numeric identifier for your organization
+            and is separate from your account number. Your organization&apos;s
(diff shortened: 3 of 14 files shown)


kingsleyzissou:
Still have to add some playwright tests

src/Components/CreateImageWizard/steps/Review/ReviewStepTextLists.tsx
```diff
@@ -704,14 +704,15 @@ export const RegisterAapList = () => {
 };
 
 export const RegisterNowList = () => {
+  const orgId = useAppSelector(selectActivationKey);
```
regexowl, line 707: This should select org id I guess

playwright/helpers/wizardHelpers.ts
```diff
@@ -45,11 +45,12 @@ export const fillInDetails = async (
 
 /**
  * Select "Register later" option in the wizard
- * This function executes only on the hosted service
+ * This function executes only on if the registration
+ * step is visible (it won't be visibel for Fedora)
```
regexowl, line 49:
```suggestion
 * step is visible (it won't be visible for Fedora)
```

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
```diff
@@ -0,0 +1,123 @@
...
+};
+
+export const ManualActivationKey = () => {
+  const orgId = useAppSelector(selectOrgId);
+  const activationKey = useAppSelector(selectActivationKey);
+  const dispatch = useAppDispatch();
+
+  return (
+    <>
+      <FormGroup label={<ActivationKeyPopover label='Activation Key' />}>
```
regexowl, line 92: Sentence case is used in most of the wizard, let's use it here also for consistency sake

```suggestion
      <FormGroup label={<ActivationKeyPopover label='Activation key' />}>
```

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    <>
+      {label}
+      <Popover
+        headerContent='About Organization ID'
+        position='right'
+        minWidth='30rem'
+        bodyContent={
+          <Content>
+            The organization ID is the numeric identifier for your organization
```
regexowl, line 68: "organizaton" starts with capital O in the popover header, but with lowercase o in the content, we should probably stick to one of these :nerd_face:

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
```diff
@@ -0,0 +1,123 @@
...
+    </>
+  );
+};
+
+const OrgIdPopover = ({ label }: { label: string }) => {
+  return (
+    <>
+      {label}
+      <Popover
+        headerContent='About Organization ID'
```
regexowl, line 63: Should we include info about how/where to find the org id?

src/Components/CreateImageWizard/steps/Registration/index.tsx
```diff
@@ -76,6 +79,9 @@ const RegistrationStep = () => {
             <ActivationKeyInformation />
           </FormGroup>
         )}
+      {process.env.IS_ON_PREMISE && registrationType !== 'register-later' && (
```
regexowl, line 82: There's slightly weird behaviour - when you select "Register later" go to another step and then return to Registration the "additional connection options" checkbox array gets automagically expanded :magic_wand: 

expand

regexowl (commented):
Nice! Added some small nit picks :)

As for copy - looks good to me, we can also check with UX or with someone from docs to get their feedback.

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
kingsleyzissou: Ya I initially had this but there were a couple of places it can be done so it got a bit messy.

Will try see how we can try manage this

kingsleyzissou: I merged the two popovers together since they're so closely related. I put a popover with the same text next to both fields and the copy was taken from the red hat docs and console dot.

Also I updated the popover field slightly so it's more consistent with the [patternfly docs](https://www.patternfly.org/components/forms/form#basic)

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <Button
+            component='a'
+            target='_blank'
+            variant='link'
+            icon={<ExternalLinkAltIcon />}
+            iconPosition='right'
+            isInline
+            href={ACTIVATION_KEYS_URL}
+          >
+            Vie activation keys in Hybrid Cloud Console
```
regexowl, line 55:
```suggestion
            View activation keys in Hybrid Cloud Console
```

otherwise looks good to me!

kingsleyzissou: Agh flip

regexowl:
Oh right, looking at the unit test failure - the "additional connection options" should be hidden by default. So the original `useState` logic was correct. :thinking: The options opening on their own is a reasonably small bug so maybe we could revisit it in a follow up. Something is probably messing up with the state setter.

kingsleyzissou:
> Oh right, looking at the unit test failure - the "additional connection options" should be hidden by default. So the original `useState` logic was correct. 🤔 The options opening on their own is a reasonably small bug so maybe we could revisit it in a follow up. Something is probably messing up with the state setter.

Ya I might try look at it now, graphite decided to throw a few more useful comments at me too

src/Components/CreateImageWizard/utilities/useValidation.tsx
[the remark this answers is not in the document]
kingsleyzissou: Ah good shout, thanks will have a look at this today

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -460,7 +460,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 name='Register'
                 id='step-register'
                 key='step-register'
-                isHidden={!!process.env.IS_ON_PREMISE || !isRhel(distribution)}
+                isHidden={!isRhel(distribution)}
```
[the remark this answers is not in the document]
kingsleyzissou: I guess for now no, we don't support satellite yet for on -prem. But yah this will be relevant when we do eventually add that

avitova:
@kingsleyzissou I feel like it got a bit smarter, it is spamming throughout my PRs lately a bit more as well.

avitova:
Yeah, it did. Okay, my last question is - because I assume we do not want to expose activation keys when the user clicks "Edit blueprint", is there a plan on what to do, so that user is able to both a) keep activation keys after editing a blueprint, but also b) check "Register later"?
But overall I am good here.

kingsleyzissou:
> Yeah, it did. Okay, my last question is - because I assume we do not want to expose activation keys when the user clicks "Edit blueprint", is there a plan on what to do, so that user is able to both a) keep activation keys after editing a blueprint, but also b) check "Register later"?
> But overall I am good here.

Ya great questions. For point a) I'm not entirely sure how we want to handle this. Since this should theoretically only be visible to the user that creates the blueprint, we should be fine. But maybe we do want to hide this?

For b) there is a small back when going back and clicking register later. I will fix that in a follow up

kingsleyzissou:
So I think this might not land for the ITM release, I have done a run through and we can add the subscription details.
But it looks like the final image doesn't get subscribed. Will need to do some digging

kingsleyzissou:
I found the issue, we're creating the `ComposeRequest` with a `blueprint` rather than `customizations` field so the subscription options aren't propagated the whole way through. We need to change the behaviour and send a composer request with `customizations` set instead. This might be a big change and I think it's definitely better to wait until after the ITM release.

kingsleyzissou:
I am very blocked on this. I might close it and then re-open it once we have an idea of how we are going to implement the image-builder cli for the on-prem frontend

PR #3521, opened 2025-08-11
Launch: implement guidance for AWS (HMS-9002)

avitova:
This commit adds a launch modal for guiding users through launching an AWS instance from their image. As the launch service will be decommissioned, the flag shall be turned on, the code will later be cleaned up and the Provisioning wizard removed. [Link to mocks](https://docs.google.com/presentation/d/1tw-YKd94xMaB0k4AH5JTwJbanyH6sAgC8mDFL2yT-pg/edit?slide=id.g344c3b3318c_0_1381#slide=id.g344c3b3318c_0_1381)

Screenshot From 2025-08-15 14-51-17

https://github.com/user-attachments/assets/328d9edd-a48a-4023-bc9a-f5530199ffce

### DIFF
--- a/src/Components/ImagesTable/Instance.tsx
+++ b/src/Components/ImagesTable/Instance.tsx
@@ -54,7 +54,9 @@ import {
   isOciUploadStatus,
 } from '../../store/typeGuards';
 import { resolveRelPath } from '../../Utilities/path';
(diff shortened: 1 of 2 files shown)


avitova:
This PR is marked as draft because we'll still be discussing mocks, therefore the modal might get slightly changed.

PR #3527, opened 2025-08-13
Wizard: Display customized policy rules in review summary (HMS-8921)

mgold1234:
**Summary:**
Previously, when a user selected a compliance policy with tailored rules,
the review page always showed the default profile customizations instead
of the policy-specific customizations.

Root cause: OscapProfileInformation component was only using the profile
endpoint (/oscap/{distribution}/{profile}/customizations) which returns
base profile rules, not the policy endpoint
(/oscap/{policy}/{distribution}/policy_customizations) which returns
customized rules.

**Step-by-Step Reproduction**

1. Create and Configure a Compliance Policy
Register a RHEL system to your Red Hat account
Navigate to Compliance in Red Hat Console
Create a new policy or select an existing one
Associate the registered system to the policy

2. Customize the Policy Rules
Open the policy for editing
Find rules that affect image customizations and remove few of them

3. Build an Image with the Customized Policy
Navigate to Image Builder in Red Hat Console
Start creating a new image
In the OpenSCAP step, select go to compliance policy tab
Select your customized policy 
Continue through the wizard to the Review step

4. Observe the Bug
Expected behavior:

Review summary should show the customized policy rules
If you disabled a package installation rule, that package should not appear in the packages list
If you disabled a service rule, that service should not appear in the enabled/disabled services

Actual behavior (before fix):

Review summary shows the default profile rules instead of policy customizations
All default packages, services, and kernel arguments appear regardless of policy tailoring
No indication that policy customization took effect

Generated with AI
**Screen Captures**
| Before                                              | After                                   |
| --------------------------------------------------- | --------------------------------------- |
 | Screenshot 2025-08-14 at 12 35 46| Screenshot 2025-08-14 at 12 27 15|

### DIFF
--- a/src/store/backendApi.ts
+++ b/src/store/backendApi.ts
@@ -43,6 +43,12 @@ export const useLazyGetOscapCustomizationsQuery = process.env.IS_ON_PREMISE
   ? cockpitQueries.useLazyGetOscapCustomizationsQuery
   : serviceQueries.useLazyGetOscapCustomizationsQuery;
 
+export const useGetComplianceCustomizationsQuery =
+  serviceQueries.useGetOscapCustomizationsForPolicyQuery;
+
+export const useLazyGetComplianceCustomizationsQuery =
+  serviceQueries.useLazyGetOscapCustomizationsForPolicyQuery;
+
 export const useComposeBlueprintMutation = process.env.IS_ON_PREMISE
   ? cockpitQueries.useComposeBlueprintMutation
   : serviceQueries.useComposeBlueprintMutation;
--- a/src/Components/CreateImageWizard/steps/Oscap/components/OscapProfileInformation.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/components/OscapProfileInformation.tsx
@@ -8,7 +8,10 @@ import {
   Spinner,
 } from '@patternfly/react-core';
 
-import { useGetOscapCustomizationsQuery } from '../../../../../store/backendApi';
+import {
+  useGetComplianceCustomizationsQuery,
+  useGetOscapCustomizationsQuery,
+} from '../../../../../store/backendApi';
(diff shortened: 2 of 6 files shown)


src/store/backendApi.ts
```diff
@@ -43,6 +43,28 @@ export const useLazyGetOscapCustomizationsQuery = process.env.IS_ON_PREMISE
...
   : serviceQueries.useLazyGetOscapCustomizationsQuery;
 
+/**
+ * Hook to fetch OSCAP customizations for a specific compliance policy.
+ * Returns policy-specific customizations (packages, services, kernel args) that
+ * may differ from the base profile due to policy tailoring.
+ * 
+ * @returns undefined for on-premise environments (falls back to profile-based customizations)
+ */
+export const useGetOscapCustomizationsForPolicyQuery = process.env.IS_ON_PREMISE
```
kingsleyzissou, line 46: We could probably rename this since we try to differentiate between `Oscap` and `Compliance`,
something like `useGetCompliancCustomizationsQuery`

src/store/backendApi.ts
```diff
@@ -43,6 +43,28 @@ export const useLazyGetOscapCustomizationsQuery = process.env.IS_ON_PREMISE
...
+  ? undefined
```
kingsleyzissou, line 47: Since this is undefined, it might cause issues with the `on-prem` frontend. Maybe we could just use the same hook for both `on-prem` and the `hosted` service, but in the hook call itself we could add `process.env.IS_ON_PREMISE` as one of the `skip` conditions?

mgold1234:
@kingsleyzissou thank you for your review. it still in draft, can you please wait that I will change it to 'Ready for review'? :pray:

croissanne (approved):
thank you, very nice!

PR #3528, opened 2025-08-13
workflows: add api generation workflow (HMS-9052)

kingsleyzissou:
Add a workflow to re-generate api changes on a cronjob that runs every day.

JIRA: [HMS-9052](https://issues.redhat.com/browse/HMS-9052)

### DIFF
--- a/src/test/Components/CreateImageWizard/steps/Registration/Registration.test.tsx
+++ b/src/test/Components/CreateImageWizard/steps/Registration/Registration.test.tsx
@@ -422,7 +422,7 @@ describe('Registration request generated correctly', () => {
     localStorage.clear();
   });
 
-  test('register with satellite', { timeout: 20000 }, async () => {
+  test('register with satellite', { timeout: 25000 }, async () => {
     await renderCreateMode();
     await goToRegistrationStep();
     await clickRegisterSatellite();


.github/workflows/update-apis.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Run API code generation
+        run: npm run api
+
+      - name: Check if there are any changes
+        run: |
+          if [ "$(git status --porcelain)" ]; then
+            echo
+            echo "API codegen is up-to-date"
+            exit "0"
```
tkoscieln, line 39: This will only end the current step right? Not the whole action, so the PR will be created on every run. :thinking:  (only if that is not the intended behaviour though)

tkoscieln (dismissed):

.github/workflows/update-apis.yml
kingsleyzissou: We do something similar in `crc` it exits the action action too so the PR isn't created if there are no changes:
https://github.com/osbuild/image-builder-crc/blob/4395456dbe7508d1057c36083ebb2cad7bec2d35/.github/workflows/openscap.yml#L30-L36

tkoscieln: Hmm, interesting, lgmt then :smiley:

tkoscieln (approved):

.github/workflows/update-apis.yml
kingsleyzissou: We can test it as soon as this gets merged

PR #3546, opened 2025-08-18
tsconfig: Enable strict mode (HMS-9075)

### LINKED ISSUE #2702: Make Typescripter stricter
Now that the code base is essentially 100% converted to Typescript, we should start tightening up the rules. Let's look into enabling strict mode and other possible rules like this one https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes

regexowl:
Fixes #2702

This enables strict mode in tsconfig and removes rules that are already covered under it.

"downlevelIteration" was removed as it should be covered by `"target": "es2021"`

New config options are:
- "esModuleInterop" - enable commonjs and es module compatibility
- "forceConsistentCasingInFileNames" - enforces consistent casing across imports
- "skipLibCheck" - skips type check of declaration files, speeds up compilation
- "resolveJsonModule" - allows importing json files as modules (we probably currently don't use this, but I felt it's kinda neat, can remove it though)
- "incremental" - incremental compilation, speeds up builds

There were also issues with some types, the second commits addresses those and updates the types to match their calls.

JIRA: [HMS-9075](https://issues.redhat.com/browse/HMS-9075)


regexowl:
New issues were introduces in the meantime, that's kinda why we need the strict mode :sweat_smile:

PR #3548, opened 2025-08-18
Wizard: add support for shared EPEL repos (HMS-5986)

xbhouse:
Adds the "community" origin when listing repositories to include the shared EPEL repos. This is gated by a feature flag, shared EPEL repos are only available in stage for now. 

Jira: [HMS-5986](https://issues.redhat.com/browse/HMS-5986)


regexowl:
> yep, if a user has added the epel repos previously, there will be duplicates until we fully migrate to the shared / community epel repos. the duplicate repos should only be in prod for a short period of time, so the label is there to differentiate them. we also have a ticket to add a warning that this older repo will stop being snapshotted ([HMS-5987](https://issues.redhat.com/browse/HMS-5987))
> 
Got it, thanks for the explanation!

> i'm not sure we should allow selecting both repos at the same time though, that's a good point. i'll add some logic to disable the checkbox of the other epel repo if one is selected

Nice!

regexowl (approved):
Works great, thank you! :heart:

PR #3554, opened 2025-08-20
Launch: implement guidance for Oracle (HMS-9070)

avitova:
This commit adds launch modal for guiding users through launching a Oracle instance from their image. It provides a link to Oracle's cloud and a link for importing the image on the user's side.

The actual implementation:
Screenshot From 2025-08-20 11-41-30

vs Mocks:
image


avitova:
This Launch Modal is a bit different than what I did for other clouds, but I like this approach much better. So I guess that when cleaning up the code in the future, I will choose this way of doing this - using the component directly in the Images Table.

kingsleyzissou (approved):

PR #3557, opened 2025-08-20
Hooks: extract auth.getUser to its own hook (HMS-9085)

kingsleyzissou:
This code was being called in multiple places and was causing issues with the on-prem frontend. Extract the logic to a single hook and only get the `userData` for the hosted frontend.

JIRA: [HMS-9085](https://issues.redhat.com/browse/HMS-9085)


mgold1234 (approved):
/lgtm

PR #3563, opened 2025-08-21
Wizard: remove Sources from Azure step (HMS-9092)

avitova:
:speaking_head:  :mega:  Yo yo yo! Only a few more days and we'll do it. :scissors: Still few small things, and then a small cleanup once we turn the flag on. :partying_face: 

As part of the Launch decommissioning, Sources are getting rid of the Launch option. Based on mocks provided from UX, we need to also get rid of some parts that mention sources directly in our Wizard, and that will be useless. This commit conditionally removes sources based on a feature flag that we'll turn on once the decommissioning happens.


src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -437,6 +437,13 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 name='Azure'
                 id='wizard-target-azure'
                 key='wizard-target-azure'
+                status={
+                  !isAzureTenantGUIDValid(azureTenantId) ||
+                  !isAzureSubscriptionIdValid(azureSubscriptionId) ||
+                  !isAzureResourceGroupValid(azureResourceGroup)
+                    ? 'error'
+                    : 'default'
+                }
```
avitova, line 446: We did not have this, and I believe this might help navigate users that used to have sources for Azure target  realize that they need to change this. I also believe we might want to add this to other Targets, so that there is an error in case something is missing.

mgold1234 (commented):
/lgtm after fixing the confilcts :)

mgold1234 (approved):
/lgtm

PR #3579, opened 2025-08-27
Wizard: hide azure sources integration (HMS-9092)

avitova:
This commit conditionally hides the azure sources integration. It also sets the default value of share method to 'manual' in order not to break many things. This is a very temporary state, and we'll very soon after turning the flag on, make these changes permanent and will deal with more logic in the code.


regexowl (approved):
Works nicely! Bye sources :wave:

avitova:
Thank you! :partying_face:

PR #3591, opened 2025-08-29
playwright/CI: Fix slack notification condition in nightly action

tkoscieln:
The conditions previously were dependent on the playwright step only and fetched invalid `result` steps property. This also caused that when the `playwright-tests` step failed, the _on failure_ notification got skipped as this is standard Github Action behaviour. Changed it to on `success()` and on `failure()`, since they are simpler and fit this use case better, because they are directly controlled by the outcome of previous jobs.


regexowl (approved):
The code changes look good to me so let's gooo :boom::rocket:

PR #3593, opened 2025-08-30
api: regenerate api code generation

schutzbot:
Update api code generation

PR #3595, opened 2025-08-31
Add new API endpoint and update Release component

mgold1234:
add release to prevent error that we see here -https://github.com/osbuild/image-builder-frontend/pull/3593


regexowl (approved):
Thank you!

PR #3604, opened 2025-09-02
Wizard: clean up analytics to track image creation failure or success (HMS-9219)

ksiekl:
This PR removes the old analytics (because only failed cloud builds were tracked), and adds a new one with info about the image build (both successful and failed ones, all of the different types).

### DIFF
--- a/src/Components/ImagesTable/ImagesTable.tsx
+++ b/src/Components/ImagesTable/ImagesTable.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import {
   Alert,
@@ -305,6 +305,9 @@ const ImagesTableRow = ({ compose, rowIndex }: ImagesTableRowPropTypes) => {
   const [pollingInterval, setPollingInterval] = useState(
     STATUS_POLLING_INTERVAL,
   );
+  const lastTrackedStatusRef = useRef<string | null>(null);
+  const { analytics, auth } = useChrome();
+  const { userData } = useGetUser(auth);
 
   const { data: composeStatus } = useGetComposeStatusQuery(
     {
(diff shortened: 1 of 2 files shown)


src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -324,6 +327,49 @@ const ImagesTableRow = ({ compose, rowIndex }: ImagesTableRowPropTypes) => {
...
 
+  useEffect(() => {
+    const currentStatus = composeStatus?.image_status.status;
+    if (!process.env.IS_ON_PREMISE && currentStatus) {
+      if (lastTrackedStatusRef.current === null) {
+        lastTrackedStatusRef.current = currentStatus;
+        return;
+      }
+
+      const builtIsCompleted =
```
avitova, line 338:
```suggestion
      const buildIsCompleted =
```

avitova (commented):
Looks good to me, I just have one small nitpick. I cannot test, sorry.

src/Components/ImagesTable/ImagesTable.tsx
ksiekl: Thanks for noticing that!

src/Components/ImagesTable/ImagesTable.tsx
```diff
@@ -324,6 +327,49 @@ const ImagesTableRow = ({ compose, rowIndex }: ImagesTableRowPropTypes) => {
...
+                }
+              : {}),
+            account_id: userData?.identity.internal?.account_id || 'Not found',
+          },
+        );
+      } else if (statusChanged) {
+        lastTrackedStatusRef.current = currentStatus;
+      }
+    }
+  }, [analytics, userData, compose, composeStatus]);
```
croissanne, line 371: Wait, do we want to have an analytic tracking for every single compose? Won't this generate rate limiting problems?

croissanne: ooh ic, yea sorry didnt read it carefully enough. lgtm then

croissanne (approved):

src/Components/ImagesTable/ImagesTable.tsx
croissanne: I'd still keep an eye on it though.

ksiekl: Thanks! Definitely, the analytics are growing, so we need to think about that

PR #3614, opened 2025-09-04
Wizard: fix root not removable when duplicate

avitova:
We allow adding more root mountpoints via import feature, which should not be possible. This fix allows users to remove extra root partition, but keeps it in the wizard so that users know about it.


kingsleyzissou (dismissed):
Nice!

ksiekl (dismissed):

avitova:
Yeah, the validation happens for some reason only on click of the next button. 😄

mgold1234:
need to fix the conflict other then that look good to me :)

src/Components/CreateImageWizard/steps/FileSystem/components/FileSystemTable.tsx
```diff
@@ -153,6 +153,10 @@ const FileSystemTable = () => {
     });
   };
 
+  const rootPartitionsCount = partitions.filter(
```
mgold1234, line 156: it runs on every render, even when the partitions array hasn't changed, wdyt to maybe change it to 
```
const rootPartitionsCount = useMemo(
  () => partitions.filter((p) => p.mountpoint === '/').length,
  [partitions]
);
```

avitova: That's a good point, thanks! I added the useMemo:)

.github/workflows/playwright.yml
avitova, line 1: Can you ack the changes you requested? @tkoscieln

avitova:
TY!:)

PR #3618, opened 2025-09-05
Wizard: enable OpenSCAP for RHEL10

kingsleyzissou:
This was previously disabled since some of the profiles are still in draft. However, we've been asked to make all the profiles available including the drafts.


avitova (approved):
Nice! Looks good.

PR #3636, opened 2025-09-11
store/cockpit: use ibcli for building image (HMS-9389)

kingsleyzissou:
This PR switches the image building from osbuild-composer to ibcli. It does this by making use of cockpits `long-running-process`. Under the hood, cockpit creates a one-shot systemd service that executes the image build. We then monitor the systemd service and save the result to a file which we then use for reporting a compose status.

JIRA: [HMS-9389](https://issues.redhat.com/browse/HMS-9389)


kingsleyzissou:
I found a bug in ibcli:
https://github.com/osbuild/image-builder-cli/issues/306

And pushed a fix here:
https://github.com/osbuild/image-builder-cli/pull/307

kingsleyzissou:
Lol

```
Last login: Thu Sep 18 09:43:10 2025 from 10.0.2.2
[systemd]
Failed Units: 4
  cockpit-image-builder-052e427f-4f0a-427c-b4b5-e835a78ef30c.service
  cockpit-image-builder-a978c59a-1a1d-4f34-8085-28387a12981b.service
  cockpit-image-builder-e9f6191c-f69a-4532-b524-adedc4338f0d.service
  cockpit-image-builder-fd9ba52b-b9c2-4065-a922-47b0c903d1e4.service
```

src/store/cockpit/cockpitApi.ts
```diff
@@ -386,29 +391,63 @@ export const cockpitApi = contentSourcesApi.injectEndpoints({
...
+              const process = new LongRunningProcess(
+                `cockpit-image-builder-${uuid}.service`,
+                updateComposeStatus(composeDir),
+              );
+
+              // this is a workaround because the process
+              // can't be started when in `init` state
+              process.state = 'stopped';
+
+              process.run(['bash', '-ec', cmd.join(' ')]).catch(async () => {
```
mvo5, line 478: it seems cockpit works with arrays of strings, so would `process.run(cmd)` work? The only other change required would be HOME= needs to move into the "options" of process.run(), something like `progress.run(cmd, environ={"HOME": ${user.home}})` (assuming python syntax here, I have no clue about typescript). As long as we control all inputs its problably fine as is but once there is user controlled input this turns into a command injection quickly and it seems we can just start with using the array of strings which makes command injections (or files with spaces) a non-issue.


It seems this could then also use the same trick that https://github.com/cockpit-project/cockpit/blob/main/examples/long-running-process/index.js#L20 is using, i.e. just monitor the jounal to get the output of ibcli, with that we could add `image-builder build --progress=json` (we currently only have verbose,term,debug there) and then you could render your own progress this way. When we design the json we need to make sure that its self-contained, i.e. that we must assume that at any point a client could connect or reconect and that the client will not have state. This is not a property of the underlying osbuild json stream which will often use "context" which needs to be captured and remembered by the consumer of the json (but no worries, just spelling it out here, doing this will not be hard especially if we start simple).

mvo5 (commented):
Thank you! One quick inline command about the file spawning. 

I also wonder what we can/could do to make testing easier for you? It seems for the most basic testing just having a way to override the location of image-builder and providing a mock script that e.g. fails with an error or that produceses some fake json would be the easiest. I almost wonder if ibcli provide this in its git repo. Like `image-builder build failing-image` and the mock would behave depedning on file type (i.e. for  failing-image fails, for slow-image take a long time etc). But maybe I'm overthinking this :)

src/test/mocks/cockpit/index.ts
```diff
@@ -28,6 +25,28 @@ export default {
       if (args.length && args[0] === 'uname') {
         resolve('x86_64');
       }
+      if (args.length && args[0] === 'image-builder' && args[1] === 'list') {
```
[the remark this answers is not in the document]
kingsleyzissou: I think it's because args.length is non-zero :)

src/store/cockpit/helpers/imageType.ts
```diff
@@ -0,0 +1,19 @@
...
+    : imageType;
+
+  // this is a list of types that we know we need to translate
+  const lookup: Record<string, string> = {
+    qcow2: 'guest-image',
+    ami: 'aws',
+    gce: 'gcp',
+    vhd: 'azure',
+    vmdk: 'vsphere',
+    ova: 'vsphere-ova',
```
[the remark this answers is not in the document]
kingsleyzissou: Not yet. No gcp and azure yet either. Getting there slowly

kingsleyzissou:
> @kingsleyzissou Good question! I did not notice that I did not have image-builder cli installed at all. Makes me think of 1. we should add some check or installation script into our makefile for setting up cockpit, 2. maybe have a way to detect that? 🤔

Ya that's a good point, I added it to the specfile but not to the makefile. I'll have a look at that

src/store/cockpit/cockpitApi.ts
kingsleyzissou: Ya I didn't like that either and adding HOME= is not ideal. But I ran into two issues here:
1. the list of strings didn't work as expected and it was really hard to debug why
2. adding the  HOME as an env variable in the options also never worked and got ignored.

But those are both valid points and I will dig into it deeper and try figure out why that wasn't working.

kingsleyzissou: With the json progress that seems nice and we could probably implement that in the getComposeStatus endpoint

kingsleyzissou: Oh sorry I actually missed this comment. I came back to it today in preparation for Friday's call.

I filed a feature request in cockpit:
https://github.com/cockpit-project/cockpit/issues/22454

kingsleyzissou:
> Thank you for exploring this! Personally I really like this approach. But curious to hear what ppl think.

Thanks! Yeah I've scheduled a meeting for tomorrow (sorry I know it's friday) for us to discuss.

PR #3640, opened 2025-09-12
cockpit: use upstream cockpit types

kingsleyzissou:
Since we've made our typescript checks stricter, we can rely on the cockpit
types from upstream. This will prevent drift between the types. The one downside
to this is that we may have to keep the cockpit `pkg/lib` directory in the repository.


kingsleyzissou:
We can re-open this later down the line if we need

PR #3656, opened 2025-09-17
README: emphasize to use `npm clean-install`

schuellerf:
Emphasize to use `npm clean-install` and explain why.

Always preferring the long form `clean-install` over `ci` as this is more readable for new developers and is the same effort for copy & paste.

Also fixes the TOC.


ochosi (approved):
LGTM

PR #3664, opened 2025-09-22
Wizard: remove deprecated satellite flag

avitova:
We already have satellite in prod for some time, no need for the flag.


regexowl (approved):
Thanks!

PR #3675, opened 2025-09-24
Wizard: Firewall revamp (HMS-9399)

ksiekl:
Change the Firewall step according to mocks


regexowl:
> Thanks @regexowl !
> 
>     1. Yes, that description was not defined so I used the old one, thinking when UX team updates it, it can be a little PR on its own.
> 

Sounds good :+1: 

>     2. Oh yeah, I didn't realize that since it wasn't in this step in Figma, thanks for noticing! I can change it here ❤️
> 

Thanks! :heart: 
>     3. Every step currently has PW test, so I'm not sure what you mean. Would you like to see the specific test cases that are currently in vitest moved to the existing PW test? I can definitely do that! 🫡

Yup yup: *"As part of this, ~~jest~~ vitest tests should be removed, new playwright tests should be written, or existing playwright tests should be updated to handle the new UI."*

The plan is to move away from vitest and msw and replace the unit tests by playwright completely during the revamp.

avitova:
Thanks!
1. The placeholder says `Enter port (e.g., 8080/tcp, 443/udp)`, but the error says `Expected format: <port/port-name>:<protocol>. Example: 8080:tcp, ssh:tcp`. They both state different formats, and even though we accept both, I would put same examples into the error message and placeholder.
2. I think I see different mocks?
image

ksiekl:
@avitova good catch, thanks! I changed the error message and placeholder text, and to your second point - there is a discrepancy between Figma mocks and ticket description in [HMS-9399](https://issues.redhat.com/browse/HMS-9399)

PR #3679, opened 2025-09-25
Wizard: Increase limit for Compliance policies in selector to 50

tkoscieln:
This was previously set to 10 and there is no other way how to see other policies than the first 10 returned by Compliance API.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
@@ -96,6 +96,7 @@ const PolicySelector = () => {
     isSuccess: isSuccessPolicies,
   } = usePoliciesQuery({
     filter: `os_major_version=${majorVersion}`,
+    limit: 50,
   });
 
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/steps/Oscap/components/PolicySelector.tsx
```diff
@@ -96,6 +96,7 @@ const PolicySelector = () => {
     isSuccess: isSuccessPolicies,
   } = usePoliciesQuery({
     filter: `os_major_version=${majorVersion}`,
+    limit: 50,
```
regexowl, line 99: Sounds good! Should we also sort the policies ascending by title by adding `sort_by` param?

tkoscieln: Makes sense, will add it

tkoscieln: Added :saluting_face:

PR #3682, opened 2025-09-26
playwright/CI: Remove Currents reporter from Gitlab runs (HMS-9314)

tkoscieln:
Remove Currents from CI runs for Cockpit plugin in Gitlab, because we use four runners per run and each one generates a report, which leads to increased spending for Currents that can be avoided since we can track test results using the results from hosted tests that are mostly the same.

### DIFF
--- a/schutzbot/playwright_tests.sh
+++ b/schutzbot/playwright_tests.sh
@@ -79,8 +79,6 @@ sudo podman run \
      -e "CI=true" \
      -e "PLAYWRIGHT_USER=admin" \
      -e "PLAYWRIGHT_PASSWORD=foobar" \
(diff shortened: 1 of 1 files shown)


tkoscieln:
@kingsleyzissou sorry for creating confusion, I mean test statistics in Currents, not the test results and artifacts themselves :disappointed: 
>I have a question about this, will it still be possible to download the the traces or videos?

Yes, all of the test artifacts are still stored in AWS, at the end of every Gitlab job log you have a direct link to bucket they are stored in, something like `Please login to xxxx AWS account and visit https://xxxx to access job artifacts.` That includes traces and videos too.

kingsleyzissou (approved):

tkoscieln:
cc @marusak

PR #3690, opened 2025-09-29
Wizard: don't clear first boot systemd service (HMS-9306)

kingsleyzissou:
If the `custom-first-boot` service is set, we should make sure not to clear it when selecting an OpenSCAP profile or compliance policy.


kingsleyzissou:
Thanks ya I was really struggling with the stage yesterday. Was so slow so it was hard to test properly. Will have a better look today

PR #3694, opened 2025-10-01
playwright: Fix missing expect import in imageBuilding.ts

tkoscieln:
Static analysis failed to notice the expect import was missing here, because vitest imports expect globally.


ochosi (approved):
LGTM!

PR #3695, opened 2025-10-01
Wizard: reintroduce registration step for cockpit (HMS-9195)

kingsleyzissou:
We can re-enable the registration step for cockpit by allowing users to pass in an activation and org id pair so that they can build an image that is automatically registered.

I had to create a new PR, but this is based on https://github.com/osbuild/image-builder-frontend/pull/3519


src/store/cockpit/cockpitApi.ts
```diff
@@ -503,14 +503,35 @@ export const cockpitApi = contentSourcesApi.injectEndpoints({
...
 
+              // we need a customizations object rather than a blueprint
+              // since we need to set the `subscription` options for
+              // registering an image. This is not available with blueprint
+              // customizations
+              const customizations = mapHostedToOnPrem(
+                blueprint as CreateBlueprintRequest,
+              ).customizations as Customizations;
+
+              if (blueprint.customizations.subscription) {
```
croissanne, line 514: What if we change `mapHostedToOnPrem` to return a `ComposeRequest` or even just `Customizations` as defined in `src/store/cockpit/composerCloudApi.ts` ? Then we can move this block into that function as well.

 The only place `mapHostedToOnPrem` is used is here afaict, so it's really just a helper function.

croissanne (commented):
Thank you, completely agree with using ComposeRequest instead of the blueprint thing. Also agree with the choice to have register later be the default.

One question otherwise lgtm.

avitova (commented):
I keep getting an error:
```
cockpitApi.queries.getOscapCustomizations({"distribution":"fedora-41","profile":"xccdf_org.ssgproject.content_profile_cis"}).data.filesystem.0.size`. Value
```
But also I needed to do a hack and enable cockpit registration step for Fedora, maybe you do not have this, and this is irrelevant. I guess I cannot test.

kingsleyzissou:
> But also I needed to do a hack and enable cockpit registration step for Fedora, maybe you do not have this, and this is irrelevant. I guess I cannot test.

Ah right, weird error since the `CIS` profile isn't available for OpenSCAP on Fedora. Maybe that's why you're getting that error.

PR #3696, opened 2025-10-01
Users UI revamp (HMS-9400)

avitova:
I tried to address the changes from the UI Revamp in this PR, as well as make this step as much bulletproof and clear as possible. As a next step, I'd also like to add debouncing, so that users do not see error on the first keystroke in the field.
Cursor helped me write tests, and then I spent quite some time adjusting it, but I believe we should aim for something like this, since we are removing unit tests. :)
Once you are reviewing, please, do so carefully. Most of the functionality on Users is changed.

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -35,6 +35,7 @@ type ValidationInputProp = TextInputProps &
     ) => void;
     isRequired?: boolean;
     warning?: string;
+    forceErrorDisplay?: boolean;
   };
 
 type ErrorMessageProps = {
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -671,16 +671,32 @@ export function useUsersValidation(): UsersStepValidation {
...
       users[index].password,
       environments.includes('azure'),
     ).isValid;
-    const passwordError =
-      users[index].password && !isPasswordValid ? 'Invalid password' : '';
+    if (users[index].password && !isPasswordValid) {
+      userErrors.userPassword = 'Invalid password';
+    }
+
+    const sshKeyError = validateSshKey(users[index].ssh_key);
```
kingsleyzissou, line 694: I know this isn't really relevant for this PR. I would really like to take a look at how we're handling validation
in the frontend and maybe streamline this a bit :)

avitova: Yes!!! We now have more places for ssh keys and secrets. This is very similar for CAs. For SSH keys and CAs, I would like to see a pattern of accepting almost anything, but raising a warning. And I really think that having errors as well as warning can be useful in our validation.

avitova:
Good call with the button...I thought it was just not possible, but now I have found a way to make the style the same. :)

avitova:
Okay, I also want to add some more fixes and tests for Editing, but once this is approve, we can merge this, and I will do a follow-up.

PR #3699, opened 2025-10-02
src: Remove import flag from code base (HMS-9468)

regexowl:
Import functionality was fully released, we can remove `image-builder.import.enabled` flag and gating from the code base.

JIRA: [HMS-9468](https://issues.redhat.com/browse/HMS-9468)


avitova (approved):
Thanks!

kingsleyzissou (approved):

PR #3704, opened 2025-10-07
playwright: Add env var to enable global static user on login

tkoscieln:
Adds a env var that enables usage of static user globally for all login attempts.

By turning this option on we should unclog the CI, which is currently failing to assign a subscription to dynamically created user and encountering the this screen later in the log in process.
image
We can switch back to using dynamically created users once [ETHEL-605](https://issues.redhat.com/projects/ETHEL/issues/ETHEL-605?filter=allopenissues) is resolved.


ochosi (approved):
Obviously cannot test this, but the code looks good!

.github/workflows/boot-tests-nightly.yml
```diff
@@ -72,20 +72,27 @@ jobs:
...
-          # Step 2: Attach subscriptions to the new account
-          curl -k -X POST https://account-manager-stage.app.eng.rdu2.redhat.com/account/attach \
-            -d "{\"username\": \"$PLAYWRIGHT_USER\", \"password\":\"$PLAYWRIGHT_PASSWORD\", \"sku\":[\"RH00003\"],\"quantity\": 1}"
-          # Step 3: Activate the new account by accepting Terms and Conditions
-          curl -k -X POST https://account-manager-stage.app.eng.rdu2.redhat.com/account/activate -d "{\"username\": \"$PLAYWRIGHT_USER\", \"password\":\"$PLAYWRIGHT_PASSWORD\"}"
-          # Step 4: Refresh account to update subscription pools
-          curl -k -X POST https://account-manager-stage.app.eng.rdu2.redhat.com/account/refresh -d "{\"username\": \"$PLAYWRIGHT_USER\", \"password\":\"$PLAYWRIGHT_PASSWORD\"}"
-          # Step 5: View account to check account status
-          curl -k -X GET "https://account-manager-stage.app.eng.rdu2.redhat.com/account/get?username=$PLAYWRIGHT_USER&password=$PLAYWRIGHT_PASSWORD"
+          export GLOBAL_STATIC_USER=$GLOBAL_STATIC_USER
```
avitova, line 79:
```suggestion
          echo "Global statis user before exporting $GLOBAL_STATIC_USER"
          export GLOBAL_STATIC_USER=$GLOBAL_STATIC_USER
```
Can you try this? If this is not empty, I'd try not using the same variable name here on the left and on the right? But this is just a wild guess.

PR #3721, opened 2025-10-14
playwright: Registration step tests

avitova:
This commit adds detailed test for registration that tests automatic, register later and satellite registration. It tests behaviour of insights and rhc switch buttons, as well as different error states. Finally, it adds verification that the review step has correct values. This is a parametrized test that runs the whole workflow of create, edit, export, import and delete for all register options.

### DIFF
--- a/playwright/Customizations/Registration.spec.ts
+++ b/playwright/Customizations/Registration.spec.ts
@@ -1,8 +1,13 @@
 import { expect } from '@playwright/test';
 import { v4 as uuidv4 } from 'uuid';
 
+import isRhel from '../../src/Utilities/isRhel';
 import { test } from '../fixtures/customizations';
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/steps/Registration/components/Registration.tsx
```diff
@@ -77,11 +77,12 @@ const Registration = () => {
                     <ActivationKeysList />
                   </Content>
                 )}
-              {process.env.IS_ON_PREMISE && (
-                <Content className='pf-v6-u-pb-sm'>
-                  <ManualActivationKey />
-                </Content>
-              )}
+              {process.env.IS_ON_PREMISE &&
+                registrationType.startsWith('register-now') && (
```
avitova: I changed this part to not show activation key and org id for other options, I hope that's okay. @kingsleyzissou

PR #3727, opened 2025-10-16
Wizard: address Google Cloud Platform renaming (HMS-9577)

avitova:
Google changed the name of the product from GCP to just GC. This commit addreses that change on places where it is visible in the frontend. Changes needed to do in the code logic are not addressed.

JIRA: [HMS-9577](https://issues.redhat.com/browse/HMS-9577)


avitova:
thank youuuu

PR #3731, opened 2025-10-17
Wizard: fix repeatable build behavior in pkg and repo steps

dominikvagner:
## Summary
This PR includes a few changes and fixes to correct the behavior of the wizard when a repeatable build with a snapshot date is chosen.
- This adds a different view for the 'Repositories' step that reflects if a repeatable build with a set snapshot date was chosen. It hides information about repositories that might be incorrect at the chosen date and presents a snapshot date and displays correct package count with a link to the snapshot detail.
- This fixes an issue with the 'Packages' step where you could pick additional packages that might not be available in the included repositories at the chosen snapshot date, or ones that were removed.
This would be very complicated to do solely in this frontend, so the names search APIs were extended to support a date argument and search the correct snapshots for the right content when that's provided. [[PR](https://github.com/content-services/content-sources-backend/pull/1242)]

Also fixes a bug found while testing, where the package search step would outright crash when a package group with an empty package list existed in the searched repos. (This can happen in Red Hat repos in the Conflict package groups)

> The required BE changes are now deployed in Prod, so this is ready for review now!

Issue Link: [[HMS-8645](https://issues.redhat.com/browse/HMS-8645)]

## Testing Steps
1. Create a repo with 2 snapshots (good one to use are [these](https://content-services.github.io/fixtures/yum/comps-modules/) ones), create a repo with the `v1` URL and then update it to the `v2` URL.
2. Verify the repositories and package search steps show the correct stuff, depending on the option and date chosen in the "Repeatable Build" step. _(For what to expect, feel free to look at [this](https://github.com/content-services/content-sources-backend/pull/1242/commits/3ee9e47b7293bcf9c31b5d873cc89febd19a1d34#diff-8e226cc3ada28060e50e00f92386494cf16b1e6b1bf9d986aac00f354b30b7c5) test)_
2.1 If you select a date few days in the past, the picked up snapshot will be the V1 and if you select use latest, or a date after the repo was created, it should be the second one.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/Packages.tsx
@@ -93,6 +93,7 @@ import {
   selectModules,
   selectPackages,
   selectRecommendedRepositories,
+  selectSnapshotDate,
   selectTemplate,
 } from '../../../../store/wizardSlice';
 import {
@@ -100,6 +101,7 @@ import {
   getEpelUrlForDistribution,
   getEpelVersionForDistribution,
 } from '../../../../Utilities/epel';
+import { convertStringToDate } from '../../../../Utilities/time';
 import useDebounce from '../../../../Utilities/useDebounce';
 
 export type PackageRepository = 'distro' | 'custom' | 'recommended' | '';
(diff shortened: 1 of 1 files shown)


regexowl (approved):
Looks great, thanks a lot! :rocket:

PR #3732, opened 2025-10-17
Wizard: Compliance revamp (HMS-9396)

ksiekl:
The changes are:
- rename Compliance to Security
- refactor ManageKeysButton to more abstract ManageButton
- refactor the UI according to Figma
- change the Profile Selector to show profile name instead of id
- add new component with Compliance detail
- rewrite the tests to PW

There are still bugs in the functionality - this PR is big as is, so I created tickets for the bugs under [Compliance epic](https://issues.redhat.com/browse/HMS-5958) and we can work on that in follow-ups. This PR concerns only the fe refactor for the new UI revamp.


src/Components/CreateImageWizard/steps/Oscap/index.tsx
```diff
@@ -130,80 +143,157 @@ const OscapContent = () => {
...
+                  <Popover
+                    headerContent='Details'
+                    bodyContent={<PolicyDetails />}
+                    minWidth='30'
+                  >
+                    <Button
+                      variant='secondary'
+                      icon={<InfoCircleIcon />}
+                      iconPosition='left'
+                      isDisabled={complianceType !== 'compliance'}
```
regexowl: Do we want to disable the button also when no policy is selected?

playwright/Customizations/Compliance.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      .check();
+
+    const toggle = frame.locator(
+      '[data-ouia-component-id="compliancePolicySelect"]',
+    );
+    await expect(toggle).toBeEnabled();
+    await toggle.click();
+
+    const policyOption = frame.getByRole('option', {
+      name: /placehollder/i,
```
regexowl:
```suggestion
      name: /placeholder/i,
```

Just to check why we're not selecting a specific policy here, I assume:
- there would be an overlap with compliance boot tests which already tests the integration
- and since it's a fresh user it would probably be a pain to setup any compliance policies before we'd be even able to pick any of them

is that right?

regexowl:
Also looking at the quickly decreasing coverage we should probably switch to PW coverage soon :sweat_smile: which is out of scope for this PR of course. I'll open an issue :lady_beetle:

avitova (commented):
Should we remove the description as well? It is not in mocks. However, I think it is helpful, and mocks do not show it be anywhere else. :thinking: 

https://github.com/user-attachments/assets/a53a7978-dc68-42d5-ada2-14e9cb6d469a

image

Let's leave fixing this for the later PR.

src/test/Components/CreateImageWizard/steps/FirstBoot/Firstboot.test.tsx
```diff
@@ -34,6 +34,19 @@ import {
...
+  useFlag: vi.fn((flag) => {
+    switch (flag) {
+      case 'image-builder.compliance.enabled':
+        return true;
+      default:
+        return false;
+    }
+  }),
+}));
+
```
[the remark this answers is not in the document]
ksiekl: Thanks! I didn't notice that :heart:

playwright/Customizations/Compliance.spec.ts
regexowl: Aha! Mea culpa, please ignore then. Mock-y sounding mock sounds fine to me :sweat_smile:

ksiekl:
@smckinno yes, the popover shows the comliance policy description, but what about oscap profiles? They don't have the additional "View details" button, and their description is shown only at the review step, not in the security step

smckinno:
Yeah that's a great call out. I actually just added more information to the mocks to cover the OpenSCAP section. Here is a screenshot so you don't have to go searching for it :) 
Screenshot 2025-10-29 at 9 16 23 AM

playwright/Customizations/Compliance.spec.ts
ksiekl: alrighty, thanks!

src/constants.ts
```diff
@@ -27,6 +27,8 @@ export const RELEASE_LIFECYCLE_URL =
 export const AZURE_AUTH_URL =
   'https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow';
 export const COMPLIANCE_URL = '/insights/compliance/scappolicies';
+export const OSCAP_URL =
+  'https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/security_guide/chap-compliance_and_vulnerability_scanning';
```
regexowl: Noticed the link leads to docs for RHEL 6, tried to look around for something up-to-date, maaybe this could work? https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/composing_a_customized_rhel_system_image/creating-pre-hardened-images-with-image-builder

@elianerpereira could you please help up with picking the right documentation link? Something that will help users find out more about OpenSCAP customization in Image Builder. It would go here:

Screenshot_20251030_100516

src/Components/CreateImageWizard/steps/Oscap/index.tsx
```diff
@@ -130,80 +147,173 @@ const OscapContent = () => {
...
-              Save blueprint and navigate to{' '}
-              {lightspeedEnabled ? 'Red Hat Lightspeed' : 'Insights'} Compliance
-            </AlertActionLink>
-          </Alert>
         )}
-      <Oscap />
-    </Form>
+
+        <FormGroup>
+          <Content className='pf-v6-u-pb-sm'>
```
regexowl: Sorry, should have been clearer about the gating - only the compliance radio should be gated behind `complianceEnabled`, meaning OpenSCAP should be available through all the environment, but compliance should be still unavailable in prod:stable

The step in prod:stable currently looks like this:
image

src/constants.ts
ksiekl: Shayna gave me a link to put there, so I added that :smile:

src/Components/CreateImageWizard/steps/Oscap/index.tsx
```diff
@@ -130,80 +147,173 @@ const OscapContent = () => {
...
+                  bodyContent={<OpenScapProfileDetails />}
+                  minWidth='30'
+                >
+                  <Button
+                    variant='secondary'
+                    icon={<InfoCircleIcon />}
+                    iconPosition='left'
+                    isDisabled={complianceType !== 'openscap' || !profileID}
+                  >
+                    View details
```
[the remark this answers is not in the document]
ksiekl: Oh, good catch, thank you!

regexowl (approved):
Looks great, thanks! :rocket:

PR #3748, opened 2025-10-22
Wizard: Add support for ignoring linter warnings (HMS-9208)

mgold1234:
This PR introduces the ability to ignore blueprint compliance warnings when a policy is updated after the blueprint’s creation.
It extends the existing linting workflow by allowing users to choose between:
Fix warnings automatically (update the blueprint)
Ignore all warnings

Key improvements:
*  Added an `Ignore all warnings` button with persistent state stored in localStorage
* Warnings reappear only when the policy content actually changes
*  Currently supports ignoring added requirements only (removed items are not handled in this PR)

 **Supported workflow:**
1. User creates a policy and a blueprint based on it
2. User updates the policy (adds new requirements)
3. The blueprint displays warnings for newly added items
4. User can choose between `fix warnings automatically (update the blueprint)` or`ignore all warnings`
5. New warnings reappear only when policy content changes again

there are few bugs:
Bug 1 — New compliance warnings didn’t appear after “Ignore” [HMS-9652](https://issues.redhat.com/browse/HMS-9652)
Repro:
A compliance warning for requirement A appears.
User clicks “Ignore all warnings”.
Policy is updated with a new requirement B.
Expected: A new warning for B should appear.
Actual: No warning alert appeared; the alert remained suppressed.  -> fixed 
Bug 2 — Warning didn’t reappear after add → remove cycle [HMS-9653](https://issues.redhat.com/browse/HMS-9653)
Repro:
A compliance warning for package audit appears; user clicks “Ignore all warnings”.
User edits the blueprint and adds the audit package (warning disappears).
User edits again and removes the audit package.
Expected: The audit warning should reappear after the package is removed.
Actual: The warning remained hidden due to a stale ignore entry. -> fixed 

Bug 3 - Create a blueprint with a compliance policy that requires “audit” and “firewalld”.
Remove “audit” → UI shows: “audit required by policy is not present”.
Click “Ignore all” → the audit warning is hidden.
Remove “firewalld” → UI shows: “firewall required by policy is not present”.
User want to press on “Fix automatically” button
the problem:
when user click “Fix automatically” → the system reinstalls all missing packages, including “audit” that was previously ignored.  -> will be fixed in a following up pr
 
Screenshot 2025-10-22 at 12 27 51

📎 Related: [HMS-9208](https://issues.redhat.com/browse/HMS-9208)

### DIFF
--- a/playwright/Customizations/ComplianceWarnings.spec.ts
+++ b/playwright/Customizations/ComplianceWarnings.spec.ts
@@ -1,0 +1,343 @@
+import { expect, type FrameLocator, type Page } from '@playwright/test';
+import { v4 as uuidv4 } from 'uuid';
+
+import { deleteCompliancePolicy } from '../BootTests/helpers/helpers';
+import { test } from '../fixtures/customizations';
+import { isHosted } from '../helpers/helpers';
+import { ensureAuthenticated } from '../helpers/login';
+import {
+  fillInImageOutput,
+  ibFrame,
+  navigateToLandingPage,
+} from '../helpers/navHelpers';
+import {
+  createBlueprint,
+  deleteBlueprint,
+  fillInDetails,
+  registerLater,
+} from '../helpers/wizardHelpers';
+
+// Clear the login from global setup so we can use static user
+test.use({ storageState: { cookies: [], origins: [] } });
+
+// Helper to create a simple Compliance policy (simplified version like the working test)
+const createCompliancePolicy = async (page: Page, policyName: string) => {
+  await page.goto('/insights/compliance/scappolicies');
+  await page.getByRole('button', { name: 'Create new policy' }).click();
+  await page.getByRole('option', { name: 'RHEL 10' }).click();
+  await expect(page.getByRole('gridcell').first()).toBeVisible();
+  await page.getByRole('radio', { name: 'Select row 0' }).click();
+  await page.getByRole('button', { name: 'Next', exact: true }).click();
+  await page.getByRole('textbox', { name: 'Policy name' }).fill(policyName);
+  await page.getByRole('button', { name: 'Next', exact: true }).click();
+  await page.getByRole('button', { name: 'Next', exact: true }).click();
+  await page.getByRole('button', { name: 'Next', exact: true }).click();
+  await page.getByRole('button', { name: 'Finish' }).click();
+  await page.getByRole('button', { name: 'Return to application' }).click();
+};
+
+// Simple helper to select compliance policy (like the working test)
+const selectCompliancePolicy = async (
+  frame: Page | FrameLocator,
+  policyName: string,
+) => {
+  await frame.getByRole('button', { name: 'Security' }).nth(1).click();
+  await frame
+    .getByRole('radio', { name: 'Use a custom Compliance policy' })
+    .click();
+  await frame.getByRole('button', { name: 'None' }).click();
+  await frame.getByRole('option', { name: policyName }).click();
+  await expect(frame.getByRole('button', { name: policyName })).toBeVisible();
+};
+
+// Helper to (re-)open the blueprint view from landing page (simplified)
+const openBlueprintFromLanding = async (page: Page, blueprintName: string) => {
+  await navigateToLandingPage(page);
+  const frame = (await ibFrame(page)) as Page | FrameLocator;
+  const searchInput = frame.getByRole('textbox', { name: 'Search input' });
+  await expect(searchInput).toBeVisible({ timeout: 10000 });
+  await searchInput.fill(blueprintName);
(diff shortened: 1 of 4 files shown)


playwright/Customizations/ComplianceWarnings.spec.ts
```diff
@@ -0,0 +1,327 @@
...
+  ],
+});
+
+const mkBlueprintDetails = (
+  blueprintId: string,
+  blueprintName: string,
+  state: State,
+) => {
+  const distribution = 'rhel-9';
+  const image_requests: never[] = [];
```
regexowl: We have a type for image requests, can we use that?

```suggestion
  const image_requests: ImageRequest[] = [];
```

playwright/Customizations/ComplianceWarnings.spec.ts
```diff
@@ -0,0 +1,327 @@
...
+      created_at: new Date().toISOString(),
+      blueprint_id: uuidv4(),
+      blueprint_version: 1,
+    },
+  ],
+});
+
+test.describe('Compliance warnings ignore flow', () => {
+  // Skip entire suite unless explicitly enabled with ENABLE_COMPLIANCE_IGNORE_FIX=1
+  const enableSuite = process.env.ENABLE_COMPLIANCE_IGNORE_FIX === '1';
```
regexowl: I can't see the variable being set anywhere, are the tests skipping this suite every time? Also why do we need to gate this?

playwright/Customizations/ComplianceWarnings.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+  // Find and remove the package
+  const packageRow = frame.getByRole('row').filter({ hasText: packageName });
+  await packageRow.getByRole('button', { name: 'Remove package' }).click();
+
+  // Save changes
+  await frame.getByRole('button', { name: 'Review and finish' }).click();
+  await frame.getByRole('button', { name: 'Save blueprint' }).click();
+};
+
+test.describe('Compliance warnings ignore flow', () => {
```
tkoscieln: We don't really group tests by `.describe` usually. Using just multiple `test(...)` is fine.

playwright/Customizations/ComplianceWarnings.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+
+// Helper function to create a blueprint with compliance profile that requires specific packages
+const createBlueprintWithCompliance = async (
+  frame: FrameLocator,
+  blueprintName: string,
+  profile: string,
+) => {
+  await frame.getByRole('button', { name: 'Create image blueprint' }).click();
+  await frame.getByTestId('release_select').click();
+  await frame
```
tkoscieln: There are helpers for this in `/BootTests/helpers/targetChooser.ts`

src/Components/ImagesTable/ImagesTableToolbar.tsx
```diff
@@ -37,7 +38,7 @@ import { BuildImagesButton } from '../Blueprints/BuildImagesButton';
 import { DeleteBlueprintModal } from '../Blueprints/DeleteBlueprintModal';
 import { EditBlueprintButton } from '../Blueprints/EditBlueprintButton';
 
-interface imagesTableToolbarProps {
+interface ImagesTableToolbarProps {
```
ksiekl: We're usually using type to declare the shape of data. Although this is technically correct, `interface` is usually defines a contract that an object must adhere to. I don't think this is that case.

playwright/Customizations/ComplianceWarnings.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+    await test.step('Ignore warnings', async () => {
+      await frame.locator('#blueprint_ignore_warnings').click();
+    });
+
+    await test.step('Update compliance profile to require audit package as well', async () => {
+      await frame.getByRole('button', { name: 'Edit blueprint' }).click();
+      await frame.getByRole('button', { name: 'Compliance' }).click();
+
+      // Switch to a different profile that requires both aide and audit
+      await frame.getByRole('textbox', { name: 'Type to filter' }).fill('stig');
```
tkoscieln: This most likely won't work and crash the test as well, as the testing account does not have any policy named `stig`.

tkoscieln (commented):
Hello, this is a review for the tests mainly, but it doesn't really makes sense to me to merge tests that are not being run at all. If there are bugs preventing the tests from running, then they should be fixed first as the feature is most likely not working/ready to get merged.
I added some comments, but in general there's a lot of code that could have just been copied from the already existing Compliance tests and should more follow the pattern of our other tests.
I also see from the code already that the tests won't work even after the bugs are fixed (I explained why in the review).

PR #3750, opened 2025-10-22
Zod users poc

kingsleyzissou:
- add zod as a dependency
- define zod schemas that can be used for validation
- adds some helper functions to manage the returned errors
- links up the user errors to the user step in the wizard

There are some behavioural differences that may need to be considered.

### DIFF
--- a/src/Components/CreateImageWizard/ValidatedInput.tsx
+++ b/src/Components/CreateImageWizard/ValidatedInput.tsx
@@ -38,7 +38,7 @@ type ValidationInputProp = TextInputProps &
   };
 
 type ErrorMessageProps = {
-  errorMessage: string;
+  errorMessages: string[];
 };
 
 type ValidationResult = 'default' | 'success' | 'error';
@@ -52,14 +52,21 @@ export const ValidatedInputAndTextArea = ({
   onChange,
   ariaLabel,
   inputType = 'textInput',
-  isRequired,
   warning = undefined,
 }: ValidationInputProp) => {
-  const errorMessage = stepValidation.errors[fieldName] || '';
-  const hasError = errorMessage !== '';
+  const errorMessages = stepValidation.errors[fieldName] || [];
+  const hasError = errorMessages.length > 0;
 
   const [isPristine, setIsPristine] = useState(!value);
-  const validated = getValidationState(isPristine, errorMessage, isRequired);
+  const validated = getValidationState(isPristine, errorMessages);
+
+  if (fieldName === 'name') {
+    console.log('Debug');
+    console.log(errorMessages);
+    console.log(stepValidation.errors);
+    console.log(hasError);
+    console.log(validated);
+  }
 
   const handleBlur = () => {
     if (value) {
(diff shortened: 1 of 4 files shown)


src/Schemas/helpers.ts
```diff
@@ -0,0 +1,89 @@
+import { z, ZodSchema } from 'zod/v4';
+import { $ZodError } from 'zod/v4/core';
+
+// TODO: maybe we could have a more descriptive name for this function
+// TODO: we should also make this more generic so that it works on objects too
+// i.e. it might be nice to check an array of users and make sure that the usernames
+// aren't duplicated
```
avitova: I agree. Thankfully, our objects are not as deeply nested, so this is possible. Let's check if there are no deeply nested things, such as an array of objects with a parameter of an array of objects. Some objects for Volumes look very complicated:
e.g.
```go
export type Disk = {
  /** Type of the partition table
   */
  type?: ("gpt" | "dos") | undefined;
  minsize?: Minsize | undefined;
  partitions: Partition[];
};

export type Partition = FilesystemTyped | BtrfsVolume | VolumeGroup;

export type BtrfsVolume = {
  type: "btrfs";
  /** The partition type GUID for GPT partitions. For DOS partitions, this field can be used to set the (2 hex digit) partition type. If not set, the type will be automatically set based on the mountpoint or the payload type.
   */
  part_type?: string | undefined;
  minsize?: Minsize | undefined;
  subvolumes: BtrfsSubvolume[];
};

export type BtrfsSubvolume = {
  /** The name of the subvolume, which defines the location (path) on the root volume
   */
  name: string;
  /** Mountpoint for the subvolume
   */
  mountpoint: string;
};
```

This is basically: Disk: { Partitions: [ BtrfsVolume: [ BtrfsSubvolume: name ] ]}.
And now to decide if two partitions are equal is tricky, sorting needs to be incorporated, but still... I'd try to think this through in advance before implementing such function. But maybe also this differs based on the specific step and field - sometimes we do not care that an array has two same values, sometimes we do. I am just afraid that we won't be able to create one generic function for this.

src/Schemas/helpers.ts
```diff
@@ -0,0 +1,89 @@
...
+  const result = Schema.safeParse(value);
+
+  if (result.success) {
+    return [];
+  }
+
+  return unwrapField(result.error);
+};
+
+export const unwrapForm = (error: $ZodError) => {
```
avitova: Oh niooo, I had to write a very similar function for my thesis. This is quite complex though - I know there is no different way than do this while looking complex.
Why do we need this? I cannot tell from the context, and it makes the readability a bit worse. But I am pretty sure that we can make this both readable and safe.

avitova (commented):
I like this. It is readable except helpers, but that is probably okay, we won't be really looking at helpers that much after we manage to stabilize this.
One question. Would it be possible to also define "a higher level validation"? Meaning we'd define that Satellite has to have a valid ssh command, a valid certificate? I think so, right?

PR #3770, opened 2025-10-28
package.json: replace @ltd/j-toml with smol-toml (RHEL-123840)

croissanne:
smol-toml has a simpler stringify interface, similar to the built-in
    JSON.stringify. Its last update is also more recent (3 months vs. 3
    years).
    
    The simpler stringify interface will also help with exporting blueprints
    as toml in cockpit.


regexowl (approved):
This is great, thanks!

PR #3778, opened 2025-10-29
packit.yaml: Release into fedora-all (HMS-9619)

regexowl:
This updates targets and dist_git_branches to fedora-all

JIRA: [HMS-9619](https://issues.redhat.com/browse/HMS-9619)

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -24,6 +24,7 @@ jobs:
       - centos-stream-10
       - fedora-41
       - fedora-42
+      - fedora-43
 
   - job: copr_build
     trigger: pull_request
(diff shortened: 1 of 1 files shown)


regexowl:
Not sure when should we update the version and when we should just add it to the list. Does this seem about right? ._.

croissanne (approved):
thanks!

packit.yaml
```diff
@@ -56,11 +57,11 @@ jobs:
   - job: propose_downstream
     trigger: release
     dist_git_branches:
-      - fedora-42
+      - fedora-43
```
croissanne: Oh wait, actually can you keep fedora-42 here?

packit.yaml
```diff
@@ -56,11 +57,11 @@ jobs:
...
       - fedora-rawhide
 
   - job: koji_build
     trigger: commit
     dist_git_branches:
-      - fedora-42
+      - fedora-43
```
croissanne: same here

croissanne (changes requested):
let's keep f42

croissanne (dismissed):
yesssss, thanks!

PR #3779, opened 2025-10-29
cockpit-image-builder: export blueprints (RHEL-123840)

croissanne:
blueprints: enable exporting blueprints in cockpit
    
    Export the blueprint as TOML on premise, and not as JSON. The
    expectation is that users might want to use this blueprint with
    composer-cli or image-builder-cli.
    
    Because the behaviour is different between the cockpit and hosted UIs,
    refrain from making the backend api abstraction.

### DIFF
--- a/playwright/Customizations/OpenSCAP.spec.ts
+++ b/playwright/Customizations/OpenSCAP.spec.ts
@@ -1,3 +1,6 @@
+import * as fsPromises from 'fs/promises';
+import * as path from 'path';
+
 import { expect } from '@playwright/test';
 import { v4 as uuidv4 } from 'uuid';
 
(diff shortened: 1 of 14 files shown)


playwright/helpers/helpers.ts
```diff
@@ -59,32 +59,36 @@ const ON_PREM_RELEASES = new Map([
   ['rhel-10', 'Red Hat Enterprise Linux (RHEL) 10'],
 ]);
 
+
+export const getHostDistro = (): string => {
+  return "fedora-42";
```
croissanne: ignore this, this is just cuz i'm running the pw tests from debian

playwright/Customizations/OpenSCAP.spec.ts
```diff
@@ -171,9 +171,67 @@ test('Create a blueprint with OpenSCAP customization', async ({
       .click();
   });
 
-  // This is for hosted service only as these features are not available in cockpit plugin
   await test.step('Export BP', async () => {
-    await exportBlueprint(page, blueprintName);
+    await exportBlueprint(frame, blueprintName, JSON.stringify(
```
croissanne: for openscap we might actually not want to do this, as the profile could change (and thus the tests would suddenly break without having changed anything).

PR #3798, opened 2025-11-03
Wizard: Add compliance warnings for removed item from policies (HMS-9605)

mgold1234:
Currently, we only display error messages when items are added to compliance policies, and we want to inform the user about items that are removed as well and display warning messges.
This PR adds an informational message when items are removed from a compliance policy.

**Scenario**
 1) Uer creates a blueprint with a compliance policy that includes `firewalld`.
 2) Compliance policy is updated and `firewalld` is removed from it.
**Actual:**
An informational message is displayed to notify the user that items were removed from the compliance policy.

While implementing this warning info message, I found a backend issue with how compliance data is processed.
It looks like the problem is in the buildServiceSnapshots function, which is used in both the “Fix errors automatically” (FixupBlueprint) and “Edit blueprint” (UpdateBlueprint) flows.
Right now, this function resets the compliance snapshot and causes all lint warnings to disappear by mistake.

Scenario 1 – Fix errors automatically
When the user clicks “Fix errors automatically”, the backend removes both errors and warnings.
Because of that, info alerts disappear even though the warnings weren’t fixed.
We expect that only errors message should be removed and warnings message should stay.

Scenario 2 – Edit blueprint
When the user edits any blueprint field, for example; update the blueprint name, the backend also resets all lint warnings, even if nothing related to compliance was changed.
We except that: warnings should stay unless the user actually fixes the issue that display in the message.

In short, the frontend works correctly, but the backend resets all warnings after fixing or editing a blueprint, which makes the UI look like it’s not working as expected.

The decision for those 2 bugs is that it will not fix unless real customers hit this and report an issue -https://issues.redhat.com/browse/HMS-9804

**JIRA:** 
https://issues.redhat.com/browse/HMS-9605

Will add unit tests latter

Screenshot 2025-11-10 at 11 36 21

### DIFF
--- a/playwright/BootTests/Compliance/ComplianceCIS.boot.ts
+++ b/playwright/BootTests/Compliance/ComplianceCIS.boot.ts
@@ -15,7 +15,12 @@ import {
   fillInDetails,
   registerLater,
 } from '../../helpers/wizardHelpers';
-import { deleteCompliancePolicy } from '../helpers/helpers';
+import {
+  createCompliancePolicy,
+  deleteCompliancePolicy,
+  navigateToCompliancePolicy,
+  removeCompliancePolicyRule,
+} from '../helpers/helpers';
 import {
   buildImage,
   constructFilePath,
@@ -39,7 +44,7 @@ test('Compliance step integration test - CIS', async ({ page, cleanup }) => {
 
   // Delete the blueprint compliance policy and Openstack resources after the run
   await cleanup.add(() => deleteBlueprint(page, blueprintName));
-  await cleanup.add(() => deleteCompliancePolicy(page, policyName));
+  cleanup.add(() => deleteCompliancePolicy(page, policyName));
   await cleanup.add(() => OpenStackWrapper.deleteImage(blueprintName));
   await cleanup.add(() => OpenStackWrapper.deleteInstance(blueprintName));
 
@@ -47,44 +52,7 @@ test('Compliance step integration test - CIS', async ({ page, cleanup }) => {
   // TODO: because of the empty state in Compliance service when new user has not registered a system yet
   await login(page, true);
 
-  await test.step('Create a Compliance policy', async () => {
-    await page.goto('/insights/compliance/scappolicies');
-    await page.getByRole('button', { name: 'Create new policy' }).click();
-    await page.getByRole('option', { name: 'RHEL 10' }).click();
-    await expect(
-      page.getByRole('gridcell', { name: 'ANSSI-BP-028 (enhanced)' }).first(),
-    ).toBeVisible(); // Wait for the policy type list to load
-    // Temporarily disabled - do not use filter, but manually find and select the policy
-    // await page.getByRole('textbox', { name: 'text input' }).fill(policyType);
-    // await expect(
-    //   page.getByRole('gridcell', { name: policyType }).first(),
-    // ).toBeVisible();
-    // await page.getByRole('radio', { name: 'Select row 0' }).click();
-    await page.getByRole('button', { name: 'Go to last page' }).click();
-    await expect(
-      page.getByRole('gridcell', { name: policyType }),
-    ).toBeVisible();
-    await page
-      .getByRole('row')
-      .filter({ hasText: policyType })
-      .getByRole('radio')
-      .first()
-      .click();
-
-    await page.getByRole('button', { name: 'Next', exact: true }).click();
-    await page.getByRole('textbox', { name: 'Policy name' }).fill(policyName); // Get the policy name
-    await page.getByRole('button', { name: 'Next', exact: true }).click(); // Skip "Details"
-    await page.getByRole('button', { name: 'Next', exact: true }).click(); // Skip "Systems"
-    /** TODO: Currently broken
-    // Change rule to see if tailoring works correctly
-    await page.getByRole('textbox', { name: 'text input' }).fill('bluetooth');
-    await page.getByRole('checkbox', { name: 'Select row 0' }).click(); */
-    await page.getByRole('button', { name: 'Next', exact: true }).click();
-    await page.getByRole('button', { name: 'Finish' }).click();
-    await page
-      .getByRole('button', { name: 'Return to application' })
-      .click({ timeout: 2 * 60 * 1000 }); // Wait for the policy to be created
-  });
+  await createCompliancePolicy(page, policyName, policyType, 'RHEL 10');
 
   // Navigate to IB landing page and get the frame
   await navigateToLandingPage(page);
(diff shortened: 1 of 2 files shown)


src/Components/ImagesTable/ImagesTableToolbar.tsx
```diff
@@ -61,6 +71,39 @@ const ImagesTableToolbar: React.FC<imagesTableToolbarProps> = ({
...
+
+  // Update state when blueprint changes
+  React.useEffect(() => {
+    if (!selectedBlueprintId || typeof window === 'undefined') {
+      setShowInfoWarnings(true);
+      setStoredWarnings([]);
+      return;
+    }
+
+    const stored = localStorage.getItem(`warnings-info-${selectedBlueprintId}`);
```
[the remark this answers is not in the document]
lucasgarfield: For instance, the alert about how file system customizations aren't available for the iso isn't dismissible... the only thing we dismiss afaik is the notification about new features, but that feels like a very different situation...

playwright/BootTests/helpers/helpers.ts
```diff
@@ -82,6 +82,140 @@ export const deleteRepository = async (
...
+    // Navigate to rules tab
+    const rulesTab = page.getByRole('tab', {
+      name: /rules|Rules|tailoring|Tailoring/i,
+    });
+    await expect(rulesTab.first()).toBeVisible();
+    await rulesTab.first().click();
+
+    // Enter edit mode
+    const editRulesButton = page.getByRole('button', {
+      name: /edit.*rules|Edit.*rules/i,
```
ksiekl: same here as above, I think the button name visible in the page should be enough, it's clearer which button that is when developing as well, makes the dev experience better

playwright/BootTests/helpers/helpers.ts
```diff
@@ -82,6 +82,140 @@ export const deleteRepository = async (
...
+    await expect(searchInput.first()).toBeEnabled();
+
+    // Search for the rule
+    await searchInput.first().clear();
+    await expect(searchInput.first()).toHaveValue('', { timeout: 5000 });
+    await searchInput.first().fill(ruleName);
+    await expect(searchInput.first()).toHaveValue(ruleName, { timeout: 5000 });
+
+    // Find and uncheck the rule
+    const ruleDisplayPattern = new RegExp(escapeRegExp(ruleDisplayName), 'i');
```
ksiekl: I'm a little lost, why is `escapeRegExp` called on the rule name here?

ksiekl (commented):
Looks good, just added a few comments mostly about how you're finding values in the tests.

The stage is down, and I can't check the warning manually, however, I'm a bit nervous about the bug that was closed - does the UI look like it's buggy from a user's perspective still? Because if the bug is officially closed and won't be fixed, and the UI behaves buggy I don't think we should merge this... 

And second question - does the `preserveExisting` really store the warnings? I'm a bit unsure, it doesn't look like that from the code. Sorry, I just can't check it right now on the stage, so any explanation you have will do

mgold1234:
@ksiekl thank you for all your comments, really approtiate it.
regarding the bug, user can see the warnings but if he decides to fix/ or edit the blueprint this warning will disappear
and regarding the 'preserveExisting' I recheck it again and it makes sense to remove it. thank you for that! 
https://github.com/user-attachments/assets/7c334b5d-03d1-410b-bfe8-60f98fc63438

src/Components/ImagesTable/ImagesTableToolbar.tsx
```diff
@@ -60,11 +62,50 @@ const ImagesTableToolbar: React.FC<imagesTableToolbarProps> = ({
...
+
+  const lintWarnings = React.useMemo(
+    () => blueprintDetails?.lint?.warnings ?? [],
+    [blueprintDetails],
+  );
+
+  React.useEffect(() => {
+    if (selectedBlueprintId && blueprintDetails !== undefined) {
+      dispatch(
+        setWarningsContent({
```
ondrejbudai: Note that I'm a React/Redux noob so please forgive me if this question is utterly stupid. Why do we need to store the warnings in the redux store? Cannot we just retrieve them from the blueprint details that we already fetched, and render them? I don't understand the need to keep a second copy in the local state.

playwright/Customizations/Compliance.spec.ts
```diff
@@ -83,3 +89,108 @@ test('Create a blueprint with Compliance policy selected', async ({
...
+    await expect(page.getByRole('row', { name: policyName })).toBeVisible({
+      timeout: 60000,
+    });
+  });
+
+  if (isHosted()) {
+    await page.goto('/insights/image-builder/imagewizard?release=rhel9');
+  } else {
+    await page.goto('/imageWizard?release=rhel9');
+  }
```
avitova: I am not sure that I understand this. Would this not fail on different OSes? I must admit that testing for compliance in the cockpit might be a bit more complicated. Since this PR is big enough, I suggest creating a separate ticket for testing the feature for cockpit, and skipping tests for it for now. What do you think?

playwright/Customizations/Compliance.spec.ts
```diff
@@ -83,3 +89,108 @@ test('Create a blueprint with Compliance policy selected', async ({
...
+    const searchInput = updatedFrame.getByRole('textbox', {
+      name: 'Search input',
+    });
+    await expect(searchInput).toBeVisible();
+    await searchInput.fill(blueprintName);
+    await expect(searchInput).toHaveValue(blueprintName);
+
+    const blueprintButton = updatedFrame.locator(
+      `button[id="${blueprintName}"]`,
+    );
```
avitova: If we can use a different locator than `locator`, that would be nice. I'd help here, but I currently do not have resources for testing this due to stage refresh.

regexowl (commented):
This works really nicely, thanks!

The only idea I have for the functionality is maybe adding a small spinner after the user clicked on "fix errors automatically" as it hangs for a while. But that's a nit pick that could be addressed in a follow up, so leaving things as they are now is perfectly fine.

Added comments for the tests.

playwright/Customizations/Compliance.spec.ts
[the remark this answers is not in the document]
mgold1234: @regexowl wooo this is really good idea to put the test in the boot tests, thank you!!

playwright/BootTests/Compliance/ComplianceCIS.boot.ts
```diff
@@ -207,3 +180,109 @@ test('Compliance step integration test - CIS', async ({ page, cleanup }) => {
...
+  await test.step('Verify compliance warning appears in blueprint', async () => {
+    await navigateToLandingPage(page);
+    const updatedFrame = await ibFrame(page);
+
+    const searchInput = updatedFrame.getByRole('textbox', {
+      name: 'Search input',
+    });
+    await expect(searchInput).toBeVisible({ timeout: 60000 });
+    await searchInput.fill(blueprintName);
+    await expect(searchInput).toHaveValue(blueprintName, { timeout: 10000 });
```
regexowl: Is the timeout necessary?

playwright/BootTests/Compliance/ComplianceCIS.boot.ts
```diff
@@ -34,54 +34,27 @@ test('Compliance step integration test - CIS', async ({ page, cleanup }) => {
   const blueprintName = 'compliance-cis-test-' + uuidv4();
   const policyName = 'test-policy-' + uuidv4();
   const policyType =
-    'DRAFT - CIS Red Hat Enterprise Linux 10 Benchmark for Level 2 - Workstation';
+    'CIS Red Hat Enterprise Linux 10 Benchmark for Level 1 - Server';
```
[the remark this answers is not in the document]
mgold1234: I thought maybe it is better to choose a value from the first page, to make sure it will pass. 
I will revert that

PR #3802, opened 2025-11-04
Wizard: fix interconnected ids of compliance policy & oscap profile (HMS-9596)

ksiekl:
The id of compliance policy got set for oscap id as well when selecting it in the Security step. This PR splits the former `changeCompliance` action into two - one for compliance and one for oscap to be set. There's also a minor fix in the name of a test step in the Compliance PW test just to make it more precise :)


tkoscieln (approved):
lgtm, but maybe @regexowl or @avitova could double check please?

PR #3804, opened 2025-11-04
Wizard: Fix case for Red Hat Lightspeed compliance

ochosi:
As product names should be capitalized and features or components should be lowercase, we should go with 'compliance' over 'Compliance'. Also, there was a missing space in the alert.


mgold1234 (approved):
/lgtm

PR #3822, opened 2025-11-12
tests: Fix Compliance radio not being checked

tkoscieln:
Make sure that Compliance is selected in the Security step before selecting policy.


kingsleyzissou (approved):

tkoscieln:
Turns out, the Compliance policy was not being selected in the creation due to delay in the selection itself. Added an expect to fix that and reworked the "Edit BP" step of the tests - previously it was selecting the policy again while it should have just checked it is already selected.

PR #3825, opened 2025-11-14
ghaction/release: Pass Slack bot token instead of webhook

ochosi:
We have updated our Slack integration to use a Slack bot token instead of a webhook to be able to post threaded responses or react to messages. For reference see: https://github.com/osbuild/release-action/pull/35

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
         uses: osbuild/release-action@main
         with:
           token: "${{ secrets.SCHUTZBOT_GITHUB_ACCESS_TOKEN }}"
-          slack_webhook_url: "${{ secrets.SLACK_WEBHOOK_URL }}"
+          slack_bot_token: "${{ secrets.SLACK_BOT_TOKEN }}"
 
       # upload release artefact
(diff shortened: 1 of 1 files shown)


kingsleyzissou (approved):

achilleas-k (approved):

tkoscieln (changes requested):
There's two more occurences where we use Slack webhook [here](https://github.com/osbuild/image-builder-frontend/blob/main/.github/workflows/boot-tests-nightly.yml#L115) and [here](https://github.com/osbuild/image-builder-frontend/blob/main/.github/workflows/boot-tests-nightly.yml#L139). Replacing those might be a little bit more complicated though :thinking:

ochosi:
> There's two more occurences where we use Slack webhook [here](https://github.com/osbuild/image-builder-frontend/blob/main/.github/workflows/boot-tests-nightly.yml#L115) and [here](https://github.com/osbuild/image-builder-frontend/blob/main/.github/workflows/boot-tests-nightly.yml#L139). Replacing those might be a little bit more complicated though :thinking: 

I don't think it'd be hard, but I'm wondering if it'd be worth it. The webhook works fine for single messages. I'm only using the bot token to be able to do threaded responses.

So unless we want to surface more details, e.g. which part of the nightly failed in a threaded response, it's not a requirement to migrate.

kingsleyzissou:
The manual API check is failing for other reasons now, will look at next week.

PR #3830, opened 2025-11-17
API: Comment out provisioning URL in codegen

regexowl:
This disables re-generating provisioning API schema from the original URL. The API was decommissioned and is no longer available.

A follow up PR to clean up the rest of provisioning from the codebase will be required, but this should fix an issue with failing manual changes check in the CI.


regexowl:
Closing as duplicate of https://github.com/osbuild/image-builder-frontend/pull/3829

PR #3832, opened 2025-11-18
BootTests: Check system was registered to Compliance policy (HMS-9574)

tkoscieln:
Extend the CIS Boot test to check if the system was successfully registered to the created policy.

### DIFF
--- a/playwright/BootTests/Compliance/ComplianceCIS.boot.ts
+++ b/playwright/BootTests/Compliance/ComplianceCIS.boot.ts
@@ -13,13 +13,10 @@ import {
   createBlueprint,
   deleteBlueprint,
   fillInDetails,
-  registerLater,
 } from '../../helpers/wizardHelpers';
 import {
   createCompliancePolicy,
   deleteCompliancePolicy,
-  navigateToCompliancePolicy,
-  removeCompliancePolicyRule,
 } from '../helpers/helpers';
 import {
   buildImage,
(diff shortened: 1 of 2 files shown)


playwright/BootTests/Compliance/ComplianceCIS.boot.ts
```diff
@@ -177,6 +183,21 @@ test('Compliance step integration test - CIS', async ({ page, cleanup }) => {
     expect(exitCode).toBe(0);
     expect(output).toContain('0');
   });
+
+  await test.step('Register system to insights client', async () => {
+    // System with compliance policy is by default not registered to Insights
```
ochosi: Not directly related to this test, but: Shouldn't this be something to warn about in the UI?

playwright/BootTests/Compliance/ComplianceCIS.boot.ts
```diff
@@ -177,6 +183,21 @@ test('Compliance step integration test - CIS', async ({ page, cleanup }) => {
...
+    expect(exitCode).toBe(0);
+  });
+
+  await test.step('Check system was registered to the policy', async () => {
+    await page.goto('/insights/compliance/scappolicies');
+    await page.getByRole('textbox', { name: 'text input' }).fill(policyName);
+    await expect(page.getByRole('row', { name: policyName })).toBeVisible();
+    await page.getByRole('link', { name: policyName }).click();
+    await page.getByRole('tab', { name: 'Systems' }).click();
+    await expect(page.getByText('1 - 1').nth(2)).toBeVisible();
```
ochosi: Can't we simply look for the `blueprintName` here (since this is OpenStack, that's likely the instance name) to make this less brittle?

playwright/BootTests/Compliance/ComplianceCIS.boot.ts
tkoscieln: Yes I suppose. This is compliance specific behaviour so we should propably warn about that somewhere in the Compliance step? @lucasgarfield @regexowl

regexowl: Would you say just stating this in a description of the step would work? As to not overwhelm the UI with alerts. Or would we want to go with an info alert anyway?

regexowl:
@tkoscieln Looks good! Could you please squash the two "BootTests: Check system was registered to Compliance policy" commits into one?

PR #3837, opened 2025-11-19
playwright: Add wait time and extra check

regexowl:
This adds a short wait time to the compliance boot test to allow space for the policy customizations to get applied to the image.

Extra check for the presence of package required by OpenSCAP was added. This shouldn't be the complete check of applied customizations as the other (kernel, services, ...) are missing. The check should be an indication of customizations not being applied timely.


regexowl:
Addressed in https://github.com/osbuild/image-builder-frontend/pull/3838

PR #3852, opened 2025-11-23
test/content: add repeatable build integration test (HMS-9498)

lucasgarfield:
Add end-to-end test for Image Builder and Content integration with repeatable builds using repository snapshots. The test:
- Creates a repository with snapshotting enabled
- Edits the repository to create multiple snapshots
- Uses Image Builder wizard to create a build with an older snapshot
- Verifies the snapshot URL and package on the booted system

### DIFF
--- a/playwright/BootTests/Content/ContentRepeatable.boot.ts
+++ b/playwright/BootTests/Content/ContentRepeatable.boot.ts
@@ -1,0 +1,187 @@
+import { expect } from '@playwright/test';
+import { v4 as uuidv4 } from 'uuid';
+
+import { test } from '../../fixtures/customizations';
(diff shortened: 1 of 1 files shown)


tkoscieln (commented):
LGTM, but won't approve yet as it's weird that the repos are disabled :thinking:

lucasgarfield:
Slack thread from end of last month about this, had some input from Justin, still not super clear to me whether this behavior is what we want but at least this is some additional context:
https://redhat-internal.slack.com/archives/C0235DZB0DT/p1763945650169399

Let's just get this merged once it is green and we'll call the tested behavior the correct behavior until we feel the need to change it.

PR #3853, opened 2025-11-24
dependabot: Update grouping of updates (HMS-9798)

regexowl:
There's been issues with how the dependabot updates are set up. This should hopefully resolve the issue.

JIRA: [HMS-9798](https://issues.redhat.com/browse/HMS-9798)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-      day: "monday"
-      time: "04:00"
     open-pull-requests-limit: 3
     rebase-strategy: "auto"
     ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,17 +4,14 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-      day: "monday"
-      time: "04:00"
```
regexowl: I don't think we really care about specific time, let's just check weekly.

regexowl:
> * The `ignore` block still ignores all update types for `dependency-name: "*"`, which will prevent Dependabot from ever opening PRs and makes the new grouping ineffective—consider narrowing this ignore list (e.g., to majors only) if you want updates to occur.

What are you talking about? Am I blind?

PR #3854, opened 2025-11-24
package-lock: Freshly install dependencies (HMS-9793)

regexowl:
This is the result of removing `node_modules` and `package-lock.json` and running `npm i` to bring our lock file in sync with `package.json`.

The changes were caused by `lock file maintenance` PRs that updated transitive deps without bumping direct ones.

JIRA: [HMS-9793](https://issues.redhat.com/browse/HMS-9793)


regexowl:
This would also resolve following warnings in `npm i` output

```
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported
npm warn deprecated abab@2.0.6: Use your platform's native atob() and btoa() methods instead
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
```

PR #3855, opened 2025-11-25
src: Fix rendering validation for `ValidatedInput` (HMS-9797)

regexowl:
Since the `isPristine` value got set to `true` every time the field was empty, the Next button usually got correctly disabled, but no error was rendered.

Currently the `isPristine` doesn't get re-set when input is deleted, meaning there can be a green check mark when the value is not required, but if it is required, the error message gets rendered properly. I believe lingering green checkmark will be much less confusing for the user than disabled Next buttons without any indication of error.

JIRA: [HMS-9797](https://issues.redhat.com/browse/HMS-9797)


ksiekl (approved):
Looks great! Thank you 💟

PR #3859, opened 2025-11-26
Wizard: refactor ManageButton since it became widely used

ksiekl:
The former ManageButton is used in multiple different components so I took it out to the utilities folder and renamed :black_cat:


regexowl (approved):
10/10, no notes. Thanks!

PR #3864, opened 2025-11-26
Wizard: fix snapshot date assignment logic in Wizard (HMS-9629)

katarinazaprazna:
When a user enables repeatable builds (snapshot mode) and leaves the default value in the `DatePicker` without interacting with it, the `onChange` handler doesn't fire. This means:
- If the default value is the current date in YYYY-MM-DD format, it never gets converted to RFC3339/ISO format
- If the default value is an empty string, it remains as an empty string in the state

The `onChange` handler, which is responsible for parsing and converting the date to the proper format, only runs on user interaction - not when the field contains its initial default value.

### Description

- Ensure the content-sources `snapshots/for_date` API receives properly formatted RFC3339 dates when users don't interact with the `DatePicker`
- Ensure the default snapshot date uses midnight UTC when toggling "Use Latest" off, matching the behavior of `changeSnapshotDate` - not mixing midnight and current timestamp formats

### Notes

Related backend PR which updates the API documentation https://github.com/osbuild/image-builder-crc/pull/1734

### DIFF
--- a/src/Components/CreateImageWizard/utilities/requestMapper.ts
+++ b/src/Components/CreateImageWizard/utilities/requestMapper.ts
@@ -626,7 +626,10 @@ const getImageRequests = (
       type: uploadTypeByTargetEnv(type),
       options: getImageOptions(type, state),
     },
-    snapshot_date: !useLatest && !template ? snapshotDate : undefined,
+    // Only include snapshot_date when using snapshot mode (not latest or template)
+    // Convert empty strings to undefined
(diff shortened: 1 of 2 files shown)


katarinazaprazna:
Ah, I see! Good catch. @ksiekl, I've made that adjustment now :)

ksiekl (approved):
Heh, not my catch, but I'll happily take the credit 😄 Thank you for fixing this ❤️

PR #3878, opened 2025-12-02
gitlab: set the correct tag for rhos runners

ondrejbudai:
schutzbot uses tags for limiting the number of concurrent PRs on different platforms. This is needed because we have different quotas on AWS vs. OpenStack. See
https://github.com/osbuild/gitlab-ci-config/blob/main/config.toml.

Since this project uses solely openstack, let's use the terraform/openstack tag.


mgold1234 (approved):
/lgtm

PR #3882, opened 2025-12-03
src: Add 'auto-lvm' option to partitioning mode and re-enable tests (HMS-9789)

regexowl:
This adds a default 'auto-lvm' radio to the partitioning mode. As the partitioning mode can also be undefined a new checkbox was added. The checkbox allows the user to choose whether they'd like to specify partitioning mode used in the image or not.

There were also some disabled tests in the Disk playwright suite:
- test checking for correctly selected partitioning mode radios
- "Review exported BP" - on-prem export wasn't previously working, since the issue was fixed the test was re-enabled and used fixtures were updated.

JIRA: [HMS-9789](https://issues.redhat.com/browse/HMS-9789)

### DIFF
--- a/src/Components/CreateImageWizard/steps/FileSystem/components/PartitioningMode.tsx
+++ b/src/Components/CreateImageWizard/steps/FileSystem/components/PartitioningMode.tsx
@@ -24,7 +24,7 @@ const PartitioningMode = () => {
             </Label>
           </>
         }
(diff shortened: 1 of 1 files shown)


ochosi (approved):
I tested this in Cockpit: all the images built successfully, and the compose requests looked as expected. Nice work!

PR #3883, opened 2025-12-03
Wizard: add explanation of which services are expected in firewall (RHIN-2168)

ksiekl:
Previously, we had ambiguity in the Firewall step - not specifying which services could be added to the inputs. I added the explanation to the description:

image

### DIFF
--- a/src/Components/CreateImageWizard/steps/Firewall/components/Services.tsx
+++ b/src/Components/CreateImageWizard/steps/Firewall/components/Services.tsx
@@ -25,7 +25,7 @@ const Services = () => {
       <FormGroup label='Enabled services'>
         <LabelInput
           ariaLabel='Add enabled service'
-          placeholder='Enter service name'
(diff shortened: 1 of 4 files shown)


src/Components/CreateImageWizard/steps/Firewall/components/Services.tsx
```diff
@@ -47,6 +54,13 @@ const Services = () => {
           stepValidation={stepValidation}
           fieldName='disabledServices'
         />
+        <HelperText>
+          <HelperTextItem variant='default'>
+            Use firewalld service names (e.g., <code>ssh</code>) rather than
+            systemd unit names. List available services with{' '}
```
mgold1234: maybe write: 
'To see all available services, run firewall-cmd --get-services'
instead of 
'List available services with{' '}<code>firewall-cmd --get-services</code>.'

PR #3885, opened 2025-12-04
ImagesTable: expose compose progress information (HMS-9341)

croissanne:
This introduces progress information when available. More detailed progress (including messages), still need to be implemented and forwarded from composer.

---

image

### DIFF
--- a/src/Components/ImagesTable/Status.tsx
+++ b/src/Components/ImagesTable/Status.tsx
@@ -90,7 +90,8 @@ export const AwsDetailsStatus = ({ compose }: ComposeStatusPropTypes) => {
         />
       );
     }
-
+    case 'building':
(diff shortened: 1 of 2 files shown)


achilleas-k (commented):
I endorse this wholeheartedly.  But I'll let people who know the repo better do the actual reviews and approvals.

ochosi:
> Thank you, this works great in service!
> 
> ~Tried to check it in cockpit also, it went from in progress directly to "Ready". First time I though I've missed it, but the second time I've stared really intently and I'm quite sure no progress was seen.~
> 
> Nevermind, @ochosi correctly pointed out I didn't test with up-to-date composer version. Still need to do that

This totally works in cockpit-image-builder as is (see screenshot below).
image

PR #3898, opened 2025-12-08
konflux: Disable dependency bumps (HMS-9861)

regexowl:
As we're managing dependency updates via dependabot, there's no need to generate them via konflux as well.

JIRA: [HMS-9861](https://issues.redhat.com/browse/HMS-9861)


kingsleyzissou (approved):
+1000

PR #3914, opened 2025-12-12
playwright: Update policy name (HMS-9879)

regexowl:
The policy was taken out of draft, this reflects the change. A small to check to see if the policy is already visible before attempting to click on it was also added.

JIRA: [HMS-9879](https://issues.redhat.com/browse/HMS-9879)


regexowl:
The compliance boot test should pass, but the content boot tests are expected to fail, there are some issues with snapshots on the content side.

ochosi (approved):
LGTM!

regexowl:
> @regexowl can we merge it?

Well, no. Even if we were not blocked on content, the compliance check is still failing for some reason :eyes:

Issue #3913, opened 2025-12-12
[BUG] SPUR: The help icon isn't centered (HMS-9878)


ochosi:
**Bug description**
The help icon isn't centered in the button in Image Builder's header. (see screenshot)

<img width="257" height="78" alt="Image" src="https://github.com/user-attachments/assets/db2c6abe-67f7-44ec-a241-51b9c6b08e93" />

The culprit is this custom CSS: https://github.com/osbuild/image-builder-frontend/blob/main/src/Components/sharedComponents/ImageBuilderHeader.scss#L6

Upon removing it, the icon is centered again.

**Reproduction steps**
1. Go to the service and look at the icon on the LandingPage.

**Expected behavior**
The icon should be centered.

JIRA: [HMS-9878](https://issues.redhat.com/browse/HMS-9878)

PR #3915, opened 2025-12-12
ImageBuildHeader: center help icon (HMS-9878)

kingsleyzissou:
The custom css was causing the help icon not to be centered inside the button. This was affecting both on-prem & hosted.

Fixes: https://github.com/osbuild/image-builder-frontend/issues/3913


regexowl (approved):
Works nicely! Could we get rid of the entire `ImageBuilderHeader.scss` file? I'm trying to compare how the page looks with and without the styling and can't see a difference.

But happy to leave that for a follow, it might not be in scope.

regexowl (approved):
:rocket:

PR #3917, opened 2025-12-12
Wizard: add langpack if locale is selected (HMS-8766)

ksiekl:
Adds langpack based on a user's locale selection. User selects a locale, the code looks if such a langpack for the locale exists (ignoring C langpack), and if it does, it quietly adds it to the request. There is no sign of the package in the "Additional packages" step, but I did add and information for the user in the "Locale" step, we can remove that as well if deemed unnecessary.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Locale/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Locale/index.tsx
@@ -1,13 +1,48 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 
-import { Content, Form, Title } from '@patternfly/react-core';
+import { Content, Form, Spinner, Title } from '@patternfly/react-core';
+
+import { useSearchLanguagePacks } from '@/store/api/distributions';
 
 import KeyboardDropDown from './components/KeyboardDropDown';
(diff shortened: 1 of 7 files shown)


src/Components/CreateImageWizard/steps/Locale/index.tsx
```diff
@@ -1,13 +1,97 @@
...
+      dispatch(setLocaleLangpacks([]));
+      return;
+    }
+
+    if (!process.env.IS_ON_PREMISE && distroUrls.length === 0) {
+      return;
+    }
+
+    let cancelled = false;
+    const run = async () => {
```
kingsleyzissou: This is quite a bit of business logic inside a component, maybe we should extract it to its own hook?
Like a `useSearchLanguagePacks` hook or something like that? 

We could also put that hook in the store, similar to what we did with `store/distributions/hooks` just for the sake of consistency.

regexowl:
@ksiekl Looks good! Can you please resolve the conflicts?

ksiekl:
thank you @regexowl I'll be off for the week in a while but I'll be checking if it can be rebased on thu and fri :saluting_face:

PR #3924, opened 2025-12-16
LandingPage: Update "Service unavailable" alert with a payload (HMS-9913)

regexowl:
This allows to render more information about outage via parsing an unleash payload.

JIRA: [HMS-9913](https://issues.redhat.com/browse/HMS-9913)

### DIFF
--- a/src/Components/LandingPage/ServiceUnavailableAlert.tsx
+++ b/src/Components/LandingPage/ServiceUnavailableAlert.tsx
@@ -1,13 +1,29 @@
 import React from 'react';
 
 import { Alert } from '@patternfly/react-core';
+import { useVariant } from '@unleash/proxy-client-react';
 
 const ServiceUnavailableAlert = () => {
+  const variant = useVariant('image-builder.service-unavailable');
+  const payload = variant.payload
+    ? JSON.parse(variant.payload.value)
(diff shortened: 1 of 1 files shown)


src/Components/LandingPage/ServiceUnavailableAlert.tsx
```diff
@@ -1,13 +1,29 @@
 import React from 'react';
 
 import { Alert } from '@patternfly/react-core';
+import { useVariant } from '@unleash/proxy-client-react';
 
 const ServiceUnavailableAlert = () => {
+  const variant = useVariant('image-builder.service-unavailable');
+  const payload = variant.payload
+    ? JSON.parse(variant.payload.value)
+    : undefined;
```
ochosi: I assume `undefined` results in the payload not being rendered and prevents rendering issues?

ochosi (dismissed):
Love it - thank you <3

src/Components/LandingPage/ServiceUnavailableAlert.tsx
regexowl: Indeed. Had to add some optional chaining for good measure, but the default alert if the payload is not defined should look like this:
image

PR #3929, opened 2025-12-17
Wizard: Add groups section to Users step (HMS-8784)

mgold1234:
This added the new "Groups" section to the Users step, allowing users to define groups prior to
assigning users to them. This prevents image build failures caused by missing group definitions.

implement it as per as the mock from Shayna:

Screenshot 2025-12-11 at 12 49 09

JIRA: [HMS-8784](https://issues.redhat.com/browse/HMS-8784)

### DIFF
--- a/playwright/Customizations/Users.spec.ts
+++ b/playwright/Customizations/Users.spec.ts
@@ -99,7 +99,7 @@ test('Create a blueprint with Users customization', async ({
     ).toBeVisible();
     await expect(
       frame.getByText(
-        'Each group will automatically be assigned an ID number.',
+        'Define groups before assigning users to them. Each group will be created on systems using this image. Each group will automatically be assigned an ID number.',
       ),
     ).toBeVisible();
 
@@ -117,9 +117,18 @@ test('Create a blueprint with Users customization', async ({
     const gidInputs = frame.getByPlaceholder('Auto-generated');
     await expect(gidInputs.nth(0)).toHaveValue(/^\d+$/);
     await expect(gidInputs.nth(1)).toHaveValue(/^\d+$/);
+    // Verify GID inputs are disabled (read-only)
+    await expect(gidInputs.nth(0)).toBeDisabled();
+    await expect(gidInputs.nth(1)).toBeDisabled();
   });
 
   await test.step('Test group validation errors', async () => {
+    // Get Groups section container for scoped assertions
+    const groupsSection = frame
+      .getByRole('heading', { name: 'Groups' })
+      .first()
+      .locator('..');
+
     const groupNameInputs = frame.getByPlaceholder('Set group name');
     const addGroupButton = frame.getByRole('button', { name: 'Add group' });
     await expect(addGroupButton).toBeVisible();
@@ -130,20 +139,20 @@ test('Create a blueprint with Users customization', async ({
     await expect(newGroupInput1).toBeVisible();
     await expect(newGroupInput1).toBeEnabled();
 
-    await expect(frame.getByText('Invalid group name')).toBeHidden();
+    await expect(groupsSection.getByText('Invalid group name')).toBeHidden();
     await newGroupInput1.fill('@invalid');
     await expect(newGroupInput1).toHaveValue('@invalid');
-    await expect(frame.getByText('Invalid group name')).toBeVisible();
+    await expect(groupsSection.getByText('Invalid group name')).toBeVisible();
 
     const addGroupButtons = frame.getByRole('button', { name: 'Add group' });
     await expect(addGroupButtons.last()).toBeDisabled();
 
     await newGroupInput1.clear();
-    await expect(frame.getByText('Invalid group name')).toBeHidden();
+    await expect(groupsSection.getByText('Invalid group name')).toBeHidden();
     await expect(newGroupInput1).toHaveValue('');
     await newGroupInput1.fill('develop');
     await expect(newGroupInput1).toHaveValue('develop');
-    await expect(frame.getByText('Invalid group name')).toBeHidden();
+    await expect(groupsSection.getByText('Invalid group name')).toBeHidden();
     await expect(addGroupButtons.last()).toBeEnabled();
 
     await newGroupInput1.clear();
@@ -151,7 +160,7 @@ test('Create a blueprint with Users customization', async ({
     await newGroupInput1.fill('admins');
     await expect(newGroupInput1).toHaveValue('admins');
     await expect(
-      frame.getByText('Group name already exists').first(),
+      groupsSection.getByText('Group name already exists').first(),
     ).toBeVisible();
     await expect(addGroupButtons.last()).toBeDisabled();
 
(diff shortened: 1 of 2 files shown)


regexowl:
Quick observations:
- since "Groups" seems to be a sort of pre-requisite to "Users" would it make sense to put it above users on the step?
- the step can be also renamed to "Users and groups" or smth to reflect that there's additional functionality
- group ID input should be probably disabled as we're auto-generating the value and the user being able to click into it, but not edit it seems a bit weird
- on a bit more general note - should the GID be editable by user or is that something that is not usually handled by the user?
- Schutzbot tests are failing

Also a group can be assigned to a user without it being created first, would we be able to at least partially safeguard that or maybe add some floating info like "You've assigned user to a custom group that is not defined in 'Groups', make sure it will exist on the system"?

I assume there would be a default set of system groups that will work just fine, groups that will be created with a package installation and groups that need to be specified in "Groups" in order for them to work. @achilleas-k would we be able to formulate at least a list for the first case of "system defined groups" based on distro/target/... or not really?

src/Components/CreateImageWizard/steps/Users/components/GroupRow.tsx
```diff
@@ -0,0 +1,127 @@
...
+            >
+              Add group
+            </Button>
+          )}
+        </FormGroup>
+      </GridItem>
+      <GridItem span={4}>
+        <FormGroup label='Group ID' className='pf-v6-u-pb-md'>
+          <TextInput
+            value={hasName && groupGid ? groupGid.toString() : ''}
```
regexowl: Isn't the GID stored in the state? Shouldn't we be able to just load it from there? Can group without a name have a GID?

src/store/wizardSlice.ts
```diff
@@ -1192,6 +1220,73 @@ export const wizardSlice = createSlice({
         state.groups.splice(index, 1);
       }
     },
+    addUserGroup: (state, action: PayloadAction<string>) => {
+      const trimmedName = action.payload.trim();
+      const existingGroupIndex = state.userGroups.findIndex(
+        (group) => group.name === trimmedName,
+      );
```
regexowl: This looks like "add" and "update" function merged into one, making it more complicated than it necessarily needs to be. "Add group" should just push a new group object to the user groups array. Then there could be "update group by id/gid" action, that will take the name as an argument and update the correct index with it. Or am I missing something?

src/store/wizardSlice.ts
```diff
@@ -1192,6 +1220,73 @@ export const wizardSlice = createSlice({
...
+      }
+    },
+    removeUserGroupById: (state, action: PayloadAction<string>) => {
+      const index = state.userGroups.findIndex(
+        (group) => group.id === action.payload,
+      );
+      if (index !== -1) {
+        state.userGroups.splice(index, 1);
+      }
+    },
```
regexowl: This is good :+1:

src/Components/Blueprints/helpers/onPremToHostedBlueprintMapper.tsx
```diff
@@ -144,14 +147,29 @@ export const mapOnPremToHosted = async (
...
           : undefined,
       users:
         users !== undefined || user_keys !== undefined
           ? [...(users ? users : []), ...(user_keys ? user_keys : [])]
           : undefined,
-      groups: blueprint.customizations?.groups,
+      groups: (() => {
+        // Merge both groups and group arrays if both are present to prevent data loss
+        const allGroups = [
+          ...(blueprint.customizations?.groups || []),
```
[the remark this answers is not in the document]
mgold1234: @regexowl hope this will make sense why both group and groups?
Different platform formats:
On-Prem (Cockpit): Uses customizations.group (singular) in TOML format
Hosted (Cloud API): Uses customizations.groups (plural) in JSON format
We need both to:
Support importing blueprints from on-prem to hosted
Maintain backwards compatibility with existing on-prem blueprints
Prevent data loss when blueprints contain both fields
The code merges both arrays and deduplicates, ensuring no data is lost during import/export between platforms.
For new blueprints: Use groups - it's the current standard. The group field is deprecated but still supported for compatibility.

src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
```diff
@@ -17,6 +17,7 @@ const UserInfo = () => {
     users.length === 0
       ? [
           {
+            id: 'temp-user',
```
regexowl: Why is the extra id needed for user?

playwright/fixtures/data/exportBlueprintContents.ts
```diff
@@ -142,6 +142,18 @@ languages = [ "C.UTF-8" ]`;
 export const exportedUsersBP = (blueprintName: string): string => {
   return `name = "${blueprintName}"
 
+[[customizations.group]]
+name = "developers"
+gid = <GID_PLACEHOLDER>
```
regexowl: Why not user a "real" GID?

playwright/Customizations/Users.spec.ts
```diff
@@ -134,11 +287,13 @@ test('Create a blueprint with Users customization', async ({
...
     await expect(
-      frame.getByText('Username already exists').first(),
+      frame.getByText('User name already exists').first(),
     ).toBeHidden();
     await usernameInputs.nth(3).fill('admin1');
+    // Trigger validation by blurring the input
+    await usernameInputs.nth(3).blur();
     await expect(
-      frame.getByText('Username already exists').first(),
+      frame.getByText('User name already exists').first(),
```
regexowl: Same here, no space in username.

src/store/wizardSlice.ts
```diff
@@ -1220,6 +1234,37 @@ export const wizardSlice = createSlice({
...
+        while (existingGids.has(nextGid) && nextGid <= 65534) {
+          nextGid++;
+        }
+        if (nextGid <= 65534) {
+          state.userGroups[index].gid = nextGid;
+        }
+      }
+    },
+    removeUserGroup: (state, action: PayloadAction<number>) => {
+      state.userGroups.splice(action.payload, 1);
```
regexowl: The splicing feels a bit "random". Either we should check if the index we want to splice exists in the user groups array and splice after it's found. Or `.filter` method could work, similar to `removeUser`.

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -970,6 +979,17 @@ const getModules = (state: RootState) => {
   return undefined;
 };
 
+const getUserGroups = (state: RootState) => {
+  const userGroups = selectUserGroups(state)
+    .filter((grp) => grp.name && grp.name.trim().length > 0)
```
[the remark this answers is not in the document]
regexowl: Got it, makes perfect sense, thanks!

src/Components/Blueprints/helpers/onPremToHostedBlueprintMapper.tsx
```diff
@@ -207,12 +223,36 @@ export const mapHostedToOnPrem = (
   };
 
   if (blueprint.customizations.packages) {
-    result.packages = blueprint.customizations.packages.map((pkg) => {
-      return {
-        name: pkg,
-        version: '*',
-      };
+    // Separate regular packages from package groups (those starting with @)
```
regexowl: Unnecessary comment, clear from the code

mgold1234:
decided to split that into smaller prs

PR #3930, opened 2025-12-17
Wizard: Rename groups function in the store

mgold1234:
This PR renames group-related functions in the store to avoid confusion. We are adding a new "Groups" section to the Users step, which currently results in three different group-related sections: 
1) User groups
2) Groups
3) Package groups

This change clarifies the naming to better distinguish between them.

### DIFF
--- a/src/store/wizardSlice.ts
+++ b/src/store/wizardSlice.ts
@@ -1167,7 +1167,10 @@ export const wizardSlice = createSlice({
         state.enabled_modules.splice(index, 1);
       }
     },
-    addGroup: (state, action: PayloadAction<GroupWithRepositoryInfo>) => {
+    addPackageGroup: (
+      state,
+      action: PayloadAction<GroupWithRepositoryInfo>,
+    ) => {
       const existingGrpIndex = state.groups.findIndex(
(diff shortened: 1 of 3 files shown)


src/store/wizardSlice.ts
```diff
@@ -1560,8 +1566,8 @@ export const {
   setUserPasswordByIndex,
   setUserSshKeyByIndex,
   setUserAdministratorByIndex,
-  addUserGroupByIndex,
-  removeUserGroupByIndex,
+  assignGroupToUserByIndex,
```
kingsleyzissou: Nitpick: I think we should maybe stick with `add` rather than `assign`. Just for consistency.

Since we're tidying this up though, I think `ByIndex` is also kind of ambiguous. Is it the user index or the group index? Maybe we could do something like `addGroupToIndexedUser` or maybe more verbose (and clearer) `addGroupToUserByUserIndex`.

kingsleyzissou (commented):
Love this change, it really makes sense to clear this up. I just have a nitpick and some thoughts about one of the function names.

regexowl (approved):
Nice! Thanks :heart:

PR #3933, opened 2025-12-18
validation: Don't allow adding groups that match the username (HMS-9933)

### LINKED ISSUE #3932: [BUG] Groups validation should exclude a user's group (HMS-9933)
**Bug description**
The useradd command will fail when creating a user and explicitly adding the user to a group of the same name. 
E.g. `useradd --groups 'wheel,myuser' --password '123456' myuser` will fail with exit status 6, because the 'myuser' group does not exist.

**We've seen multiple build failures like this today.**

Hence, we need to validate this in the frontend (and ideally on the image-builder-crc API layer, too) and not allow adding a group with the same name as the user.

**Reproduction steps**
1. Go to the service and create a blueprint
2. In the user step, add a user and add a group with the same name
3. Go to review
4. Build
5. Observe that the build will fail

**Expected behavior**
We should not allow customers to add a group that doesn't exist, in this case the group with the name of the newly-created user.

ochosi:
A group that has the same name as the user is automatically created by `useradd`. Adding it explicitly leads to a build error, because `useradd` will exist with status 6.

The underlying problem is of course bigger: whenever a user refers to a group that doesn't exist, the build will fail.

Fixes #3932

**Note:** This will have to be amended to also check the user-specified groups once #3929 gets merged, because then customers can add groups explicitly, thus preventing the build error.

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -834,7 +834,16 @@ export function useUsersValidation(): UsersStepValidation {
     }
 
     const duplicateGroups = getListOfDuplicates(users[index].groups);
-    if (invalidGroups.length > 0 || duplicateGroups.length > 0) {
+    const groupMatchingUsername =
+      users[index].name && users[index].groups.includes(users[index].name)
+        ? users[index].name
+        : '';
+
+    if (
+      invalidGroups.length > 0 ||
+      duplicateGroups.length > 0 ||
+      groupMatchingUsername
+    ) {
       const groupsErrors = [];
       if (invalidGroups.length > 0) {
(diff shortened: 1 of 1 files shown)


regexowl:
Just to check - `tom` cannot be a part of group called `tom`, but `jerry` can be a part of `tom` group?

image

ochosi:
> Just to check - `tom` cannot be a part of group called `tom`, but `jerry` can be a part of `tom` group?
> image

@regexowl This is tricky and your comment is spot on.
What your screenshot shows wouldn't work. However, just removing the group `tom` from the user `tom` and keeping it for user `jerry` would work. This is even dependent on the order of users in the list.

It seems me, that the case that customers would feel the need to explicitly define the user group there is comparatively common so I'd try to catch that one (as per this PR).

There's the additional can of worms of any other non-existing group that we won't be able to fix this way.

PR #3934, opened 2025-12-18
packit: Drop Fedora 41 because it's EOL

ochosi:
Currently testing-farm jobs for Fedora 41 spin forever in the frontend pull requests, which is a good reminder that this version is EOL since November.
So let's drop it!


regexowl (approved):
Thank you :heart:

PR #3938, opened 2025-12-19
Wizard: Deny if exact match (HMS-9937)

regexowl:
The mount points from the deny list need to be really denied if they're also and exact match.

JIRA: [HMS-9937](https://issues.redhat.com/browse/HMS-9937)


regexowl:
How to test - try to use a "deny" path as a mountpoint (like `/bin` or `/var/lock`), it shouldn't pass now. Previously it passed unless something was appended to it.

avitova (approved):
lgtm:)

PR #3942, opened 2025-12-22
src: fix error caused by undefined possible value

avitova:
The value passed should not be undefined but let's make this more safe and check during runtime for he compose status.


mgold1234:
@avitova can you please add some details to reproduce this error  to the description?

PR #3944, opened 2025-12-22
Wizard: Add group value on blur, Add user, and Next actions

### LINKED ISSUE #3943: [BUG] - Group field value not added on blur, Add user, or Next actions (HMS-9944)
**Bug description**
Group field value not added on blur, Add user, or Next actions

## Summary
When typing a value in the "Groups" field in the Users customization step, the value is only added to the list when pressing Enter. The value should also be added when:
- Blurring the field (clicking outside)
- Clicking "Add user" button
- Clicking "Next" button

**Reproduction steps**
1. Navigate to the Create Image Wizard
2. Go to the "Users" step
3. Type a group name in the "Groups" field (e.g., "admin")
4. **Do NOT press Enter**
5. Try one of the following:
   - Click outside the field (blur)
   - Click "Add user" button
   - Click "Next" button
6. Observe that the group value is NOT added to the list

## Expected Behavior
The group value should be automatically added to the list when:
- The user blurs the field (clicks outside)
- The user clicks "Add user" button
- The user clicks "Next" button
**Expected behavior**

## Actual Behavior
The group value is only added when explicitly pressing Enter. If the user forgets to press Enter or expects the value to be saved automatically, the value is lost.

JIRA: [HMS-9944](https://issues.redhat.com/browse/HMS-9944)

mgold1234:
This commit fixes a bug where group field value was only added to the list on Enter key press. Now the value is added to the list also when:
- Blurring the field (onBlur)
- Clicking "Add user" button
- Clicking "Next" button

Fixes: https://github.com/osbuild/image-builder-frontend/issues/3943

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 
 import {
   Button,
@@ -13,6 +13,7 @@ import {
 } from '@patternfly/react-core';
 import { WizardStepType } from '@patternfly/react-core/dist/esm/components/Wizard';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
+import { useStore } from 'react-redux';
 import { useNavigate, useSearchParams } from 'react-router-dom';
 
 import { useCustomizationRestrictions } from '@/store/api/distributions';
@@ -41,6 +42,8 @@ import Gcp from './steps/TargetEnvironment/Gcp';
 import TimezoneStep from './steps/Timezone';
 import UsersStep from './steps/UsersAndGroups';
 import {
+  computeUserGroupsValidation,
+  computeUsersValidation,
   useAAPValidation,
   useAzureValidation,
   useDetailsValidation,
@@ -72,6 +75,7 @@ import {
   RHEL_9,
 } from '../../constants';
 import { useGetUser } from '../../Hooks';
+import type { RootState } from '../../store';
 import { selectIsOnPremise, selectPathResolver } from '../../store/envSlice';
 import { useAppDispatch, useAppSelector } from '../../store/hooks';
 import './CreateImageWizard.scss';
@@ -80,6 +84,7 @@ import {
   changeArchitecture,
   changeDistribution,
   changeTimezone,
+  commitAllPendingGroupInputs,
   initializeWizard,
   selectAwsAccountId,
   selectAwsShareMethod,
@@ -87,6 +92,7 @@ import {
   selectAzureResourceGroup,
   selectAzureSubscriptionId,
   selectAzureTenantId,
+  selectBlueprintMode,
   selectDistribution,
   selectGcpAccountType,
   selectGcpEmail,
@@ -95,6 +101,8 @@ import {
   selectImageTypes,
   selectIsImageMode,
   selectTimezone,
+  selectUserGroups,
+  selectUsers,
 } from '../../store/wizardSlice';
 import { getHostArch, getHostDistro } from '../../Utilities/getHostInfo';
 import isRhel from '../../Utilities/isRhel';
@@ -190,6 +198,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   const resolvePath = useAppSelector(selectPathResolver);
   const navigate = useNavigate();
   const dispatch = useAppDispatch();
+  const store = useStore();
   const imageTypes = useAppSelector(selectImageTypes);
   const [searchParams] = useSearchParams();
 
@@ -311,6 +320,31 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
     usersStepAttemptedNext,
   ]);
 
+  // Commit any pending group input and validate the users step from fresh state
+  const commitAndValidateUsersStep = useCallback(() => {
+    dispatch(commitAllPendingGroupInputs());
+    const state = store.getState() as RootState;
+    const users = selectUsers(state);
+    const userGroups = selectUserGroups(state);
+    const blueprintMode = selectBlueprintMode(state);
+    const environments = selectImageTypes(state);
+    const freshUsersValidation = computeUsersValidation(
+      users,
+      userGroups,
+      blueprintMode,
+      environments,
+    );
+    const freshGroupsValidation = computeUserGroupsValidation(userGroups);
+    if (
+      freshUsersValidation.disabledNext ||
+      freshGroupsValidation.disabledNext
+    ) {
+      setUsersStepAttemptedNext(true);
+      return false;
+    }
+    return true;
+  }, [dispatch, store]);
+
   let startIndex = 1; // default index
   const JUMP_TO_REVIEW_STEP = 24;
 
@@ -319,7 +353,6 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   }
 
   const [wasRegisterVisited, setWasRegisterVisited] = useState(false);
-  const [wasUsersVisited, setWasUsersVisited] = useState(false);
   const lastTrackedStepIdRef = useRef<string | undefined>();
 
   useEffect(() => {
@@ -355,8 +388,6 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
         }
       } else if (step.id === 'step-register' && step.isVisited) {
         setWasRegisterVisited(true);
-      } else if (step.id === 'wizard-users' && step.isVisited) {
-        setWasUsersVisited(true);
       }
     }, [step.id, step.isVisited]);
 
@@ -528,25 +559,15 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
             isHidden={!restrictions.users.required}
             navItem={CustomStatusNavItem}
             status={
-              wasUsersVisited
-                ? usersValidation.disabledNext ||
-                  userGroupsValidation.disabledNext
-                  ? 'error'
-                  : 'default'
+              usersStepAttemptedNext &&
+              (usersValidation.disabledNext ||
+                userGroupsValidation.disabledNext)
+                ? 'error'
                 : 'default'
             }
             footer={
               <CustomWizardFooter
-                beforeNext={() => {
-                  if (
-                    usersValidation.disabledNext ||
-                    userGroupsValidation.disabledNext
-                  ) {
-                    setUsersStepAttemptedNext(true);
-                    return false;
-                  }
-                  return true;
-                }}
+                beforeNext={commitAndValidateUsersStep}
                 disableNext={
                   usersValidation.disabledNext ||
                   userGroupsValidation.disabledNext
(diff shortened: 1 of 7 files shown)


regexowl:
Maybe we could add some kind of guidance or a helper text to let people know that they can add a group on Enter keydown? To let people know about things without doing changes they might not want. I don't know, just more thoughts...

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -63,12 +63,17 @@ const LabelInput = ({
   };
 
   const addItem = (value: string) => {
-    if (list?.includes(value) || requiredList?.includes(value)) {
+    const trimmedValue = value.trim();
```
regexowl: Nice! :heart:

regexowl (changes requested):
Thanks for double checking the behaviour with the UX team! :heart:

The on blur and on add a user behaviour looks great. Let's just tweak the Next button behaviour: _"trigger validation when "Next" is clicked and if there are any errors to render them and block the Next button"_

src/Components/CreateImageWizard/steps/Users/components/UserInfo.tsx
```diff
@@ -8,8 +8,38 @@ import { useAppSelector } from '../../../../../store/hooks';
 import { selectUsers } from '../../../../../store/wizardSlice';
 import { useUsersValidation } from '../../../utilities/useValidation';
 
-const UserInfo = () => {
+type UserInfoProps = {
+  usersPristine: boolean;
+  onInputErrorsChange?: ((hasAnyError: boolean) => void) | undefined;
```
regexowl: I'm wondering if the `onInputErrorsChange` could also be removed from the passed props to make things simpler. Looking at the filesystem again, the difference between that validation and users would be that this works with an input directly, meaning outside of state.

Just thinking out loud - would it make sense to create an optional value for group input in the store so it can easily get to the validator? It would need to be handled in the store, but it might mean less `useEffect`s and keeping up with refs. What do you think?

src/Components/CreateImageWizard/steps/Users/utilities/isUserDefined.ts
```diff
@@ -0,0 +1,18 @@
+import { UserWithAdditionalInfo } from '../../../../../store/wizardSlice';
+
+/**
+ * Check if a user is "defined" - has started filling any field.
+ * A user is considered defined if they have any of:
+ * - password
+ * - SSH key
+ * - groups
+ * - current group input (typing in progress)
+ */
```
regexowl: Unnecessary comment, the info is evident from the function itself.

regexowl (changes requested):
Ah, I see. I've started reviewing, but there seems to be older users step than we have currently. It's really weird that the PR doesn't warn about conflicts, but there was a bigger change to the way the users table is rendered.

Before (and what I see in this PR):
image

And after (what is currently on stage after SPUR changes):
image

Could you please rebase so the code is up-to-date with main? I'll continue with the review after the rebase.

regexowl:
Noticed some weird behaviour. When there's more groups the "Add user" and "Next" button don't "work" any more unless the new group is added first. So realistically - the user would have to click the Next or Add button twice to do the action they wanted to do:

[Screencast_20260212_100920.webm](https://github.com/user-attachments/assets/b691514f-67e4-4a34-96d9-9fdebb8176fa)

Can you also please squash the commits so the `.gitignore` changes cancel each other?

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -125,8 +125,12 @@ export const CustomWizardFooter = ({
         <Button
           variant='primary'
           onClick={() => {
+            if (document.activeElement instanceof HTMLElement) {
+              document.activeElement.blur();
+            }
             if (!beforeNext || beforeNext()) goToNextStep();
           }}
+          onMouseDown={(e) => e.preventDefault()}
```
lucasgarfield: `preventDefault` on `mousedown` suppresses native focus management. This breaks keyboard navigation — users tabbing to this button won't get expected focus behavior, and screen readers may not announce it correctly.

This exists to prevent the browser's natural blur from firing before `onClick`, but it's treating a symptom. If input commitment moves to `beforeNext`, this workaround is unnecessary. Same issue on the "Review and finish" button at line 163.

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -527,11 +532,10 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
             isHidden={!restrictions.users.required}
             navItem={CustomStatusNavItem}
             status={
-              wasUsersVisited
-                ? usersValidation.disabledNext ||
-                  userGroupsValidation.disabledNext
-                  ? 'error'
-                  : 'default'
+              usersStepAttemptedNext &&
```
lucasgarfield: This changes when the nav item shows error status:
- **Before:** after *visiting* the step (`wasUsersVisited`)
- **After:** after *attempting to proceed* (`usersStepAttemptedNext`)

This isn't mentioned in the PR description. It also creates an inconsistency — the Registration step still uses `wasRegisterVisited` for the same purpose.

If intentional, call it out in the description and consider aligning Registration to the same pattern.

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -46,33 +50,58 @@ const LabelInput = ({
...
   inlineChips = false,
+  onInputErrorChange,
+  currentInputValue,
+  onInputValueChange,
+  addOnBlur = false,
 }: LabelInputProps) => {
   const dispatch = useAppDispatch();
 
-  const [inputValue, setInputValue] = useState('');
+  const useReduxState =
```
lucasgarfield: Two problems:

1. `useReduxState` — the `use` prefix is reserved for hooks. This is a boolean. `isReduxControlled` or similar.

2. This component now switches between controlled (Redux) and uncontrolled (local state) modes at runtime based on prop presence. That creates two divergent code paths and requires non-null assertions (`currentInputValue!` on line 63, `onInputValueChange!` on lines 85 and 152) to satisfy TypeScript.

Fix with discriminated union types:
```tsx
type ControlledProps = {
  currentInputValue: string;
  onInputValueChange: (value: string) => UnknownAction;
};
type UncontrolledProps = {
  currentInputValue?: never;
  onInputValueChange?: never;
};
type LabelInputProps = (ControlledProps | UncontrolledProps) & { /* shared */ };
```
No runtime checks, no `!` assertions — TypeScript enforces that both are present or both absent.

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -46,33 +50,58 @@ const LabelInput = ({
...
+
   const [onStepInputErrorText, setOnStepInputErrorText] = useState('');
-  let [invalidImports, duplicateImports] = ['', ''];
 
-  if (stepValidation.errors[fieldName]) {
-    [invalidImports, duplicateImports] =
-      stepValidation.errors[fieldName].split('|');
-  }
+  const validationErrorText = stepValidation.errors[fieldName] || '';
+  const validationErrors = React.useMemo(
```
lucasgarfield: This `useMemo` is splitting a short error string — the memoization overhead likely exceeds the computation cost. Compute inline.

Also, line 67 uses `||` for the default value. Prefer `??` — `||` coerces falsy values like `0` and `''`, which can mask bugs in other contexts. Consistency matters.

The pipe-delimited error format itself is fragile (if an error message ever contains `|`, parsing breaks), but that's pre-existing.

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -46,33 +50,58 @@ const LabelInput = ({
...
+    () =>
+      validationErrorText
+        ? validationErrorText.split(/\s*\|\s*/).filter((e) => e.trim())
+        : [],
+    [validationErrorText],
+  );
+
+  useEffect(() => {
```
lucasgarfield: This `useEffect` syncs derived error state to the parent via callback. Two problems:

1. If `onInputErrorChange` isn't memoized with `useCallback`, this effect fires every render. If the parent updates state in response, you get an infinite loop.

2. The parent's error view lags one render behind reality (`useEffect` fires after render).

Better approach: if validation state is in Redux, both parent and child read from Redux directly — no callback, no sync lag. `hasError` is derived, not synchronized.

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -46,33 +50,58 @@ const LabelInput = ({
...
+      dispatch(onInputValueChange!(value));
+    } else {
+      setLocalInputValue(value);
+    }
     setOnStepInputErrorText('');
   };
 
-  const addItem = (value: string) => {
-    if (list?.includes(value) || requiredList?.includes(value)) {
+  const addItem = (value: string): boolean => {
```
lucasgarfield: `addItem` validates, sets error state, dispatches to Redux, calls parent callbacks, and returns a boolean. The return value is unused at every call site.

Separate validation from action:
```tsx
const validateItem = (value: string): string | null => {
  const trimmed = value.trim();
  if (!trimmed) return null;
  if (list?.includes(trimmed)) return `${item} already exists.`;
  if (!validator(trimmed)) return getErrorMessage(fieldName);
  return null;
};

const addItem = (value: string) => {
  const error = validateItem(value);
  if (error) {
    setOnStepInputErrorText(error);
    return;
  }
  dispatch(addAction(value.trim()));
  setOnStepInputErrorText('');
};
```
Validation is now testable in isolation. `addItem` does one thing.

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -130,14 +166,30 @@ const LabelInput = ({
...
   const handleRemoveItem = (e: React.MouseEvent, value: string) => {
     dispatch(removeAction(value));
   };
 
   const errors = [];
   if (onStepInputErrorText) errors.push(onStepInputErrorText);
-  if (invalidImports) errors.push(invalidImports);
-  if (duplicateImports) errors.push(duplicateImports);
+  // Filter out validation errors that are already shown in onStepInputErrorText
+  const filteredValidationErrors = validationErrors.filter((error) => {
```
lucasgarfield: Deduplicating errors via `includes('already exists')` / `includes('Group already exists')` — this will break silently if anyone changes the error wording, and it only handles one specific case.

If both `onStepInputErrorText` and `stepValidation.errors` can produce the same error, the error sources aren't coordinated. Either use error codes for identity, or have a single source of truth so deduplication isn't needed.

lucasgarfield (changes requested):
The bug fix is needed — group values shouldn't be silently dropped on Next. But the approach has problems.

The core issue: this PR coordinates input commitment through DOM events (`document.activeElement.blur()` + `onMouseDown preventDefault`) instead of explicit state management. This creates a race condition where `beforeNext` reads stale validation state, and breaks keyboard accessibility.

**What needs to change:**
- Replace DOM manipulation with explicit input commitment in `beforeNext`
- Remove `onMouseDown preventDefault` (accessibility regression)
- Fix the controlled/uncontrolled mode switching in `LabelInput`
- Remove `useEffect` that syncs derived error state to parent
- Add tests for the new blur behavior

The `??` refactoring and input trimming are fine. Details inline.

regexowl (changes requested):
Ran through the code and added some comments.

Also noticed two possible bugs:
- Invalid groups get silently dropped on "Add user" and "Next"
- Duplicate groups get silently dropped on "Add user" and "Next"

PR #3948, opened 2025-12-23
devDeps: Bump vitest from 3.2.4 to 4.0.16 (HMS-9957)

regexowl:
This bumps vitest from 3.2.4 to 4.0.16

The way constructors are handled changed in between versions 3 and 4, so a small update to `MockResizeObserver` was needed.

JIRA: [HMS-9957](https://issues.redhat.com/browse/HMS-9957)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,11 +7,6 @@
     "sourceMap": true,
     "allowJs": true,
     "moduleResolution": "node",
-    "types": [
-      "node",
-      "vitest/globals",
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "allowJs": true,
     "moduleResolution": "node",
     "types": [
+      "node",
```
regexowl: vitest 4.x doesn't include node globals implicitly

PR #3960, opened 2026-01-06
src: Bump jsdom and update tests where needed

regexowl:
`ElementInternals` got fixed in jsdom v27.0.0, meaning icons, descriptions and similar parts of the elements get properly separated now and are no longer a part of the accessible name.


regexowl:
This PR is a bit of a mess as is and I won't have time to get to it soon. Closing for not and making a note to revisit this.

PR #3965, opened 2026-01-07
Wizard: add description to AAP step

lzap:
We got a call from a customer who did not know when AAP configuration happens. If this is a build time or first boot configuration. This should clarify it.

---

And it is fair, we do not have pretty much anything on that page that explains it.

image

Review carefully, I copy-pasted this from Firstboot step and I have no idea what I am doing. Hopefully, tests will not blow up.


lzap:
Jesus Christ, bots are giving thumbs up and downs themselves. What is happening... :-D

PR #3983, opened 2026-01-13
Wizard: Fix indentation of registration list on review step (HMS-10037)

regexowl:
The indent was smaller, this should fix the issue.

Before:
image

After:
image

JIRA: [HMS-10037](https://issues.redhat.com/browse/HMS-10037)


regexowl:
> * Consider whether applying `pf-v6-u-min-width` only to this `dt` is intentional, or if similar `dt` elements in the review step should be updated for consistent alignment and spacing.

Yeah, yeah, yeah, I can add it to other `.dt`s as well.

PR #3986, opened 2026-01-13
cockpit: Replace blueprint mode radios with a toggle group (HMS-10042)

regexowl:
This updates the blueprint mode component to use a toggle group instead of previously suggested radio buttons.

JIRA: [HMS-10042](https://issues.redhat.com/browse/HMS-10042)


kingsleyzissou (approved):
Looks 👌

PR #3994, opened 2026-01-16
cockpit: Fix on edit crash due to getLatestRelease

regexowl:
This ensures the `getLatestRelease` runs only when there's a distribution included in the request.

How to to reproduce the issue:
1. create a blueprint with image mode selected
2. click on Edit blueprint

Current behaviour:
- the app crashes

After the fix:
- the Edit wizard for the blueprint opens


regexowl:
I don't remember if seeing the bootc ref was supposed to be a part of today's demo... so just to be sure I guess. Although the building wouldn't work as the reference is not in the API yet... so yeah...

PR #3995, opened 2026-01-16
packit: Add RHEL10 x86 & ARM COPR builds

ochosi:
Since we're releasing this into RHEL10, it'd be useful to have the RPMs ready for it for easy testing.


kingsleyzissou (approved):
🚀

PR #3996, opened 2026-01-16
playwright: Add static user to compliance and registration

tkoscieln:
Add static user to tests that require it. Compliance needs it because of the zero state and registration due to activation key.


kingsleyzissou:
> @kingsleyzissou sorry for that, I started using lazygit and I keep forgetting to edit the commit message after squashing 😄 I also changed the static user in registration, it is now only used for the auto reg test instead of all three

Pro tip I learned from @achilleas-k, do a fixup rather than a squash it's the same but doesn't copy the unwanted commit message :)

PR #3998, opened 2026-01-16
Wizard: remove the shadow line in Landing page (HMS-10024)

ksiekl:
The random looking shadow of a section is removed to make UI less confusing. Unfortunately, the shadow is a PF6 feature, so we either have to use CSS styling directly to remove it, or CSS styling directly to not use the PF "sticky top" class, and creating our own sticky class. two solutions, same result :cry: 
Also changed the Images to Image builder in the first step.
ahjb5n


mgold1234:
wooo this one is soooo tricky!!!

mgold1234 (approved):
/lgtm

PR #4011, opened 2026-01-21
Wizard: Remove validation for non-required empty input (HMS-10081)

regexowl:
This updates the pristinity of the validated input to reset when there are no errors (=== value is not required) and the input is empty.

Important thing to check is that there is still an error message when the input is empty, but the value is required.

Before:
pristine-before

After:
pristine-after

JIRA: [HMS-10081](https://issues.redhat.com/browse/HMS-10081)


regexowl:
Updated the value condition as per the AI review.

kingsleyzissou (approved):

PR #4014, opened 2026-01-21
Wizard: Add groups section to Users step (HMS-8784)

mgold1234:
Adds a new "Groups" section to the Users step, allowing users to define groups prior to
assigning users to them. This prevents image build failures caused by missing group definitions.

Future work (to be done in follow-up PRs):
- Add validation in the Users section to check if assigned groups exist in the Groups section
- Add support for on-prem blueprints
- Write tests for the new groups section functionality. Tests are not being written at this time, as they would cause failures in the on-prem environment. This will be supported in a future update.
Screenshot 2026-01-21 at 22 52 33

JIRA: [HMS-8784](https://issues.redhat.com/browse/HMS-8784)

### DIFF
--- a/playwright/Customizations/Users.spec.ts
+++ b/playwright/Customizations/Users.spec.ts
@@ -44,13 +44,13 @@ test('Create a blueprint with Users customization', async ({
   await navigateToLandingPage(page);
   const frame = ibFrame(page);
 
-  await test.step('Navigate to Users step', async () => {
+  await test.step('Navigate to Groups and users step', async () => {
     await fillInImageOutput(frame);
     await registerLater(frame);
   });
 
   await test.step('Create initial valid users', async () => {
-    await frame.getByRole('button', { name: 'Users' }).click();
+    await frame.getByRole('button', { name: 'Groups and users' }).click();
 
     // Create admin user with correct password and wheel group
     await frame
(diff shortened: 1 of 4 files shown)


regexowl (commented):
Looks good, left a few comments :speech_balloon:

src/Components/CreateImageWizard/steps/Users/components/GroupInfo.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { useUserGroupsValidation } from '../../../utilities/useValidation';
+
+type GroupInfoProps = {
+  attemptedNext?: boolean | undefined;
+};
+
+const GroupInfo = ({ attemptedNext = false }: GroupInfoProps) => {
+  const dispatch = useAppDispatch();
+  const groups = useAppSelector(selectUserGroups);
+  const groupsValidation = useUserGroupsValidation();
```
kingsleyzissou: But basically there is a case where someone might create a group and a user with the same name but not assign the group to the user which will cause the build to fail. This will be fixed in osbuild (I believe) but it's not fixed yet for us.

So for example these two screenshots are functionally the same, but one passes validation while the other doesn't:

Image
Image

This should maybe be checked in the `users` validation rather than here, that might be simpler

src/Components/CreateImageWizard/steps/Users/components/GroupRow.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const onRemoveGroup = () => {
+    dispatch(removeUserGroup(index));
+  };
+
+  const handleGroupNameChange = (
+    _e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+    value: string,
+  ) => {
+    if (groupCount === 0) {
```
kingsleyzissou: See the comment below. But this won't be needed if we disable removing the group when there is only one left.

The other concern with this is that it's kind of setting a group name with an empty string as a name in the state.

kingsleyzissou (approved):
Thanks!

PR #4024, opened 2026-01-23
cockpit: Show registration for image mode (HMS-10093)

regexowl:
Since distribution can be undefined, it doesn't pass the `isRhel` check and the step with its review was incorrectly hidden when in image mode.

JIRA: [HMS-10093](https://issues.redhat.com/browse/HMS-10093)


regexowl:
https://github.com/osbuild/image-builder-frontend/pull/4044 changes the logic of filtering steps. If we ever need this, we should implement it in an up-to-date way. Closing this one.

PR #4047, opened 2026-01-28
Wizard: changing the AWS card for a checkbox in cockpit (HMS-10107)

ksiekl:
The card before didn't have a logo and wasn't formatted properly. Now it is a simple checkbox like the private clouds options.

image

These are just my 2 cents of how it could look - any notes on it are welcome :)
There's also the option to add the logo to some public folder and drag it out from there I guess, but I didn't like that idea.... but now cockpit doesn't match hosted service, which is also not ideal.


ksiekl:
Ah :woman_facepalming: for sure, I can add them as well in this PR, that's a quick little change

ksiekl:
@regexowl added them, there are no tests for them in the cockpit right now, but I would leave that as as a work for additional PR

PR #4053, opened 2026-01-30
tests: fix react router warning message (HMS-10131)

kingsleyzissou:
React router is switching an internal mechanism for handling router state updates, we don't use `React.Lazy`, so we should be fine. See: See: https://reactrouter.com/6.30.3/upgrading/future#v7_starttransition

JIRA: [HMS-10131](https://issues.redhat.com/browse/HMS-10131)


regexowl:
I understand it the same way, is there a mechanism that will help us to not forget or should we create an issue or something like that?

PR #4057, opened 2026-02-01
Wizard: Improve groups input in User step (HMS-10134)

mgold1234:
This commit prevent awkward textbox movement when adding or removing groups.

- Render group chips inside the input field instead of below it
- Make group chips compact to reduce visual clutter.
- Limit the number of visible chips and add proper overflow handling using:
“X more” / “Show less” toggling.

https://github.com/user-attachments/assets/b7cce09d-20b9-443c-addf-6e77524bc90f

JIRA: [HMS-10134](https://issues.redhat.com/browse/HMS-10134)


regexowl (changes requested):
Two nitpicks, otherwise looks good to me :)

PR #4068, opened 2026-02-04
Wizard: Add groups section to review step (HMS-8784)

mgold1234:
Display user groups in the review step alongside users. The groups section appears links back to the users step via "Revisit step" button.

Screenshot 2026-02-04 at 12 17 17
JIRA: https://issues.redhat.com/browse/HMS-8784

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
@@ -20,6 +20,7 @@ import {
   FirewallList,
   FirstBootList,
   FSCList,
+  GroupsList,
   HostnameList,
   ImageOutputList,
   KernelList,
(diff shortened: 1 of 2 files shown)


src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
```diff
@@ -381,6 +387,23 @@ const Review = () => {
           <UsersList />
         </ExpandableSection>
       )}
+      {userGroups.length > 0 && (
```
regexowl: See https://github.com/osbuild/image-builder-frontend/pull/4068/changes#r2763437972

```suggestion
      {!restrictions.users.shouldHide && userGroups.length > 0 && (
```

We can implement a separate check if we ever need to separate groups and users to individual steps in the future.

PR #4072, opened 2026-02-04
Wizard: Add groups to user review, arrange in tables (HMS-10147)

regexowl:
This adds "Groups" fields to the Users review list, so the groups added to the user get properly rendered.

For better overview of the users and groups the previously used description lists were replaced with a table as per revamp mocks.

JIRA: [HMS-10147](https://issues.redhat.com/browse/HMS-10147)


regexowl:
> @regexowl I am not sure I like this line

Fair enough! I've checked our revamp mocks and we actually already have a pattern there. Replaced the stream of description lists with a table, just need to check if any other tests need to be udated.

PR #4073, opened 2026-02-04
Wizard: Add validation to warn about undefined group (HMS-8784)

mgold1234:
Warn users when they assign a user to a group that is neither a known system group nor defined in the Groups section.
This helps users avoid unexpected behavior, as undefined groups may not exist at build time.

Note: This is a warning, not an error, as the build system handles group conflicts automatically.

Screenshot 2026-02-05 at 11 25 10

JIRA: HMS-8784

### DIFF
--- a/playwright/Customizations/UsersAndGroups.spec.ts
+++ b/playwright/Customizations/UsersAndGroups.spec.ts
@@ -96,6 +96,14 @@ test('Create a blueprint with Users customization', async ({
       name: 'Administrator',
     });
     await expect(adminCheckboxes.nth(1)).not.toBeChecked();
+
+    // The 'developers' group is not defined in the Groups section and is not
+    // a system group, so a warning should appear
+    await expect(
+      frame.getByText(/User assigned to undefined group\(s\): developers/),
+    ).toBeVisible();
+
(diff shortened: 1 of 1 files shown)


regexowl:
Yup, it's a bit more complex. I'd say this PR makes sense as is. We're unable to see a list of groups created with installation of packages without depsolving. Which leaves two possible paths:
1. required group isn't defined -> user doesn't know -> build failure
2. required group isn't defined -> user sees warning -> adds group -> build passes

There was previously an issue while duplicating groups, but I believe @achilleas-k already resolved that in backend (can't find the comment though, so Achilleas please feel free to correct me :sweat_smile:)

Also with the undefined groups being non-blocking warning... I think this should be working fine with our current constraints.

mgold1234:
@regexowl as I understand from @kingsleyzissou -
`Ya we updated the docs recently too. So duplicate groups shouldn't be an issue . Having said that, we might not yet have the newer version where this is the case though.`
I think it is better to warn them about the missing groups instead of being silent.
and as @smckinno said we will discuss about the other complicated things in the next UX meeting

regexowl (approved):
Nice catch!

PR #4076, opened 2026-02-05
Wizard: Fix wizard height (HMS-10158)

regexowl:
The wizard started to shrink between steps again. This seems to be helping?

JIRA: [HMS-10158](https://issues.redhat.com/browse/HMS-10158)


regexowl:
Ok, rescoped the css. This seems to be working the best - the navigation is scrollable, but the body of Wizard isn't unless the step is higher than the visible height. Using just `80vh` leaves space for the header.

PR #4078, opened 2026-02-06
spec: update nodejs BuildRequires

yselkowitz:
As of Fedora 44, the packaging scheme for Node.js is changing. This is compatible with both the old and new schemes.

https://pagure.io/packaging-committee/pull-request/1521
https://fedoraproject.org/wiki/Changes/NodejsAlternativesSystem


croissanne (approved):
thanks!

PR #4080, opened 2026-02-06
Wizard: Add conditional search instructions for Additional Packages step (HMS-10161)

mgold1234:
Show different search instructions based on environment (on-premise vs hosted) to guide users on the available search capabilities.
Screenshot 2026-02-06 at 14 46 13

JIRA: [HMS-10161](https://issues.redhat.com/browse/HMS-10161)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/index.tsx
@@ -5,12 +5,14 @@ import { Content, Form, Title } from '@patternfly/react-core';
 import PackageRecommendations from './PackageRecommendations';
 import Packages from './Packages';
 
+import { selectIsOnPremise } from '../../../../store/envSlice';
 import { useAppSelector } from '../../../../store/hooks';
 import { selectDistribution } from '../../../../store/wizardSlice';
 import isRhel from '../../../../Utilities/isRhel';
 
 const PackagesStep = () => {
   const distribution = useAppSelector(selectDistribution);
+  const isOnPremise = useAppSelector(selectIsOnPremise);
   return (
     <Form>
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/steps/Packages/index.tsx
```diff
@@ -19,6 +21,20 @@ const PackagesStep = () => {
       <Content>
         Blueprints created with Images include all required packages.
       </Content>
+      <Content>
+        {isOnPremise ? (
+          <>
+            Search by specifying the whole package name, or glob using asterisk
+            wildcards (*) before or after the package name.
```
regexowl: The 'exact match' wording makes sense here, because we're telling the user to either use an exact match, or search on partial match using a wildcard.

```suggestion
          Search for exact matches by specifying the whole package name, or glob
          using asterisk wildcards (*) before or after the package name.
```

src/test/Components/CreateImageWizard/steps/Packages/Packages.test.tsx
```diff
@@ -232,10 +234,56 @@ describe('Step Packages', () => {
       name: /Search packages/i,
     });
     await screen.findByText(
-      /Search for exact matches by specifying the package name to add additional packages to your image/i,
+      /Search for additional packages to add to your image/i,
     );
   });
 
+  test('shows on-premise wildcard search instructions when running on-premise', async () => {
```
regexowl: It would make sense to check for the service part of the description in this file (the paragraph about package groups). That could be just one line when the step renders.

On-prem is tested in the Playwright tests, so the on-prem text should be checked there.

PR #4084, opened 2026-02-08
Wizard: Hide empty user groups from review step (HMS-10164)

mgold1234:
# Summary

Hide empty user groups from review step display

## Overview

Previously, the review step would display an empty Groups table with "None None" values even when no groups were configured. This created visual clutter and confused users about whether groups were actually configured. This PR filters out empty groups from the review display while preserving the initial empty group in the form, maintaining a smooth user experience when adding groups.

## Architectural Changes

None

## Key Changes

- Updated `ReviewStep.tsx` to filter empty groups in the display condition (`userGroups.filter((g) => g.name).length > 0`)
- Modified `GroupsList` component to filter out groups without names before rendering
- Removed unnecessary ternary operators in `GroupsList` since filtered groups are guaranteed to have names
- Preserved the initial empty group in the wizard state to allow immediate group entry without clicking "Add group"

## Breaking Changes

This PR is fully backward compatible.

## Testing

Manually tested the review step with:
- No groups configured (section is hidden)
- One or more valid groups (section displays correctly)
- Mix of empty and filled groups (only filled groups appear)
- Form interaction remains unchanged (initial empty group field available)

JIRA: [HMS-10164](https://issues.redhat.com/browse/HMS-10164)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/ReviewStep.tsx
@@ -61,6 +61,7 @@ import {
   selectTimezone,
   selectUserGroups,
   selectUsers,
+  UserGroup,
 } from '../../../../store/wizardSlice';
 import SecurityInformation from '../Oscap/components/SecurityInformation';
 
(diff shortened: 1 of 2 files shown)


regexowl (approved):
Looks great, thanks!

PR #4096, opened 2026-02-10
Wizard: Add registration to customization restrictions system (HMS-10177)

mgold1234:
this commit add 'registration' as a customization type to enable automatic hiding
of the Registration wizard step for image types that don't support it.

Background:
The wizard has a mechanism to hide unsupported customizations based on
selected image types. 'registration' was not included in this
system, causing the Registration step to be shown for image types that
don't support it (e.g., network-installer).

JIRA:  (HMS-10177)

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -562,9 +562,7 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
                 name='Register'
                 id='step-register'
                 key='step-register'
-                isHidden={
-                  !isRhel(distribution) || restrictions.registration.shouldHide
-                }
+                isHidden={restrictions.registration.shouldHide}
                 navItem={CustomStatusNavItem}
                 status={
                   wasRegisterVisited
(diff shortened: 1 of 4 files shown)


src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
```diff
@@ -222,6 +229,18 @@ const TargetEnvironment = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  useEffect(() => {
+    if (restrictions.registration.shouldHide) {
```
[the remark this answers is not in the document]
mgold1234: I created small hook that will be responsible to that

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -249,6 +249,8 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
   const distribution = useAppSelector(selectDistribution);
   const timezone = useAppSelector(selectTimezone);
 
+  const isRegistrationAvailable = !restrictions.registration.shouldHide;
```
kingsleyzissou: Small nitpick: we could probably just move this in-line to be consistent with the rest.
But this shouldn't block the PR

src/test/store/distributions/hooks.test.tsx
```diff
@@ -118,6 +121,41 @@ describe('useCustomizationRestrictions hook logic', () => {
     });
   });
 
+  describe('non-RHEL distribution restrictions', () => {
```
kingsleyzissou: Thanks for adding tests too!

PR #4104, opened 2026-02-11
Cockpit/ImagesTable: Populate Image Mode OS column

ochosi:
Currently, the `OS` column is empty for image mode. This commit changes that and extracts the major and minor version from the bootc ref.
That not only fixes the ImagesTable but also fixes our "Launch" capability via Cockpit Machines, which relies on information from the `OS` column.


regexowl:
> > @kingsleyzissou Not sure if @ochosi is online this week. We can't do the squash ourselves, can we?
> 
> Was just thinking this now while making coffee. I have this branch checked out now, let me do that and then you can review

Thanks! The code looked good to me, didn't test how it runs, but since you did I think we're good :heavy_check_mark:

PR #4111, opened 2026-02-13
Wizard: Hide "Other" label when it's the only category (HMS-10198)

regexowl:
This hides the "Other" form group label if the only supported environments are from that category.

image

JIRA: [HMS-10198](https://issues.redhat.com/browse/HMS-10198)


achilleas-k (approved):
Yay!

PR #4112, opened 2026-02-13
Add customization support labels to wizard steps (HMS-10201)

kingsleyzissou:
## Summary
Add customization support labels to wizard steps that indicate whether each selected image type supports the current customization, helping users understand compatibility before proceeding.
 
### Changes
- Refactor `computeRestrictions` hook to extract reusable `isCustomizationSupported` helper function
- Add `simpleTargetNames` constant for display-friendly image type names
- Create `useImageTypeCustomizationSupport` hook to check customization support per image type
- Add `CustomizationLabels` shared component that renders support status as PatternFly labels
- Integrate customization labels into all wizard steps (AAP, FileSystem, Firewall, FirstBoot, Hostname, Kernel, Locale, Oscap, Packages, Registration, Repositories, Services, Snapshot, Timezone, Users)

JIRA: [HMS-10201](https://issues.redhat.com/browse/HMS-10201)

### DIFF
--- a/playwright/Customizations/OpenSCAP.spec.ts
+++ b/playwright/Customizations/OpenSCAP.spec.ts
@@ -35,33 +35,20 @@ test('Create a blueprint with OpenSCAP customization', async ({
   await navigateToLandingPage(page);
   const frame = ibFrame(page);
 
-  await test.step('WSL only disables selector', async () => {
+  await test.step('WSL + Installer shows WSL is not supported', async () => {
+    // Back to Image output and add Installer
     await frame.getByRole('button', { name: 'Create image blueprint' }).click();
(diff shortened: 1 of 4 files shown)


src/Components/CreateImageWizard/steps/Users/index.tsx
```diff
@@ -17,6 +18,7 @@ const UsersStep = ({ attemptedNext }: UsersStepProps) => {
 
   return (
     <Form>
+      <CustomizationLabels customization='users' />
```
regexowl: I wonder if we could connect the "customization" prop to the step name. Might be worth to revisit this during the revamp. Absolutely non-blocking now

PR #4113, opened 2026-02-13
Wizard: hide openscap for network-installer (HMS-10200)

kingsleyzissou:
## Summary
Hides OpenSCAP compliance options in the wizard when the network installer image type is selected, as OpenSCAP is not supported for this image type.

### Changes
- Add `useCustomizationRestrictions` hook to check image type restrictions
- Conditionally hide compliance policy and OpenSCAP profile selectors based on restrictions

JIRA: [HMS-10200](https://issues.redhat.com/browse/HMS-10200)


regexowl:
Nice! Is this to allow FIPS in network-installer? The `isHidden` prop for the step will need to be updated so Security renders when the network-installer target is selected.

PR #4122, opened 2026-02-17
ImagesTable: change permission of artifact dir on launch

croissanne:
While cockpit-files does not need the artifact directory to be world-readable, cockpit-machines does need it when creating a VM.


regexowl (approved):
🚀

PR #4225, opened 2026-03-23
Wizard: commit pending LabelInput value on blur

### LINKED ISSUE #3943: [BUG] - Group field value not added on blur, Add user, or Next actions (HMS-9944)

mgold1234:
Previously, LabelInput only committed typed values when the user pressed Enter. This meant pending input was silently discarded when clicking away, navigating to the next wizard step, or interacting with other controls. Adding an onBlur handler ensures the pending value is validated and added automatically when the input loses focus, providing consistent behavior across all LabelInput consumers (kernel arguments, user groups, NTP servers, ports, firewall and systemd services).

Fixes: https://github.com/osbuild/image-builder-frontend/issues/3943

### DIFF
--- a/src/Components/CreateImageWizard/LabelInput.tsx
+++ b/src/Components/CreateImageWizard/LabelInput.tsx
@@ -126,10 +126,17 @@ const LabelInput = ({
     setOnStepInputErrorText('');
   };
 
+  const commitInputValue = (value: string) => {
+    const trimmed = value.trim();
+    if (trimmed) {
+      addItem(trimmed);
+    }
+  };
+
   const handleKeyDown = (e: React.KeyboardEvent, value: string) => {
     if (e.key === 'Enter') {
       e.preventDefault();
-      addItem(value);
+      commitInputValue(value);
     }
   };
 
@@ -154,6 +161,7 @@ const LabelInput = ({
           onChange={onTextInputChange}
           value={inputValue}
           onKeyDown={(e) => handleKeyDown(e, inputValue)}
+          onBlur={() => commitInputValue(inputValue)}
         >
           {totalItems > 0 && (
             <LabelGroup


kingsleyzissou (commented):
I've tried this out and some of the inputs don't work now. For example pressing enter on `groups`, `ntp servers`, firewall `ports`.

Having said that, I'm not 100% this is something we want. Like this might mean that some half typed input might get added and that might not be desired.

Like I'm thinking of a case were a user is trying to figure out the format of the firewall options maybe tabs out and a half-inputted value gets submitted and validated.

mgold1234:
as I understand @lucasgarfield as an idea how to solve it and he will push commit to that pr.
so leave it open for now

regexowl:
Discussed at the UX/dev sync, we'll close this for now and will revisit after few weeks when we decide what to do with the pending values.

PR #4234, opened 2026-03-25
Wizard: Revamp first boot step ( HMS-9522)

mgold1234:
Add a FileUpload component above the CodeEditor for drag-and-drop
script uploads. Enhance the CodeEditor with custom controls for
copy, download, and revert actions.

- Add FileUpload with drag-and-drop support above CodeEditor
- Add custom CodeEditor controls (copy, download, revert)
- Add revert functionality to restore initial or uploaded script
- Add helper text clarifying supported script types
- Remove registration ordering text from step description
- Move inline styles to SCSS file
- Add unit tests for new functionality
JIRA: [HMS-9522](https://redhat.atlassian.net/browse/HMS-9522)

[HMS-9522]: https://redhat.atlassian.net/browse/HMS-9522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
+++ b/src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
@@ -1,9 +1,15 @@
-import React from 'react';
+import React, { useRef, useState } from 'react';
 
-import { CodeEditor, Language } from '@patternfly/react-code-editor';
+import {
+  CodeEditor,
+  CodeEditorControl,
+  Language,
+} from '@patternfly/react-code-editor';
 import {
   Alert,
   Content,
+  DropEvent,
+  FileUpload,
   Form,
   FormGroup,
   FormHelperText,
@@ -11,6 +17,7 @@ import {
   HelperTextItem,
   Title,
 } from '@patternfly/react-core';
+import { UndoIcon } from '@patternfly/react-icons';
 
 import { useFirstBootValidation } from '@/Components/CreateImageWizard/utilities/useValidation';
 import { CustomizationLabels } from '@/Components/sharedComponents/CustomizationLabels';
@@ -25,11 +32,24 @@ import {
 } from '@/store/slices/wizard';
 import { useFlag } from '@/Utilities/useGetEnvironment';
 
+// Inline <style> needed because sassPrefix wraps SCSS in .imageBuilder,
+// but the wizard Modal renders in a portal outside that wrapper.
+const editorStyles = `
+.first-boot-editor .pf-v6-c-code-editor__header-content {
+  background-color: var(--pf-t--global--background--color--primary--default);
+}
+.first-boot-editor .pf-v6-c-code-editor__controls {
+  flex: 1;
+  justify-content: flex-end;
+}
+.first-boot-editor button[aria-label='Upload code'] {
+  display: none;
+}`;
+
 const detectScriptType = (scriptString: string): Language => {
   const lines = scriptString.split('\n');
 
   if (lines[0].startsWith('#!')) {
-    // Extract the path from the shebang
     const path = lines[0].slice(2);
 
     if (path.includes('bin/bash') || path.includes('bin/sh')) {
@@ -44,7 +64,6 @@ const detectScriptType = (scriptString: string): Language => {
       return Language.yaml;
     }
   }
-  // default
   return Language.shell;
 };
 
@@ -57,6 +76,59 @@ const FirstBootStep = () => {
   const isWizardRevampEnabled = useFlag('image-builder.wizard-revamp.enabled');
 
   const Wrapper = isWizardRevampEnabled ? React.Fragment : Form;
+
+  const initialScriptRef = useRef(selectedScript);
+  const [filename, setFilename] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [uploadedScript, setUploadedScript] = useState('');
+
+  const handleScriptChange = (newScript: string) => {
+    if (!selectedScript && !!newScript) {
+      dispatch(addEnabledService(FIRST_BOOT_SERVICE));
+    } else if (!!selectedScript && !newScript) {
+      dispatch(removeEnabledService(FIRST_BOOT_SERVICE));
+    }
+    // In case the user is on windows
+    dispatch(setFirstBootScript(newScript.replace(/\r\n/g, '\n')));
+  };
+
+  const handleFileInputChange = (_event: DropEvent, file: File) => {
+    setFilename(file.name);
+  };
+
+  const handleDataChange = (_event: DropEvent, value: string) => {
+    const normalizedValue = value.replace(/\r\n/g, '\n');
+    setUploadedScript(normalizedValue);
+    handleScriptChange(value);
+  };
+
+  const handleClear = () => {
+    setFilename('');
+    setUploadedScript('');
+    handleScriptChange('');
+  };
+
+  const handleFileReadStarted = () => {
+    setIsLoading(true);
+  };
+
+  const handleFileReadFinished = () => {
+    setIsLoading(false);
+  };
+
+  const handleRevert = () => {
+    handleScriptChange(uploadedScript || initialScriptRef.current);
+  };
+
+  const customControls = [
+    <CodeEditorControl
+      key='revert-button'
+      icon={<UndoIcon />}
+      aria-label='Revert changes'
+      onClick={handleRevert}
+      tooltipProps={{ content: 'Revert changes' }}
+    />,
+  ];
 
   return (
     <Wrapper>
@@ -91,26 +163,43 @@ const FirstBootStep = () => {
         </Content>
       </Alert>
       <FormGroup>
+        <FileUpload
+          id='first-boot-script-upload'
+          type='text'
+          filename={filename}
+          filenamePlaceholder='Drag and drop a file or upload'
+          onFileInputChange={handleFileInputChange}
+          onDataChange={handleDataChange}
+          onReadStarted={handleFileReadStarted}
+          onReadFinished={handleFileReadFinished}
+          onClearClick={handleClear}
+          isLoading={isLoading}
+          allowEditingUploadedText={false}
+          browseButtonText='Upload'
+          hideDefaultPreview
+        />
+      </FormGroup>
+      <FormGroup>
         <CodeEditor
-          isUploadEnabled
-          isDownloadEnabled
-          isCopyEnabled
-          isLanguageLabelVisible
+          className='first-boot-editor'
           language={language}
-          onCodeChange={(code) => {
-            if (!selectedScript && !!code) {
-              dispatch(addEnabledService(FIRST_BOOT_SERVICE));
-            } else if (!!selectedScript && !code) {
-              dispatch(removeEnabledService(FIRST_BOOT_SERVICE));
-            }
-            // In case the user is on windows
-            dispatch(setFirstBootScript(code.replace('\r\n', '\n')));
-          }}
+          onCodeChange={(code) => handleScriptChange(code)}
           code={selectedScript}
           height='35vh'
-          emptyStateButton='Browse'
+          isCopyEnabled
+          isDownloadEnabled
+          isUploadEnabled
+          emptyStateBody='Drag a file here, upload files, or start from scratch.'
+          emptyStateButton='Upload files'
           emptyStateLink='Start from scratch'
+          customControls={customControls}
         />
+        <style>{editorStyles}</style>
+        <HelperText className='pf-v6-u-pt-sm'>
+          <HelperTextItem>
+            Supports bash shell, python, or Ansible playbooks
+          </HelperTextItem>
+        </HelperText>
         {errors.script && (
           <FormHelperText>
             <HelperText>
--- a/src/Components/CreateImageWizard/steps/FirstBoot/tests/FirstBoot.test.tsx
+++ b/src/Components/CreateImageWizard/steps/FirstBoot/tests/FirstBoot.test.tsx
@@ -1,14 +1,17 @@
 import { screen } from '@testing-library/react';
 
-import { initialState } from '@/store/slices/wizard';
+import { FIRST_BOOT_SERVICE } from '@/constants';
 import { createUser, waitForAction } from '@/test/testUtils';
 
-import { openCodeEditor, renderFirstBootStep, uploadScript } from './helpers';
+import { renderFirstBootStep, uploadScript } from './helpers';
 
 const VALID_SCRIPT = `#!/bin/bash
 echo "Hello, World!"`;
 
 const SCRIPT_WITHOUT_SHEBANG = `echo "Hello, World!"`;
+
+const CRLF_SCRIPT = '#!/bin/bash\r\necho "line1"\r\necho "line2"';
+const LF_SCRIPT = '#!/bin/bash\necho "line1"\necho "line2"';
 
 describe('FirstBoot Component', () => {
   describe('Rendering', () => {
@@ -35,11 +38,13 @@ describe('FirstBoot Component', () => {
       ).toBeInTheDocument();
     });
 
-    test('displays start from scratch button initially', async () => {
+    test('displays helper text for supported script types', async () => {
       renderFirstBootStep();
 
       expect(
-        await screen.findByRole('button', { name: /Start from scratch/i }),
+        await screen.findByText(
+          /Supports bash shell, python, or Ansible playbooks/i,
+        ),
       ).toBeInTheDocument();
     });
   });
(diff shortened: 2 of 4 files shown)


src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -88,27 +172,43 @@ const FirstBootStep = () => {
...
           code={selectedScript}
           height='35vh'
-          emptyStateButton='Browse'
+          isUploadEnabled
+          emptyStateBody='Drag a file here, upload files, or start from scratch.'
+          emptyStateButton='Upload files'
           emptyStateLink='Start from scratch'
+          customControls={customControls}
         />
+        <style>{`.first-boot-editor .pf-v6-c-code-editor__controls { flex: 1; } .first-boot-editor .pf-v6-c-code-editor__header { background-color: #fff !important; } .first-boot-editor .pf-v6-c-code-editor__main { background-color: #fff !important; } .first-boot-editor button[aria-label="Upload code"] { display: none; }`}</style>
```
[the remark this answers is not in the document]
mgold1234: The inline <style> tag is intentional here. This project's webpack config uses sassPrefix: '.imageBuilder' which wraps all SCSS rules from src/ under .imageBuilder { ... }. This changes the selector specificity and prevents the styles from correctly overriding PatternFly's default styling (grey background, button positioning). The inline <style> bypasses webpack's SCSS processing and injects directly into the DOM, which is the only reliable way to override these PatternFly styles in this setup.

src/Components/CreateImageWizard/steps/FirstBoot/tests/FirstBoot.test.tsx
```diff
@@ -84,39 +96,39 @@ describe('FirstBoot Component', () => {
...
     });
 
-    test('does not show registration message when using register-later', async () => {
-      renderFirstBootStep({
-        registration: {
-          ...initialState.registration,
-          registrationType: 'register-later',
-        },
-      });
+    test('displays upload button', async () => {
```
regexowl, line 107: This duplicates a previous 'displays upload button' test, doesn't it?

src/Components/CreateImageWizard/steps/FirstBoot/tests/FirstBoot.test.tsx
```diff
@@ -35,11 +38,22 @@ describe('FirstBoot Component', () => {
...
     });
 
-    test('displays start from scratch button initially', async () => {
+    test('displays upload button', async () => {
+      renderFirstBootStep();
+
+      const uploadButtons = await screen.findAllByRole('button', {
+        name: /Upload/i,
+      });
+      expect(uploadButtons.length).toBeGreaterThanOrEqual(1);
```
regexowl: The assertion is a bit loose to my liking. We should check for a specific number of the 'Upload' buttons to make sure a button is there, but there are no unexpected buttons rendered.

src/Components/CreateImageWizard/steps/FirstBoot/tests/FirstBoot.test.tsx
```diff
@@ -84,39 +96,39 @@ describe('FirstBoot Component', () => {
...
 
-  describe('Registration Message', () => {
-    test('shows registration message when not using register-later', async () => {
-      renderFirstBootStep({
-        registration: {
-          ...initialState.registration,
-          registrationType: 'register-now-insights',
-        },
-      });
+  describe('FileUpload', () => {
```
regexowl:
```suggestion
  describe('File Upload Component', () => {
```

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
regexowl, line 1: The file upload component doesn't work well in dark mode, the background should be probably transparent or smth like that.

Image

regexowl: The custom controls are not visible.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -88,27 +172,43 @@ const FirstBootStep = () => {
...
           code={selectedScript}
           height='35vh'
-          emptyStateButton='Browse'
+          isUploadEnabled
+          emptyStateBody='Drag a file here, upload files, or start from scratch.'
+          emptyStateButton='Upload files'
           emptyStateLink='Start from scratch'
+          customControls={customControls}
         />
+        <style>{`.first-boot-editor.pf-v6-c-code-editor { background-color: #fff !important; } .first-boot-editor .pf-v6-c-code-editor__controls { flex: 1; } .first-boot-editor .pf-v6-c-code-editor__header { background-color: #fff !important; } .first-boot-editor .pf-v6-c-code-editor__header-content { background-color: #fff !important; } .first-boot-editor .pf-v6-c-code-editor__tab { background-color: #fff !important; } .first-boot-editor .pf-v6-c-code-editor__main { background-color: #fff !important; } .first-boot-editor button[aria-label="Upload code"] { display: none; }`}</style>
```
regexowl: The styling is not very readable here. Could you please try to move it into a `.scss` file?

regexowl: The "drag and drop" toolbar should be connected to the file upload component with one border if possible.

Image

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -88,27 +172,43 @@ const FirstBootStep = () => {
...
-              dispatch(removeEnabledService(FIRST_BOOT_SERVICE));
-            }
-            // In case the user is on windows
-            dispatch(setFirstBootScript(code.replace('\r\n', '\n')));
-          }}
+          onCodeChange={(code) => handleScriptChange(code)}
           code={selectedScript}
           height='35vh'
-          emptyStateButton='Browse'
+          isUploadEnabled
```
regexowl, line 191: The upload button is enabled here, but then hidden in the styling with `button[aria-label="Upload code"] { display: none; }` which doesn't make much sense.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -44,20 +49,102 @@ const detectScriptType = (scriptString: string): Language => {
...
+        onClick={handleCopy}
+        tooltipProps={{ content: 'Copy to clipboard' }}
+      />
+      <CodeEditorControl
+        key='download-button'
+        icon={<DownloadIcon />}
+        aria-label='Download'
+        onClick={handleDownload}
+        tooltipProps={{ content: 'Download' }}
+      />
```
regexowl: Both copy and download buttons are supported directly by the `CodeEditor` component via props `isCopyEnabled` and `isDownloadEnabled`, meaning we could enable them there, then remove them from the custom control together with their functions and we could use their functionality directly as implemented by Patternfly.

The only custom control would be the revert, which currently works great, so no issues with that.

What would need to be done is to override the position of the buttons in the styling, but since we're doing a lot of overrides anyway it doesn't sound like a problem to add one more. Something like `justify-content: flex-end` for the controls.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
regexowl, line 191: `isCopyEnabled` and `isDownloadEnabled` can be added here and removed from the custom controls.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -44,20 +49,102 @@ const detectScriptType = (scriptString: string): Language => {
...
+
+  const handleDownload = (code: string) => {
+    const blob = new Blob([code], { type: 'text/plain' });
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement('a');
+    a.href = url;
+    a.download = filename || 'first-boot-script.sh';
+    a.click();
+    URL.revokeObjectURL(url);
+  };
```
regexowl: This would be handled by the default PatternFly controls functionality.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -70,9 +157,6 @@ const FirstBootStep = () => {
       <Content component={isWizardRevampEnabled ? 'small' : 'p'}>
         Add a custom script to be executed when the image boots for the first
         time.
-        {registrationType !== 'register-later' && (
-          <> The first boot script will run after registration is done.</>
-        )}
```
regexowl: I think this should stay even though it isn't visible in the mocks.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -88,27 +172,43 @@ const FirstBootStep = () => {
           privacy.
         </Content>
       </Alert>
+      <FormGroup>
+        <FileUpload
+          id='first-boot-script-upload'
+          type='text'
+          value=''
```
regexowl: Why is the value there if it's left empty?

regexowl (changes requested):
Thank you for working on this. Did first pass and added some comments.

src/Components/CreateImageWizard/steps/FirstBoot/FirstBoot.scss
```diff
@@ -0,0 +1,15 @@
+.first-boot-editor.pf-v6-c-code-editor {
+  background-color: transparent !important;
```
regexowl: The styling override shouldn't be needed.

src/Components/CreateImageWizard/steps/FirstBoot/FirstBoot.scss
```diff
@@ -0,0 +1,15 @@
...
+    flex: 1;
+    justify-content: flex-end;
+  }
+
+  .pf-v6-c-code-editor__header,
+  .pf-v6-c-code-editor__header-content,
+  .pf-v6-c-code-editor__tab,
+  .pf-v6-c-code-editor__main {
+    background-color: transparent !important;
+  }
```
regexowl: The styling override shouldn't be needed.

src/Components/CreateImageWizard/steps/FirstBoot/FirstBoot.scss
```diff
@@ -0,0 +1,15 @@
+.first-boot-editor.pf-v6-c-code-editor {
+  background-color: transparent !important;
+
+  .pf-v6-c-code-editor__controls {
+    flex: 1;
+    justify-content: flex-end;
+  }
```
regexowl: This doesn't seem to be aligning the controls to the right.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -88,27 +148,42 @@ const FirstBootStep = () => {
...
-          }}
+          onCodeChange={(code) => handleScriptChange(code)}
           code={selectedScript}
           height='35vh'
-          emptyStateButton='Browse'
+          isCopyEnabled
+          isDownloadEnabled
+          emptyStateBody='Drag a file here, upload files, or start from scratch.'
+          emptyStateButton='Upload files'
           emptyStateLink='Start from scratch'
```
regexowl, line 194: When I open a fresh Wizard modal, the empty state is not there now. The tests seem to be failing on it as well. It goes directly to this:

Image

regexowl (changes requested):
Thanks for the changes! Looks much better already :)

Added a few more comments, but otherwise looks really nice.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
mgold1234: @regexowl Moved the styles into a formatted const editorStyles template literal. Can't use a .scss file because the project's sassPrefix: '.imageBuilder' wraps all SCSS rules in .imageBuilder { }, but the wizard Modal renders in a portal outside that wrapper, so the styles wouldn't apply.

mgold1234: The FileUpload and CodeEditor are separate PatternFly components with their own border styling. Connecting them into a single visual block would require overriding PF's internal border styles on both

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
mgold1234: I tried using isCopyEnabled/isDownloadEnabled, but PF hides the built-in copy/download buttons when the editor is in empty state (isUploadEnabled + no code). Since we need isUploadEnabled for the empty state to render, custom controls are the only way to keep copy/download always visible.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
mgold1234: isUploadEnabled is required for the CodeEditor's empty state to render — without it, PF skips the empty state entirely. The built-in upload button is hidden because we have our own FileUpload component above.

mgold1234: same

src/Components/CreateImageWizard/steps/FirstBoot/FirstBoot.scss
mgold1234: This override is needed for dark mode , without it the CodeEditor has an opaque background that doesn't adapt to the theme.

src/Components/CreateImageWizard/steps/FirstBoot/FirstBoot.scss
mgold1234: same

src/Components/CreateImageWizard/steps/FirstBoot/FirstBoot.scss
mgold1234: can you please check again? the controls are now aligning to the right correctly.

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
regexowl: I'm fine with the icons being hidden in an empty state in favour of not having to implement custom handling functions. Let's keep the isCopyEnabled/isDownloadEnabled and the isUploadEnabled as you've mentioned that one is needed for the empty state to render.

It makes sense for them to be hidden as there is nothing to copy/download if the code editor is empty.

src/Components/CreateImageWizard/steps/FirstBoot/FirstBoot.scss
regexowl: Well, yeah, but if you look at the code editor in dark mode, it has a white background directly on the PatternFly page: https://www.patternfly.org/components/code-editor

So I'd say let's remove the override, so it follows how are things supposed to be rendered according to PF.

src/test/Components/CreateImageWizard/steps/FirstBoot/Firstboot.test.tsx
```diff
@@ -72,14 +72,6 @@ const selectSimplifiedOscapProfile = async () => {
...
 };
 
-const openCodeEditor = async (): Promise<void> => {
-  const user = userEvent.setup();
-  const startBtn = await screen.findByRole('button', {
-    name: /Start from scratch/i,
-  });
-  await waitFor(() => user.click(startBtn));
-};
-
```
regexowl, line 82: Won't this remove a test that helped us find out that the empty state is missing?

src/Components/CreateImageWizard/steps/FirstBoot/tests/helpers.tsx
```diff
@@ -18,13 +17,6 @@ export const renderFirstBootStep = (
   return renderWithRedux(<FirstBootStep />, wizardStateOverrides);
 };
 
-export const openCodeEditor = async (user: UserEventInstance) => {
-  const startBtn = await screen.findByRole('button', {
-    name: /Start from scratch/i,
-  });
-  await clickWithWait(user, startBtn);
-};
-
```
regexowl, line 27: Same here, isn't this just removing a test that would alert us there's something wrong with the empty state?

src/Components/CreateImageWizard/steps/FirstBoot/index.tsx
```diff
@@ -25,11 +32,21 @@ import {
 } from '@/store/slices/wizard';
 import { useFlag } from '@/Utilities/useGetEnvironment';
 
+// Inline <style> needed because sassPrefix wraps SCSS in .imageBuilder,
+// but the wizard Modal renders in a portal outside that wrapper.
```
regexowl, line 36: Ok, we can revisit this later.

regexowl (commented):
Two comments about the tests, otherwise good to go :)

regexowl (approved):
Thank you! ✨

PR #4249, opened 2026-03-30
Wizard: add image mode on hosted service (HMS-10361)

ksiekl:
Add the "Image mode" toggle for hosted service, get the possible bootc containers from the CRC API using the GET distributions endpoint.
image

The architecture is set for now, so I didn't use the dropdown we have for it in package mode.

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import {
   Content,
@@ -9,9 +9,11 @@ import {
 import { BuildIcon, RepositoryIcon } from '@patternfly/react-icons';
 
 import { Distributions } from '@/store/api/backend';
+import { selectIsOnPremise } from '@/store/slices/env';
 import {
   changeBlueprintMode,
   changeDistribution,
+  selectDistribution,
   selectIsImageMode,
 } from '@/store/slices/wizard';
 
@@ -22,10 +24,14 @@ import { getHostDistro } from '../../../../../Utilities/getHostInfo';
 
 const BlueprintMode = () => {
   const dispatch = useAppDispatch();
+  const isOnPremise = useAppSelector(selectIsOnPremise);
   const isImageMode = useAppSelector(selectIsImageMode);
+  const distribution = useAppSelector(selectDistribution);
   const [defaultDistro, setDefaultDistro] = useState<Distributions>(RHEL_10);
+  const previousDistro = useRef<Distributions>(RHEL_10);
 
   useEffect(() => {
+    if (!isOnPremise) return;
     const fetchDefaultDistro = async () => {
       try {
         const distro = await getHostDistro();
@@ -38,7 +44,7 @@ const BlueprintMode = () => {
     };
 
     fetchDefaultDistro();
-  }, []);
+  }, [isOnPremise]);
 
   return (
     <FormGroup label='Image type' isRequired>
@@ -50,7 +56,11 @@ const BlueprintMode = () => {
           isSelected={!isImageMode}
           onChange={() => {
             dispatch(changeBlueprintMode('package'));
-            dispatch(changeDistribution(defaultDistro));
+            dispatch(
+              changeDistribution(
+                isOnPremise ? defaultDistro : previousDistro.current,
+              ),
+            );
           }}
           aria-describedby='blueprint-mode-description'
         />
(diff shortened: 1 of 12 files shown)


src/Components/CreateImageWizard/steps/ImageOutput/index.tsx
```diff
@@ -49,19 +59,22 @@ const ImageOutputStep = () => {
         Select the release, architecture, and a target environment to build your
         image. Learn more about <DocumentationButton />
       </Content>
-      {isOnPremise &&
-        // The distribution is 'image-mode' when the blueprint is in image mode
-        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-        !distribution?.startsWith('fedora') && <BlueprintMode />}
-      {isImageMode && <ImageSourceSelect />}
+      {isImageModeEnabled && !(distribution as string).startsWith('fedora') && (
```
kingsleyzissou, line 62: I'd be super tempted to drop the Fedora gating here... but we can tackle that at a later time

src/Components/CreateImageWizard/steps/ImageOutput/components/ImageSourceSelect.tsx
```diff
@@ -228,4 +234,120 @@ const ImageSourceSelect = () => {
...
+  const dispatch = useAppDispatch();
+  const arch = useAppSelector(selectArchitecture);
+  const imageSource = useAppSelector(selectImageSource);
+  const [isOpen, setIsOpen] = useState(false);
+
+  const {
+    data: distributions,
+    isLoading,
+    isError,
+  } = useGetDistributionsQuery({ kind: 'bootc', arch });
```
kingsleyzissou, line 247: I think we should refactor the on-premise code to call the same hook so that the code is shared and we only have one component. This would reduce the duplication. But this is quite a big refactor in the cockpit stuff. 

I can handle it in a follow up (it could get quite messy).

src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
```diff
@@ -60,6 +70,9 @@ const BlueprintMode = () => {
           buttonId='blueprint-mode-image'
           isSelected={isImageMode}
           onChange={() => {
+            if (!isOnPremise) {
+              previousDistro.current = distribution as Distributions;
```
kingsleyzissou, line 74: Nitpick: it would probably be more stable to create an item for this in the wizard slice and save the previous distro in the store.

It's would also be fine to share this logic with on premise since the previousDistro should be the defaultDistro anyways.

But this shouldn't block the PR.

kingsleyzissou (approved):
The changes look good. The only thing I noticed is that the review step numbers in the wizard are off now with the hidden customisations:

Image

We can handle that in a follow up though

avitova:
Looks good, however, I was not able to build an image:
```
Image could not be built
Status code 400: distribution not available
```

avitova:
Also, import somehow does not work for me - I have "Create blueprint" disabled if I import this json:
```{
  "bootc": {
    "reference": "rhel/10.1-ec2"
  },
  "content_sources": [],
  "customizations": {
    "groups": [
      {
        "gid": 1000,
        "name": "group1"
      },
      {
        "gid": 1001,
        "name": "group2"
      },
      {
        "gid": 1002,
        "name": "group3"
      }
    ],
    "users": [
      {
        "hasPassword": true,
        "name": "user1"
      }
    ]
  },
  "description": "",
  "distribution": "image-mode",
  "metadata": {
    "exported_at": "2026-04-14 10:23:28.249933721 +0000 UTC",
    "is_on_prem": false,
    "parent_id": "527cdcef-c255-4dcb-8528-72e7d739310a"
  },
  "name": "image-mode-x86_64-20260414-1210"
}```

ksiekl:
@avitova thanks, no image is able to be built rn, my latest information from last week is that there is no image mode compose handler in the backend prepared. I'll checkout the import tho

avitova:
Makes sense, thanks! I am happy if you merge and solve the import in a follow-up. It is behind a flag anyways.

src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
ksiekl: Thanks! I'll look at it, that should be a quick fix I hope. The rest of the things you mention I'll fix in a follow up

ksiekl:
I talked with @avitova and since this buggy behavior on import happens very selectively and only in the old wizard, which should be out really soon, I would merge this PR and check out that behavior in a follow up as well. What do you think @kingsleyzissou ?

kingsleyzissou:
Ship it!

ksiekl:
Alrighty! Will do a follow up on the other mentioned review comments! Thanks!

lzap:
> @avitova thanks, no image is able to be built rn, my latest information from last week is that there is no image mode compose handler in the backend prepared. I'll checkout the import tho

I can confirm and I just picked a jira task to finish with CRC configuration so I am gonna do CRC changes now.

https://redhat.atlassian.net/browse/HMS-10356

PR #4261, opened 2026-04-01
Wizard: Packages revamp (HMS-9397)

regexowl:
This updates the base behaviour and tests.

The step is not yet 1:1 with the mocks and some cleanup will be still needed, but the PR is already quite big.


kingsleyzissou (dismissed):
This one is a bit too big to review line-by-line. But it looks pretty solid. I've gone through and verified it and tested this in the frontend and it does what it says it does.

There is one small nitpick, but I guess that could be a follow up

avitova (commented):
We have some tests under `src/test/Components/CreateImageWizard/steps/Packages`. I think the plan was also to remove those if the coverage is sufficient?

src/Components/CreateImageWizard/steps/Packages/components/Packages.tsx
```diff
@@ -640,140 +71,60 @@ const Packages = () => {
...
-              <SearchInput
-                type='text'
-                placeholder='Search packages'
-                aria-label='Search packages'
-                data-testid='packages-search-input'
-                value={searchTerm}
-                onChange={handleSearch}
-                onClear={() => handleClear()}
-              />
+              <FormGroup label='Package type'>
```
avitova, line 74:
```suggestion
              <FormGroup label='Package Type'>
```

avitova (approved):
I think I am good here, just a tiny suggestion so that we are aligned with mocks, but it can be done in a follow-up. Since the weekend is coming, I will just approve and let's merge on Monday if you feel like it.

kingsleyzissou (approved):
Approved from me too, follow up is completely fine since this is such a large PR

src/Components/CreateImageWizard/steps/Packages/components/Packages.tsx
regexowl: Thank you! :heart:

I'll need to check with the UX about using title case vs sentence case. I believe the code base is currently around 80% sentence case, so I just rolled with that for now. We could update all the case issues in a bulk when there's a final verdict.

PR #4266, opened 2026-04-02
Wizard: Add allowed characters for kernel arguments (HMS-9508)

regexowl:
This updates the validation for the kernel arguments to contain semicolon and backslash as allowed characters.


avitova (approved):
Nice:)

PR #4283, opened 2026-04-07
Add Claude-powered nightly boot test failure analysis

lucasgarfield:
When nightly boot tests fail, Claude Code (via Vertex AI) analyzes the Playwright report and posts a classified analysis to Slack. Failures are categorized as FLAKE, EXTERNAL, or BUG. Flakes are automatically retried up to 3 times.

The workflow is split into 4 separate jobs to isolate secrets:

  boot-tests → notify-slack    → post-analysis
             → analyze-failures ↗

The analyze-failures job (where Claude runs) only has access to Vertex AI credentials. Slack tokens and other secrets are physically isolated on separate runners — this is enforced by the OS, not by Claude's tool restrictions, so even if Claude tried to read /proc/self/environ or similar it would find nothing sensitive.

Notable design decisions:

- The failure notification now uses chat.postMessage (bot token) instead of an incoming webhook so we can thread Claude's analysis as a reply and update the message if a retry succeeds.

- Claude only runs on the first attempt. Retries post a short "also failed" thread reply instead of re-analyzing — same conclusion, saves tokens and time.

- If a retry succeeds, the original "FAILED" message is edited to show "PASSED on retry" so the channel isn't left with a stale alert.

- Claude is told to skip .webm video files in the report. It was reading them (adding ~15 min) but the same flake signal is available from retry counts, error messages, and commit history.

- boot-test-instructions.md contains the full prompt with classification rules and report template, kept separate from the workflow for readability and easier iteration.

### DIFF
--- a/boot-test-instructions.md
+++ b/boot-test-instructions.md
@@ -1,0 +1,82 @@
+# Boot Test Analyst
+
+You are a CI failure analyst. Your job is to investigate failed GitHub Actions runs
+of Playwright end-to-end tests, classify them, and produce a structured report.
+
+Test results are available in one of two formats:
+
+1. **Preferred: `test-results.json`** — Playwright's JSON reporter output in the
+   working directory. This is a structured, machine-readable file containing all
+   test names, statuses, error messages, stack traces, durations, and retry info.
+   Start here if it exists.
+
+2. **Fallback: `playwright-report/`** — Playwright's HTML report directory. Only
+   used for older runs that predate the JSON reporter. Look at the data files
+   inside, not `index.html`.
+
+**Do not read video files (.webm).** They are large and slow to process. You can
+determine flakiness from retry counts, error messages, and commit history without
+watching recordings.
+
+## Available context
+
+You have access to:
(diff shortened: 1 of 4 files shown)


boot-test-instructions.md
tkoscieln: This is not a review comment, but a thought 😄 we could think about making this more reusable, sort of a sample template that could be copied over to other repos

tkoscieln (commented):
Thank you, looks and works great! 🔧 Left some general comments and pointed out one or two issues I found.

tkoscieln (approved):
Let's re-try the BootTests just to be sure (even though I think those failures are unrelated), but otherwise lgtm

lucasgarfield:
Here's a run proving that the changes work:
https://github.com/osbuild/image-builder-frontend/actions/runs/24431049714

I'm going top go ahead and merge because the current boot test failure is related to an intentional breakage from @regexowl (see https://github.com/osbuild/image-builder-frontend/pull/4305).

PR #4293, opened 2026-04-08
Wizard: Move AAP step to Registration step (HMS-9521)

tkoscieln:
This PR moves the AAP step into Registration step. Standalone AAP step was removed entirely. Includes a small improvement to Playwright hosted login function to remove unnecessary 30 second wait when Consent popover resets SSO login form.

---
**Before**: 
image

---
**After**:
image


avitova (commented):
https://github.com/user-attachments/assets/b2caef72-8461-446b-b086-189e550b7fe2

Looks okay, I think. I just do not like this behaviour. I think we should wait for user to insert values before showing an error.
Otherwise ok, thanks!:)

tkoscieln:
@avitova's validation comment will be adressed in a [followup](https://redhat.atlassian.net/browse/HMS-10516) so we don't block the revamp on this

avitova (approved):

kingsleyzissou (approved):

PR #4295, opened 2026-04-09
Wizard: fix Snapshot date bug when importing a bp  (HMS-10323)

ksiekl:
Fixed a bug with snapshot date when importing a blueprint - the exported blueprint format (BlueprintExportResponse) stores snapshot_date at the top level, not in image_requests. 

On a side note, in the second commit, I renamed a variable called `request` that had a `...Response` type, which was a bit confusing.

### DIFF
--- a/playwright/BootTests/Content/helpers.ts
+++ b/playwright/BootTests/Content/helpers.ts
@@ -54,8 +54,8 @@ export const deleteRepository = async (
         .getByLabel('Kebab toggle')
         .click();
       await page.getByRole('menuitem', { name: 'Delete' }).click();
-      // Wait until the repo is loaded in the delete modal
-      await expect(page.getByRole('gridcell').first()).not.toBeEmpty();
+      // Wait until the delete confirmation modal is fully rendered
+      await expect(page.getByRole('dialog')).toBeVisible();
       await page.getByRole('button', { name: 'Delete' }).click();
     },
     { box: true },
--- a/playwright/Customizations/RepeatableBuild.spec.ts
+++ b/playwright/Customizations/RepeatableBuild.spec.ts
@@ -1,5 +1,5 @@
-// import * as fsPromises from 'fs/promises';
-// import * as path from 'path';
+import * as fsPromises from 'fs/promises';
+import * as path from 'path';
 
 import { expect } from '@playwright/test';
 import { v4 as uuidv4 } from 'uuid';
@@ -9,7 +9,6 @@ import {
   navigateToRepositories,
 } from '../BootTests/Content/helpers';
 import { test } from '../fixtures/customizations';
-// import { exportedLocaleBP } from '../fixtures/data/exportBlueprintContents';
 import { isHosted } from '../helpers/helpers';
 import { ensureAuthenticated } from '../helpers/login';
 import {
@@ -20,12 +19,12 @@ import {
 import {
   createBlueprint,
   deleteBlueprint,
-  // exportBlueprint,
+  exportBlueprint,
   fillInDetails,
-  // fillInImageOutputGuest,
-  //importBlueprint,
+  fillInImageOutputGuest,
+  importBlueprint,
   registerLater,
-  // verifyExportedBlueprint,
+  verifyExportedBlueprint,
 } from '../helpers/wizardHelpers';
 
 test('Create a blueprint with Repeatable build customization', async ({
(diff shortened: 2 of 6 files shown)


playwright/helpers/helpers.ts
```diff
@@ -24,6 +24,16 @@ export const togglePreview = async (page: Page) => {
...
+export const enablePreview = async (page: Page) => {
+  const toggleSwitch = page.locator('#preview-toggle');
+
+  if (!(await toggleSwitch.isChecked())) {
+    await toggleSwitch.click();
+    // eslint-disable-next-line playwright/no-wait-for-timeout
+    await page.waitForTimeout(4000);
+  }
+};
+
```
regexowl, line 36: The preview shouldn't need to be enabled as the same functionality is expected both from the old and new wizard. The steps themselves should look and work the same, right?

playwright/BootTests/Content/helpers.ts
```diff
@@ -54,8 +54,8 @@ export const deleteRepository = async (
         .getByLabel('Kebab toggle')
         .click();
       await page.getByRole('menuitem', { name: 'Delete' }).click();
-      // Wait until the repo is loaded in the delete modal
-      await expect(page.getByRole('gridcell').first()).not.toBeEmpty();
+      // Wait until the delete confirmation modal is fully rendered
+      await expect(page.getByRole('dialog')).toBeVisible();
```
regexowl, line 58: I'd expect this shouldn't make much difference as waiting for `await expect(page.getByRole('gridcell').first()).not.toBeEmpty();` should assure the modal is fully loaded, but 🤷

playwright/Customizations/RepeatableBuild.spec.ts
```diff
@@ -40,15 +36,25 @@ test('Create a blueprint with Repeatable build customization', async ({
...
-  // Here we want to be sure that the repository is deleted due to URL exclusivity per repository
-  await deleteRepository(page, repositoryUrl);
-
   // Delete the blueprint after the run fixture
   cleanup.add(() => deleteBlueprint(page, blueprintName));
   cleanup.add(() => deleteRepository(page, repositoryName));
 
   await ensureAuthenticated(page);
 
+  const previewToggle = page.locator('#preview-toggle');
```
regexowl: Isn't this duplicate to the `enablePreview` call? Also see the comment about testing in preview vs stable.

playwright/Customizations/RepeatableBuild.spec.ts
```diff
@@ -127,32 +140,38 @@ test('Create a blueprint with Repeatable build customization', async ({
   });
 
   await test.step('Check Repeatable build with content template', async () => {
-    await frame.getByRole('button', { name: 'Repeatable build' }).click();
+    await frame.getByRole('button', { name: 'Base settings' }).click();
```
regexowl, line 137: This PR combines changes necessary to fix the bug with the migration of these tests to the new wizard modal. Which I think made it increasingly harder to implement.

The import should work the same, regardless on whether it's in the old or new wizard. My vote would go to implementing the bug fix and to continuing testing against stable for now. The playwright test suite should be pointed against the new wizard in a separate task.

src/store/api/contentSources/hosted/contentSourcesApi.ts
regexowl, line 1: These changes are already in via an API update. A rebase should clean them up :)

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -668,43 +668,62 @@ const getMetadata = (metadata: BlueprintMetadata) => {
...
+
   return {
     wizardMode,
-    blueprintMode: isImageModeDistribution(request.distribution)
-      ? 'image'
-      : 'package',
-    metadata: getMetadata(request.metadata),
+    // API needs to be updated first
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    blueprintMode: blueprint.distribution ? 'package' : 'image',
```
regexowl: Is this change necessary or something that sneaked in?

regexowl (commented):
The fix works nicely, left some comments about changes that are possibly non-relevant to the fix and about testing.

The tests will need to be switched to run against V3 for the revamp work, but it shouldn't be relevant to this PR. The functionality of the individual steps between both the old and new Wizard should be the same, including importing. Maybe I'm overlooking something?

playwright/helpers/helpers.ts
ksiekl: Well, not really, the steps' names are different, and any clicking of "Next" or "Back" buttons changes the behavior

playwright/Customizations/RepeatableBuild.spec.ts
ksiekl: Oh, well.... the RepeatableBuild test was commented out and didn't really run, but after uncommenting, it didn't even pass on old. And I though since I have to rewrite it, I might as well do it for the new one. So rolling it back to test against old wizard does seems like a lot of work, and kinda useless since we are going to the new wizard in a week

ksiekl: The import does work the same, I didn't change the test logic, just clicking new buttons in the new wizard

src/Components/CreateImageWizard/utilities/requestMapper.ts
ksiekl: Ugh, a snak in, sorry, I put that back how it was

playwright/Customizations/RepeatableBuild.spec.ts
regexowl: > So rolling it back to test against old wizard does seems like a lot of work, and kinda useless since we are going to the new wizard in a week

Yeah, fair enough since it's already implemented. We should make a note to clean up the toggle switchers after, but let's roll with stage:preview then :+1:

ksiekl: I'm putting it in my notes, won't be able to do it this sprint but I'll keep an eye on it later

playwright/BootTests/Content/helpers.ts
ksiekl: The delete confirmation modal no longer renders a table or grid inside it, so Playwright couldn't find any element matching getByRole('gridcell') :disappointed:

regexowl (dismissed):
Nice, all green, thanks! 🍏

regexowl:
@ksiekl Merging https://github.com/osbuild/image-builder-frontend/pull/4328 came with a conflict, can you please resolve it? This will be good to merge after :rocket:

regexowl (approved):
Perfect, thank you!

kingsleyzissou (approved):

PR #4299, opened 2026-04-13
Wizard: Revamp package recommendations (HMS-9397)

regexowl:
This revamps package recommendations, removes the old component, updates analytics and tests.


src/Components/CreateImageWizard/steps/Packages/components/Packages.tsx
regexowl, line 1: I'm fetching the recommendations here instead of in the search itself because fetching it there led to "Maximum update depth exceeded..." warning. Unsure why that was, but this seems to work without the loop 🤷

regexowl:
> * The recommendations `useEffect` that calls `fetchRecommendedPackages` fires on every `packages`/`distribution` change without any cancellation or error handling, which can lead to race conditions and unhandled rejections if the user edits selections quickly; consider tracking the latest request or using `abort`/`unwrap().catch(...)` to avoid updating analytics with stale responses.
> 
> * There is repeated construction of derived values like `distribution.replace('-', '')` and repeated analytics payload fields across several handlers; factoring these into small helpers/constants would make the code easier to maintain and reduce the risk of subtle inconsistencies between events.

Yeah, both reasonable I guess :eyes:

src/test/Components/CreateImageWizard/steps/Packages/Packages.test.tsx
```diff
@@ -491,20 +466,23 @@ describe('Packages request generated correctly', () => {
...
           ...blueprintRequest.customizations,
-          packages: expectedAllPackageRecommendations,
+          packages: expectedSinglePackageRecommendation,
         },
       };
 
       expect(receivedRequest).toEqual(expectedRequest);
     });
 
     test('deselecting a package recommendation removes it from the request', async () => {
```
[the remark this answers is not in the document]
regexowl: Added some.

regexowl:
Well, hope I didn't mess up the rebase even more :crossed_fingers:

src/Components/CreateImageWizard/steps/Packages/components/Packages.tsx
```diff
@@ -56,6 +90,161 @@ const Packages = () => {
...
+                modelVersion: response.data.modelVersion,
+              },
+            );
+          }
+        } catch {
+          // error state handled by isErrorRecommendations
+        }
+      })();
+    }
+  }, [packages, distribution, isOnPremise]);
```
mgold1234, line 167: analytics, isBeta, and fetchRecommendedPackages are used inside the effect but aren't listed in the dependency array. This can lead to stale closure issues. In practice, fetchRecommendedPackages from RTK Query is a stable reference and analytics likely is too, but it would be better to either add them to the array or add an explicit eslint-disable comment explaining why they're intentionally omitted — similar to what's done for the mount-only useEffect at line 106.

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
```diff
@@ -558,6 +604,10 @@ const PackageSearch = ({
   };
 
   const getPackageDescription = (pkg: IBPackageWithRepositoryInfo) => {
+    if (pkg.isRecommendation) {
```
mgold1234: this part is dead code. The ternary at line 853 already checks pkg.isRecommendation and returns the hardcoded "Suggested based on your selections..." string before getPackageDescription is ever called:


description={
  pkg.isRecommendation
    ? 'Suggested based on your selections. Enabled by RHEL Lightspeed'
    : getPackageDescription(pkg)  // only called when isRecommendation is false
}
So the if (pkg.isRecommendation) check inside getPackageDescription can never be true. maybe consider removing it wdyt?

mgold1234:
@regexowl I started to look over the code, it looks very good.
I notice that when I  am searching package the recomeded package doesnt show up anymore, is it ok?
Screenshot 2026-04-16 at 15 23 16

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
```diff
@@ -507,12 +521,31 @@ const PackageSearch = ({
     isSearchingOtherRepos,
   ]);
 
+  const transformedRecommendations = useMemo(() => {
+    if (
+      packageType === 'groups' ||
+      !recommendations.length ||
+      !debouncedSearchTerm
```
mgold1234: This means the user must have an active search to see recommendations, even if they've already selected packages and recommendations have been fetched. If the user clears the search field and reopens the dropdown, they'll only see "Start typing to search packages" with no recommendations.

In the old UX, the recommendations panel was always visible below the table regardless of search state. Is this behavior change intentional?

src/Components/CreateImageWizard/steps/Packages/components/Packages.tsx
regexowl: The deps are unstable, causing an infinite loop when in the array. Added a comment.

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
regexowl: Nice catch, removed!

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
regexowl: Fair enough, recommendations are not really dependent on the search term, but rather on selected packages. Update the logic so they render even without a search term.

regexowl:
> @regexowl I started to look over the code, it looks very good.
> I notice that when I am searching package the recomeded package doesnt show up anymore, is it ok?

You'll need to search for packages, that do have recommendations. Try adding `blktrace` to selected. Once you do the recommendations should be fetched.

image

mgold1234:
@regexowl now I get it, the thing is that I need to close the dropdown and open it again after choosing packages from the list

https://github.com/user-attachments/assets/ef73e1e9-276a-4c7b-9db1-2d8f2c2b9772

regexowl:
> @regexowl now I get it, the thing is that I need to close the dropdown and open it again after choosing packages from the list

Yes, the dropdown is closing automatically with a package being selected.

The endless re-fetching is a bit annoying as it takes quite some time, but since it depends on the up-to-date list of selected packages we can't really try to prefetch the results :(

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
```diff
@@ -75,6 +79,11 @@ type PackageSearchProps = {
   setIsSelectingGroup: (value: GroupWithRepositoryInfo | undefined) => void;
   activeStream: string;
   setActiveStream: (value: string) => void;
+  recommendations: PackageRecommendation[];
```
mgold1234, line 82: PackageSearch already had 8 props before this PR, and now receives 5 more (13 total). This is getting unwieldy. Consider grouping the recommendation-related props into a single object:


type RecommendationsConfig = {
  items: PackageRecommendation[];
  isLoading: boolean;
  isError: boolean;
  onSelected: (name: string) => void;
  onDropdownOpened: (hasRecs: boolean) => void;
};
Not a blocker, but worth noting as accumulating technical debt.

src/Components/CreateImageWizard/steps/Packages/components/Packages.tsx
```diff
@@ -11,27 +11,50 @@ import {
...
 import PackageSearch from './PackageSearch';
 import PackagesTable from './PackagesTable';
 import RepositoryModal from './RepositoryModal';
 
+import { excludeEUSReposFilter } from '../../../../../Components/CreateImageWizard/steps/Repositories/components/Repositories';
 import {
+  AMPLITUDE_MODULE_NAME,
   ContentOrigin,
   EPEL_10_REPO_DEFINITION,
 } from '../../../../../constants';
```
mgold1234: The deep relative imports (../../../../../) should use the @/ alias. Per the project's code style guidelines, relative imports are intended for ./ and ../ only. Since this code was moved from the deleted PackageRecommendations component, this is a good opportunity to clean them up:


// before
import { AMPLITUDE_MODULE_NAME } from '../../../../../constants';
import { useAppSelector } from '../../../../../store/hooks';

// after
import { AMPLITUDE_MODULE_NAME } from '@/constants';
import { useAppSelector } from '@/store/hooks';

regexowl: True, I'll open a follow up to fix the imports in the entire packages step.

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
regexowl: Good point, but the object would just hide the issue. I could try to clean up the props a bit in a follow up. I'd like to revisit the Packages step code in general.

src/Components/CreateImageWizard/steps/Packages/components/Packages.tsx
regexowl: https://github.com/osbuild/image-builder-frontend/pull/4321

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
kingsleyzissou: Ya I think we should eventually come back to this and clean it up, but I guess to get us over the line for summit we can do this at a later stage

kingsleyzissou (approved):
I've tested it out and it looks good.

I think this component is doing a lot! And we've already commented about how we should revisit it at some point. I think I'd like to tackle the fetching/sorting of all the packages at some point which should help in tidying up the component. But that's definitely not in scope for this PR and possible for after summit.

regexowl:
> I've tested it out and it looks good.
> 
> I think this component is doing a lot! And we've already commented about how we should revisit it at some point. I think I'd like to tackle the fetching/sorting of all the packages at some point which should help in tidying up the component. But that's definitely not in scope for this PR and possible for after summit.

Thank you! Agree that we should come back to this after the deliverables are finished.

PR #4311, opened 2026-04-14
Wizard: Don't include default timezone and locale for unsupported image types

jlsherrill:
Don't include default timezone and locale if the image type doesn't support it.   To test this, simply build a network installer and try to boot it, prior to this it would fail to fully boot.

Fixes HMS-10456

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -22,12 +22,16 @@ import {
   addImageType,
   changeArchitecture,
   changeDistribution,
+  changeKeyboard,
   changeTimezone,
+  clearLanguages,
   initializeWizard,
   selectDistribution,
   selectImageSource,
   selectImageTypes,
   selectIsImageMode,
+  selectKeyboard,
+  selectLanguages,
   selectTimezone,
 } from '@/store/slices/wizard';
 
@@ -232,6 +236,8 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
 
   const distribution = useAppSelector(selectDistribution);
   const timezone = useAppSelector(selectTimezone);
+  const languages = useAppSelector(selectLanguages);
+  const keyboard = useAppSelector(selectKeyboard);
 
   // Image Output
   const targetEnvironments = useAppSelector(selectImageTypes);
@@ -306,6 +312,13 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
       return;
     }
 
+    if (restrictions.timezone.shouldHide) {
+      if (timezone) {
+        dispatch(changeTimezone(''));
+      }
+      return;
+    }
+
     const defaultTimezone =
       distribution === RHEL_10 || targetEnvironments.includes('azure')
         ? DEFAULT_TIMEZONE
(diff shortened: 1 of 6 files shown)


kingsleyzissou (commented):
I was hoping to get to the `requestMapper` as part of my refactoring... it definitely needs some TLC.

src/store/api/distributions/constants.ts
```diff
@@ -69,7 +69,7 @@ export const DISTRO_DETAILS: Record<string, ImageTypeInfo> = {
   },
   'network-installer': {
     name: 'network-installer',
-    supported_blueprint_options: ['locale', 'fips'],
+    supported_blueprint_options: ['fips'],
```
kingsleyzissou, line 72: `locale` is supported though:
https://github.com/osbuild/images/blob/aa49243b220a9d5cdb14b7db7355555bf7d315e1/data/distrodefs/rhel-10/imagetypes.yaml#L2460-L2463

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -1139,6 +1140,26 @@ const getModules = (state: RootState) => {
...
 
+const noTypeSupportsOption = (
+  state: RootState,
+  option: 'timezone' | 'locale',
+): boolean => {
+  const imageTypes = selectImageTypes(state);
+  return (
+    imageTypes.length > 0 &&
+    imageTypes.every((type) => {
+      if (!(type in DISTRO_DETAILS)) {
```
kingsleyzissou, line 1151: The `DISTRO_DETAILS` was intended to be a temporary thing for an abstraction layer to help us with hiding stuff until Tom's endpoint had landed. I think the plan is to switch over to that and then we won't need `DISTRO_DETAILS` any more.

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -306,6 +312,13 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
       return;
     }
 
+    if (restrictions.timezone.shouldHide) {
```
kingsleyzissou, line 315: Would it be simpler to maybe just return early here and leave timezone as undefined (or alternatively set it to undefined explicitly)? That should simplify the request mapper code too then? We would only really need to check if timezone is defined or not there

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -306,6 +312,13 @@ const CreateImageWizard = ({ isEdit }: CreateImageWizardProps) => {
...
+      if (timezone) {
+        dispatch(changeTimezone(''));
+      }
+      return;
+    }
+
```
kingsleyzissou, line 321: Also, I think we might need to double check that NTP servers are cleared out

src/Components/CreateImageWizard/utilities/requestMapper.ts
```diff
@@ -1148,6 +1169,10 @@ const getTimezone = (state: RootState) => {
     return undefined;
   }
 
+  if (noTypeSupportsOption(state, 'timezone')) {
+    return undefined;
+  }
+
   if (!timezone && ntpservers?.length === 0) {
```
kingsleyzissou, line 1176: Since we set timezone to an empty string `''`, this block here should catch this and return undefined. So the above `noTypeSupportOption` might be overkill

we just need to make sure that we clear NTP servers too - I added a comment below

src/store/api/distributions/constants.ts
jlsherrill: oh good point, it was originally broken, but that is fixed

PR #4312, opened 2026-04-15
Wizard: Fix bug re-initializing registration (HMS-10540)

regexowl:
How to reproduce the bug:
1. set registration to 'register-later'
2. return to the first step of the wizard
3. check the registration again

Current behaviour:
- registration will be set as 'register-now' again

Desired behaviour:
- the 'register-later' option stays selected


regexowl:
This is present both in the old and new wizard.

kingsleyzissou (approved):

PR #4313, opened 2026-04-15
Wizard: Enable navigation when the first step is valid (HMS-10541)

regexowl:
Since only the first step has really required fields, it should be safe to enable the left hand navigation as soon as it's correctly filled in.


playwright/Basic/imageMode.spec.ts
regexowl, line 1: Had to add some checks, because the playwright tests were faster than the validation.

kingsleyzissou (approved):

tkoscieln (approved):
Lgtm 🚀

PR #4315, opened 2026-04-16
Wizard: Add SearchableSelect component and refactor KeyboardDropDown (HMS-9517)

mgold1234:
Replace the Select/typeahead dropdown in the Keyboard section of the
Locale step with a new reusable SearchableSelect component that uses
Menu/MenuContainer/SearchInput from PatternFly.

The SearchableSelect component provides a cleaner dropdown with a
dedicated search input field, replacing the inline typeahead pattern.
It will also be used in a follow-up PR to redesign the Languages
dropdown.

JIRA: HMS-9517

### DIFF
--- a/src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
+++ b/src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
@@ -1,25 +1,11 @@
-import React, { useEffect, useState } from 'react';
+import React, { useMemo, useState } from 'react';
 
-import {
-  Button,
-  FormGroup,
-  HelperText,
-  HelperTextItem,
-  MenuToggle,
-  MenuToggleElement,
-  Select,
-  SelectList,
-  SelectOption,
-  TextInputGroup,
-  TextInputGroupMain,
-  TextInputGroupUtilities,
-} from '@patternfly/react-core';
-import { TimesIcon } from '@patternfly/react-icons';
+import { FormGroup, HelperText, HelperTextItem } from '@patternfly/react-core';
 
 import { changeKeyboard, selectKeyboard } from '@/store/slices/wizard';
 
 import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
-import sortfn from '../../../../../Utilities/sortfn';
+import SearchableSelect from '../../../../sharedComponents/SearchableSelect';
 import { useLocaleValidation } from '../../../utilities/useValidation';
 import { keyboardsList } from '../data/keyboardsList';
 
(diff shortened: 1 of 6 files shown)


src/Components/CreateImageWizard/steps/Locale/components/KeyboardDropDown.tsx
```diff
@@ -30,125 +16,26 @@ const KeyboardDropDown = () => {
...
-          ) : (
-            <SelectOption isDisabled>
-              {`No results found for "${filterValue}"`}
-            </SelectOption>
-          )}
-        </SelectList>
-      </Select>
+        placeholder='Select a keyboard'
+        onSelect={handleSelect}
+        isFullWidth
```
regexowl: Shouldn't be full width according to mocks.

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -0,0 +1,163 @@
...
+      <span
+        style={{
+          display: 'block',
+          width: '100%',
+          whiteSpace: 'nowrap',
+          overflow: 'hidden',
+          textOverflow: 'ellipsis',
+        }}
+        title={selectedLabel || selected || undefined}
+      >
```
regexowl: Same here, this feels unnecessary

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    return [...options]
+      .filter((opt) =>
+        opt.label.toLowerCase().includes(searchValue.toLowerCase()),
+      )
+      .sort((a, b) => sortfn(a.label, b.label, searchValue));
+  }, [searchValue, options]);
+
+  const handleSelect = (
```
regexowl, line 54: There is currently no way to unselect a selected keyboard. Maybe we should add a logic where clicking on a selected keyboard option again unselects it. This would allow user to clear the selection if they change their mind.

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -0,0 +1,163 @@
...
+            }}
+          />
+        </MenuSearchInput>
+      </MenuSearch>
+      <Divider />
+      <MenuContent maxMenuHeight='300px'>
+        <MenuList>
+          {filteredOptions.length > 0 ? (
+            filteredOptions.map((option) => (
+              <MenuItem key={option.value} itemId={option.value}>
```
regexowl: The active selection doesn't have a checkmark next to the name in the option.

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+  };
+
+  const toggle = (
+    <MenuToggle
+      ref={toggleRef}
+      onClick={() => setIsOpen(!isOpen)}
+      isExpanded={isOpen}
+      {...(isFullWidth !== undefined && { isFullWidth })}
+      {...(toggleClassName !== undefined && { className: toggleClassName })}
```
regexowl: Are these props really needed? The search toggle shouldn't be full width.

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -0,0 +1,163 @@
...
+      style={
+        isFullWidth
+          ? {
+              width: '100%',
+              minWidth: '100%',
+            }
+          : undefined
+      }
```
regexowl: We might want to set the style in a similar way to templates search dropdown with adjusted values. Maybe setting a minWidth would be enough to make sure the field doesn't shrink?

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+  const [searchValue, setSearchValue] = useState('');
+
+  const toggleRef = useRef<HTMLButtonElement>(null);
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  const selectedLabel = useMemo(() => {
+    const match = options.find((opt) => opt.value === selected);
+    return match?.label;
+  }, [options, selected]);
```
regexowl: Value and label should be identical for the keyboard options, is this needed?

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const filteredOptions = useMemo(() => {
+    if (!searchValue) {
+      return options;
+    }
+
+    return [...options]
```
regexowl: The spread might be unnecessary I think, filter should return a new array anyway. But this is a super small thing and could be more of a stylistic choice, so not a blocker.

```suggestion
    return options
```

regexowl (changes requested):
Thanks! Added some comments.

src/Components/sharedComponents/SearchableSelect.tsx
mgold1234: ammm no

regexowl:
@mgold1234 Works great, thank you! Could you please clean up how are the files split into the commits, so the commit messages reflect the changes that are included in them?

Or just squash the commits into one, that would also work perfectly fine in this case :)

regexowl:
Ah, now I see what the issue with the tests is, I've added some tests in https://github.com/osbuild/image-builder-frontend/pull/4335, could you please update locators in them so they pass?

regexowl (approved):
Thank you!

PR #4316, opened 2026-04-16
Wizard: Locale languages redesign (HMS-9517)

mgold1234:
## Summary
- Replace the multi-select typeahead language dropdown with individual rows, each with its own SearchableSelect and remove button
- Add an "Add language" button to insert new language rows
- Depends on #PR1 (locale-searchable-select) which introduces the SearchableSelect component

## Test plan
- [ ] Verify adding a language via "Add language" button works
- [ ] Verify removing a language via the remove button works
- [ ] Verify language search and selection within each row
- [ ] Verify pre-populated languages render correctly on edit
- [ ] Run unit tests: `npm run test:unit`
- [ ] Run Playwright E2E: locale customization flow
JIRA: HMS-9517

### DIFF
--- a/src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
+++ b/src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
@@ -1,31 +1,22 @@
-import React, { useEffect, useState } from 'react';
+import React, { useMemo, useState } from 'react';
 
 import {
   Button,
   FormGroup,
   HelperText,
   HelperTextItem,
-  Label,
-  LabelGroup,
-  MenuToggle,
-  MenuToggleElement,
-  Select,
-  SelectList,
-  SelectOption,
-  TextInputGroup,
-  TextInputGroupMain,
-  TextInputGroupUtilities,
 } from '@patternfly/react-core';
-import { TimesIcon } from '@patternfly/react-icons';
+import { MinusCircleIcon, PlusCircleIcon } from '@patternfly/react-icons';
 
 import {
   addLanguage,
   removeLanguage,
+  replaceLanguage,
   selectLanguages,
 } from '@/store/slices/wizard';
 
 import { useAppDispatch, useAppSelector } from '../../../../../store/hooks';
-import sortfn from '../../../../../Utilities/sortfn';
+import SearchableSelect from '../../../../sharedComponents/SearchableSelect';
 import { useLocaleValidation } from '../../../utilities/useValidation';
 import { languagesList } from '../data/languagesList';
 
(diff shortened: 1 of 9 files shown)


src/Components/sharedComponents/SearchableSelect.scss
```diff
@@ -0,0 +1,11 @@
+.searchable-select-full-width {
```
kingsleyzissou: It doesn't look like this file gets used

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -132,6 +156,8 @@ const SearchableSelect = ({
       onOpenChangeKeys={['Escape']}
     />
   );
+
+  return menuContainer;
```
kingsleyzissou: Nitpick: You could just return `<MenuContainer ... />` directly without creating a new variable

src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
```diff
@@ -47,9 +38,59 @@ const parseLanguageOption = (language: string) => {
...
+          selected={selectedLanguage}
+          placeholder='Select a language'
+          onSelect={onSelect}
+          isFullWidth
+        />
+      </div>
+      <Button
+        variant='plain'
+        onClick={onRemove}
+        aria-label='Remove language'
```
kingsleyzissou: Nitpick: we're mapping over this and creating multiple rows with multiple remove buttons, they're all getting the same `aria-label`

src/store/slices/wizard/index.ts
```diff
@@ -1415,6 +1415,19 @@ export const wizardSlice = createSlice({
         }
       }
     },
+    replaceLanguage: (
```
kingsleyzissou, line 1418: I think technically there is nothing stopping us from adding duplicate languages. This is fine for now but maybe we should have some kind of validation error for this 

But this could be done as a follow up... or after summit (but we should create an issue to track it)

mgold1234: https://redhat.atlassian.net/browse/HMS-10599

src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
```diff
@@ -47,9 +38,59 @@ const parseLanguageOption = (language: string) => {
...
+          ([raw]) =>
+            !existingLanguages.includes(raw) || raw === selectedLanguage,
+        )
+        .map(([raw, parsed]) => ({ value: raw, label: parsed })),
+    [existingLanguages, selectedLanguage],
+  );
+
+  return (
+    <div className='pf-v6-u-display-flex pf-v6-u-align-items-center pf-v6-u-gap-sm pf-v6-u-mb-sm'>
+      <div style={{ width: '50%', minWidth: '200px', maxWidth: '400px' }}>
```
ksiekl: It's not a good practise to use px to set components' styling, because it can display badly on different screens. I would prefer if the styling was not needed at all, but if it really is necessary, please use percentages instead.

src/Components/sharedComponents/SearchableSelect.tsx
```diff
@@ -24,21 +24,30 @@ type SearchableSelectProps = {
...
+  isFullWidth?: boolean;
 };
 
 const SearchableSelect = ({
   options,
   selected,
   placeholder = 'Select an option',
+  searchPlaceholder = 'Search by name',
   onSelect,
+  isFullWidth = false,
```
ksiekl, line 38: I didn't spot a place where you set this to true, ever, so maybe it's not needed if it's always set to false?

ksiekl (commented):
Thanks, this looks good! ❤️ Just a little styling issue and nitpicks, no need to change this one if you don't want to - the space between the text and the "Add language" button looks too small for my taste, not sure if the mocks show it like this, but I would prefer to add a bit more breathing room between these two: 

Image

src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
```diff
@@ -59,169 +100,70 @@ const LanguagesDropDown = () => {
...
-
-  const onToggleClick = () => {
-    setIsOpen(!isOpen);
-  };
-
-  const onClearButtonClick = () => {
-    setInputValue('');
-    setFilterValue('');
+  const handleRemoveLanguage = (language: string) => {
+    dispatch(removeLanguage(language));
```
kingsleyzissou, line 124: So I did another test and noticed that we're not checking for the length of the languages...
meaning you can remove all the languages. We probably wan't to disable this button if there is only one language left.

Happy for this to be done in a follow up though

src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
```diff
@@ -47,9 +38,59 @@ const parseLanguageOption = (language: string) => {
...
+          selected={selectedLanguage}
+          placeholder='Select a language'
+          onSelect={onSelect}
+          isFullWidth
+        />
+      </div>
+      <Button
+        variant='plain'
+        onClick={onRemove}
+        aria-label={`Remove language ${selectedLanguage || ''}`}
```
kingsleyzissou: Nitpick, if the language is undefined we will end up with a label that has an empty space at the end:
`"Remove language "`

src/Components/sharedComponents/SearchableSelect.tsx
mgold1234: isFullWidth is used and set to true in LanguagesDropDown, PolicySelector, ProfileSelector, and TimezoneDropDown.

src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
mgold1234: opened thiss one https://redhat.atlassian.net/browse/HMS-10609

kingsleyzissou (approved):

mgold1234:
thank you @kingsleyzissou and @regexowl

PR #4294, opened 2026-04-08
Wizard: revamp the review step (HMS-10416)

kingsleyzissou:
## Summary
Replaces the monolithic Review step (~2,000 lines across 3 files with 18 `useState` calls) with a modular card-based architecture. Each review section is now an independent component that hides when unconfigured, matching the updated UX requirements.

## Changes
- Decompose Review step into feature-scoped cards (ImageOverview, Registration, Security, Content, RepeatableBuild, AdvancedSettings) with shared primitives (ReviewGroup, StatusItem, LabelMapper, FlexColumn)
- Extract `useSecuritySummary` hook to centralize OpenSCAP/compliance data fetching across Kernel, Services, and Review steps
- Extract `useTargetEnvironmentCategories` hook to share cloud/format categorization between ImageOutput and Review
- Add derived wizard selectors (`selectFirewallEnabled`, `selectAapEnabled`, `selectAllRepositoryIds`, `selectFSConfigurationsCount`, `selectImageMinSize`) and batch repository name resolution into a single API call
- Add comprehensive unit tests for all new cards and update Playwright e2e tests to use role-based selectors

PR #4318, opened 2026-04-16
Wizard/Review: address follow-up comments from review step revamp

kingsleyzissou:
## Summary

Addresses unresolved review comments from #4294 (review step revamp) covering text corrections, accessibility, navigation bugs, and a restrictions fix.

## Changes
- Fix text: rename kernel heading to 'Args', comma-separate user groups, use American spelling 'Organization ID', and format snapshot dates with `timestampToDisplayString`
- Add `aria-label` to all `LabelGroup` components via a new required `ariaLabel` prop on `LabelMapper`, plus direct labels on `SecurityDetails` and `RepositoryDetails`
- Use `ExclamationCircleIcon` for the danger variant in `StatusItem` to match the PF6 status icon pattern
- Fix edit button navigation: correct the wizard step ID typo (`advance-settings-step` → `advanced-settings-step`) and point RepeatableBuild/Security cards to `base-settings-step`
- Hide registration card for non-RHEL distros by adding `aap` to the non-RHEL guard in `isCustomizationSupported`

### DIFF
--- a/playwright/Customizations/RepeatableBuild.spec.ts
+++ b/playwright/Customizations/RepeatableBuild.spec.ts
@@ -106,7 +106,7 @@ test('Create a blueprint with Repeatable build customization', async ({
       .filter({ hasText: 'Enable repeatable build' });
     await expect(repeatableBuildCard).toBeVisible();
     await expect(repeatableBuildCard.getByText('Enabled')).toBeVisible();
-    await expect(repeatableBuildCard.getByText('2025-12-24')).toBeVisible();
+    await expect(repeatableBuildCard.getByText('Dec 24, 2025')).toBeVisible();
   });
 
   await test.step('Check Repeatable build step behaviour with 1 repo', async () => {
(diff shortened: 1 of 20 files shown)


kingsleyzissou:
I skipped some of the patternfly recommendations since that actually changes the layout of the `users` & `firewall` items... we could maybe discuss those changes with UX

regexowl (approved):
Looks great!

I wonder if we could scroll the relevant section into the view for registration, repeatable build and security 🤔 Might add it to the JIRA

kingsleyzissou:
> Looks great!
> 
> I wonder if we could scroll the relevant section into the view for registration, repeatable build and security 🤔 Might add it to the JIRA

Oh that's a good point, definitely would be nice to have

regexowl:
> Oh that's a good point, definitely would be nice to have

Looking into it, seems quite straightforward. Will hopefully open a small PR shortly.

PR #4319, opened 2026-04-16
vitest: Make LandingPage tests co-located and remove dependency on msw (HMS-10548)

regexowl:
Move the landing page tests to a co-located directory.


kingsleyzissou (approved):
Neat

PR #4328, opened 2026-04-17
Wizard: image-mode changes property name (HMS-10554)

ksiekl:
In regards to new changes in CRC:
- the `image_name` prop was changed to `reference` 
- the distro now no longer expects `image-mode`, but the real distro.
- The bootc request also expects only one reference, so I constrained the target env in image mode only to radio buttons
- added a ref for the architecture so that when the user clicks between the package\image mode, they don't mess up

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
@@ -8,13 +8,15 @@ import {
 } from '@patternfly/react-core';
 import { BuildIcon, RepositoryIcon } from '@patternfly/react-icons';
 
-import { RHEL_10 } from '@/constants';
+import { RHEL_10, X86_64 } from '@/constants';
 import { Distributions } from '@/store/api/backend';
 import { useAppDispatch, useAppSelector } from '@/store/hooks';
 import { selectIsOnPremise } from '@/store/slices/env';
 import {
+  changeArchitecture,
   changeBlueprintMode,
   changeDistribution,
+  selectArchitecture,
   selectDistribution,
   selectIsImageMode,
 } from '@/store/slices/wizard';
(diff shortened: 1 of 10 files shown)


src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
```diff
@@ -71,9 +78,14 @@ const BlueprintMode = () => {
...
             if (!isOnPremise) {
               previousDistro.current = distribution as Distributions;
+              previousArch.current = architecture;
             }
             dispatch(changeBlueprintMode('image'));
-            dispatch(changeDistribution('image-mode'));
+            if (isOnPremise) {
+              dispatch(changeDistribution('image-mode'));
+            } else {
+              dispatch(changeArchitecture(X86_64));
```
regexowl, line 87: Is the arch hardcoded because image mode only supports x86 for now?

regexowl (approved):
Looks perfect, thank you! 🚀 

One tiny bug I've noticed is that if you select an image mode unsupported target while in package mode and then switch over to image mode, the target will stay stale in the store. We could either clear all `imageTypes` when switching between the modes (it's a big change so it could make sense), or we can filter out unsupported targets. Both would work just fine.

But that would be easy to address in a follow, so let's merge this and unblock other PRs.

regexowl:
> One tiny bug I've noticed is that if you select an image mode unsupported target while in package mode and then switch over to image mode, the target will stay stale in the store. We could either clear all `imageTypes` when switching between the modes (it's a big change so it could make sense), or we can filter out unsupported targets. Both would work just fine.

Opened a follow up https://github.com/osbuild/image-builder-frontend/pull/4333

src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
ksiekl: Yes, that's exactly the reason

ksiekl:
@regexowl thank you! :heart:

PR #4335, opened 2026-04-20
vitest: Add "onSubmit" tests to text inputs (HMS-10462)

regexowl:
This adds new form submission tests to ensure that entering inside a text input doesn't trigger the default form action and reloads the wizard.


kingsleyzissou (approved):
Nice

PR #4339, opened 2026-04-21
playwright: Make repeatable build test less flaky

regexowl:
The hosted tests don't have to depend on the integration UI. Replace the repo creation via the UI with an API creation.


regexowl:
I'll probably revisit later. But we should make the hosted tests independent on the content integration.

PR #4340, opened 2026-04-21
Wizard/FSC: Update the filesystem configuration step as per mocks (HMS-9395)

regexowl:
This updates the step as per the revamp mocks.

There will be follow up PRs, so included here are not
- volume group revamp
- splitting FSC tests to unit and integration tests / any playwright migration
- spacing
- final documentation button

VG revamp and spacing follow up: https://github.com/osbuild/image-builder-frontend/pull/4341


ksiekl (approved):
Thanks, looks good to me :)
I have just a little nitpick that can be resolved later, but this field should accept only numbers: 

Image

regexowl:
> Thanks, looks good to me :) I have just a little nitpick that can be resolved later, but this field should accept only numbers:
> Image

Great catch, I'll fix that in a follow up. Thank you! :heart:

PR #4341, opened 2026-04-22
Wizard/FSC: Move volume group to expandable and fix step spacing (HMS-9395)

regexowl:
This updates the spacing on the step and moves the content of the volume group card to an expandable. The "Add LVM volume group" and remove group buttons were replaced by the mechanic of opening/closing the expandable as discussed with the UX team.

Rebased on https://github.com/osbuild/image-builder-frontend/pull/4340

### DIFF
--- a/playwright/Customizations/Disk.spec.ts
+++ b/playwright/Customizations/Disk.spec.ts
@@ -70,7 +70,7 @@ test('Create a blueprint with Disk customization', async ({
     ).toBeVisible();
     await expect(
       frame.getByRole('button', {
-        name: /add lvm volume group/i,
+        name: /Volume Group Manager/i,
       }),
     ).toBeVisible();
   });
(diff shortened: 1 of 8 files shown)


regexowl:
> * In `VolumeGroups`, the `onToggle` handler always adds a new LVM volume group when expanding, even if one already exists; consider creating a VG only when none is present (e.g., `if (isExpanded && !vg)`) to avoid duplicate groups when the step loads with an existing VG.

Fixed

regexowl:
> * The `VolumeGroups` component now effectively only uses `volumeGroups[0]`; if multiple LVM volume groups are possible, you may want to either enforce a single-VG constraint explicitly or adapt the UI to handle multiple groups instead of silently discarding the others.

Only one VG is currently supported, added a comment.

ksiekl (approved):
Thanks! This looks like such donkey work 😓 💪 
I have a very little, very niche nitpick that probably no user will ever notice; feel free to ignore it.

When narrowing the window, there is a moment when the Mount point name and the input fields are not aligned well:

Image

regexowl:
> When narrowing the window, there is a moment when the Mount point name and the input fields are not aligned well:

Thank you! Nice catch, I'll open an issue for this so we can revisit it later.

PR #4345, opened 2026-04-22
Wizard: Resolve auto-scrolling issues (HMS-10579)

regexowl:
The `autoFocus` in the registration step made the wizard scroll down to registration both when switching between steps and on state changes. This resolves the issue and makes the Wizard start at the top of the step.


ochosi (approved):
Works like a charm ✨ 
What a simple fix!

kingsleyzissou (approved):

kingsleyzissou:
Admin merged this because the playwright tests are broken and this is a trivial fix

PR #4348, opened 2026-04-22
Wizard/Repositories: Update as per newest mocks (HMS-10579)

regexowl:
This adds the required label, lock icon to the required repositories and updates the copy on the step as per recent mocks.


avitova (dismissed):
Nice! I would appreciate it if the placeholder on the search field was visible completely and not juts its part - meaning we'd have a bigger input field. But not a blocker.
Image

regexowl:
> Nice! I would appreciate it if the placeholder on the search field was visible completely and not juts its part - meaning we'd have a bigger input field. But not a blocker.

Good point, added a commit to make the input wider.

kingsleyzissou (approved):

PR #4350, opened 2026-04-22
src: Open new Wizard while on `/imagewizard` (HMS-10579)

regexowl:
This renders the `LandingPage` a open the wizal in modal when on `/imagewizard` subpath.

### DIFF
--- a/src/Components/LandingPage/LandingPage.tsx
+++ b/src/Components/LandingPage/LandingPage.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import {
   PageSection,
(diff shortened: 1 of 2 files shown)


avitova (commented):
I compared the previous behaviour and do not see a difference. Why are we changing this?

regexowl:
> I compared the previous behaviour and do not see a difference. Why are we changing this?

At the moment `/imagewizard` opens to the old wizard. We need it to open the new one.

Since the new one is just a modal that opens over the landing page there was no routing necessarily needed and the subpath would go away with deprecation of the old wizard. But as @ochosi pointed out the documentation and download links and such use the `/imagewizard`subroute, so we have to keep the support for it.

mgold1234 (approved):
/lgtm

PR #4351, opened 2026-04-22
Wizard: change analytics tracking (HMS-10493)

ksiekl:
This commit:
- added the analytics tracking to the new wizard modal
- removed the security step tracking, since it is not a standalone step anymore
- every other tracking was retained during revamp so all else is gucci

### DIFF
--- a/src/Components/CreateImageWizard3/CreateImageWizard3.tsx
+++ b/src/Components/CreateImageWizard3/CreateImageWizard3.tsx
@@ -14,8 +14,10 @@ import {
   WizardStep,
   WizardStepType,
 } from '@patternfly/react-core';
+import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
 import { useSearchParams } from 'react-router-dom';
 
+import { useGetUser } from '@/Hooks';
 import { useGetBlueprintQuery } from '@/store/api/backend';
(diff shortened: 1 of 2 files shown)


mgold1234:
@ksiekl can you please fix the conflict and I will review it after

src/Components/CreateImageWizard3/CreateImageWizard3.tsx
```diff
@@ -315,6 +355,8 @@ const CreateImageWizard3 = () => {
     >
       <Wizard
         onClose={handleClose}
+        onStepChange={handleStepChange}
+        title='Create image wizard'
```
regexowl: The title should be removed. When it's there, a tooltip with "Create image wizard" renders anywhere the mouse hovers over the modal:

Image

regexowl (commented):
Perfect, thanks a lot for reviewing the tracking events! ❤️ 

One small nitpick, otherwise it's good to go 🚀

src/Components/CreateImageWizard3/CreateImageWizard3.tsx
ksiekl: Ah, naasty, that slipped past me, sorry!

regexowl (approved):
🚀

PR #4353, opened 2026-04-23
Wizard/Timezone: SPUR fixes (HMS-10579)

regexowl:
This:
- renames the section to "Time" so there are not two "Timezone" headings right after each other.
- updates spacing between the title and the step content
- updates helper text below the timezone dropdown


mgold1234 (approved):
/lgtm

PR #4355, opened 2026-04-23
Wizard/OpenSCAP: Show error only when the packages are installed

regexowl:
The error that profiles can't be fetched is rendered even when the OpenSCAP packages are not installed. Let's hide it and show it only when the packages are already available and something goes wrong.


tkoscieln (approved):
lgtm 🚀

PR #4369, opened 2026-04-28
on-prem: fix compose status retrieval to handle missing files (HMS-10350)

### LINKED ISSUE #4200: OnPrem: Get compose status could exit early and ignore compose status (HMS-10350)
The following lines could potentially cause an error that throws an error causing the loop to break and only one item being read
if the file is not found or if the file somehow gets corrupt. This means the loop will exit early and we might not get all the compose statuses we're looking for.

https://github.com/osbuild/image-builder-frontend/blob/435d3d977a9ab25cf2ad5cdb7a2402487f0f1fd7/src/store/cockpit/cockpitApi.ts#L743-L745

## Expected Behavior
The loop should continue searching other blueprint directories if the compose file 
isn't found in the current one.

## Suggested Fix
Wrap the file read in a try-catch and continue the loop on error.

```typescript
for (const bpEntry of entries) {
  try {
    const request = await cockpit
      .file(path.join(blueprintsDir, bpEntry[0], queryArg.composeId))
      .read();
    return {
      data: {
        image_status: data.image_status,
        request: JSON.parse(request),
      },
    };
  } catch {
    // File not found in this blueprint directory, continue searching
    continue;
  }
}
```

JIRA: [HMS-10350](https://issues.redhat.com/browse/HMS-10350)

kingsleyzissou:
## Summary
`getComposeStatus` and `readComposes` crashed when a compose file was missing or corrupt on disk, because `cockpit.file().read()` was called without error handling. This also meant `getComposeStatus` only ever checked the first blueprint directory.

## Changes
- Extract `safeReadJsonFile<T>` helper that wraps cockpit file reads in try-catch, returning `null` on error with a `console.warn` for debugging
- Fix `getComposeStatus` to search all blueprint directories instead of returning on the first iteration
- Fix `readComposes` to skip unreadable compose files instead of aborting the entire listing
- Add unit tests for both `safeReadJsonFile` (3 tests) and `readComposes` skip-and-continue behavior (4 tests)

Closes #4200

### DIFF
--- a/src/store/api/backend/onprem/composerApi/composes.ts
+++ b/src/store/api/backend/onprem/composerApi/composes.ts
@@ -8,12 +8,14 @@ import { OnPremBuilder, onPremQueryHandler } from '@/store/api/shared';
 import {
   getBlueprintsPath,
   readComposes,
+  safeReadJsonFile,
   toComposerComposeRequest,
 } from './helpers';
 
 import {
   ComposeBlueprintApiArg,
   ComposeBlueprintApiResponse,
+  ComposeRequest,
   ComposeResponse,
   ComposesResponseItem,
   GetBlueprintComposesApiArg,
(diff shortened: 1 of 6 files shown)


regexowl (approved):
Niiice 🚀

PR #4370, opened 2026-04-28
Wizard: Update spacing on the base settings step (HMS-10579)

regexowl:
This updates the spacing between titles and content of the substeps.

PR #4371, opened 2026-04-28
Wizard: Update spacing between the main title of the step and the step description (HMS-10579)

regexowl:
This updates the spacing, Base settings are addressed in #4370, so they were left out of this PR.


regexowl:
> * Consider whether the nested `<Content>` inside another `<Content>` is necessary for spacing, as this double wrapping may introduce unintended layout/spacing changes compared to using a single `Content` with modifiers or utility classes.

Considered, the nesting is necessary.

kingsleyzissou (approved):

PR #4378, opened 2026-04-29
Wizard: Security changes (HMS-10579)

mgold1234:
- Capitalize "Compliance" in radio button label
- Move FIPS toggle below the compliance/OpenSCAP section
- 
JIRA: HMS-10579


regexowl:
The third 'None' radio should be implemented at the same time as the 'None' options get removed from the dropdowns. That way the user still has an option to clearly select no policy/compliance.

regexowl:
Maybe the addition of the third radio and removal of 'None' option could be split to a separate PR.

regexowl (approved):
Thanks!

PR #4379, opened 2026-04-29
Wizard: Fix host initialization (HMS-10611)

regexowl:
This makes the host distro/arch overwrite the initial state for both create and import mode of the Wizard.


tkoscieln (approved):
lgtm 🚀

kingsleyzissou (approved):

PR #4380, opened 2026-04-29
Wizard: exclude EUS repos from repo search

xbhouse:
This filters out the EUS repos in the `RepositorySearch` component since these aren't fully supported by IB yet. Also adds the helper for filtering out EUS repos to the `repositoriesUtilities` to avoid circular dependencies.


regexowl (approved):
Thank you!

PR #4382, opened 2026-04-30
Wizard: Add 'No additional policy or profile' radio and remove 'None' from dropdowns (HMS-10579)

mgold1234:
Add a radio button to explicitly opt out of compliance, replacing the
'None' option that was previously embedded in both the policy and
profile dropdowns. Default compliance type is now 'none'.

JIRA : HMS-10579

### DIFF
--- a/src/Components/CreateImageWizard/steps/Oscap/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/index.tsx
@@ -128,8 +128,8 @@ const OscapContent = () => {
     dispatch(changeFips(checked));
   };
 
-  const handleTypeChange = (complianceType: string) => {
-    dispatch(changeComplianceType(complianceType as ComplianceType));
+  const handleTypeChange = (complianceType: ComplianceType) => {
+    dispatch(changeComplianceType(complianceType));
 
     // Avoid showing profile information when switching between types by
     // clearing the compliance data.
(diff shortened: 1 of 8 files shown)


src/Components/CreateImageWizard/steps/Oscap/index.tsx
ksiekl: This is just a nitpick, but since we don't have "None" as an actual option anymore, this doesn't need to be a string but could be the actual type:
```
const handleTypeChange = (type: ComplianceType) => {
  dispatch(changeComplianceType(type));
  // ...
};
```
But this doesn't affect the code much, it would be just a little cleaner.

ksiekl (dismissed):
Thanks, this looks nice! I noticed a little thing that's been bugging me - in the dark mode the placeholder text for the profile selector uses white text color when it should be a grey one like in the policy:

Image
But that can be a follow up 😄 
Other than that, just one small nitpick.

mgold1234:
@ksiekl thank you for the review will fix the dark mode in a follow up pr thanks again

mgold1234:
https://redhat.atlassian.net/browse/HMS-10638

ksiekl:
Huh, the tests seem related to this PR

ksiekl:
Great, thank you! :sparkles:

ksiekl (dismissed):

PR #4385, opened 2026-04-30
Wizard/ Oscap: Lock FIPS toggle when enabled by compliance profile (HMS-10579)

mgold1234:
Some compliance policies and OpenSCAP profiles (e.g., STIG) require
FIPS to be enabled. When selecting such a profile, the FIPS toggle is
now automatically turned on and locked, preventing the user from
disabling it. Clearing the profile selection or switching compliance
types unlocks the toggle.

Added `enabledByProfile` to the FIPS state to track whether FIPS was
set by a profile. When locked, the helper text explains that FIPS is
required by the selected profile.
JIRA: HMS-10579

### DIFF
--- a/src/Components/CreateImageWizard/steps/Oscap/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Oscap/index.tsx
@@ -33,6 +33,7 @@ import {
   useGetOscapCustomizationsQuery,
   useGetOscapProfilesQuery,
 } from '@/store/api/backend';
+import { useSecuritySummary } from '@/store/api/backend/hooks';
 import { usePoliciesQuery } from '@/store/api/compliance';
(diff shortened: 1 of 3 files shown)


src/store/slices/wizard/index.ts
```diff
@@ -1724,8 +1726,12 @@ export const wizardSlice = createSlice({
         (_, index) => index !== action.payload,
       );
     },
-    changeFips: (state, action: PayloadAction<boolean>) => {
-      state.fips.enabled = action.payload;
+    changeFips: (
```
kingsleyzissou: I'm not sure this is the approach we should go with. There is the `useSecuritySummary` hook and we can just extend that to return whether or not fips is required and call that hook in the index hook where we implement the fips toggle

kingsleyzissou (approved):
Perfect thanks 🫶

PR #4387, opened 2026-04-30
Wizard: visually split required and optional packages

kingsleyzissou:
## Summary
OpenSCAP-required packages are now visually distinguished from user-added packages in the packages table, making it clear which packages cannot be removed.

## Changes
- Required packages render at the top of the table with a disabled remove button, tooltip, and lock icon; user-added packages appear below with the standard remove button
- `RemovePackageButton` extended with an `isRequired` prop that disables the button and swaps to a lock icon, matching the existing `RemoveRepositoryButton` pattern
- `RemoveRepositoryButton` updated to also show a lock icon when disabled for visual consistency across wizard steps
- Added shared `createOscapHandler` test utility for mocking OpenSCAP customizations API responses
- Three new tests covering DOM ordering, disabled button state, and fallback behavior without an OpenSCAP profile

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/components/PackagesTable.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/components/PackagesTable.tsx
@@ -12,6 +12,7 @@ import {
 } from '@patternfly/react-table';
 import { orderBy } from 'lodash';
 
+import { useSecuritySummary } from '@/store/api/backend';
 import { ApiRepositoryCollectionResponseRead } from '@/store/api/contentSources';
 import { useAppDispatch, useAppSelector } from '@/store/hooks';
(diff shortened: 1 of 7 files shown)


regexowl (approved):
Works great, thanks!

Noticed one small bug - the required packages can be removed if searched for and their option is clicked in the search. But it could be a tiny follow up, so not blocking on that.

kingsleyzissou:
> Works great, thanks!
> 
> Noticed one small bug - the required packages can be removed if searched for and their option is clicked in the search. But it could be a tiny follow up, so not blocking on that.

Oh ya I noticed that too, I was on the fence about whether to solve it here or a follow up. I'll tackle that next

PR #4388, opened 2026-04-30
Wizard: Add breathing room between steps

regexowl:
This adds margin above dividers between steps to give the user a bit more visual separation between the different topics.


mgold1234 (approved):
/lgtm

PR #4389, opened 2026-04-30
Wizard: add installer to image mode (HMS-10601)

ksiekl:
The new installer for image mode just dropped!
I have two things that I'd like opinions on :arrow_down: 
image

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
@@ -24,6 +24,7 @@ import { selectIsOnPremise } from '@/store/slices';
 import {
   addImageType,
   changeImageTypes,
+  changeIsoPayloadReference,
   changeRegistrationType,
   reinitializeAws,
   reinitializeAzure,
@@ -33,6 +34,7 @@ import {
   selectDistribution,
   selectImageTypes,
   selectIsImageMode,
+  selectIsoPayloadReference,
 } from '@/store/slices/wizard';
 import { useFlag } from '@/Utilities/useGetEnvironment';
 
@@ -123,6 +125,31 @@ const TargetEnvironment = () => {
       dispatch(changeImageTypes([environments[0]]));
     }
   }, [isImageMode, environments, dispatch]);
+
+  const isoPayloadReference = useAppSelector(selectIsoPayloadReference);
+  useEffect(() => {
+    if (!isImageMode || !environments.includes('bootable-container-iso')) {
+      return;
+    }
+    const entry = bootcDistributions?.find(
+      (d) => d.type === 'bootable-container-iso' && d.distro === distribution,
+    );
+    const refs = entry?.iso_payload_references;
+    if (!refs || refs.length === 0) {
+      return;
+    }
+    if (isoPayloadReference && refs.includes(isoPayloadReference)) {
+      return;
+    }
+    dispatch(changeIsoPayloadReference(refs[0]));
+  }, [
+    isImageMode,
+    environments,
+    bootcDistributions,
+    distribution,
+    isoPayloadReference,
+    dispatch,
+  ]);
 
   const isOnlyNetworkInstallerSelected =
     environments.length === 1 && environments.includes('network-installer');
@@ -536,6 +563,25 @@ const TargetEnvironment = () => {
               }}
             />
           ))}
+        {miscFormats.includes('bootable-container-iso') && isImageMode && (
+          <Radio
+            className='pf-v6-u-mb-sm pf-v6-u-ml-lg'
+            id='radio-bootable-container-iso'
+            name='target-environment'
+            label='Container installer (.iso)'
+            aria-label='Container installer'
+            description={
+              <Content component='small' style={{ maxWidth: TEXT_WRAP_WIDTH }}>
+                Install a bootable container to bare metal or virtual machines
+                using the guided Anaconda installer.
+              </Content>
+            }
+            isChecked={environments.includes('bootable-container-iso')}
+            onChange={() =>
+              handleSelectSingleEnvironment('bootable-container-iso')
+            }
+          />
+        )}
         {miscFormats.includes('network-installer') &&
           (isImageMode ? (
             <Radio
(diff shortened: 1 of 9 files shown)


src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
```diff
@@ -540,6 +567,32 @@ const TargetEnvironment = () => {
...
+            className='pf-v6-u-mb-sm pf-v6-u-ml-lg'
+            id='radio-bootable-container-iso'
+            name='target-environment'
+            label='Container installer (.iso)'
+            aria-label='Container installer'
+            description={
+              <Content component='small' style={{ maxWidth: TEXT_WRAP_WIDTH }}>
+                An installable ISO for deploying a bootable container to bare
+                metal or virtual machines.
+              </Content>
```
ksiekl, line 577: I'm not really vibing with this description but I didn't come up with anything better.... any ideas?

src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
```diff
@@ -540,6 +567,32 @@ const TargetEnvironment = () => {
...
+            }
+            isChecked={environments.includes('bootable-container-iso')}
+            onChange={() =>
+              handleSelectSingleEnvironment('bootable-container-iso')
+            }
+            body={
+              environments.includes('bootable-container-iso') ? (
+                <FormGroup label='Embedded OS image'>
+                  {isoPayloadReference}
+                </FormGroup>
```
ksiekl: This is currently just one value, that's always used as a payload for this particular bootc image type. Eventually, it could be a selector, and the user could pick from multiple options. But right now, they don't and I wonder if we should show this to the user at all.

src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
lucasgarfield: Lucas:

I ran this through the Patternfly MCP with some additional context, here's what it came up with, I think it is solid but it made me realize that we may want to align all the other image descriptions with the same reasoning.

Opus 4.6:

Suggested revision

  > Install a bootable container to bare metal or virtual machines using the guided Anaconda installer.  

What's good: It's concise, uses plain language, and correctly uses sentence case.

  What could improve (per PatternFly guidelines):

  1. Lead with what the user gets, not what the artifact is. PatternFly says to lead with the benefit and use action-oriented language. Right now it describes the file format ("an installable
   ISO") rather than what the user will accomplish.
  2. Use second person. PatternFly strongly prefers "you/your" to center the user. The current draft is impersonal.
  3. Mention the attended installer. You flagged this as key missing context — the Anaconda installer is what distinguishes this from other deployment methods, and it's what the user will
  actually experience.
  4. Sentence fragment punctuation. If this is a fragment (no subject), no trailing period. If it's a full sentence, it needs one. Currently it's a fragment with no period, which is fine —
  just be consistent with the other descriptions.

  This version:
  - Leads with the action ("Install") — active voice, action-oriented
  - Centers the user (imperative mood implies "you")
  - Mentions Anaconda as the distinguishing detail
  - Uses "guided" instead of "attended" — plainer language for users who may not know the attended/unattended distinction
  - Stays concise (one line)

src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
lucasgarfield: I didn't run the code but in the screenshot, the container that appears here is `rhel-10-qcow2:latest` even though the iso has been selected. 

That might just be a little bug but it makes me wonder - is it really true that there is only one value that will be used as a payload? It seems like it would be `rhel-10-qcow2:latest` for the qcow and `rhel-10-installer:latest` for the installer.  Or `rhel-10-ec2:latest` for the AMI, etc... Basically whatever we have in github.com/osbuild/bootc-foundry/

I feel like this should actually not be indented but rather live somewhere visible (the top?) and always be displayed.

ksiekl: Although the iso is selected, it has another reference `isoPayloadReference` in it, so that's what you are seeing. It's what the frontend gets from the backend, not a bug. Also, we decided to hide this embedded value for the user anyways, so I'm deleting this little section

src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
ksiekl: Thanks, this sounds good! And yes, we probably should check out the other descriptions as well

src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
lucasgarfield: Weird, I still don't quite understand why the isoPayloadReference is for the qcow2 but I'll take a look at the backend - I must be missing something. In any case, and given our tight timeline, just deleting the section makes good sense to me.

src/store/slices/wizard/index.ts
```diff
@@ -831,6 +843,9 @@ export const wizardSlice = createSlice({
     },
     changeImageTypes: (state, action: PayloadAction<ImageTypes[]>) => {
       state.imageTypes = action.payload;
+      if (!action.payload.includes('bootable-container-iso')) {
```
croissanne, line 848: this feels a little bit magic, maybe a small comment explaining why?

croissanne (dismissed):
LGTM, small suggestion as I do think dropping the payload inside of the wizard store is a bit magic.

src/store/slices/wizard/index.ts
ksiekl: Added the comment :smile:

croissanne (approved):
perfect thank you

lzap:
It was removed again.

ksiekl:
That snapshot test is never going to pass I think, we have to figure that one out first

PR #4392, opened 2026-05-01
Wizard/Image output: Replace arch for image mode with disabled select (HMS-10579)

regexowl:
The arch was previously just a simple text for image mode, now it uses the `<ArchSelect>` component and disables it.


kingsleyzissou (approved):
Nice

PR #4383, opened 2026-04-30
Wizard/ImageOutput: filter supported bootc environments by distro (HMS-10618)

croissanne:
Not all distributions support all bootc image types. For instance Hummingbird only supports the guest-image. Filter out unsupported target environments by filtering on distribution.

PR #4402, opened 2026-05-04
Wizard: save bootc reference across edits of bp

ksiekl:
The bootc reference was being wiped when editing the bp, now the state stays the same across edits.

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
@@ -94,17 +94,13 @@ const TargetEnvironment = () => {
   const isFetching = isArchFetching || isBootcFetching;
   const isError = isArchError || isBootcError;
 
-  const supportedEnvironments = useMemo(
-    () =>
-      skipArchitectures
-        ? [
-            ...new Set(
-              bootcDistributions?.map((d) => d.type) ?? EMPTY_ENVIRONMENTS,
-            ),
-          ]
-        : archEnvironments,
-    [skipArchitectures, bootcDistributions, archEnvironments],
-  );
+  const supportedEnvironments = useMemo(() => {
+    if (!skipArchitectures) return archEnvironments;
+
+    return [
+      ...new Set(bootcDistributions?.map((d) => d.type) ?? EMPTY_ENVIRONMENTS),
+    ];
+  }, [skipArchitectures, bootcDistributions, archEnvironments]);
 
   const dispatch = useAppDispatch();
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
```diff
@@ -94,17 +96,18 @@ const TargetEnvironment = () => {
...
-    [skipArchitectures, bootcDistributions, archEnvironments],
-  );
+  const supportedEnvironments = useMemo(() => {
+    if (!skipArchitectures) return archEnvironments;
+
+    const selectedDistro = bootcDistributions?.find(
+      (d) => d.reference === imageSource,
+    );
+    const filtered = selectedDistro
+      ? bootcDistributions?.filter((d) => d.name === selectedDistro.name)
```
croissanne: This won't work currently I think, you'll get the same error I'm getting here https://github.com/osbuild/image-builder-frontend/pull/4383 (which is the same idea but slightly different approach).

I'm going to fix this in https://github.com/osbuild/image-builder-crc/pull/1817. So I'd just drop this filtering from the PR for now.

ksiekl: The test currently with this filtering passes but you're right that since we made the filtering in the crc already, we probably shouldn't double it in the frontend as well

src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
```diff
@@ -128,6 +124,14 @@ const TargetEnvironment = () => {
...
   }, [isImageMode, environments, dispatch]);
 
+  useEffect(() => {
+    if (!isImageMode || environments.length === 0) return;
+    const valid = environments.filter((e) => supportedEnvironments.includes(e));
+    if (valid.length < environments.length) {
+      dispatch(changeImageTypes(valid));
+    }
+  }, [isImageMode, environments, supportedEnvironments, dispatch]);
+
```
croissanne: Hmm now there's 2 useEffects which will change the image types on environemnts, isImageMode change. 

Also I'm not actually sure what this does. If the targets in the blueprint are no longer supported in the api this removes them? This is a situation that shouldn't occur imo, as this would mean a breaking change from the API.

ksiekl: Ahh sorry :facepalm: yes, this was supposed to be gone as well

croissanne (approved):

croissanne:
thank you!

PR #4408, opened 2026-05-05
Wizard: add error banner to users in image mode (HMS-10639)

ksiekl:
At least one user is necessary in Image mode, this commit adds a banner with that alert if a user tries to continue through the wizard without any user set.

I didn't want to have an error shown immediately as a user would click "Image mode". So now, the User step is marked as required with an asterix and if someone tries to click "Next" without one, only then will the alert banner appear:
spacing


ksiekl:
Closing this since we actuallly want to make Users optional

PR #4411, opened 2026-05-05
playwright: use static repository for RepeatableBuild test

kingsleyzissou:
Replace per-run repository creation with a persistent static repo that is created once and reused across test runs. This eliminates the backend indexing race condition that caused flaky search results and avoids concurrency conflicts between parallel test runs.

Add a generic ensureRepositoryExists helper that searches by name and creates the repo only if it doesn't exist, handling concurrent creation gracefully.


croissanne (dismissed):

croissanne (dismissed):

kingsleyzissou:
Repeatable build now passes. I had to skip the repo checks.
Re-running the tests because there were some other flakes

regexowl (approved):
Thank you!

PR #4417, opened 2026-05-06
ImagesTable: Show OS label for hosted bootc images

ochosi:
The OS column previously showed "Image mode" for all bootc composes from the hosted service because the reference parser only recognized the official `registry.redhat.io` references we use on premises. This commit adds a second parser for the hosted `quay.io/…/image-builder-bootc-foundry` pattern so RHEL 10 and Fedora Hummingbird builds display meaningful OS labels in the service.

image


ochosi:
We should probably go with https://github.com/osbuild/image-builder-frontend/pull/4419 as a more light-weight implementation, avoiding all the ref parsing (I just wasn't sure initially if we could rely on the distribution always being set).

ochosi:
Closed in favor of #4419

PR #4418, opened 2026-05-06
playwright: Update selectDistro

regexowl:
selectDistro can fail since it's looking for an exact match instead of a partial match. The distribution field continues with "Full support..." description.


regexowl:
Nope, different timeout

PR #4420, opened 2026-05-06
Wizard: Add tooltip to disabled network installer checkbox (HMS-10579)

mgold1234:
When other image types are selected, the network installer checkbox is disabled without explanation. Add a tooltip explaining that it cannot be combined with other image types.

Screenshot 2026-05-18 at 12 33 20
JIRA: (HMS-10579)

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
@@ -563,7 +563,9 @@ const TargetEnvironment = () => {
               label={createLabelWithTooltip(
                 'Network',
                 'Installer (.iso)',
-                'This is a lightweight image that differs from a standard "full" ISO by requiring an active network connection to pull the latest software directly from package repositories, as no OS packages are stored locally on the image.',
+                isOtherEnvironmentSelected
+                  ? 'Network installer cannot be combined with other image types'
+                  : 'This is a lightweight image that differs from a standard "full" ISO by requiring an active network connection to pull the latest software directly from package repositories, as no OS packages are stored locally on the image.',
               )}
               aria-label='Network installer checkbox'
               isChecked={environments.includes('network-installer')}


src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
```diff
@@ -68,6 +68,46 @@ const createLabelWithTooltip = (
...
+  isOtherEnvironmentSelected,
+  onToggle,
+}: NetworkInstallerCheckboxProps) => {
+  const checkbox = (
+    <Checkbox
+      className='pf-v6-u-mb-sm pf-v6-u-ml-lg'
+      id='checkbox-network-installer'
+      isLabelWrapped
+      name='Network - Installer'
+      label={createLabelWithTooltip(
```
kingsleyzissou: First of all, I don't think we should create a new component for this, we can just do it all in-line.

Secondly, maybe we could just use the exiting `createLabelWithTooltip` function and change the contents depending on whether the other environments are selected or not.

```typescript
label={
  !isOtherEnvironmentSelected
    ? createLabelWithTooltip(...) :
    : createLabelWithTooltip(
      'Network', 
      'Installer (.iso)', 
      'Network installer cannot be combined with other image types'),
}
```

kingsleyzissou (changes requested):
I think this is quite a big diff for a small change, I made a suggestion inline

src/Components/CreateImageWizard/steps/Oscap/components/ProfileSelector.tsx
```diff
@@ -170,11 +168,11 @@ const ProfileSelector = ({
     }
   }, [filterValue, profileDetails, isOpen]);
 
-  const handleToggle = () => {
```
kingsleyzissou, line 173: I'm not sure if the changes to this file were intentional. I think they're unrelated to what this PR is trying to do.

kingsleyzissou (commented):
The changes to the tooltip look good. Can we just drop the unrelated changes and then we can get this merged? 🙏

src/Components/CreateImageWizard/steps/ImageOutput/components/TargetEnvironment.tsx
```diff
@@ -563,7 +563,9 @@ const TargetEnvironment = () => {
               label={createLabelWithTooltip(
                 'Network',
                 'Installer (.iso)',
-                'This is a lightweight image that differs from a standard "full" ISO by requiring an active network connection to pull the latest software directly from package repositories, as no OS packages are stored locally on the image.',
+                isOtherEnvironmentSelected
```
kingsleyzissou, line 566: Nice, this is even cleaner than my suggestion!

kingsleyzissou (approved):
Amazing, thanks!

PR #4421, opened 2026-05-06
Wizard/FSC: Bump default root partition size to 10 GiB in advanced mode

ochosi:
We want consistent defaults between basic and advanced partitioning, but also avoid failing builds because of a root partition that is too small.


regexowl (approved):
:rocket:

kingsleyzissou (approved):

lzap (approved):

regexowl:
Admin merging, the test failure are just flakes

PR #4422, opened 2026-05-06
fix: Allow repo dropdown to open on input when empty

jlsherrill:
Removes inputValue condition from onInputClick so the dropdown opens when clicking the search box, even when empty, allowing users to see unfiltered repositories immediately.

Fixes HMS-10642


jlsherrill:
image

regexowl (approved):
Works nicely, thanks!

PR #4424, opened 2026-05-07
playwright: Safeguard target selecting in OpenSCAP test

regexowl:
This replaces direct use of `selectDistro` with its safeguarded alternative `fillInImageOutput` to resolve issue with playwright trying to check selectors that didn't render yet.

PR #4427, opened 2026-05-07
Wizard: hide all customizations for bootable-container-iso

croissanne:
This enables the mocked distribution details for image mode. At most filesystem and users is allowed, but any image type can choose to support less. In the case of the bootable container iso, no customizations should be supported.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/index.tsx
@@ -63,11 +63,13 @@ const ReviewStep = () => {
         restrictions={restrictions}
         oscapPackages={security.packages}
       />
-      <AdvancedSettingsOverview
-        restrictions={restrictions}
-        oscapKernelArgs={security.kernel.append}
-        oscapServices={security.services}
-      />
+      {Object.values(restrictions).some((r) => !r.shouldHide) && (
+        <AdvancedSettingsOverview
+          restrictions={restrictions}
+          oscapKernelArgs={security.kernel.append}
+          oscapServices={security.services}
+        />
+      )}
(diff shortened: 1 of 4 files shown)


regexowl:
Went through the code and tested the changes manually - looks good to me!

Stage is wildly uncooperative today so the hosted tests may still be failing, but looking at the previous failures those should be just flakes.

Opened two fixes, that should help with the stability of the tests:
- https://github.com/osbuild/image-builder-frontend/pull/4424
- https://github.com/osbuild/image-builder-frontend/pull/4429

ochosi:
Quickly tried this out and LGTM, only small wrinkle (that I wouldn't block on for now) is that an empty Advanced settings section is shown on the review step (which is probably something to fix unrelated to this PR).
image

kingsleyzissou (commented):
Changes look good. Maybe we should hide the advanced settings if there is nothing selected.

This is the only image type where this would be the case though... and it might be more work than it seems

croissanne:
> Changes look good. Maybe we should hide the advanced settings if there is nothing selected.
> 
> This is the only image type where this would be the case though... and it might be more work than it seems



> Changes look good. Maybe we should hide the advanced settings if there is nothing selected.
> 
> This is the only image type where this would be the case though... and it might be more work than it seems

naw i have somtehing locally for that

croissanne:
you mean on the review page right?

regexowl (approved):
🚀 :shipit:

kingsleyzissou (approved):

regexowl:
The Oscap failure is expected and unrelated, everything else is green - admin merging

PR #4429, opened 2026-05-08
playwright: Update image mode to work with RHEL source

regexowl:
AWS is unsupported for Fedora Hummingbird, but the tests were selecting first option to test the AWS on, which is by default Fedora Hummingbird.


kingsleyzissou (approved):

regexowl:
Admin merging, the only failure was OpenSCAP that is addressed in https://github.com/osbuild/image-builder-frontend/pull/4424

PR #4432, opened 2026-05-10
Wizard: Match GCP principal input width to account type select (HMS-10579)

mgold1234:
The principal/domain input field was wider than the account type dropdown. Constrain its width to 50% to align with the select.
Screenshot 2026-05-10 at 19 39 56

JIRA: HMS-10579

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/Gcp/index.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/Gcp/index.tsx
@@ -92,6 +92,7 @@ const Gcp = () => {
       <FormGroup
         label={accountType === 'domain' ? 'Domain' : 'Principal'}
         isRequired
+        style={{ maxWidth: '50%' }}
       >
         <ValidatedInput
           aria-label='google principal'


src/Components/CreateImageWizard/steps/ImageOutput/components/Gcp/index.tsx
```diff
@@ -92,6 +92,7 @@ const Gcp = () => {
       <FormGroup
         label={accountType === 'domain' ? 'Domain' : 'Principal'}
         isRequired
+        className='pf-v6-u-w-50'
```
avitova: This is what [Azure uses](https://github.com/osbuild/image-builder-frontend/blob/0a1f40206653c0f85ed4f5ab734ff736a64d0d2b/src/Components/CreateImageWizard/steps/ImageOutput/components/Azure/components/TenantIdInput.tsx#L18), and it works for me locally, maybe it would be wiser so that 
```suggestion
        style={{ maxWidth: '50%' }}
```

avitova (dismissed):
See my comment, but this fixes the issue too, so I don't want to block your PR on this.

regexowl (approved):
Thanks!

PR #4434, opened 2026-05-11
Wizard/Packages: Fix recommendations tracking

regexowl:
There's and issue where "Recommended Package Added" event fired without "Package Recommendations Shown" in edit mode.

This fixes the issue by updating the dependency of the "shown" event from `transformedRecommendations.length` to `transformedRecommendations`. This way the useEffect isn't ignored when the API returns a different set of recommendations with the same count of pkgs as the previous one (which can easily happen as the rec limit is set to 5).


mgold1234 (approved):
/lgtm

PR #4437, opened 2026-05-12
Wizard: fix validation in AAP step

ksiekl:
The default state consisted of various errors, now the behavior is more controlled
aap

### DIFF
--- a/src/Components/CreateImageWizard/utilities/useValidation.tsx
+++ b/src/Components/CreateImageWizard/utilities/useValidation.tsx
@@ -303,13 +303,11 @@ export function useAAPValidation(): StepValidation {
   if (!callbackUrl && !hostConfigKey && !tlsCertificateAuthority) {
     return { errors: {}, disabledNext: false };
   }
-  if (!callbackUrl || callbackUrl.trim() === '') {
-    errors.callbackUrl = 'Ansible Callback URL is required';
-  } else if (!isValidUrl(callbackUrl)) {
+  if (callbackUrl && !isValidUrl(callbackUrl)) {
     errors.callbackUrl = 'Callback URL must be a valid URL';
   }
 
-  if (!hostConfigKey || hostConfigKey.trim() === '') {
+  if (hostConfigKey !== undefined && hostConfigKey.trim() === '') {
     errors.hostConfigKey = 'Host Config Key is required';
   }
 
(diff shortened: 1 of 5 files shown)


src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -305,11 +305,12 @@ export function useAAPValidation(): StepValidation {
   }
   if (!callbackUrl || callbackUrl.trim() === '') {
     errors.callbackUrl = 'Ansible Callback URL is required';
-  } else if (!isValidUrl(callbackUrl)) {
+  } else if (/^https?:\/\/.*\./.test(callbackUrl) && !isValidUrl(callbackUrl)) {
```
regexowl: This was slightly confusing behaviour - the insecure checkbox turns disabled after `http://` is typed in. We could probably validate and disable the checkbox earlier on `http:`.

Even though the behaviour is technically correct, disabling right after it's obvious the user doesn't want `http` might make things a bit clearer.

regexowl: The same with hiding the CA input - it isn't hidden until the full `http://` is typed in and the insecure checkbox is checked.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -305,11 +305,12 @@ export function useAAPValidation(): StepValidation {
...
     errors.callbackUrl = 'Callback URL must be a valid URL';
   }
 
-  if (!hostConfigKey || hostConfigKey.trim() === '') {
+  const urlIsValid = !!callbackUrl && isValidUrl(callbackUrl);
+  if (urlIsValid && (!hostConfigKey || hostConfigKey.trim() === '')) {
```
regexowl: Unsure about handling of the should/shouldn't show validation errors here. 

When I type any value into host config key field and remove it right away I won't see any validation error (even though the key is required) unless I've previously filled in a valid URL.

Also the URL error hiding and showing based on interaction with the key field feels weird.

https://github.com/user-attachments/assets/87b00883-ca19-4bdf-a853-f3952b6a7555

regexowl (commented):
Thanks for looking into this! Added a few comment about when the errors are being triggered.

src/Components/CreateImageWizard/utilities/useValidation.tsx
ksiekl: Okay, here is the thing:
1. The AAP is optional, however, if one field is filled out, the rest should be as well. Therefore, it causes cross-coupling with the other fields in this step. I made it look like each field is validated independently, but it complicated the code, and it's now quite ugly with a bunch of `ref` workarounds.
 2. I made the host config key to be quietly deleted if it is the only field that the user fills out, since it doesn't make any sense to have it if we don't have the URL.
 3. There is a logical problem with the footer - I think that since each step has its own footer, it causes problems. Right now, the moment a user selects any target env, the Next and Review buttons are enabled, and the user can send a partially filled-out AAP step :cry: 
 
 All in all, it's quite a complicated mess... Here is how the step interacts now:
 
aap

I think it's cleaner and smoother, but at what cost :sweat: I'm really crashing out over this one, any ideas?

regexowl:
> I think it's cleaner and smoother, but at what cost 😓 I'm really crashing out over this one, any ideas?

Just thinking out loud - maybe we could separate the validation for the individual fields into their own functions and then evaluate whether the Next button should be disabled by combining them? But that also doesn't sound ideal...

Would have to look into this more, unfortunately won't get to it today.

ksiekl:
Yes, I was thinking that validation could do with some redesign. However, this task is too big given that I am leaving for a long time, so I'm going to close this PR and keep it open in the SPUR doc. If no one checks it out, I can look at this again when I come back :smile:

PR #4438, opened 2026-05-12
deps: Remove lodash dependency

regexowl:
The dependency is rather large and had several vulnerabilities recently. We should be able to make do without it.


croissanne (approved):
very nice

PR #4439, opened 2026-05-13
playwright: Update repo filter locator

regexowl:
This updates the locator to update the tests after changes on the content side.


ochosi (approved):
Confirmed the change between stage and prod. Stage already has the updated `aria-label` values, hence the boot tests fail.

This seems to be the pull request in question. https://github.com/content-services/content-sources-frontend/pull/989

What I noticed there is that the content team doesn't use `getByRole` with the name but `getByPlaceholderText` instead. Might be worth discussing with @dominikvagner to standardize on the same method.

PR #4440, opened 2026-05-13
deps: Remove `classnames`

regexowl:
I believe this dep isn't needed and it was also last updated 2 years ago. Might as well drop it.


regexowl:
`@redhat-cloud-services/frontend-components` has `classnames` listed as a dependency, but it should be safe to remove it from our direct deps.

avitova (approved):

PR #4442, opened 2026-05-14
Deps: Update lodash

lucasgarfield:
Resolves CVE-2025-13465 and CVE-2026-4800.


lucasgarfield:
Note that this is against the `rhel-10.2` branch, not main. Hoping to get this into 10.2 before GA (which is now delayed until at least May 19).

kingsleyzissou (commented):
Wait is this for cockpit? It might be a non-issue. We closed a similar cve before because this just runs locally

lucasgarfield:
@kingsleyzissou you're correct, this should be a non-issue, however I'd like to use this as an opportunity to get practice doing a point release and z-stream for educational purposes.

kingsleyzissou (approved):

PR #4443, opened 2026-05-14
Playwright: Fix accept cookies

lucasgarfield:
The addLocatorHandler for the TrustArc cookie consent wasn't reliably firing when targeting buttons inside .contentFrame(), so trigger on the iframe element itself and click the accept button inside the handler.


lucasgarfield:
I opened this hoping to resolve an issue on my 10.2 branch pull request - it turns out to not be needed for main. :facepalm:

PR #4444, opened 2026-05-18
Wizard: Filter out empty users from review step (HMS-10691)

mgold1234:
When a user was added and then removed, an empty user object remained in the Redux store. 
The review step displayed it because it only checked users.length === 0, showing a blank row with no meaningful data.

Filter out users that have no name, password, SSH key, or stored password before rendering the review section.

ISSUE IN GITHUB: 4423
JIRA: 4423

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/components/Users.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/components/Users.tsx
@@ -3,13 +3,13 @@ import React from 'react';
 import { Flex, Truncate } from '@patternfly/react-core';
 
 import { useAppSelector } from '@/store/hooks';
-import { selectUsers } from '@/store/slices';
+import { selectNonEmptyUsers } from '@/store/slices/wizard';
 
 import { FlexColumn, ReviewGroup, StatusItem } from '../../shared';
 import { Hideable } from '../../types';
 
 export const Users = ({ shouldHide }: Hideable) => {
-  const users = useAppSelector(selectUsers);
+  const users = useAppSelector(selectNonEmptyUsers);
 
   if (shouldHide || users.length === 0) {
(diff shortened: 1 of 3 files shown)


src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/components/Users.tsx
```diff
@@ -3,15 +3,15 @@ import React from 'react';
...
 import { useAppSelector } from '@/store/hooks';
-import { selectUsers } from '@/store/slices';
+import { selectNonEmptyUsers } from '@/store/slices/wizard';
 
 import { FlexColumn, ReviewGroup, StatusItem } from '../../shared';
 import { Hideable } from '../../types';
 
 export const Users = ({ shouldHide }: Hideable) => {
-  const users = useAppSelector(selectUsers);
+  const nonEmptyUsers = useAppSelector(selectNonEmptyUsers);
```
kingsleyzissou:
```suggestion
const users = useAppSelector(selectNonEmptyUsers);
```

kingsleyzissou: `nonEmptyUsers` gets a bit confusing, `users` should be enough and the selector tells us what is happening

src/store/slices/wizard/index.ts
```diff
@@ -559,6 +559,16 @@ export const selectUsers = (state: RootState) => {
   return state.wizard.users;
 };
 
+export const selectNonEmptyUsers = (state: RootState) => {
```
kingsleyzissou, line 562: Yes! Love this 🫶

kingsleyzissou (commented):
One minor comment, but this looks great

kingsleyzissou (approved):
Great stuff

PR #4445, opened 2026-05-18
tests: Fix Compliance distro selector

tkoscieln:
SSIA

### DIFF
--- a/playwright/BootTests/Compliance/ComplianceCIS.boot.ts
+++ b/playwright/BootTests/Compliance/ComplianceCIS.boot.ts
@@ -51,7 +51,7 @@ test('Compliance step integration test - CIS', async ({ page, cleanup }) => {
   await test.step('Create a Compliance policy', async () => {
     await navigateToCompliance(page);
     await page.getByRole('button', { name: 'Create new policy' }).click();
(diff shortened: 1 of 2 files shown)


mgold1234 (approved):
/lgtm

avitova (commented):
It almost seems like the [getByRole](https://github.com/osbuild/image-builder-frontend/blob/0a1f40206653c0f85ed4f5ab734ff736a64d0d2b/playwright/helpers/targetChooser.ts#L55) for distros stopped working? Should we use getByText?
Because the failure is for different reasons than what you are fixing, it seems.

tkoscieln:
> It almost seems like the [getByRole](https://github.com/osbuild/image-builder-frontend/blob/0a1f40206653c0f85ed4f5ab734ff736a64d0d2b/playwright/helpers/targetChooser.ts#L55) for distros stopped working? Should we use getByText? Because the failure is for different reasons than what you are fixing, it seems.

That's a different issue https://github.com/patternfly/patternfly-react/issues/12422 and it has to be fixed in Patternfly https://github.com/patternfly/patternfly-react/pull/12424. Since we started using Modal, I have discovered a bug where the Modal sometimes sets `aria-hidden` to `true` for dropdowns, which is what PW uses for navigation.

PR #4457, opened 2026-05-25
Wizard: Disable language remove button when only one remains (HMS-10609)

mgold1234:
Users could remove all languages in the Locale step, leaving the list empty. Disable the remove button when only one language remains to ensure at least one language is always selected. The button re-enables once a second language is added.
JIRA: HMS-10609

### DIFF
--- a/src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
+++ b/src/Components/CreateImageWizard/steps/Locale/components/LanguagesDropDown.tsx
@@ -122,6 +122,9 @@ const LanguagesDropDown = () => {
 
   const handleRemoveLanguage = (language: string) => {
     dispatch(removeLanguage(language));
+    if (languages.length <= 1) {
+      setShowNewRow(true);
+    }
   };
 
   return (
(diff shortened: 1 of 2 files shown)


regexowl (commented):
Ok, here we'll need to balance two things:
1. the input (and the only language) shouldn't be probably removable as it indeed looks a bit strange
2. no languages selected is a valid choice and we need to allow it

Now when the user selects a few languages, but then changes their mind and decides to remove them all again, they will be stuck with one non-removable language and no way to un-select the language, select "None" or smth like that.

mgold1234:
@regexowl wdyt about this solution -  when the user removes the last remaining language, instead of leaving an empty list with no way to add a language back, we will automatically show an empty selection row with the "Select a language" placeholder. This lets users either pick a new language or leave the selection empty (which is a valid choice). The language is removed from the store, so 0 languages is properly supported.

regexowl:
> @regexowl wdyt about this solution - when the user removes the last remaining language, instead of leaving an empty list with no way to add a language back, we will automatically show an empty selection row with the "Select a language" placeholder. This lets users either pick a new language or leave the selection empty (which is a valid choice). The language is removed from the store, so 0 languages is properly supported.

Sounds good to me!

regexowl (approved):
Nice!

PR #4458, opened 2026-05-25
Fix noisy test warnings in wizard steps (HMS-10715)

kingsleyzissou:
## Summary
Fix noisy test warnings and resolve CI Playwright install hang caused by a Node 24.x regression.

## Changes
- Pin Node.js to `24.15.0` in CI runners to work around a regression in newer Node 24.x that causes `npx playwright install` to hang after downloading the browser
- Memoize `selectNonEmptyUsers` with `createSelector` to prevent unnecessary rerenders that caused act warnings in tests
- Replace `user.click()` with `clickWithWait()` in Locale and Oscap tests to resolve act warnings from unmicrotasked state updates

### DIFF
--- a/.github/actions/setup-node-cached/action.yml
+++ b/.github/actions/setup-node-cached/action.yml
@@ -7,7 +7,7 @@ runs:
     - name: Set up Node.js
       uses: actions/setup-node@v6
       with:
-        node-version: 24
+        node-version: 24.15.0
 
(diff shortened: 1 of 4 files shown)


ochosi (approved):

ochosi:
Sucks that we have to pin the Node version but that seems to do the trick...

PR #4468, opened 2026-05-27
Packit: Only propose downstream releases for main

lucasgarfield:
If we need to do a point release for a RHEL Z-stream, we don't want to propose downstream releases for Fedora.


lucasgarfield:
I can't think of a reason why we would ever propose releases for anything other than main?

ondrejbudai (approved):

kingsleyzissou (approved):

PR #4426, opened 2026-05-07
test: Clean up test output

regexowl:
This cleans up test input by:
- removing nested forms from the code base
- resolving "missing act()" warnings

PR #4469, opened 2026-05-27
vitest: Clean up the test output (HMS-10742)

regexowl:
This updates tests to clean up the test output and fixes the ImagesTable test in `npm run test:cockpit`.

Rebased on https://github.com/osbuild/image-builder-frontend/pull/4426

JIRA: [HMS-10742](https://issues.redhat.com/browse/HMS-10742)

[HMS-10742]: https://redhat.atlassian.net/browse/HMS-10742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
@@ -1,12 +1,13 @@
-import React, { MutableRefObject, useEffect, useRef } from 'react';
+import React, { useEffect } from 'react';
 
 import {
+  Button,
   Content,
   ContentVariants,
   FormGroup,
-  FormGroupLabelHelp,
   Popover,
 } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
 
 import { ValidatedInput } from '@/Components/CreateImageWizard/ValidatedInput';
 import { CDN_PROD_URL } from '@/constants';
@@ -20,14 +21,9 @@ import {
   selectOrgId,
 } from '@/store/slices/wizard';
 
-const ManualRegistrationPopover = ({
-  ref,
-}: {
-  ref: MutableRefObject<null>;
-}) => {
+const ManualRegistrationPopover = () => {
   return (
     <Popover
-      triggerRef={ref}
       headerContent='About Activation Keys & Organization ID'
       position='right'
       minWidth='30rem'
@@ -46,9 +42,12 @@ const ManualRegistrationPopover = ({
         </Content>
       }
     >
-      <FormGroupLabelHelp
-        ref={ref}
+      <Button
+        icon={<HelpIcon />}
+        variant='plain'
+        size='sm'
         aria-label='About Activation Keys & Organization ID'
+        hasNoPadding
       />
     </Popover>
   );
@@ -58,8 +57,6 @@ export const ManualActivationKey = () => {
   const dispatch = useAppDispatch();
   const orgId = useAppSelector(selectOrgId);
   const activationKey = useAppSelector(selectActivationKey);
-  const orgIdRef = useRef(null);
-  const activationKeyRef = useRef(null);
 
   useEffect(() => {
     dispatch(changeServerUrl('subscription.rhsm.redhat.com'));
@@ -71,7 +68,7 @@ export const ManualActivationKey = () => {
       <FormGroup
         className='pf-v6-u-mb-md'
         label={'Activation key'}
-        labelHelp={<ManualRegistrationPopover ref={activationKeyRef} />}
+        labelHelp={<ManualRegistrationPopover />}
         isRequired
       >
         <ValidatedInput
@@ -88,7 +85,7 @@ export const ManualActivationKey = () => {
       </FormGroup>
       <FormGroup
         label={'Organization ID'}
-        labelHelp={<ManualRegistrationPopover ref={orgIdRef} />}
+        labelHelp={<ManualRegistrationPopover />}
         isRequired
       >
         <ValidatedInput
--- a/src/test/Components/ImagesTable/ImagesTable.test.tsx
+++ b/src/test/Components/ImagesTable/ImagesTable.test.tsx
@@ -32,9 +32,13 @@ describe('Images Table', () => {
     expect(headerCells[2]).toHaveTextContent('Updated');
     expect(headerCells[3]).toHaveTextContent('OS');
     expect(headerCells[4]).toHaveTextContent('Target');
-    expect(headerCells[5]).toHaveTextContent('Version');
-    expect(headerCells[6]).toHaveTextContent('Status');
-    expect(headerCells[7]).toHaveTextContent('Instance');
+    if (!process.env.IS_ON_PREMISE) {
+      expect(headerCells[5]).toHaveTextContent('Version');
+      expect(headerCells[6]).toHaveTextContent('Status');
+      expect(headerCells[7]).toHaveTextContent('Instance');
+    } else {
+      expect(headerCells[5]).toHaveTextContent('Status');
+    }
 
     const imageNameValues = mockComposes.map((compose) =>
       compose.image_name ? compose.image_name : compose.id,
--- a/src/test/fixtures/repositories.ts
+++ b/src/test/fixtures/repositories.ts
@@ -610,7 +610,7 @@ const generateLinks = (
 const generateFillerRepos = (num: number): ApiRepositoryResponse[] => {
   const repos = new Array(num).fill(undefined).map((_, i) => {
     return {
-      uuid: '9cf1d45d-aa06-46fe-87ea-121845cc6bbb',
+      uuid: `filler-repo-${i}`,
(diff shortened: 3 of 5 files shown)


regexowl:
This is mostly an idea. Usually some "missing act()" warnings sneak in without noticing and the output becomes a mess. If we fail on those, the problems could be addressed directly in the PR.

src/test/Components/ImagesTable/ImagesTable.test.tsx
```diff
@@ -10,7 +10,7 @@ describe('Images Table', () => {
   });
 
   const user = userEvent.setup();
-  test('render ImagesTable', async () => {
+  test.skipIf(process.env.IS_ON_PREMISE)('render ImagesTable', async () => {
```
regexowl: I don't think this ever worked since we mock the composes only for service and not on-prem. Checked older runs and the test was failing, but the run was marked as a successful one. Zero idea why.

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
```diff
@@ -20,14 +21,9 @@ import {
   selectOrgId,
 } from '@/store/slices/wizard';
 
-const ManualRegistrationPopover = ({
-  ref,
-}: {
-  ref: MutableRefObject<null>;
-}) => {
+const ManualRegistrationPopover = () => {
```
regexowl, line 24: The ref was causing issues in the test output. The popover implementation can be simplified without losing functionality.

regexowl: Check for example https://github.com/osbuild/image-builder-frontend/actions/runs/26509587349/job/78070585286 to see the error

src/test/fixtures/repositories.ts
```diff
@@ -609,8 +611,9 @@ const generateLinks = (
 
 const generateFillerRepos = (num: number): ApiRepositoryResponse[] => {
   const repos = new Array(num).fill(undefined).map((_, i) => {
+    const uuid = uuidv4();
     return {
-      uuid: '9cf1d45d-aa06-46fe-87ea-121845cc6bbb',
+      uuid,
```
regexowl: The filler repos were generated with a static uuid. Since we use it as a key, it needs to be unique.

src/Components/CreateImageWizard/steps/Repositories/components/Repositories.tsx
```diff
@@ -364,7 +364,7 @@ const Repositories = () => {
         ),
       );
     }
-  }, [templateUuid, reposInTemplate]);
+  }, [dispatch, isTemplateSelected, reposInTemplate]);
```
regexowl: The out-of-date dependencies were causing a loop, resulting in a "maximum depth reached" warning.

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
kingsleyzissou: Ya this was annoying me, was going to try and jump on this next, but seems like you beat me to it

src/Components/CreateImageWizard/steps/Registration/components/ManualActivationKey.tsx
```diff
@@ -46,9 +42,12 @@ const ManualRegistrationPopover = ({
...
-      <FormGroupLabelHelp
-        ref={ref}
-        aria-label='About Activation Keys & Organization ID'
+      <Button
+        icon={<HelpIcon />}
+        variant='plain'
+        size='sm'
+        aria-label='Credentials path info'
+        hasNoPadding
       />
```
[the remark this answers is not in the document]
regexowl: Fair point, fixed.

regexowl:
> * The change to `generateFillerRepos` now produces a fresh UUID with `uuidv4()` for each repo, which can introduce non-determinism into tests; consider using a deterministic ID based on the index (e.g. `filler-repo-${i}`) so test output and snapshots remain stable while still avoiding duplicate keys.

Also addressed.

src/test/Components/ImagesTable/ImagesTable.test.tsx
mgold1234: why does this test produce console errors in the on-premise environment? Would it be worth fixing the root cause rather than skipping the test entirely?

src/Components/CreateImageWizard/steps/Repositories/components/Repositories.tsx
```diff
@@ -364,7 +364,9 @@ const Repositories = () => {
         ),
       );
     }
-  }, [templateUuid, reposInTemplate]);
+    // This should run only when the template UUID changes
```
mgold1234: Would it be possible to restructure this to avoid the eslint-disable? For example, using a ref to capture reposInTemplate so the effect can safely include all its dependencies.

regexowl:
Sorry for the pushes, can't reproduce the unit test failure locally so it's a bit of a pain to debug...

src/test/mocks/cockpit/fsinfo.ts
```diff
@@ -5,7 +5,7 @@ import path from 'path';
 import { mockBlueprintIds } from '../../fixtures/blueprints';
 import { mockComposes } from '../../fixtures/composes';
 
-const bpDir = '/default/.cache/cockpit-image-builder/';
+const bpDir = '/default/.local/state/cockpit-image-builder';
```
regexowl, line 8: We had mocks after all, but they were not being loaded properly.

src/Components/CreateImageWizard/steps/Repositories/components/Repositories.tsx
```diff
@@ -332,7 +344,11 @@ const Repositories = () => {
   ]);
 
   useEffect(() => {
-    if (isTemplateSelected && reposInTemplate.length > 0) {
+    if (
+      isTemplateSelected &&
+      reposInTemplate.length > 0 &&
+      !isReposInTemplateLoading
```
regexowl: This should wait for the templates to finish loading before triggering and doing any operations.

src/Components/CreateImageWizard/steps/Repositories/components/Repositories.tsx
```diff
@@ -244,29 +244,41 @@ const Repositories = () => {
...
-      previousReposData.length < initialSelectedState.size
+      previousLoading ||
+      !previousSuccess ||
+      previousReposData.length >= initialSelectedState.size
     ) {
-      const prevSet = new Set(previousReposData.map(({ uuid }) => uuid));
-      const itemsToRemove = [...initialSelectedState]
-        .filter((uuid) => !prevSet.has(uuid))
-        .map((uuid) => ({ uuid })) as ApiRepositoryResponseRead[];
-      removeSelected(itemsToRemove);
```
regexowl, line 258: This is a side effect that was re-triggering the useMemo, ending in a loop. I've split the logic to run `removeSelected` outside of the useMemo.

regexowl:
Ok, I think this is finally ready for review. Found some more loops while trying to clean up the test output.

src/Components/CreateImageWizard/steps/Repositories/components/Repositories.tsx
```diff
@@ -244,29 +244,41 @@ const Repositories = () => {
...
+  useEffect(() => {
+    if (hasRemovedUnavailable || unavailableRepoIds.length === 0) {
+      return;
+    }
+
+    removeSelected(
+      unavailableRepoIds.map((uuid) => ({
+        uuid,
+      })) as ApiRepositoryResponseRead[],
+    );
```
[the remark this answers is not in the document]
regexowl: Yes, we have a bunch of incomplete dependency arrays and I'd prefer to split resolving those into a separate PR as its a bag of cats on its own.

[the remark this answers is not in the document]
regexowl: Or maybe not, what do I know. Shouldn't cause a re-render ig, giving it a try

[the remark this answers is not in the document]
regexowl: Ha, nope! Of course, because that would re-trigger the effect

kingsleyzissou (dismissed):
Let's go 🚀

regexowl:
Ugh, even more loops. Looks like a relevant failure, so back to the draft it goes ._.

regexowl:
Hmm, I'll pivot a bit. Will update the PR to not touch the loops and to not fail on warn/error. Just cleaning up the warnings and errors for now. And I'll open a fresh one that will focus on the loops only where I could push without pinging everyone all the time...

regexowl:
And this will be the follow up https://github.com/osbuild/image-builder-frontend/pull/4489

kingsleyzissou (approved):

PR #4470, opened 2026-05-27
[10.2]Packit: Only propose downstream releases for main

lucasgarfield:
If we need to do a point release for a RHEL Z-stream, we don't want to propose downstream releases for Fedora.


lucasgarfield:
We will need this for the 10.2 branch as well as main.

ondrejbudai (approved):

PR #4472, opened 2026-05-27
ImagesTable: Fix image/package mode labels in blueprint view (HMS-10739)

ochosi:
Right now, the image/package mode colorful labels are shown in the unfiltered images table view. As soon as the view is filtered to a single blueprint, the colorful labels disappear. This is due to a lack of separation of concerns in the code, that lumps the logic of rendering a name as a clickable link (to the blueprint) and rendering the labels together.

This commit separates those concerns and always shows the colorful labels, as expected.


regexowl (approved):
Thank you!

PR #4481, opened 2026-05-28
Remove dummy 'image-mode' distribution (HMS-10745)

kingsleyzissou:
## Summary
Remove the `'image-mode'` dummy distribution that was used as a workaround for on-prem bootc builds. Now that on-prem has a proper `getDistributions` query, blueprint mode detection relies on the `bootc` field instead of a magic distribution string.

## Changes
- Remove `IMAGE_MODE` constant, `isImageMode`/`asDistribution` type guards, and all `Distributions | 'image-mode'` union types
- Unify on-prem API arg types with hosted (`GetArchitecturesApiArg`, `GetOscapProfilesApiArg`, etc.) instead of custom composer variants
- Extract `prepareBlueprintForSave` to strip distribution for bootc blueprints at the persistence boundary, replacing inline `IMAGE_MODE` checks
- Update `toComposerComposeRequest` to clear distribution based on `bootc` presence rather than matching a magic string
- Add unit tests for `mapRequestToState`, `mapBlueprintExportToState`, and `prepareBlueprintForSave`

Depends on #4480

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/BlueprintMode.tsx
@@ -22,7 +22,6 @@ import {
   selectDistribution,
   selectIsImageMode,
 } from '@/store/slices/wizard';
-import { asDistribution } from '@/store/typeGuards';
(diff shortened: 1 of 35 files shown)


regexowl (approved):
Looks good to me 🧹 🚀

PR #4482, opened 2026-05-29
Clear timezone and locale for unsupported image types (HMS-10456)

kingsleyzissou:
## Summary
Use the existing restrictions system to clear wizard state for timezone
and locale when the selected image type doesn't support them (e.g.,
network-installer), preventing stale customization data from leaking
into blueprint requests.

## Changes
- Add `clearTimezone` and `clearLocale` reducers to atomically reset related state (timezone + NTP servers, languages + keyboard)
- Add `useEffect` hooks in both wizard versions to dispatch clearing actions when `restrictions.*.shouldHide` is true, with guards to skip edit mode and avoid redundant dispatches
- Add `timezone.shouldHide` assertion for network-installer in the restrictions hook tests
- Add integration tests for network-installer: checkbox mutual exclusion, wizard step reduction, and blueprint request verification with empty `customizations`
- Add unit tests for the new reducers covering reset and no-op cases

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -34,6 +34,8 @@ import {
   changeProxy,
   changeServerUrl,
   changeTimezone,
+  clearLocale,
+  clearTimezone,
   initializeWizard,
   loadWizardState,
   selectDistribution,
@@ -121,6 +123,7 @@ const CreateImageWizard = () => {
   const { userData } = useGetUser(auth);
   const hasTrackedInitialStepRef = useRef(false);
   const hasTrackedWizardOpenedRef = useRef(false);
+  const isHostDistroDetected = useRef(!isOnPremise);
 
   const {
     data: blueprintDetails,
@@ -247,6 +250,7 @@ const CreateImageWizard = () => {
     const initializeHostDistro = async () => {
       const distro = await getHostDistro();
       dispatch(changeDistribution(distro));
+      isHostDistroDetected.current = true;
     };
 
     const initializeHostArch = async () => {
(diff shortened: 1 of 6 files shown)


kingsleyzissou:
https://github.com/osbuild/image-builder-frontend/pull/4311 got closed, I cherry-picked the changes from that and made some changes

src/Components/CreateImageWizard3/CreateImageWizard3.tsx
```diff
@@ -273,7 +280,24 @@ const CreateImageWizard3 = () => {
...
+  ]);
+
+  useEffect(() => {
+    if (mode !== 'create') {
+      return;
+    }
+
+    if (restrictions.locale.shouldHide) {
+      dispatch(clearLocale());
+    }
```
jlsherrill: you mentioned locale was still supported, right?

kingsleyzissou: Correct, I'll tag you where I introduced the bug

src/store/api/distributions/constants.ts
```diff
@@ -73,7 +73,7 @@ export const DISTRO_DETAILS: Record<string, ImageTypeInfo> = {
   },
   'network-installer': {
     name: 'network-installer',
-    supported_blueprint_options: ['locale', 'fips'],
+    supported_blueprint_options: ['fips'],
```
kingsleyzissou: @jlsherrill this diff was the problem

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -305,7 +314,24 @@ const CreateImageWizard = () => {
...
+
+  useEffect(() => {
+    if (mode !== 'create') {
+      return;
+    }
+
+    if (restrictions.locale.shouldHide) {
+      dispatch(clearLocale());
+    }
+  }, [mode, dispatch, restrictions.locale.shouldHide]);
```
regexowl: Would it make sene to handle both timezone and locale in one `useEffect`? Since they're related we could just put everything in one place, but not a blocker.

regexowl (dismissed):
Looks good! One minor nitpick, but not blocking on that

src/Components/CreateImageWizard/CreateImageWizard.tsx
kingsleyzissou: Might as well address this now

kingsleyzissou:
> Looking at the test failures, those seem to be relevant.

I had a suspicion but I haven't had the chance to look into it properly yet

regexowl (approved):
Nice!

PR #4489, opened 2026-06-01
vitest: Fail on console errors and warnings (HMS-10806)

regexowl:
Some changes introduce warning and error console logs in the test output. This will make the test run fail when there's any "mess" and the problems can be fixed in place instead of retroactively when the test output becomes unreadable.

"Maximum update depth exceeded" warning was silences for now and will be revisited later. There's just new act warning and stuff coming in almost every week.

JIRA: [HMS-10806](https://issues.redhat.com/browse/HMS-10806)

[HMS-10806]: https://redhat.atlassian.net/browse/HMS-10806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/setup.ts
+++ b/src/test/setup.ts
@@ -30,13 +30,18 @@ vi.stubGlobal('ResizeObserver', MockResizeObserver);
 vi.mock('@redhat-cloud-services/frontend-components/useChrome', () => ({
   default: () => ({
     auth: {
-      getUser: async () => ({
-        identity: {
-          internal: {
-            org_id: 5,
-          },
-        },
-      }),
+      getUser: () =>
+        new Promise((resolve) => {
+          setTimeout(() => {
+            resolve({
+              identity: {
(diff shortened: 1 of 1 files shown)


src/test/setup.ts
```diff
@@ -30,13 +30,18 @@ vi.stubGlobal('ResizeObserver', MockResizeObserver);
...
   default: () => ({
     auth: {
-      getUser: async () => ({
-        identity: {
-          internal: {
-            org_id: 5,
-          },
-        },
-      }),
+      getUser: () =>
```
regexowl, line 33: This is needed to resolve an error that creeped in while deprecating the old wizard.

kingsleyzissou (approved):
🚀

PR #4490, opened 2026-06-01
api: Connect customization validation to CRC (HMS-10672)

tkoscieln:
This replaces `distributionDetailsApi.ts` mock that had hardcoded values for customization validation for each target with actual API response from CRC endpoint `/distributions/{distro}`. It's used with filtering to retrieve supported customizations by a target for specific distro and arch from metadata the endpoint returns. These metadata are propagated to CRC endpoint from osbuild/image-builder (formerly osbuild/images) definitions. Includes a mapping between backend naming and frontend naming for customizations.

---

~~Currently blocked on https://github.com/osbuild/osbuild-composer/pull/5163 getting released and updated to in CRC - this will enable backend to accept aliases in filtering for image types and architecture and will also fix a bug when images with no partition table (WSL,...) were not returned~~


croissanne:
While I agree with this, the timing is not ideal. We want to get rid of composer and this adds another dependency on it. Could we keep the mock on prem for now?

tkoscieln:
@croissanne but don't we have to migrate Cloud API 1:1 to image-builder anyway? :thinking:

croissanne:
> @croissanne but don't we have to migrate Cloud API 1:1 to image-builder anyway? 🤔

Not quite, if we go the socket-activated unit which spawns singleshot systemd-run stuff we'll still need extra parameters to tell image-builder where to save everything. Similarly, for compose statuses we will not be able to rely on ib-cli, but rather from files on disk.

That said, this particular functionality we could probably add endpoints to this socket-activated shim which translates it into `image-builder describe` commands. I wasn't aware this was already implemented fully, so if `image-builder describe` can do exactly what you need here, then it should be fine. But keep in mind there that the actual api code here will look very different (the response we might get back from image-builder might have a different structure, and we need to implement `--format json` in image-builder too).

So maybe it's ok actually, because describe could support it.

tkoscieln:
Except for aliases missing from the describe command ouput (translated to JSON using `yq`), everything else is there :tada: Yes, the structure is pretty different. But the CRC endpoint depends on the current Cloud API too, it's only a passthrough for it, so in order to avoid discrepancy between hosted and on prem we can then overhaul the API for both of them when we switch to the socket-activated shim? :thinking: 

<details><summary>JSON output of image-builder describe</summary><p>

```json
{
  "distro": "rhel-9.9",
  "type": "qcow2",
  "arch": "x86_64",
  "os_version": "9.9",
  "bootmode": "hybrid",
  "partition_type": "gpt",
  "default_filename": "disk.qcow2",
  "build_pipelines": [
    "build"
  ],
  "payload_pipelines": [
    "os",
    "image",
    "qcow2"
  ],
  "packages": {
    "build": {
      "include": [
        "coreutils",
        "dosfstools",
        "glibc",
        "grub2-pc",
        "platform-python",
        "policycoreutils",
        "pqrpm",
        "python3",
        "python3-iniparse",
        "qemu-img",
        "rpm",
        "selinux-policy-targeted",
        "systemd",
        "xfsprogs",
        "xz"
      ],
      "exclude": []
    },
    "os": {
      "include": [
        "@core",
        "authselect-compat",
        "chrony",
        "cloud-init",
        "cloud-utils-growpart",
        "cockpit-system",
        "cockpit-ws",
        "dnf-utils",
        "dosfstools",
        "dracut-config-generic",
        "efibootmgr",
        "grub2",
        "grub2-efi-x64",
        "grub2-pc",
        "insights-client",
        "kernel",
        "nfs-utils",
        "oddjob",
        "oddjob-mkhomedir",
        "psmisc",
        "python3-jsonschema",
        "qemu-guest-agent",
        "redhat-release",
        "redhat-release-eula",
        "rsync",
        "selinux-policy-targeted",
        "shim-x64",
        "subscription-manager-cockpit",
        "tar",
        "tcpdump",
        "tuned",
        "xfsprogs"
      ],
      "exclude": [
        "aic94xx-firmware",
        "alsa-firmware",
        "alsa-lib",
        "alsa-tools-firmware",
        "biosdevname",
        "dnf-plugin-spacewalk",
        "dracut-config-rescue",
        "fedora-release",
        "fedora-repos",
        "firewalld",
        "iprutils",
        "ivtv-firmware",
        "iwl100-firmware",
        "iwl1000-firmware",
        "iwl105-firmware",
        "iwl135-firmware",
        "iwl2000-firmware",
        "iwl2030-firmware",
        "iwl3160-firmware",
        "iwl3945-firmware",
        "iwl4965-firmware",
        "iwl5000-firmware",
        "iwl5150-firmware",
        "iwl6000-firmware",
        "iwl6000g2a-firmware",
        "iwl6000g2b-firmware",
        "iwl6050-firmware",
        "iwl7260-firmware",
        "langpacks-*",
        "langpacks-en",
        "libertas-sd8787-firmware",
        "nss",
        "plymouth",
        "rng-tools",
        "udisks2"
      ]
    }
  },
  "partition_table": {
    "uuid": "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
    "type": "gpt",
    "partitions": [
      {
        "size": 1048576,
        "type": "21686148-6449-6E6F-744E-656564454649",
        "bootable": true
      },
      {
        "size": 209715200,
        "type": "C12A7328-F81F-11D2-BA4B-00A0C93EC93B",
        "payload": {
          "type": "vfat",
          "label": "ESP",
          "mountpoint": "/boot/efi",
          "fstab_options": "defaults,uid=0,gid=0,umask=077,shortname=winnt",
          "fstab_passno": 2
        }
      },
      {
        "size": 1073741824,
        "type": "BC13C2FF-59E6-4262-A352-B275FD6F7172",
        "payload": {
          "type": "xfs",
          "label": "boot",
          "mountpoint": "/boot",
          "fstab_options": "defaults"
        }
      },
      {
        "size": 2147483648,
        "type": "0FC63DAF-8483-4772-8E79-3D69D8477DE4",
        "payload": {
          "type": "xfs",
          "label": "root",
          "mountpoint": "/",
          "fstab_options": "defaults"
        }
      }
    ]
  },
  "blueprint": {
    "supported_options": [
      "distro",
      "packages",
      "modules",
      "groups",
      "enabled_modules",
      "containers",
      "customizations.cacerts",
      "customizations.directories",
      "customizations.disk",
      "customizations.dnf",
      "customizations.files",
      "customizations.filesystem",
      "customizations.partitioning_mode",
      "customizations.fips",
      "customizations.firewall",
      "customizations.user",
      "customizations.sshkey",
      "customizations.group",
      "customizations.hostname",
      "customizations.kernel",
      "customizations.locale",
      "customizations.openscap",
      "customizations.repositories",
      "customizations.rpm",
      "customizations.services",
      "customizations.timezone",
      "customizations.rhsm",
      "customizations.sshd",
      "name",
      "version",
      "description"
    ]
  }
}
```
</p></details>

<details><summary>JSON output of CRC endpoint (which is passthrough for Cloud API)</summary><p>

```json
{
  "architectures": {
    "x86_64": {
      "image_types": {
        "qcow2": {
          "aliases": [
            "guest-image"
          ],
          "base_partition_table": {
            "partitions": [
              {
                "fs_type": "vfat",
                "label": "ESP",
                "minsize": "209715200",
                "mountpoint": "/boot/efi"
              },
              {
                "fs_type": "xfs",
                "label": "boot",
                "minsize": "1073741824",
                "mountpoint": "/boot"
              },
              {
                "fs_type": "xfs",
                "label": "root",
                "minsize": "2147483648",
                "mountpoint": "/"
              }
            ],
            "type": "gpt"
          },
          "boot_mode": "hybrid",
          "default_size": 10737418240,
          "exports": [
            "qcow2"
          ],
          "filename": "disk.qcow2",
          "mime_type": "application/x-qemu-disk",
          "name": "qcow2",
          "ostree_ref": "",
          "partition_type": "gpt",
          "payload_package_sets": [
            "blueprint"
          ],
          "supported_blueprint_options": [
            "distro",
            "packages",
            "modules",
            "groups",
            "enabled_modules",
            "containers",
            "customizations.cacerts",
            "customizations.directories",
            "customizations.disk",
            "customizations.dnf",
            "customizations.files",
            "customizations.filesystem",
            "customizations.partitioning_mode",
            "customizations.fips",
            "customizations.firewall",
            "customizations.user",
            "customizations.sshkey",
            "customizations.group",
            "customizations.hostname",
            "customizations.kernel",
            "customizations.locale",
            "customizations.openscap",
            "customizations.repositories",
            "customizations.rpm",
            "customizations.services",
            "customizations.timezone",
            "customizations.rhsm",
            "customizations.sshd",
            "name",
            "version",
            "description"
          ]
        }
      },
      "name": "x86_64"
    }
  },
  "codename": "",
  "module_platform_id": "platform:el9",
  "name": "rhel-9.6",
  "os_version": "9.6",
  "product": "Red Hat Enterprise Linux",
  "releasever": "9"
}

```
</p></details>

src/store/api/backend/index.ts
```diff
@@ -81,6 +81,12 @@ export const useGetComposeStatusQuery = process.env.IS_ON_PREMISE
   ? composerQueries.useGetComposeStatusQuery
   : serviceQueries.useGetComposeStatusQuery;
 
+export const useGetDistributionQuery = (
+  process.env.IS_ON_PREMISE
+    ? composerQueries.useGetDistributionQuery
+    : serviceQueries.useGetDistributionQuery
+) as typeof serviceQueries.useGetDistributionQuery;
```
regexowl: Is the `typeof` necessary?

regexowl (dismissed):
Thanks for looking into this! Started to review, but didn't finish yet since I've noticed the on-prem customization filtering no longer works with these changes. Can you please take a look at that?

kingsleyzissou (commented):
This is looking great, I have some suggestions.

I think ideally we would get rid of the the `distributions` subdirectory since now both backends should support this.

I added a derived endpoint in this PR https://github.com/osbuild/image-builder-frontend/pull/4597. Maybe we could do something similar. Have an endpoint in there for managing the distribution details and normalizing the results?

That is quite an architectural change though. But it would get rid of the need for custom hooks. We could move the constants and types into `store/api/backend` which would be quite nice and then there would be a single RTK query hook for both.

Another nice to have, but not completely necessary would be to break the commits up and build on them one commit at a time rather than one massive commit.

But great job on this 🙂

src/store/api/backend/index.ts
tkoscieln: Removed entirely

tkoscieln:
@kingsleyzissou not sure if your review is still relevant if we reverted the changes back to the hardcoded values for cockpit? Seems like cockpit and service will divert here and it will remain that way since describe command in image-builder-cli does contain the same info, but in very different format :thinking:

kingsleyzissou (approved):
Nice, the changes look good

PR #4492, opened 2026-06-01
Wizard: Fix kernel alert rendering (HMS-10579)

regexowl:
The alert needs to be rendered outside of the Content block to be formatted properly.

Before:
image

After:
image


regexowl:
> * The `{' '}` inserted after the `</Content>` closing tag is unnecessary and will render an extra space node; consider removing it to keep the JSX clean.

Valid

kingsleyzissou (approved):

PR #4496, opened 2026-06-01
[10.2] release: pass tag to release-action for workflow_dispatch

lucasgarfield:
The release-action now accepts a tag input, which is needed when triggered via workflow_dispatch since GITHUB_REF is a branch, not a tag.


lucasgarfield:
See: https://github.com/osbuild/release-action/pull/46

avitova (approved):
Nice!

PR #4501, opened 2026-06-03
Wizard: Prevent Users columns from overflowing the card on small screens (HMS-10579)

mgold1234:
Wizard: Prevent Users columns from overflowing the card on small screens 
Screenshot 2026-06-03 at 15 13 06

JIRA: (HMS-10579)


regexowl (approved):
Thanks!

PR #4505, opened 2026-06-03
[10.2] playwright: fix tmt in dist-git

lucasgarfield:
Recently testing-farm tests were failing and a fix was merged in 3507013e where TMT_TREE is used instead of TMT_SOURCE.

This got the CI working, but broke the tests when running in dist-git.

We need to use TMT_SOURCE when in CI, and TMT_SOURCE_DIR when in dist-git. This commit adds the dist-git case back while keeping TMT_SOURCE for Packit.

### DIFF
--- a/schutzbot/playwright_tests.sh
+++ b/schutzbot/playwright_tests.sh
@@ -2,11 +2,16 @@
 set -euo pipefail
 
 PW_WORKERS=4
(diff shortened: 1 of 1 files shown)


croissanne (approved):

lucasgarfield:
This will go in with https://github.com/osbuild/image-builder-frontend/pull/4509

PR #4508, opened 2026-06-04
Add PlatformContext to replace build-time hook switching

kingsleyzissou:
## Summary
Introduce a `PlatformContext` to replace build-time `process.env.IS_ON_PREMISE` ternaries with a runtime React context, enabling both hosted and
 on-prem platform implementations to coexist in the same bundle and be testable without separate builds.

## Changes
   - Add `PlatformContext`, `usePlatform()` hook, and `PlatformProvider` with a `PlatformHooks` type contract enforcing compile-time exhaustiveness
 via `satisfies`
   - Implement `hostedPlatform` and `onPremPlatform` as module-level singletons providing platform-specific RTK Query hooks, feature flags, and
 environment helpers
   - Migrate all 40+ consumer components from direct barrel-file imports to `usePlatform()` destructuring
   - Remove `process.env.IS_ON_PREMISE` ternary switching from backend and content sources barrel files, replacing with clean re-exports
   - Add `Proxy`-based `mockPlatform` fixture for tests and update all test utilities (`renderWithRedux`, `renderLandingPage`) to derive platform
 from Redux state
 
 ## Motivation for these changes

So this might be a hard sell, but I was looking at ways to reduce the `process.env.IS_ON_PREMISE` calls.
But it's fundamentally about decoupling the two frontends. There are a couple of reasons for this but the most important one is:
- if we forget to add a ternary here, the on premise frontend could end up with a hosted hook and things would fail and we might only realise much later

This approach establishes a type contract here that could help us catch weird bugs at compilation rather than at run time. We can more effectively isolate things so hosted only gets hosted queries & mutations, environment flags and 
so on. And on-prem would only get what it knows about.

In the future, it would also make it easier to switch to a mono repo and isolate cockpit and the service even further (this is an entirely separate discussion altogether though).

## Notes

I haven't done a `useFlag` migration yet, that can be done in a follow up along with a general tidy up.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,6 +92,53 @@ Imports are enforced alphabetically by ESLint. Group order:
 - Playwright tests for E2E coverage of critical flows
 - The `TZ=UTC` prefix is applied automatically by npm scripts
 
+## Platform Context
+
+The app runs on two platforms — **hosted** (console.redhat.com) and **on-premises** (Cockpit plugin). Platform-specific hooks, API slices, and utilities are abstracted behind a React context so components never branch on the environment directly.
+
+### Architecture
+
+| File | Purpose |
+| --- | --- |
+| `src/context/platform/types.ts` | `PlatformHooks` type — the contract for `queries`, `mutations`, `env`, and `api` |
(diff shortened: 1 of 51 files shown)


kingsleyzissou:
Closing this in favour of the webpack alias approach

PR #4509, opened 2026-06-05
[rhel-10.2] store/cockpit: create bp dir before copying from .cache

croissanne:
The backwards compatibility script fails in case the target .state directory doesn't exist. Ensure the state directory exists.


regexowl (approved):

lucasgarfield (approved):
lgtm

PR #4511, opened 2026-06-05
[rhel-10.2] cockpit/spec: fix version

croissanne:
The last post release version bump resulted in a patch release, we just need another minor one.


lucasgarfield (approved):
lgtm

PR #4524, opened 2026-06-10
Adopt Prettier as a standalone formatter (HMS-10815)

kingsleyzissou:
## Summary

Separates formatting from linting by adopting Prettier as a standalone tool, replacing the `eslint-plugin-prettier` integration.
This gives faster formatting, cleaner ESLint output, and lets each tool do what it does best.

## Changes

- Add Prettier config (`.prettierrc.json`, `.prettierignore`) and `format`/`format:check` npm scripts
- Replace `eslint-plugin-prettier` with `eslint-config-prettier` so ESLint no longer runs Prettier rules — it just defers to
Prettier on style
- Remove `npm-run-all2` dependency; simplify `lint` and `verify` scripts
- Reformat the codebase with the new standalone Prettier setup
- Enable ESLint caching on the default `lint` command

## Notes
- The code generation picks up the prettierconfig and formats the code accordingly, so I put these changes in with the first commit
- Splitting Prettier out of the lint pass and adding ESLint caching brings linting from ~24s down to ~7s with a warm
cache.

JIRA: [HMS-10815](https://issues.redhat.com/browse/HMS-10815)

[HMS-10815]: https://redhat.atlassian.net/browse/HMS-10815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


regexowl:
The boot test "download" failures are not relevant to changes in this PR. I'm ruminating over them in the frontend channel ._.

package.json
```diff
@@ -100,9 +100,11 @@
     "whatwg-fetch": "3.6.20"
   },
   "scripts": {
-    "lint": "npm-run-all lint:*",
-    "lint:js": "eslint src playwright",
+    "lint": "eslint src playwright --fix --cache",
```
regexowl, line 103: Could we deduplicate `lint` and `lint:js:fix`? Looks like they're doing pretty much the same thing.

package.json
```diff
@@ -118,7 +120,7 @@
     "build": "fec build",
     "build:cockpit": "webpack --config cockpit/webpack.config.ts",
     "api": "bash api/codegen.sh",
-    "verify": "npm-run-all build lint test",
+    "verify": "npm run build && npm run lint && npm run format && npm run test",
```
regexowl, line 122: Nice! Getting rid of the `npm-run-all` dep makes perfect sense 🧹

.prettierrc.json
```diff
@@ -0,0 +1,9 @@
+{
```
regexowl, line 1: Just comparing with the previously used list from ESLint config - these two are missing:
- tsxSingleQuote: true,
- tsSingleQuote: true,

are they left out on purpose?

regexowl (commented):
Looks good! Just a few questions

regexowl:
Also rebased, the boot tests failure should be gone now

package.json
kingsleyzissou: Ya happy to remove, I just kept it there for backwards compatibility

.prettierrc.json
kingsleyzissou: I should have commented on that, turns out those don't actually exist lol (or they got removed at somepoint)

regexowl: Lol, fair enough :D

package.json
regexowl: If we're using it somewhere we can keep it. :+1:  Just noticed they're very similar

regexowl (dismissed):
Let's go! :shipit:

package.json
kingsleyzissou: I couldn't find any references. We could maybe add a deprecation warning for a bit or just remove it and see if anything breaks?

.prettierrc.json
kingsleyzissou: I added a comment to the commit message (since it's a json file)

kingsleyzissou:
I quickly fixed those issues

regexowl (approved):
Nice!

PR #4526, opened 2026-06-10
Support non-RHEL bootc images in on-prem image mode (HMS-10816)

kingsleyzissou:
Support non-RHEL bootc images in on-prem image mode

## Summary

Broadens on-prem image mode support beyond RHEL to accept arbitrary bootc containers (Fedora, CentOS Stream, Hummingbird, custom images). Previously, strict label checks silently filtered out anything that wasn't a Red Hat–published container.

## Changes

- Loosen `PodmanLabels` type and remove the hard-coded registry filter so non-RHEL bootc images are no longer silently dropped
- Add `inferDistro` helper to detect the distro from image labels and references, with named support for RHEL, Fedora, CentOS Stream, and Hummingbird
- Use container metadata `.Architecture` for architecture detection instead of the user-defined label, normalizing kernel arch format (`amd64`/`arm64`) to RPM format (`x86_64`/`aarch64`)
- Add test coverage for `inferDistro`, `filterBootcImages`, `toBootcDistro`, and the `ImageSourceSelect` component with non-RHEL images


src/store/api/backend/onprem/composerApi/helpers/podmanImages.ts
```diff
@@ -40,7 +40,13 @@ export const filterBootcImages = (arch: string | undefined) => {
       return false;
     }
 
-    if (image.Labels.architecture !== arch) {
+    // Some bootc images don't include an architecture label.
+    // When missing, we let them through and fall back to the
+    // host architecture in toBootcDistro.
+    if (
+      image.Labels.architecture !== undefined &&
```
kingsleyzissou: The comment explains the reasoning for this a little bit.
But, for example, I pulled down `quay.io/fedora/fedora-bootc` which I think uses a manifest list but it pulled down the correct arch for my machine. When you inspect this image, it doesn't have arch set at all.
Maybe we could just assume that the user pulled down the correct arch.

This could be a footgun though. So we might want to drop this commit.

kingsleyzissou: I guess to clarify this further, some images don't have `architecture` set in their metadata. You just pull the image and it goes through the manifest list and pulls the correct image for you for your host's architecture. How do we want to handle this edge case?

Do we want to ignore those images? Or do we want to just assume that they match the host arch?
I have gone for the second option here

kingsleyzissou: @ondrejbudai thoughts?

kingsleyzissou: Silly me, this should be fixed now, we will use `.Architecture` instead

ochosi (dismissed):
I will admit I skipped reviewing the tests, but the code and the logic looks good as is! Ship it!

ochosi:
I have locally tested this properly now and here are some remarks:
1. This change isn't obvious to test, because Image Mode isn't shown on Fedora yet.
2. The architecture detection didn't work for me, as it isn't in the labels but only visible through `podman inspect`.
3. The distribution version doesn't seem to get extracted successfully, so `quay.io/fedora/fedora-bootc:44` only shows up as "Fedora" in the dropdown.

I think this can be merged as is though. I've created a small follow-up that addresses these points based on your branch. See #4532

kingsleyzissou:
> I have locally tested this properly now and here are some remarks:
> 
> 1. This change isn't obvious to test, because Image Mode isn't shown on Fedora yet.
> 
> 2. The architecture detection didn't work for me, as it isn't in the labels but only visible through `podman inspect`.
> 
> 3. The distribution version doesn't seem to get extracted successfully, so `quay.io/fedora/fedora-bootc:44` only shows up as "Fedora" in the dropdown.
> 
> 
> 
> I think this can be merged as is though. I've created a small follow-up that addresses these points based on your branch. See #4532

1. Ya that's by design at the moment. I actually think we should drop the constraint after this PR merges

2. In what sense? Like the arch was empty or it didn't show any image at all? I think we just take the host arch

3. This is really, really tricky and it's our best guess. We can try check the tag to see if it's possibly a version rather than latest or a sha... but this is different for each distro we support. So maybe this is something we should decide on before we merge

ochosi:
> 1. Ya that's by design at the moment. I actually think we should drop the constraint after this PR merges

Agreed. Just wanted to point it out for anyone else testing the PR 🙂 Same goes for my branch, it's far from polished and is only designed to help test this PR for now.

> 2. In what sense? Like the arch was empty or it didn't show any image at all? I think we just take the host arch

Without the patch on my branch I didn't see any image at all. I have a Thinkpad with F44 and pulled the amd64 container image. (Well, that is, I didn't specify anything and just podman pulled, which I expected to work.)

> 3. This is really, really tricky and it's our best guess. We can try check the tag to see if it's possibly a version rather than latest or a sha... but this is different for each distro we support. So maybe this is something we should decide on before we merge 

I think we could make that a UX/PM decision after this PR lands and show a bunch of different examples and how they'd show up. Showing the full ref may be a good idea in many cases.

kingsleyzissou:
> > 1. Ya that's by design at the moment. I actually think we should drop the constraint after this PR merges
> 
> Agreed. Just wanted to point it out for anyone else testing the PR 🙂 Same goes for my branch, it's far from polished and is only designed to help test this PR for now.
> 
> > 2. In what sense? Like the arch was empty or it didn't show any image at all? I think we just take the host arch
> 
> Without the patch on my branch I didn't see any image at all. I have a Thinkpad with F44 and pulled the amd64 container image. (Well, that is, I didn't specify anything and just podman pulled, which I expected to work.)
> 
> > 3. This is really, really tricky and it's our best guess. We can try check the tag to see if it's possibly a version rather than latest or a sha... but this is different for each distro we support. So maybe this is something we should decide on before we merge
> 
> I think we could make that a UX/PM decision after this PR lands and show a bunch of different examples and how they'd show up. Showing the full ref may be a good idea in many cases.

~Oh right, sorry I missed the follow up PR there that you had sent (was on mobile). I see what the issue is now, there is a race between getting the host arch (that's possibly not defined yet)~

Meh it's simpler than that. The Friday afternoon commit broke things 😓 I rushed it. The label *should* be there, when you inspect manually it should be there

kingsleyzissou:
Okay I've fixed this PR now, it was too broken to merge as it was. The containers should display now. We have to chain `podman images` and `podman inspect` to get what we need. Unfortunate, but not much more we can do.

We can remove the fedora constraint for cockpit in a follow up like discussed.

And as you mentioned we will have to revisit the versions, it's tricky because the fedora bootc containers have an empty version string and only this:

```json
"RepoTags": [
   "quay.io/fedora/fedora-bootc:44"
],
```

And this (which seems very fragile):

```json
"org.opencontainers.image.version": "44.20260614.0",
```

src/store/api/backend/onprem/composerApi/helpers/podman/normalizeArch.ts
```diff
@@ -0,0 +1,9 @@
+const archMap: Record<string, string> = {
+  amd64: 'x86_64',
+  arm64: 'aarch64',
+  x86_64: 'x86_64',
+  aarch64: 'aarch64',
+};
+
+export const normalizeArch = (arch: string | undefined): string | undefined =>
```
regexowl, line 8: Hmm, is the `normalizeArch` used anywhere outside of the tests? I guess the arch should be normalized in `filters.ts`?

kingsleyzissou: Hmm it should be used in the toBootcDistro function. Maybe I missed that

kingsleyzissou: Yep I accidentally removed that

ochosi:
I just re-tested this and it works well now!

The arch detection is fine and as you said, the tag parsing may be a future can of :worm:

ochosi (approved):

src/Components/CreateImageWizard/steps/ImageOutput/tests/mocks/fixtures.ts
```diff
@@ -141,3 +141,34 @@ export const mockBootcDistributionsMultipleTypes: BootcDistributionItem[] = [
...
+export const mockBootcDistributionsMixed: BootcDistributionItem[] = [
+  {
+    distro: 'rhel-10',
+    name: 'Red Hat Enterprise Linux (RHEL) 10',
+    type: 'guest-image',
+    arch: 'x86_64',
+    reference: 'registry.redhat.io/rhel10/rhel-bootc:rhel-10',
+  },
+  {
+    distro: 'fedora-42',
```
ochosi, line 154: nitpick: I'd bump this to F44

PR #4527, opened 2026-06-11
Wizard: prevent duplicate languages in Locale step (HMS-10599)

mgold1234:
`replaceLanguage` did not guard against duplicates, and `LanguageRow`
used `key={lang}` which caused React warnings when duplicates arrived
from an imported blueprint.

- Add duplicate guard to `replaceLanguage`
- Fix list key to handle duplicate entries
- Add tests for duplicate language validation

### Why is this needed?

The duplicate language validation and error message were already
implemented in a previous commit, but had no test coverage.
Additionally, `replaceLanguage` did not prevent creating duplicates
at the reducer level, and React would log warnings about duplicate
keys when a blueprint with duplicate languages was loaded from the API.

These changes harden the existing validation by preventing duplicates
at the source and ensuring the rendering handles them gracefully.

### To reproduce

Create a blueprint via API with duplicate languages:
```js
fetch('/api/image-builder/v1/blueprints', {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({
    name: "test-duplicate-languages",
    distribution: "rhel-9",
    image_requests: [{ architecture: "x86_64", image_type: "guest-image",
      upload_request: { type: "aws.s3", options: {} } }],
    customizations: { locale: { languages: ["en_US.UTF-8", "en_US.UTF-8", "C.UTF-8"] } }
  })
}).then(r => r.json()).then(console.log)
```
Then open the blueprint in the wizard — the Locale step shows
"Duplicated languages: en_US.UTF-8" and blocks the Next button.

JIRA: HMS-10599

### DIFF
--- a/src/Components/CreateImageWizard/steps/Locale/tests/Locale.test.tsx
+++ b/src/Components/CreateImageWizard/steps/Locale/tests/Locale.test.tsx
@@ -1,6 +1,6 @@
-import { screen } from '@testing-library/react';
+import { act, screen } from '@testing-library/react';
 
-import { initialState } from '@/store/slices/wizard';
+import { initialState, replaceLanguage } from '@/store/slices/wizard';
 import { clickWithWait, createUser } from '@/test/testUtils';
 
 import {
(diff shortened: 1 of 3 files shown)


src/store/slices/wizard/index.ts
```diff
@@ -1452,6 +1452,13 @@ export const wizardSlice = createSlice({
       action: PayloadAction<{ oldLanguage: string; newLanguage: string }>,
     ) => {
       if (state.locale.languages) {
+        const isDuplicate = state.locale.languages.some(
+          (lang, i) =>
+            lang === action.payload.newLanguage &&
+            state.locale.languages![i] !== action.payload.oldLanguage,
+        );
+        if (isDuplicate) return;
+
```
kingsleyzissou: This is a little bit hard to follow, maybe we could check the old language first?
Something like:

```suggestion
        const index = state.locale.languages.findIndex(
         (lang) => lang === action.payload.oldLanguage,
        );
        if (index === -1) return;
        
        const isDuplicate = state.locale.languages.some(
         (lang, i) => i !== index && lang === action.payload.newLanguage,
        );
        if (isDuplicate) return;
 ```

File: src/Components/CreateImageWizard/steps/Locale/tests/Locale.test.tsx [hunk 2]
```diff
@@ -348,6 +348,38 @@ describe('Locale Component', () => {
     });
   });
 
+  describe('Validation', () => {
```
kingsleyzissou, line 387: The new test covers the duplicate-blocked case, but there's no test that a valid (non-duplicate) replacement actually succeeds

src/Components/CreateImageWizard/steps/Locale/tests/Locale.test.tsx
```diff
@@ -348,6 +348,57 @@ describe('Locale Component', () => {
...
+
+    test('replaceLanguage succeeds for valid non-duplicate replacement', async () => {
+      const { store } = renderLocaleStep({
+        locale: {
+          languages: ['en_US.UTF-8', 'de_DE.UTF-8'],
+          keyboard: '',
+        },
+      });
+
+      await waitFor(() => {
```
kingsleyzissou: Nitpick: `store.dispatch` is synchronous, were you getting errors here or something? This shouldn't block the PR but maybe we can fix this in a follow up?

src/Components/CreateImageWizard/steps/Locale/tests/Locale.test.tsx
```diff
@@ -348,6 +348,57 @@ describe('Locale Component', () => {
...
+        store.dispatch({
+          type: 'wizard/replaceLanguage',
+          payload: { oldLanguage: 'de_DE.UTF-8', newLanguage: 'nl_NL.UTF-8' },
+        });
```
kingsleyzissou: Nitpick: we should rather use the `replaceImage` action rather than doing this method of dispatch:

```suggestion
         store.dispatch(
           replaceLanguage({ oldLanguage: 'de_DE.UTF-8', newLanguage: 'en_US.UTF-8' })
         );
```

src/Components/CreateImageWizard/steps/Locale/tests/Locale.test.tsx
```diff
@@ -348,6 +348,57 @@ describe('Locale Component', () => {
...
+        locale: {
+          languages: ['en_US.UTF-8', 'de_DE.UTF-8'],
+          keyboard: '',
+        },
+      });
+
+      store.dispatch({
+        type: 'wizard/replaceLanguage',
+        payload: { oldLanguage: 'de_DE.UTF-8', newLanguage: 'en_US.UTF-8' },
+      });
```
kingsleyzissou: Nitpick: we should rather use the `replaceImage` action rather than doing this method of dispatch:

```suggestion
         store.dispatch(
           replaceLanguage({ oldLanguage: 'de_DE.UTF-8', newLanguage: 'en_US.UTF-8' })
         );
```

kingsleyzissou (dismissed):
Looks good. I added some nitpicks inline, but we could do those in a follow up.

kingsleyzissou (approved):

PR #4530, opened 2026-06-12
Replace barrel-file ternaries with webpack platform aliases and usePlatformFeatures

kingsleyzissou:
## Summary

Replace the 19 `process.env.IS_ON_PREMISE` ternaries in barrel files with build-time webpack alias resolution, and introduce a `usePlatformFeatures` hook that translates the raw `isOnPremise` boolean into semantic capability flags for component-level branching.

## Changes

- **Webpack alias swap for hook resolution** — Paired alias files (`index.hosted.ts` / `index.onprem.ts`) replace the ternary-laden barrel files in `backend/`, `contentSources/`, and `useGetEnvironment`. Each webpack config resolves `@/store/api/backend` to the correct platform file at build time, so consumers don't change and the other platform's code never enters the module graph.

- **`usePlatformFeatures` hook** — A thin hook that maps `isOnPremise` to named capabilities (`canCrossArchBuild`, `showDevelopmentReleases`, `securitySectionLabel`, etc.). Migrated 5 components as proof-of-concept; remaining components can adopt incrementally.

- **Cleanup of `process.env.IS_ON_PREMISE` in shared hooks** — `contentSources/hooks.tsx` and `backend/hooks.tsx` now use the existing `selectIsOnPremise` selector instead of reading the env var directly.

- **Test coverage** — Export-surface tests for both alias files, alias-resolution tests verifying vitest wiring, `usePlatformFeatures` unit tests, and updated component tests covering both platform behaviors via `preloadedState`.

## Notes

Build-time platform file resolution is a well-established pattern across the ecosystem:

- **React Native** — Metro bundler resolves `.ios.ts` / `.android.ts` extensions automatically per target platform
- **React Native Web** — extends the convention with `.web.ts` for browser targets
- **Node.js conditional exports** — `package.json` `exports` field resolves different entry points based on conditions (`"browser"`, `"node"`, `"import"`)
- **Preact** — the entire `preact/compat` story is a `resolve.alias` swap for React
- **Webpack docs** — `resolve.alias` is explicitly recommended for environment-specific builds


kingsleyzissou:
Accidetnally created this on osbuild instead of my fork

PR #4532, opened 2026-06-13
Image mode fedora

ochosi:
Based on #4526 

This pull request extends #4526 by essentially three changes:
1. Enable "Image Mode" on Fedora (this was previously blocked by a filter)
2. Derive the architecture by means of `podman inspect`
3. Extract the distribution version to show "Fedora 44" instead of "Fedora"


ochosi:
Most of what this PR tried to do is done in the original PR now, so I'll close it out.

PR #4538, opened 2026-06-16
Nest wizard slice state under submodule keys

kingsleyzissou:
## Summary

Restructure the wizard Redux slice by nesting flat state keys under logical submodule namespaces (`details`, `output`, `filesystem`, `system`, `content`, `cloudProviders`, `registration`, `compliance`). This improves state encapsulation, makes selector boundaries explicit, and sets the foundation for extracting submodule slices independently.

## Changes

- Nest all wizard state keys under domain-specific submodule keys, updating selectors, reducers, and all consumer references across 57 files
- Extract `index.ts` into `slice.ts` with a barrel re-export to keep import paths stable while improving internal file organization
- Convert the slice from a default export to a named export (`wizardSlice`)
- Rename `selectGroups` to `selectPackageGroups` to disambiguate from user groups now at `system.groups`
- Add dedicated per-submodule test files (`compliance.test.ts`, `details.test.ts`, `output.test.ts`) with structural invariant assertions

## Notes

This kind of front loads a lot of the changes. The idea is then to break the nested items into file-based modules, eg:
```
|- src/store/wizard/filesystem
|- src/store/wizard/content
|- src/store/wizard/compliance
...
```

The idea is that this would break up the wizard slice since it has become a bit of a god object. That would be too much change in one go, so that can be a follow up and the changes there would be more contained.

This is pretty mechanical change and most of the diffs are just changes to test fixtures.


kingsleyzissou:
This was just meant to be a reference PR. I'm going to close it and break it up into smaller ones

PR #4539, opened 2026-06-16
Wizard slice: phase 0 — file restructure and filesystem nesting (HMS-10833)

kingsleyzissou:
## Summary

Prepare the wizard Redux slice for modularization by restructuring the file layout and nesting the filesystem state under its own key. This is phase 0 of a larger effort to break up the wizard slice, which has become a god object.

## Changes

- Extract `index.ts` into `slice.ts` with a barrel re-export to support adding sibling modules without growing a single file
- Convert the wizard reducer from a default export to a named export (`wizardReducer`) for consistency with the rest of the module's exports
- Nest `fscMode`, `disk`, `fileSystem`, and `partitioning_mode` under a `filesystem` key, updating selectors, reducers, and all consumer references
- Update filesystem test fixtures and component tests to match the new state shape

## Notes

This front loads the structural changes needed to break the wizard slice into file-based submodules, e.g.:
```
|- src/store/wizard/filesystem
|- src/store/wizard/content
|- src/store/wizard/compliance
...
```

Only the filesystem domain is nested in this PR to keep the diff reviewable. Follow-up PRs will nest the remaining domains and then extract each into its own file-based module.

The diff is mostly mechanical — the bulk of the line count is test fixture updates and selector path changes.

JIRA: [HMS-10833](https://issues.redhat.com/browse/HMS-10833)

[HMS-10833]: https://redhat.atlassian.net/browse/HMS-10833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kingsleyzissou:
See https://github.com/osbuild/image-builder-frontend/pull/4538 for reference

regexowl (approved):
Wow, thanks for looking into this! Breaking up the wizard slice will be a massive improvement ❤️

kingsleyzissou:
> Wow, thanks for looking into this! Breaking up the wizard slice will be a massive improvement ❤️

Thanks!! 🙏  I guess for later down the line we could then move validation into the respective submodule if that makes sense? So filesystem could have a dedicated file for filesystem related validation. Should make things easier to work with :)

regexowl:
> Thanks!! 🙏 I guess for later down the line we could then move validation into the respective submodule if that makes sense? So filesystem could have a dedicated file for filesystem related validation. Should make things easier to work with :)

Yes, sounds good!

We'll be also later changing how the validation works - the Next button should be always enabled and just send you to an invalid input when clicked. Meaning the validation will be more closely coupled with the step, moving it under the step's directory makes perfect sense.

PR #4540, opened 2026-06-16
Cockpit: Add container registry status and login (HMS-10831)

ochosi:
This commit introduces a simple way of verifying that a user is logged in to `registry.redhat.io` through a simple and fairly light-weight run-time test shelling out to `podman search`. This will not only verify that the user is logged in, but also that the registry is reachable.
A subsequent call to `podman login` will help narrow down the failure causes that we can then communicate to the user.

The final implementation of this feature should likely be gated to RHEL.

## Screencast
Here's a screen recording of the feature in action.

https://github.com/user-attachments/assets/92fc7596-590f-4e44-9fef-9719cc20e440

## Local testing
If you would like to test this branch on Fedora, you need to ungate Image Mode:

```
diff --git a/src/Components/CreateImageWizard/steps/ImageOutput/index.tsx b/src/Components/CreateImageWizard/steps/ImageOutput/index.tsx
index 5f554dc3..9341182d 100644
--- a/src/Components/CreateImageWizard/steps/ImageOutput/index.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/index.tsx
@@ -73,7 +73,7 @@ const ImageOutputStep = () => {
           Learn more about <DocumentationButton />.
         </Content>
       </Content>
-      {!(distribution as string).startsWith('fedora') && <BlueprintMode />}
+      <BlueprintMode />
       {isImageMode && <ImageSourceSelect />}
       {!isImageMode && (
         <>
```

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/RegistryAuthStatus.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/RegistryAuthStatus.tsx
@@ -1,0 +1,112 @@
+import React from 'react';
+
+import {
+  Button,
+  Content,
+  Flex,
+  FlexItem,
+  Label,
+  Spinner,
+} from '@patternfly/react-core';
+import {
+  CheckCircleIcon,
+  ExclamationTriangleIcon,
+} from '@patternfly/react-icons';
+
+import type { RegistryAuthStatus as ApiRegistryAuthStatus } from '@/store/api/backend/onprem';
+
+type RegistryAuthState =
+  | ApiRegistryAuthStatus
+  | { status: 'checking' }
+  | { status: 'network-error'; error: string };
+
+type RegistryAuthStatusProps = {
+  authState: RegistryAuthState;
+  onLoginClick: () => void;
+};
+
+const RegistryAuthStatus = ({
+  authState,
+  onLoginClick,
+}: RegistryAuthStatusProps) => {
+  return (
+    <div className='pf-v6-u-mb-sm'>
+      {authState.status === 'checking' && (
+        <Flex
+          alignItems={{ default: 'alignItemsCenter' }}
+          gap={{ default: 'gapSm' }}
+        >
+          <FlexItem>
+            <Spinner size='sm' aria-label='Checking registry authentication' />
+          </FlexItem>
+          <FlexItem>
+            <Content>Checking registry authentication...</Content>
+          </FlexItem>
+        </Flex>
+      )}
+      {authState.status === 'authenticated' && (
+        <Label
+          color='green'
+          variant='outline'
+          icon={
+            <CheckCircleIcon color='var(--pf-t--global--icon--color--status--success--default)' />
+          }
+        >
+          Logged in to registry.redhat.io as {authState.username}
+        </Label>
+      )}
+      {authState.status === 'not-logged-in' && (
+        <Flex
+          alignItems={{ default: 'alignItemsCenter' }}
+          gap={{ default: 'gapSm' }}
+        >
+          <FlexItem>
+            <Label
+              color='orange'
+              variant='outline'
+              icon={
+                <ExclamationTriangleIcon color='var(--pf-t--global--icon--color--status--warning--default)' />
+              }
+            >
+              Not logged in to registry.redhat.io
+            </Label>
+          </FlexItem>
+          <FlexItem>
+            <Button variant='link' onClick={onLoginClick} isInline>
+              Log in
+            </Button>
+          </FlexItem>
+        </Flex>
+      )}
+      {authState.status === 'auth-failed' && (
+        <Flex
+          alignItems={{ default: 'alignItemsCenter' }}
+          gap={{ default: 'gapSm' }}
+        >
+          <FlexItem>
+            <Label
+              color='orange'
+              variant='outline'
+              icon={
+                <ExclamationTriangleIcon color='var(--pf-t--global--icon--color--status--warning--default)' />
+              }
+            >
+              Authentication failed for {authState.username} - credentials may
+              be expired
+            </Label>
+          </FlexItem>
+          <FlexItem>
+            <Button variant='link' onClick={onLoginClick} isInline>
+              Log in
+            </Button>
+          </FlexItem>
+        </Flex>
+      )}
+      {authState.status === 'network-error' && (
+        <Content>Unable to reach registry.redhat.io: {authState.error}</Content>
+      )}
+    </div>
+  );
+};
+
+export default RegistryAuthStatus;
--- a/src/store/api/backend/onprem/composerApi/registry.ts
+++ b/src/store/api/backend/onprem/composerApi/registry.ts
@@ -1,0 +1,54 @@
+import cockpit from 'cockpit';
+
+import { OnPremBuilder, onPremQueryHandler } from '@/store/api/shared';
+
+import type { RegistryAuthStatus, RegistryLoginApiArg } from '../types';
+
+export const registryEndpoints = (builder: OnPremBuilder) => ({
+  getRegistryAuthStatus: builder.query<RegistryAuthStatus, void>({
+    queryFn: onPremQueryHandler(async () => {
+      let username: string;
+      try {
+        const result = await cockpit.spawn(
+          ['podman', 'login', '--get-login', 'registry.redhat.io'],
+          { superuser: 'require' },
+        );
+        username = (result as string).trim();
+      } catch {
+        return { status: 'not-logged-in' };
+      }
+
+      try {
+        await cockpit.spawn(
+          ['podman', 'search', 'registry.redhat.io/rhel10', '--limit', '1'],
+          { superuser: 'require' },
+        );
+      } catch {
+        return { status: 'auth-failed', username };
+      }
+
+      return { status: 'authenticated', username };
+    }),
+  }),
+
+  registryLogin: builder.mutation<RegistryAuthStatus, RegistryLoginApiArg>({
+    queryFn: onPremQueryHandler(
+      async ({ queryArgs: { username, password } }) => {
+        await cockpit
+          .spawn(
+            [
+              'podman',
+              'login',
+              '--username',
+              username,
+              '--password-stdin',
+              'registry.redhat.io',
+            ],
+            { superuser: 'require', err: 'message' },
+          )
+          .input(password);
+        return { status: 'authenticated', username };
+      },
+    ),
+  }),
+});
--- a/src/store/api/backend/onprem/index.ts
+++ b/src/store/api/backend/onprem/index.ts
@@ -14,10 +14,12 @@ export {
   useGetComposeStatusQuery,
   useGetOscapCustomizationsQuery,
   useGetOscapProfilesQuery,
+  useGetRegistryAuthStatusQuery,
   useGetWorkerConfigQuery,
   useLazyExportBlueprintCockpitQuery,
   useLazyGetBlueprintsQuery,
   useLazyGetOscapCustomizationsQuery,
+  useRegistryLoginMutation,
   useUpdateBlueprintMutation,
   useUpdateWorkerConfigMutation,
 } from './composerApi';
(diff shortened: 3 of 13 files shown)


src/Hooks/useRegistryLogin.ts
```diff
@@ -0,0 +1,116 @@
...
+  | { status: 'network-error'; error: string };
+
+type UseRegistryLoginReturn = {
+  authState: RegistryAuthState;
+  isLoggingIn: boolean;
+  checkLoginStatus: () => void;
+  login: (username: string, password: string) => Promise<void>;
+};
+
+export const useRegistryLogin = (): UseRegistryLoginReturn => {
```
kingsleyzissou: While we definitely need this, I think it's implemented at the wrong layer. These are essentially the equivalent of api calls so we should be putting them in `store/api` somewhere. I'm not sure whether we should put it in `store/api/backend` or create a new module for it though since this endpoint doesn't exist in the hosted frontend.

This would also give us rtk query hooks that handle `loading`, `error` and `success` cases too.

src/Hooks/useRegistryLogin.ts
```diff
@@ -0,0 +1,116 @@
...
+  const [authState, setAuthState] = useState<RegistryAuthState>({
+    status: 'checking',
+  });
+  const [isLoggingIn, setIsLoggingIn] = useState(false);
+
+  const checkLoginStatus = useCallback(() => {
+    setAuthState({ status: 'checking' });
+
+    cockpit
+      .spawn(['podman', 'search', 'registry.redhat.io/rhel10', '--limit', '1'], {
```
kingsleyzissou: I think we should change this to `async` code too:

```
await cockpit.spawn(...)
```

I've added higher order error functions that will catch any of the errors there and simplify all the `.then` and `.catch` cases which should resolve the "promise hell" here and all the nesting.

src/Hooks/useRegistryLogin.ts
```diff
@@ -0,0 +1,116 @@
...
+            ],
+            { superuser: 'require', err: 'message' },
+          )
+          .input(password);
+
+        checkLoginStatus();
+      } catch (err) {
+        const cockpitError = err as Error & { exit_status?: number };
+
+        if (cockpitError.exit_status === 125) {
```
kingsleyzissou: We would probably need to add a case for this in the error handler if we switched it to an rtk query mutation.

kingsleyzissou (commented):
Nice, seems to be working well. I think we should push the login and auth status down the stack and into the store as mentioned inline. This would also give us a lot of the lifecycle management benefits there too which is a bonus

src/Hooks/useRegistryLogin.ts
ochosi: Replaced the nested `.then`/`.catch` chains with `async`/`await` using the existing `onPremQueryHandler` wrapper.

src/Hooks/useRegistryLogin.ts
ochosi: The `getRegistryAuthStatus` query now defensively wraps all `cockpit.spawn` calls in try/catch blocks. This is necessary because `podman search` can succeed even without stored credentials (anonymous/cached access), so the subsequent `podman login --get-login` call could throw. All failure paths now resolve to meaningful status values (not-logged-in or auth-failed) rather than propagating as RTK Query errors.

src/Hooks/useRegistryLogin.ts
ochosi: Moved the registry auth logic from a custom React hook (`useRegistryLogin`) into `src/store/api/backend/onprem/composerApi/registry.ts` as proper RTK Query endpoints. As you suggested, this gives us automatic loading/error/success state management, cache invalidation via tags, and keeps the API call logic in the `store/api/` layer where it belongs.

ochosi:
@kingsleyzissou Thanks for the review! I've reworked the registry login logic into the RTK Query API layer (responded to your comments directly).

The behavior/UX is unchanged for now. As soon as we're happy with the architecture, I'll rework the modal as per some suggestions from Shayna.

src/store/api/backend/onprem/index.ts
```diff
@@ -14,12 +14,15 @@ export {
...
   useGetWorkerConfigQuery,
   useLazyExportBlueprintCockpitQuery,
   useLazyGetBlueprintsQuery,
   useLazyGetOscapCustomizationsQuery,
+  useRegistryLoginMutation,
   useUpdateBlueprintMutation,
   useUpdateWorkerConfigMutation,
 } from './composerApi';
 
+export type { RegistryAuthStatus } from './composerApi/registry';
```
kingsleyzissou: Nitpick: this type could probably jsut get moved to the `./types` directory for this module

src/store/api/backend/onprem/composerApi/registry.ts
```diff
@@ -0,0 +1,83 @@
...
+      // since some registries allow anonymous search.
+      try {
+        const result = await cockpit.spawn(
+          ['podman', 'login', '--get-login', 'registry.redhat.io'],
+          { superuser: 'require' },
+        );
+        return { status: 'authenticated', username: (result as string).trim() };
+      } catch {
+        return { status: 'not-logged-in' };
+      }
```
kingsleyzissou, line 28: I know this is trying to check if the user is already logged in. But maybe we should move this check first.
If it fails, we fail early and we fail fast. This removes the need for the nested `try/catch`

Then we could try the search to see if the user is logged in. I guess in my mind the flow could be:

 1. --get-login fails → no credentials stored → not-logged-in
 2. --get-login succeeds → credentials exist, but are they valid? → podman search → succeeds → authenticated
 3. --get-login succeeds → credentials exist → podman search → fails → auth-failed

Something like:

```suggestion
     let username: string;
     try {
       const result = await cockpit.spawn(
         ['podman', 'login', '--get-login', 'registry.redhat.io'], ...
       );
       username = (result as string).trim();
     } catch {
       return { status: 'not-logged-in' };
     }
  
     try {
       await cockpit.spawn(
         ['podman', 'search', 'registry.redhat.io/rhel10', '--limit', '1'], ...
       );
     } catch {
       return { status: 'auth-failed', username };
     }
  
     return { status: 'authenticated', username };
 ```

File: src/store/api/backend/onprem/composerApi/registry.ts [hunk 6]
```diff
@@ -0,0 +1,83 @@
...
+            .input(password);
+        } catch (loginError: unknown) {
+          const err = loginError as Error & { exit_status?: number };
+          if (err.exit_status === 125) {
+            return { status: 'auth-failed', username };
+          }
+          throw loginError;
+        }
+
+        return { status: 'authenticated', username };
```
kingsleyzissou: Nitpick: I think we could drop the `try/catch` here for this one. If podman login fails, onPremQueryHandler catches it, converts it to `OnPremError`, and the modal displays it through `loginMutationError`

```suggestion
        await cockpit.spawn([...], { superuser: 'require', err: 'message' }).input(password);
        return { status: 'authenticated', username };
```

src/store/api/backend/onprem/composerApi/registry.ts
```diff
@@ -0,0 +1,83 @@
+import cockpit from 'cockpit';
+
+import { OnPremBuilder, onPremQueryHandler } from '@/store/api/shared';
+
+export type RegistryAuthStatus =
+  | { status: 'authenticated'; username: string }
+  | { status: 'not-logged-in' }
+  | { status: 'auth-failed'; username: string };
```
kingsleyzissou: Nitpick: since this is exported and used elsewhere, we should probably put it in with the `./types` folder for this module.

src/store/api/backend/onprem/composerApi/registry.ts
```diff
@@ -0,0 +1,83 @@
...
+
+type RegistryLoginApiArg = {
+  username: string;
+  password: string;
+};
```
kingsleyzissou: Nitpick: I would also probably move this type alongside the other `ApiArg` custom types.

src/Components/CreateImageWizard/steps/ImageOutput/components/RegistryAuthStatus.tsx
```diff
@@ -0,0 +1,112 @@
...
+  CheckCircleIcon,
+  ExclamationTriangleIcon,
+} from '@patternfly/react-icons';
+
+type RegistryAuthState =
+  | { status: 'checking' }
+  | { status: 'authenticated'; username: string }
+  | { status: 'not-logged-in' }
+  | { status: 'auth-failed'; username: string }
+  | { status: 'network-error'; error: string };
```
kingsleyzissou, line 21: Nitpick: I think there is a bit of duplication with this type, could probably combine it with the ones in the api one and import it from there instead.

kingsleyzissou (commented):
Nice looks a lot better with the changes and moving it to the rtk query function. Just a couple of nitpicks inline

src/store/api/backend/onprem/composerApi/registry.ts
ochosi: As far as I could find out, `podman login --get-login` doesn't actively check anything beyond having an authentication file set up for a registry locally. So if you wanted to pull from any registry, you'd still have to `podman login` first.

The reason to frontload `podman search` is that it quickly tells you a couple of things:
1. Is the user already logged in (for real)
2. Is the registry reachable (network)

If both conditions are satisfied, we can show a checkmark and move on.
If the registry isn't reachable, we can now show a network error. Checking the auth file via `--get-login` still makes sense as far as I could tell, though (this is mostly about auth caching).

kingsleyzissou: Ya that's true, but we definitely won't be able to search if there is no authfile (as far as I can tell).
But that's fine, if we want to stick with this strategy I would maybe extract the try/catch blocks into their own helpers (they could go into the [podman helpers directory](src/store/api/backend/onprem/composerApi/helpers/podman)):

```typescript
async function canSearch(): Promise<boolean> { ... }
async function getStoredUsername(): Promise<string | null> { ... }
```

And then in the function body it would just simplify to:
```typescript
const searchOk = await canSearch();
const username = await getStoredUsername();

if (!username) {
  return { status: 'not-logged-in' };
}

if (!searchOk) {
  return`` { status: 'auth-failed', username };
}

return { status: 'authenticated', username };
```

src/store/api/backend/onprem/composerApi/registry.ts
```diff
@@ -0,0 +1,83 @@
...
+      try {
+        await cockpit.spawn(
+          ['podman', 'search', 'registry.redhat.io/rhel10', '--limit', '1'],
+          { superuser: 'require' },
+        );
+      } catch {
+        // Search failed — determine if it's an auth issue or something else
+        // by checking whether credentials are stored locally.
+        try {
+          const result = await cockpit.spawn(
```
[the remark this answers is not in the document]
ochosi: This was also noted by the human reviewer, hence marking as resolved.

src/store/api/backend/onprem/composerApi/registry.ts
ochosi: Let's actually stick to your original suggestion. It makes sense to me now. (Needed to think it through again)

ochosi:
FWIW Snyk complained about hardcoded credentials. I waived that finding because it's fake credentials as part of a test. See: https://github.com/osbuild/image-builder-frontend/pull/4540/changes#diff-aa68acb8fd0b677fc15d583661799981bf325e65a7b0b3af93c1fcf02975fd0bR119

kingsleyzissou (approved):
Amazing! Thanks for this. Not sure if we need to wait for anything to land in cockpit podman?

ochosi:
> Amazing! Thanks for this. Not sure if we need to wait for anything to land in cockpit podman?

No, the linking to Cockpit Podman is not part of this branch/pull request. So we can move ahead with this.
I just linked to it when I thought it would move more quickly :innocent:

PR #4544, opened 2026-06-17
Wizard slice: phase 1 — nest next set of state domains (HMS-10846)

kingsleyzissou:
## Summary

Continue the wizard slice modularization started in #4539 by nesting more state domains under their own keys.

## Changes

- Nest `cloudProviders` state (`aws`, `azure`, `gcp`) under a `cloudProviders` key with updated selectors, reducers, and consumer references
- Nest `registration` state (`activationKey`, `registrationType`) under a `registration` key
- Nest `compliance` state (`oscap` profile and policy fields) under a `compliance` key with new dedicated tests
- Nest `content` state (`packages`, `repositories`, `customRepositories`, etc.) under a `content` key
- Update all component tests and fixtures to match the new state shapes

## Notes

Follow-up to #4539 (phase 0), which restructured the file layout and nested the `filesystem` domain.

The diff is mostly mechanical — the bulk of the line count is selector path changes and test fixture updates.

JIRA: [HMS-10846](https://issues.redhat.com/browse/HMS-10846)

[HMS-10846]: https://redhat.atlassian.net/browse/HMS-10846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/setup.ts
+++ b/src/test/setup.ts
@@ -102,6 +102,9 @@ beforeAll(() => {
     if (message.includes('Maximum update depth exceeded')) {
       return;
     }
+    if (message.includes('which is more than the warning threshold of 32ms')) {
+      return;
+    }
     throw new Error(`Console warning:\n${message}`);
   });
 });
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/Aws/tests/AwsTarget.test.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/Aws/tests/AwsTarget.test.tsx
@@ -132,10 +132,13 @@ describe('AWS Component', () => {
 
     test('displays error when field is cleared after having a value', async () => {
       renderAwsStep({
-        aws: {
-          accountId: '123456789012',
-          shareMethod: 'manual',
-          source: undefined,
+        cloudProviders: {
+          ...initialState.cloudProviders,
+          aws: {
+            accountId: '123456789012',
+            shareMethod: 'manual',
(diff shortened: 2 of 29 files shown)


src/test/setup.ts
```diff
@@ -102,6 +102,9 @@ beforeAll(() => {
     if (message.includes('Maximum update depth exceeded')) {
       return;
     }
+    if (message.includes('which is more than the warning threshold of 32ms')) {
```
regexowl, line 105: Ah, right, is that the middleware error? Should be fine to ignore indeed.

regexowl (approved):

src/test/setup.ts
kingsleyzissou: Whoops that was a workaround that I left in and was meaning to come back to

kingsleyzissou: Oh well

regexowl: Oops, my bad. Was too eager to merge.

I remember looking into this one a while ago and arriving to a conclusion that we either have a bit too large large mock states (which I mean, unsure) or the CI runners are simply slow. And since the error appeared just occasionally, I blamed it on the runners ._.

kingsleyzissou: Ya it's fine. It started happening a lot lately when running the tests locally. Might investigate further

PR #4547, opened 2026-06-17
tests: add image mode boot test

jlsherrill:
adds a boot test for RHEL image mode qcow2

HMS-10562


regexowl (approved):
Thanks! Looks good to me

PR #4553, opened 2026-06-19
Fix blueprint import failing with duplicate custom repositories (HMS-10697)

kingsleyzissou:
## Summary

Fix blueprint import failing when custom repositories already exist in the user's content sources. The `bulk_import` endpoint errors if any repo URLs are duplicates, causing the entire repository import to fail silently. This PR pre-filters existing repos so only genuinely new ones are sent to `bulk_import`.

## Changes

- Query existing repositories by URL before calling `bulk_import`, partitioning into existing vs new repos
- Only send genuinely new repos to the `bulk_import` endpoint, avoiding the duplicate URL error
- Widen `mapToCustomRepositories` to accept both list and import response types, enabling reuse for existing repo mapping
- Export `useLazyListRepositoriesQuery` from the content sources API for imperative repo lookups
- Gracefully return already-existing repos even if the bulk import of new repos fails

### DIFF
--- a/src/Components/Blueprints/tests/ImportBlueprintModal.test.tsx
+++ b/src/Components/Blueprints/tests/ImportBlueprintModal.test.tsx
@@ -1,0 +1,337 @@
+import React from 'react';
+
+import { screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import { server } from '@/test/mocks/server';
+import { renderWithRedux } from '@/test/testUtils';
+
+import {
+  createBlueprintJson,
+  createFetchHandler,
+  existingRepo,
+  fetchMock,
+} from './mocks';
+
+import { ImportBlueprintModal } from '../ImportBlueprintModal';
+
+fetchMock.enableMocks();
+
+const mockAddNotification = vi.fn();
+vi.mock(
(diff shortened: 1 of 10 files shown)


src/Components/Blueprints/tests/ImportBlueprintModal.test.tsx
```diff
@@ -0,0 +1,337 @@
...
+  });
+
+  test('imports all repos when none already exist', async () => {
+    const newRepoUrl = 'http://brand-new.repo.example.com/x86_64/';
+    const importedRepo = {
+      uuid: 'new-uuid-1',
+      url: newRepoUrl,
+      name: 'brand-new-repo',
+      warnings: [],
+    };
```
regexowl, line 122: Just a nitpick - this appears twice in the tests, could be maybe moved to fixtures. But wouldn't block on this.

regexowl (approved):
Nice!

src/Components/Blueprints/tests/ImportBlueprintModal.test.tsx
kingsleyzissou: Good point, I'll create a quick follow up

PR #4579, opened 2026-06-28
Wizard: always enable Next and Review buttons (HMS-10658)

mgold1234:
Instead of disabling Next/Review when validation fails, keep them
  enabled and scroll to the first error on click. Create/Save buttons
  on the Review step behave the same way:
  always clickable, but navigate back to the first error step when validation fails.

  A new ValidationContext forces pristine fields to reveal their
  errors so the user sees exactly what needs fixing.

  Note: The activation key loading fix (stop blocking Next while
  the activation key query is in flight) will be submitted in a
  separate PR.
JIRA :HMS-10658

### DIFF
--- a/src/Components/CreateImageWizard/components/CustomWizardFooter.tsx
+++ b/src/Components/CreateImageWizard/components/CustomWizardFooter.tsx
@@ -7,27 +7,71 @@ import {
   WizardFooterWrapper,
 } from '@patternfly/react-core';
 import useChrome from '@redhat-cloud-services/frontend-components/useChrome';
+import { flushSync } from 'react-dom';
 
 import { AMPLITUDE_MODULE_NAME } from '@/constants';
+import { useAppDispatch } from '@/store/hooks';
+import {
+  resetForceShowErrors,
+  setForceShowErrors,
+} from '@/store/slices/wizard';
+
+import { scrollToFirstError } from '../utilities/scrollToFirstError';
 
 type CustomWizardFooterPropType = {
   disableBack?: boolean;
-  disableNext: boolean;
+  hasErrors: boolean;
   beforeNext?: () => boolean;
   isOnPremise: boolean;
 };
 
 export const CustomWizardFooter = ({
   disableBack,
-  disableNext,
+  hasErrors,
   beforeNext,
   isOnPremise,
 }: CustomWizardFooterPropType) => {
   const { goToNextStep, goToPrevStep, goToStepById, close, activeStep } =
     useWizardContext();
   const { analytics } = useChrome();
+  const dispatch = useAppDispatch();
   const reviewAndFinishBtnID = 'wizard-review-and-finish-btn';
   const cancelBtnID = 'wizard-cancel-btn';
+
+  const handleNext = () => {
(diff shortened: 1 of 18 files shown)


src/Components/CreateImageWizard/utilities/ValidationContext.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { createContext, useCallback, useContext, useState } from 'react';
+
+type ValidationContextType = {
+  forceShowErrors: boolean;
+  setForceShowErrors: () => void;
+};
+
+const ValidationContext = createContext<ValidationContextType>({
```
kingsleyzissou: I'm not sure adding context is the right choice here. Essentially all our validation is basically state
and at some point I would like to move that to the wizard slice too. That's too big a change for this
PR (don't worry that's not something I'm asking you to do).

The thing is, if we do merge this, when I do get to migrating the validation this won't be needed anymore.
I might be able to do that before I go on PTO in two weeks time, but it also may only be after that

src/Components/CreateImageWizard/components/CustomWizardFooter.tsx
```diff
@@ -7,28 +7,68 @@ import {
...
+  };
+
+  const handleReview = () => {
+    if (!isOnPremise) {
+      analytics.track(`${AMPLITUDE_MODULE_NAME} - Button Clicked`, {
+        module: AMPLITUDE_MODULE_NAME,
+        button_id: reviewAndFinishBtnID,
+        active_step_id: activeStep.id,
+      });
+    }
```
kingsleyzissou: This should probably only be fired if we don't have errors because this is going to fire every time

src/Components/CreateImageWizard/components/CustomWizardFooter.tsx
kingsleyzissou, line 86: This probably also needs `resetForceShowErrors()`, we only added it for going to the next step but we need to add it for going back too

src/Components/CreateImageWizard/utilities/ValidationContext.tsx
mgold1234: @kingsleyzissou this is great point, I will try to see see if change the implementation to wizard slice is simple

kingsleyzissou: Oh sorry I wasn't asking you to do that lol, I was thinking out loud a bit. But I see that you've made the changes. I am busy going through them now

src/store/slices/wizard/validation/slice.ts
```diff
@@ -0,0 +1,31 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+import { initialState } from './state';
+
+import { initializeWizard, loadWizardState } from '../actions';
+
+export const validationSlice = createSlice({
```
kingsleyzissou, line 7: Nice, it's a lot of boiler plate but at least it is more inline with what we do rather than introducing `useContext` into the project. Not that it's bad, it would just be another source of truth to manage

kingsleyzissou (approved):

PR #4584, opened 2026-06-29
tests: add coverage for PW + vitest (HMS-9550)

ksiekl:
Adds `Istanbul` for tracking test coverage for PW, also switches vitest from V8's coverage engine to Istanbul, so that the reports are easily merged and presented by Codecov.

- `vitest.config.ts` - changed the config to use Istanbul, the line `reportOnFailure: true,` means the coverage report is generated even if some tests fail, not sure if we want that
- `package.json` - Removed v8 we used for vitest and added Istanbul for coverage, along with the `nyc` package, which it uses to read JSON data and create the lcov report. Also added the `@jsdevtools/coverage-istanbul-loader` - FEC builds its own `module.rules` and ignores any rules you put in fec.config.js. But FEC does pass plugins through to Webpack. A webpack plugin's `apply(compiler)` method runs when webpack starts, and `compiler.options.module.rules` gives direct access to the actual rules webpack will use — so the plugin can inject the Istanbul loader there.

- `fec.config.js` contains the `IstanbulCoveragePlugin`
- `.github/workflows/playwright.yml` - added the coverage variable to start frontend with the coverage from now on,
- `playwright/fixtures/coverage.ts` - the coverage is tracked in the `window.__coverage__` object
- the coverage is checked for every PW test (`playwright/fixtures/customizations.ts`)
- `webpack.dev-coverage.config.js` - to generate reports locally


package.json
```diff
@@ -109,6 +111,7 @@
     "test:integration": "TZ=UTC vitest run src/test",
     "test:unit": "TZ=UTC vitest run --exclude='src/test/**' --fileParallelism",
     "test:cockpit": "src/test/cockpit-tests.sh",
+    "coverage:playwright": "nyc report --temp-dir .nyc_output --report-dir coverage/playwright --reporter lcov --reporter text",
```
kingsleyzissou: This is definitely a nitpick and could be done in a follow up. But maybe we could still follow the convention we have here and either prefix with `test:...` or maybe we should have an `e2e` section?

So like:
```
"e2e": "<regular playwright>",
"e2e:coverage": "<playwright with coverage>" 
```

`e2e` rather than `playwright` because our testing framework could change in the future.

kingsleyzissou (commented):
Neat, looks good, I just had one inline comment.

And then also your commit message has some `x`'s in it, I guess from squashing.

package.json
ksiekl: Oh, thanks for noticing the commit message! I switched to lazygit and have a different flow now, so this got past me, heh. And yes, I can switch it to `e2e`, I don't really think we will switch from PW anytime soon, but sure.

kingsleyzissou: Lazygit is great. You can use the keyboard shortcut f to fix the commit up into the previous one...
It's like squash without the commit message 🙂

Ya I know it's unlikely to change, but if we ever do, the command survives the tool

kingsleyzissou (approved):
Thanks!

package.json
ksiekl: thanks! didn't know that :smile:

PR #4585, opened 2026-06-29
Prepare store types for request mapper migration

kingsleyzissou:
Prepare store types for request mapper migration

## Summary

Establishes type-safe boundaries around edge image types and relocates repo conversion logic to the store layer, preparing the ground for moving the request mapper into the store.

## Changes

- Introduce `SupportedImageTypes = Exclude<ImageTypes, EdgeType>` and narrow the output slice's `imageTypes` field and action payloads to reject edge types at compile time
- Add `isEdgeType` and `isSupportedImageType` type guards with full test coverage, composing existing guards for DRY narrowing
- Filter edge types from blueprints on load in `requestMapper` so the wizard only operates on supported types
- Move `convertSchemaToIBCustomRepo` and `convertSchemaToIBPayloadRepo` from component utilities to `src/store/slices/wizard/content/utilities.ts` — pure data transforms belong at the store boundary, not in React components


regexowl (approved):
Looks good!

PR #4594, opened 2026-07-03
Wizard: Review step revamp (HMS-10785)

ksiekl:
Edit the Review step according to mocks.

Renamed `required` to `isStandalone` in `RestrictionStrategy`. The field controls whether the Users customization gets its own wizard step instead of being nested under Advanced Settings. The old name suggested a data validation constraint, but it actually determines where the section is placed in the UI. 

review-revamp

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/components/Filesystem.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/components/Filesystem.tsx
@@ -1,51 +1,171 @@
 import React from 'react';
 
-import { Label } from '@patternfly/react-core';
+import { Flex } from '@patternfly/react-core';
 
 import { useAppSelector } from '@/store/hooks';
 import {
+  selectDiskMinsize,
+  selectDiskUnit,
+  selectFilesystemPartitions,
   selectFscMode,
-  selectFSConfigurationsCount,
   selectImageMinSize,
+  selectPlainPartitions,
+  selectVolumeGroups,
 } from '@/store/slices';
 
-import { ReviewGroup } from '../../shared';
+import { FlexColumn, ReviewGroup, ReviewSection } from '../../shared';
 import { Hideable } from '../../types';
 
 export const Filesystem = ({ shouldHide }: Hideable) => {
-  const partitionType = useAppSelector(selectFscMode);
+  const mode = useAppSelector(selectFscMode);
   const minSize = useAppSelector(selectImageMinSize);
-  const configurations = useAppSelector(selectFSConfigurationsCount);
+  const diskMinsize = useAppSelector(selectDiskMinsize);
+  const diskUnit = useAppSelector(selectDiskUnit);
+  const plainPartitions = useAppSelector(selectPlainPartitions);
+  const volumeGroups = useAppSelector(selectVolumeGroups);
+  const basicPartitions = useAppSelector(selectFilesystemPartitions);
 
   if (shouldHide) {
     return null;
   }
 
+  const partitioningLabel =
+    mode === 'automatic'
+      ? 'Automatic partitioning'
+      : mode === 'advanced'
+        ? 'Advanced disk partitioning'
+        : 'Manual partitioning';
+
   return (
-    <>
+    <ReviewSection title='File system configurations'>
       <ReviewGroup
-        heading='Filesystem configurations'
-        className={partitionType !== 'basic' ? 'pf-v6-u-mb-md' : ''}
-        description={
-          <>
-            {configurations > 0 ? configurations : null}
-            <Label
-              className={configurations > 0 ? 'pf-v6-u-ml-md' : ''}
-              color='yellow'
-              isCompact
-            >
-              {partitionType === 'automatic' ? 'Automatic' : 'Manual'}
-            </Label>
-          </>
-        }
+        heading='Partitioning type'
+        description={partitioningLabel}
       />
-      {partitionType !== 'automatic' && minSize && (
+      {mode === 'basic' && minSize !== undefined && (
         <ReviewGroup
           heading='Image size (minimum)'
           description={minSize < 1 ? 'Less than 1GiB' : `${minSize} GiB`}
-          className='pf-v6-u-mb-md'
         />
       )}
-    </>
+      {mode === 'basic' && basicPartitions.length > 0 && (
+        <ReviewGroup
+          heading='Partitions'
+          description={
+            <Flex>
+              <FlexColumn
+                heading='Mount point'
+                labelKey='basic-partition-mount'
+                items={basicPartitions.map((p) => p.mountpoint)}
+              />
+              <FlexColumn
+                heading='Minimum'
+                labelKey='basic-partition-size'
+                items={basicPartitions.map((p) => p.min_size)}
+              />
+              <FlexColumn
+                heading='Unit'
+                labelKey='basic-partition-unit'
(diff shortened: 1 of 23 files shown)


src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
```diff
@@ -34,39 +40,44 @@ const AdvancedSettingsOverview = ({
   oscapKernelArgs = [],
   oscapServices,
 }: AdvancedSettingsOverviewProps) => {
+  const usersHidden =
+    restrictions.users.shouldHide || restrictions.users.required;
```
kingsleyzissou: These are kind of opposites though. If users are required, then `restrictions.users.shouldHide` should be false.

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
```diff
@@ -34,39 +40,44 @@ const AdvancedSettingsOverview = ({
...
+  const users = useAppSelector(selectNonEmptyUsers);
+  const userGroups = useAppSelector(selectNonEmptyUserGroups);
+  const hasUsers = !usersHidden && users.length > 0;
+  const hasUserGroups = !usersHidden && userGroups.length > 0;
```
kingsleyzissou: We could create derived selectors here for `hasUsers` and `hasUserGroups`

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
```diff
@@ -34,39 +40,44 @@ const AdvancedSettingsOverview = ({
...
+        <Kernel
+          shouldHide={restrictions.kernel.shouldHide}
+          oscapKernelArgs={oscapKernelArgs}
+        />
+        <Services
+          shouldHide={restrictions.services.shouldHide}
+          oscapServices={oscapServices}
+        />
+        <Firewall shouldHide={restrictions.firewall.shouldHide} />
+        {hasUserGroups && (
```
kingsleyzissou: This could get moved into the `shouldHide` prop because it's doing the same thing

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
```diff
@@ -34,39 +40,44 @@ const AdvancedSettingsOverview = ({
...
+          <ReviewSection title='Groups'>
+            <UserGroups shouldHide={false} />
```
kingsleyzissou:
```suggestion
            <UserGroups shouldHide={!hasGroups} />
```

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
```diff
@@ -34,39 +40,44 @@ const AdvancedSettingsOverview = ({
...
+          </ReviewSection>
+        )}
+        {hasUsers && (
+          <ReviewSection title='Users'>
+            <Users shouldHide={false} />
```
kingsleyzissou: Same:

```suggestion
            <User shouldHide={!hasUsers} />
```

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/components/Filesystem.tsx
```diff
@@ -1,51 +1,180 @@
...
   }
 
+  const partitioningLabel =
+    mode === 'automatic'
+      ? 'Automatic partitioning'
+      : mode === 'advanced'
+        ? 'Advanced disk partitioning'
+        : 'Manual partitioning';
+
+  const plainPartitions =
```
kingsleyzissou: This could be a derived selector with `createSelector` in the filesystem slice.

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/components/Filesystem.tsx
```diff
@@ -1,51 +1,180 @@
...
+    mode === 'advanced' ? diskPartitions.filter((p) => p.type !== 'lvm') : [];
+
+  const volumeGroups =
```
kingsleyzissou: This could also be a derived selector with createSelector in the filesystem slice.

kingsleyzissou (commented):
Nice, this is looking good. I had a few inline suggestions

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
ksiekl: Other components like Locale have their own ReviewSection internally and return null when hidden, so no
empty heading/divider renders.

Users and UserGroups can't do the same because ImageOverview also renders them, without a ReviewSection wrapper. Since both places share the same component, the ReviewSection has to stay outside in the AdvancedSettings index, and the conditional rendering stays too to avoid an empty heading and divider.

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
ksiekl: This condition was not needed at all, actually, because if there are no users or no groups, we don't want to render that whole section at all, and that's already checked by the condition above it, so I deleted this and made UserGroups Partial, so they don't need the shouldHide argument at all times. Thanks for spotting this!

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
ksiekl: same as above

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
kingsleyzissou: This was a headwrecker, but I've gone through the code and tested it and it all works

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
kingsleyzissou: Thinking about this further (and this can absolutely be a follow up since this works). Maybe we should lift the should hide to the `ReviewSection` component? And that could return the `null` there? I guess the tradeoff is we'd be passing it down through two components.

But that shouldn't block this PR. And it was a bit of a thought in terms of keeping things consistent

kingsleyzissou (approved):
Looks good.

Maybe we could possibly add a margin top to the line separator, to give a little bit more breathing room, but I know that's a bit tricky because a lot of the content is dynamic here.

Screenshot for reference (timezone in this one):

Image

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
ksiekl: Thanks for looking into it! It sure was :nerd_face:

src/Components/CreateImageWizard/steps/Review/components/AdvancedSettings/index.tsx
ksiekl: sounds nice, I'll check it out in a follow up then

ksiekl:
We could, I will also look into it in another PR, this is huge as is. Thank you for review!

PR #4597, opened 2026-07-03
Move target environment data transformations into RTK Query derived endpoints (HMS-10955)

kingsleyzissou:
## Summary

Replace inline `selectFromResult` transformers and `useMemo` normalization in `TargetEnvironment` with derived RTK Query endpoints that delegate to the existing `getArchitectures` and `getDistributions` queries. This ensures both hosted and on-prem (cockpit) backends are handled correctly.

## Changes

- Add `getArchitectureEnvironments` and `getDistributionEnvironments` derived endpoints that dispatch to the existing backend queries via `api.dispatch` with `subscribe: false`, returning a common `CategorizedEnvironments` shape
- Simplify `TargetEnvironment` by consuming the derived hooks directly instead of transforming query results in-component
- Extract `isOnlyNetworkInstallerSelected` and `isOtherEnvironmentSelected` into memoized selectors in the output slice for reuse outside the component

### DIFF
--- a/src/store/api/backend/derived.ts
+++ b/src/store/api/backend/derived.ts
@@ -1,0 +1,133 @@
+import type { EndpointBuilder } from '@reduxjs/toolkit/query/react';
+
+import {
+  isPrivateCloud,
+  isPublicCloud,
+  type MiscFormatType,
+  type PrivateCloudType,
+  type PublicCloudType,
+} from '@/store/slices/wizard';
+
+import type {
+  BootcDistributionItem,
+  DistributionItem,
+  Distributions,
+} from './hosted';
+import { imageBuilderApi } from './hosted/enhancedImageBuilderApi';
+import { composerApi } from './onprem/enhancedComposerApi';
+
+export type CategorizedEnvironments = {
+  publicClouds: PublicCloudType[];
+  privateClouds: PrivateCloudType[];
+  miscFormats: MiscFormatType[];
+  hasEnvironments: boolean;
+};
+
+export type ArchitectureEnvironmentsResult = CategorizedEnvironments;
+
+export type DistributionEnvironmentsResult = CategorizedEnvironments & {
+  distributions: BootcDistributionItem[];
+};
+
+const isBootcDistribution = (
+  item: DistributionItem | BootcDistributionItem,
+): item is BootcDistributionItem => 'distro' in item;
+
+export const categorizeEnvironments = (
+  environments: string[],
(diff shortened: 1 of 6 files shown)


regexowl:
The test failures seem relevant. I can see the "Couldn't fetch target environments" alert in cockpit when opening the Wizard.

kingsleyzissou:
> The test failures seem relevant. I can see the "Couldn't fetch target environments" alert in cockpit when opening the Wizard.

Annoying, it's a race condition. It looks like `RHEL-10` is getting sent through as the distribution

kingsleyzissou:
> > The test failures seem relevant. I can see the "Couldn't fetch target environments" alert in cockpit when opening the Wizard.
> 
> Annoying, it's a race condition. It looks like `RHEL-10` is getting sent through as the distribution

Nope, I was sending the requests to the `osbuild-cockpit` socket. Fixed it now

src/store/api/backend/derived.ts
```diff
@@ -0,0 +1,124 @@
...
+            { subscribe: false },
+          ),
+        );
+
+        if (result.error) {
+          return { error: result.error };
+        }
+
+        const distributions =
+          result.data as GetDistributionsApiResponse as BootcDistributionItem[];
```
regexowl: Same here for the `as GetDistributionsApiResponse`, it would suppress the "possibly undefined". Adding a safeguard might be nice to keep from crashing if the data can be undefined.

src/store/api/backend/derived.ts
```diff
@@ -0,0 +1,124 @@
...
+        );
+
+        if (result.error) {
+          return { error: result.error };
+        }
+
+        const architectures = result.data as GetArchitecturesApiResponse;
+        const imageTypes =
+          (architectures as ArchitectureItem[]).find((a) => a.arch === arch)
+            ?.image_types ?? [];
```
regexowl: The casting masks away possibly undefined architectures. Hmm, but could those really be undefined? Maybe there could be a safeguard added to make it a bit more obvious even if it's a case that shouldn't happen.

kingsleyzissou: Ya this is valid. I can add a typeguard here for sure

src/store/api/backend/derived.ts
kingsleyzissou: Same here, I could add a typeguard

regexowl (approved):
Thanks, looks great!

PR #4610, opened 2026-07-09
Wizard: lift shouldHide prop to the ReviewSection component

ksiekl:
moves the `shouldHide` prop higher so that components pass their combined hide condition to `ReviewSection` instead of doing their own early return


regexowl (approved):
Looks good! 🚀

PR #4611, opened 2026-07-09
webpack: Remove unused eslint-disable comment for deprecated rule

ksiekl:
The `@typescript-eslint/no-var-requires` rule was removed from the ESLint config, making the disable comment unnecessary. It was automatically cleaned up by npm run lint.


regexowl (approved):
Nice! Looks like the rule was deprecated in favor or `no-require-imports`

PR #4612, opened 2026-07-09
  Wizard: swap Users and Groups section order (HMS-10791)

mgold1234:
Show Users before Groups since users are needed almost always
  while groups are rarely used.

JIRA: HMS-10791


regexowl:
Looks good! Can you please rebase and resolve the conflict?

regexowl (approved):
Nice, thanks!

PR #4621, opened 2026-07-13
Wizard/Review: Make "Target environments" heading bigger (HMS-10792)

regexowl:
Make the heading bigger for cleaner visual hierarchy between "Target environment" and subcategories (private cloud, misc formats).

Before:
image

After:
image


ksiekl (approved):
🚀

PR #4622, opened 2026-07-13
Wizard: stop blocking Next while activation key  is loading (HMS-10658)

mgold1234:
When the activation key query is in flight, the Next button was
  blocked until the query resolved. This fix removes isFetching from
  the validation condition so navigation is not blocked while the
  query loads.

  If no activation key is selected, clicking Next now scrolls to
  the Registration section and shows a "Missing activation key"
  error inline.

JIRA: (HMS-10658)

### DIFF
--- a/src/Components/CreateImageWizard/steps/Registration/components/ActivationKeysList.tsx
+++ b/src/Components/CreateImageWizard/steps/Registration/components/ActivationKeysList.tsx
@@ -23,6 +23,7 @@ import { useAddNotification } from '@redhat-cloud-services/frontend-components-n
 import { v4 as uuidv4 } from 'uuid';
 
 import ExternalLinkButton from '@/Components/CreateImageWizard/utilities/ExternalLinkButton';
+import { useRegistrationValidation } from '@/Components/CreateImageWizard/utilities/useValidation';
 import { ACTIVATION_KEYS_URL, CDN_PROD_URL, CDN_STAGE_URL } from '@/constants';
 import {
   useCreateActivationKeysMutation,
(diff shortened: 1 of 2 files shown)


regexowl:
Thanks for looking into this!

We don't need the Next button to be disabled when the activation key is loading, but I think we should scroll the user to the Registration section and render a "missing activation key" error when there's no key selected for automatic registration option. 

We could also try to quietly load the key in the background and let the user progress to the Review, but then we would have to send them back to the first step if the key update didn't run properly or the rhsm service was unavailable

[Screencast_20260720_163213.webm](https://github.com/user-attachments/assets/e3f63bc9-af49-429a-bea2-cf2e340d4c69)

(it isn't really clear from the video, but clicking on the "Create blueprint" didn't do anything, because the blueprint isn't valid without the activation key. So if something went wrong with the key, we would leave the user without the ability to create blueprint, but also without any indication why)

src/Components/CreateImageWizard/steps/Registration/components/ActivationKeysList.tsx
```diff
@@ -312,16 +314,22 @@ const ActivationKeysList = ({ onErrorChange }: RegistrationProps) => {
...
-                analyticsStepId='step-registration'
-              >
-                Manage activation keys
-              </ExternalLinkButton>
-            </HelperTextItem>
+            {forceShowErrors && !activationKey ? (
+              <HelperTextItem variant='error'>
+                Missing activation key
+              </HelperTextItem>
+            ) : (
```
regexowl: Could we get rid of the duplication between the `useValidation` and here? We can consume the relevant error directly from the validator.

```suggestion
            {forceShowErrors && stepValidation.errors.activationKey ? (
              <HelperTextItem variant='error'>
                {stepValidation.errors.activationKey}
              </HelperTextItem>
            ) : (
```

regexowl (commented):
Works great! One last small nitpick

regexowl (approved):
Perfect, thank you!

PR #4626, opened 2026-07-15
Remove unused file (HMS-11017)

regexowl:
`SecurityInformation` must have been deprecated at some point as it's not imported anywhere. It should be safe to remove the file.

JIRA: [HMS-11017](https://issues.redhat.com/browse/HMS-11017)

[HMS-11017]: https://redhat.atlassian.net/browse/HMS-11017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


regexowl:
`knip`ed our codebase and found more. For now I only removed unused files that were obviously deprecated at some point. There are more unused exports and files, but some seem to have been forgotten about rather than being deprecated. Like check for repositories that were removed from content between the blueprint being created and edited...

Not sure if it makes sense to add `knip` with a proper config as a new dependency, so I didn't do that for now. If we take some time and clean up the codebase a smidge, we could probably refine the `knip` output to make sure stale code doesn't stay in the code base... but there's also a bunch of false positives, so it would take a bit more work.

kingsleyzissou (approved):

PR #4627, opened 2026-07-15
Wizard: always enable Create/Save buttons on Review step (HMS-10658)

mgold1234:
Keep Create and Save buttons enabled on the Review step instead of disabling them when validation errors exist. On click, navigate back to the step with errors instead of blocking.

- Refactor useIsBlueprintValid into useBlueprintValidation to expose
  firstErrorStepId
- Fix edit mode save being blocked while the async name uniqueness
  check is still in flight
- Remove shouldDisableAction helper — buttons are always enabled,
  validation runs on click via validateBeforeAction
- Replace flushSync with direct dispatch in handleValidationFail

Follows up on the "always enable Next/Review buttons" commit — same
pattern applied to the Review step.

Note: Depends on the activation key loading fix in #4622.

  JIRA: HMS-10658

### DIFF
--- a/playwright/Customizations/UsersAndGroups.spec.ts
+++ b/playwright/Customizations/UsersAndGroups.spec.ts
@@ -681,6 +681,12 @@ test('Create a blueprint with Groups customization', async ({
     await expect(frame.getByText('ops')).toBeVisible();
   });
 
+  await test.step('Verify Create button is enabled', async () => {
+    await expect(
+      frame.getByRole('button', { name: 'Create blueprint' }),
+    ).toBeEnabled();
+  });
+
   await test.step('Create and save blueprint', async () => {
     await createBlueprint(frame, blueprintName);
   });
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -113,6 +113,7 @@ import {
   useTimezoneValidation,
   useUserGroupsValidation,
   useUsersValidation,
+  WIZARD_STEP_IDS,
 } from '../CreateImageWizard/utilities/useValidation';
 
 const CreateImageWizard = () => {
@@ -366,7 +367,9 @@ const CreateImageWizard = () => {
   useEffect(() => {
     if (!isOnPremise && showWizardModal && !hasTrackedInitialStepRef.current) {
       const initialStepId =
-        mode === 'edit' ? 'review-step' : 'base-settings-step';
+        mode === 'edit'
+          ? WIZARD_STEP_IDS.REVIEW
+          : WIZARD_STEP_IDS.BASE_SETTINGS;
       const accountId = userData?.identity.internal?.account_id;
 
       analytics.track(`${AMPLITUDE_MODULE_NAME} - Step Viewed`, {
@@ -423,9 +426,9 @@ const CreateImageWizard = () => {
   ) => {
     const status = (step.id !== activeStep.id && step.status) || 'default';
 
-    const isBaseSettingsStep = step.id === 'base-settings-step';
+    const isBaseSettingsStep = step.id === WIZARD_STEP_IDS.BASE_SETTINGS;
     const hasVisitedBaseSettings = _steps.find(
-      (s) => s.id === 'base-settings-step',
+      (s) => s.id === WIZARD_STEP_IDS.BASE_SETTINGS,
     )?.isVisited;
     const canNavigate =
       (isBaseSettingsStep || !localImageSourceSelected) &&
@@ -493,7 +496,7 @@ const CreateImageWizard = () => {
       >
         <WizardStep
           name='Base settings'
-          id='base-settings-step'
+          id={WIZARD_STEP_IDS.BASE_SETTINGS}
           navItem={CustomStatusNavItem}
           status={baseSettingsHasErrors ? 'error' : 'default'}
           footer={
@@ -546,7 +549,7 @@ const CreateImageWizard = () => {
         </WizardStep>
         <WizardStep
           name='Repositories and packages'
-          id='content-step'
+          id={WIZARD_STEP_IDS.CONTENT}
           navItem={CustomStatusNavItem}
           status='default'
           isHidden={
@@ -577,7 +580,7 @@ const CreateImageWizard = () => {
         </WizardStep>
         <WizardStep
           name='Advanced settings'
-          id='advanced-settings-step'
+          id={WIZARD_STEP_IDS.ADVANCED_SETTINGS}
           navItem={CustomStatusNavItem}
           status={advancedSettingsHasErrors ? 'error' : 'default'}
           isHidden={
@@ -651,7 +654,7 @@ const CreateImageWizard = () => {
         </WizardStep>
         <WizardStep
           name='Review'
-          id='review-step'
+          id={WIZARD_STEP_IDS.REVIEW}
           navItem={CustomStatusNavItem}
           status='default'
           footer={<ReviewWizardFooter />}
--- a/src/Components/CreateImageWizard/components/ReviewWizardFooter.tsx
+++ b/src/Components/CreateImageWizard/components/ReviewWizardFooter.tsx
@@ -17,7 +17,11 @@ import {
   useCreateBPWithNotification as useCreateBlueprintMutation,
   useUpdateBPWithNotification as useUpdateBlueprintMutation,
 } from '../../../Hooks';
-import { useAppSelector } from '../../../store/hooks';
+import { useAppDispatch, useAppSelector } from '../../../store/hooks';
+import {
+  resetForceShowErrors,
+  setForceShowErrors,
+} from '../../../store/slices/wizard';
 import {
   CreateSaveAndBuildBtn,
   CreateSaveButton,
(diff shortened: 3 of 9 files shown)


regexowl:
Can you please rebase and fix the prettier issues with `npm run format` so the diff gets cleaned up from the unrelated prettier connected changes?

src/Components/CreateImageWizard/steps/Review/Footer/shouldDisableAction.ts
```diff
@@ -0,0 +1,6 @@
+// When validateBeforeAction is provided, buttons stay enabled and
+// validation runs on click; otherwise fall back to the static flag.
+export const shouldDisableAction = (
+  isDisabled: boolean,
+  validateBeforeAction?: () => boolean,
+): boolean => !validateBeforeAction && isDisabled;
```
kingsleyzissou: I think this function is not needed. It's a one liner and we end up passing it as a prop to a bunch of components.

Also there is a bug here, you're just checking if the function exists, you're not actually calling it.

src/Components/CreateImageWizard/components/ReviewWizardFooter.tsx
```diff
@@ -38,10 +45,26 @@ const ReviewWizardFooter = () => {
   const mode = useAppSelector(selectWizardModalMode);
   const blueprintId = useAppSelector(selectSelectedBlueprintId);
   const [isOpen, setIsOpen] = useState(false);
+  const { isValid, firstErrorStepId } = useBlueprintValidation();
+
+  const handleValidationFail = () => {
+    if (!firstErrorStepId) return;
+    flushSync(() => {
```
kingsleyzissou: This feels like it might be a bit overkill. It's also quite hard to understand. It forces re-renders from
what I can gather.

We could probably just put serForceShowErrors and gotToStepById in a `useEffect` hook, or did that not work?

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -146,26 +162,47 @@ export function useIsBlueprintValid(): boolean {
...
+  const advancedSettingsInvalid =
+    filesystem.disabledNext ||
+    timezone.disabledNext ||
+    locale.disabledNext ||
+    hostname.disabledNext ||
+    kernel.disabledNext ||
+    firewall.disabledNext ||
+    services.disabledNext ||
+    firstBoot.disabledNext ||
+    (!usersAreStandalone && usersHaveErrors);
```
kingsleyzissou, line 196: I think `CreateImageWizard` already has something similar to this implemented and is more complete. 
For now, maybe we should just copy that code

src/Components/CreateImageWizard/steps/Review/components/shared/ReviewCardHeader.tsx
```diff
@@ -7,13 +7,15 @@ import {
...
+import { WizardStepId } from '@/Components/CreateImageWizard/utilities/useValidation';
+
 export const ReviewCardHeader = ({
   title,
   stepId,
   sectionId,
 }: {
   title: string;
-  stepId: string;
+  stepId: WizardStepId;
```
kingsleyzissou: Uhm I think a string is fine here to be honest. We're importing a type from the validation hook.
That feels a bit odd.

Either create a constant and types in a more central place. Or just keep it simple and use a string.

I lean towards the second option because creating a central type is just scope creep.

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -1180,6 +1217,8 @@ export function useDetailsValidation(): StepValidation {
     nameError = 'Invalid blueprint name';
   } else if (isUniqueName === false) {
     nameError = 'Blueprint with this name already exists';
+  } else if (isUniqueName === null) {
+    return { errors: { name: '' }, disabledNext: false };
```
kingsleyzissou: Small thing I noticed here:

   ```diff
      } else if (isUniqueName === false) {
        nameError = 'Blueprint with this name already exists';
   +  } else if (isUniqueName === null) {
   +    return { errors: { name: '' }, disabledNext: false };
      }
 ```

Because of this early return, the function never gets to the isPending: isUniqueName === null line below it when the check is still in progress. So isPending will always come back false now, even while we're waiting on the uniqueness check.

kingsleyzissou (commented):
Quick scope question before diving into the rest. The title/description frame this as "keep Create/Save enabled on the Review step," but `CreateImageWizard.tsx`'s `canNavigate` also changes for *all* wizard nav items, not just Review:

   ```diff
   -    const canNavigate =
   -      (isBaseSettingsStep || !localImageSourceSelected) &&
   -      (mode === 'edit' ||
   -        step.isVisited ||
   -        isBaseSettingsStep ||
   -        (hasVisitedBaseSettings && !baseSettingsHasErrors));
   +    const canNavigate =
   +      mode === 'edit' ||
   +      step.isVisited ||
   +      isBaseSettingsStep ||
   +      hasVisitedBaseSettings;
 ```

This drops both the localImageSourceSelected gate and the !baseSettingsHasErrors check. If I understand this correctly, once Base Settings has been visited once, every nav item becomes clickable regardless of whether an image source was picked or whether base settings currently has errors.

Is that intentional, or a side effect of generalising the "always enabled" pattern? If it's needed for this feature to work, could we call it out explicitly in the PR description? If it's not strictly required, I'd suggest keeping the original gating here and scoping this PR to just the Review step's Create/Save buttons. It'll make the diff much easier to review and reduces the risk of an unrelated nav regression.

src/Components/CreateImageWizard/utilities/useValidation.tsx
kingsleyzissou: Actually I suspect removing this line will fix this issue completely. This case should already be handled by the code below

playwright/Customizations/UsersAndGroups.spec.ts
```diff
@@ -681,6 +681,12 @@ test('Create a blueprint with Groups customization', async ({
     await expect(frame.getByText('ops')).toBeVisible();
   });
 
+  await test.step('Verify Create button is enabled', async () => {
+    await expect(
+      frame.getByRole('button', { name: 'Create blueprint' }),
+    ).toBeEnabled();
+  });
```
regexowl, line 688: I think the test coverage could be improved. Since the `isDisabled` prop was removed from the create button, it will be always enabled and we're just asserting something that won't change.

What might be better is to add a general test that will introduce an error, try to click "Create" and check that the wizard skipped to the relevant step/error.

src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -113,6 +113,7 @@ import {
   useTimezoneValidation,
   useUserGroupsValidation,
   useUsersValidation,
+  WIZARD_STEP_IDS,
```
regexowl, line 116: Nice, introducing the constants makes perfect sense! ✨

It could have been done in a separate PR or a separate commit, since it's a refactoring that's not necessarily essential for enabling the buttons. But we can leave it as is now :)

src/Components/CreateImageWizard/components/ReviewWizardFooter.tsx
```diff
@@ -38,10 +44,24 @@ const ReviewWizardFooter = () => {
...
+  const { isValid, firstErrorStepId } = useBlueprintValidation();
+
+  const handleValidationFail = () => {
+    if (!firstErrorStepId) return;
+    dispatch(setForceShowErrors());
+    goToStepById(firstErrorStepId);
+    requestAnimationFrame(() => {
+      scrollToFirstError();
+    });
+  };
```
regexowl, line 56: This overlaps with `validateBeforeAction` and `onToggleClick` a bit. Maybe it could be cleaned up by combining `validateBeforeAction` and `handleValidationFail` into one function and have `onToggleClick` just call that. Something like:

```typescript
  const validateBeforeAction = (): boolean => {
    if (!isValid) {
      if (!firstErrorStepId) return false;
      dispatch(setForceShowErrors());
      goToStepById(firstErrorStepId);
      requestAnimationFrame(() => {
        scrollToFirstError();
      });
      return false;
    }
    return true;
  };
```

and 

``` typescript
  const onToggleClick = () => {
    if (validateBeforeAction()) {
      setIsOpen(!isOpen);
    }
  };
```

but how the code is now works great as well. So just a nitpicky suggestion, up to you :)

src/Components/CreateImageWizard/tests/CreateMode.test.tsx
```diff
@@ -33,31 +33,6 @@ describe('Create Image Wizard', () => {
...
-    });
-    const reviewNavItem = within(navigation).getByRole('button', {
-      name: /review/i,
-    });
-
-    expect(baseNavItem).toBeEnabled();
-    expect(contentNavItem).toBeDisabled();
-    expect(advancedNavItem).toBeDisabled();
-    expect(reviewNavItem).toBeDisabled();
-  });
```
regexowl, line 60: This logic isn't changed and the test still passes after the changes unless I'm missing something. Why was this removed?

src/Components/CreateImageWizard/utilities/useValidation.tsx
```diff
@@ -127,7 +128,22 @@ export type UsersStepValidation = {
...
-export function useIsBlueprintValid(): boolean {
+export const WIZARD_STEP_IDS = {
+  BASE_SETTINGS: 'base-settings-step',
+  CONTENT: 'content-step',
+  ADVANCED_SETTINGS: 'advanced-settings-step',
+  REVIEW: 'review-step',
+} as const;
+
+export type WizardStepId =
+  (typeof WIZARD_STEP_IDS)[keyof typeof WIZARD_STEP_IDS];
```
regexowl, line 139: There might be a better place for there as they're not linked to validation. Maybe constants.ts or something like that.

regexowl (commented):
This works nicely, thank you! Added some comments.

PR #4628, opened 2026-07-15
Wizard/Registration: Swap 'Register later' and 'Register with Satellite' (HMS-11023)

regexowl:
This swaps the order of the options as per recent UX sync. The 'Register later' is now at the bottom of the registration radios list.

JIRA: [HMS-11023](https://issues.redhat.com/browse/HMS-11023)

[HMS-11023]: https://redhat.atlassian.net/browse/HMS-11023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/Customizations/Registration.spec.ts
+++ b/playwright/Customizations/Registration.spec.ts
@@ -392,6 +392,11 @@ registrationModes.forEach(
                 name: 'Automatically register to Red Hat',
               }),
             ).toBeChecked();
+            if (isHosted()) {
+              await expect(
+                frame.getByPlaceholder('Select activation key'),
+              ).not.toHaveValue('');
+            }
           } else if (name === 'register-later') {
             await expect(
               frame.getByRole('radio', { name: 'Register later' }),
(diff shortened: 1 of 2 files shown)


playwright/Customizations/Registration.spec.ts
```diff
@@ -392,6 +392,11 @@ registrationModes.forEach(
                 name: 'Automatically register to Red Hat',
               }),
             ).toBeChecked();
+            if (isHosted()) {
+              await expect(
+                frame.getByPlaceholder('Select activation key'),
+              ).not.toHaveValue('');
+            }
```
regexowl, line 399: Playwright was tripping on being allowed to click through the wizard freely while the activation key was still loading. Meaning the "Save changes to blueprint" button was unexpectedly disabled.

This is not necessary for this PR, but is a general mini-fix. Happy to open a separate PR for it if desired.

kingsleyzissou (approved):

PR #4631, opened 2026-07-15
Replace async host detection with Redux store sync

kingsleyzissou:
## Summary

Replace async host detection patterns (`useEffect` + `getHostArch`/`getHostDistro`) with idiomatic Redux store sync so that the wizard slice always has the correct host architecture and distribution on-premise.

## Changes

- Add a `getHostInfo` RTK Query endpoint that calls both host detection helpers concurrently, triggered on app init in `AppCockpit`
- Sync host info into the wizard slice via a `matchFulfilled` handler (initial query) and an `initializeWizard` listener (subsequent wizard resets)
- Remove async `useEffect`-based host detection from `CreateImageWizard` and `BlueprintMode`
- Make `mapOnPremToHosted` a pure function by accepting `distro` as a parameter instead of calling `getHostDistro()` internally
- Add unit tests for the `matchFulfilled` handler and `applyHostInfo` listener

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -249,6 +249,7 @@ module.exports = defineConfig([
       ],
       '@typescript-eslint/await-thenable': 'error',
       '@typescript-eslint/no-floating-promises': 'error',
+      'no-restricted-imports': 'off',
     },
   },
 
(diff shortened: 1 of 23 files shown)


kingsleyzissou:
The test failures on gitlab seem relevant, will look into them

kingsleyzissou:
I'm going to close this and come back to it after the next RHEL release. This is nice to have, but not needed right now

PR #4634, opened 2026-07-16
Wizard/FSC: Lock partitions required by OSCAP (HMS-10790)

regexowl:
This marks partitions included in OpenSCAP profile customizations as required. The mountpoint can't be changed, the partition can't be removed and the minimum size must be at least the required minimum size for the OpenSCAP required mountpoint.

### DIFF
--- a/src/Components/CreateImageWizard/steps/FileSystem/components/Row.tsx
+++ b/src/Components/CreateImageWizard/steps/FileSystem/components/Row.tsx
@@ -1,11 +1,18 @@
 import React from 'react';
 
-import { Button, TextInput } from '@patternfly/react-core';
-import { MinusCircleIcon } from '@patternfly/react-icons';
+import { Button, TextInput, Tooltip } from '@patternfly/react-core';
+import { LockIcon, MinusCircleIcon } from '@patternfly/react-icons';
 import { Td, Tr } from '@patternfly/react-table';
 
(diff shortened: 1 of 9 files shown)


src/Components/CreateImageWizard/steps/FileSystem/components/Row.tsx
```diff
@@ -20,16 +27,53 @@ type RowPropTypes = {
...
+  const { data: oscapProfileInfo } = useGetOscapCustomizationsQuery(
+    {
+      distribution: release,
+      // @ts-expect-error skipped when undefined
+      profile: complianceProfileID,
+    },
+    {
+      skip: !complianceProfileID,
+    },
+  );
```
kingsleyzissou, line 42: Nitpick: This is 100% follow up territory if we decided to change it. I think we'd be running this query on every row... it should be fine since this is cached and we won't get a new network request. We should maybe consider passing this down via props

kingsleyzissou (approved):
🚀

src/Components/CreateImageWizard/steps/FileSystem/components/Row.tsx
regexowl: Oof, yeah, great point. I'll open a follow up to clean this up :eyes:

PR #4637, opened 2026-07-17
Wizard/FSC: Deduplicate `useGetOscapCustomizationsQuery` request (HMS-10790)

regexowl:
The request would be called for every row. RTK query would have deduplicated the requests, but this is cleaner approach, with one subscription instead of N.


regexowl:
Follow up to https://github.com/osbuild/image-builder-frontend/pull/4634

kingsleyzissou (approved):
Thanks for doing this so quickly!

PR #4642, opened 2026-07-20
Wizard: PackageTable width fix (HMS-10794)

ksiekl:
Fix the jumping width of the Packages table.
package-width


regexowl (approved):
Simple, elegant, perfection ✨

PR #4643, opened 2026-07-20
Wizard/Snapshots: Validate on progressing (HMS-10760)

regexowl:
`requiredDateOptions` were immediately rendering a missing date error on clicking the "Clear date" button. This replaces it with a custom validation that gets triggered when trying to progress in the Wizard (Next/Review).

This way the user can clear the input without being immediatelly screamed at by the validator.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Snapshot/components/Snapshot.tsx
+++ b/src/Components/CreateImageWizard/steps/Snapshot/components/Snapshot.tsx
@@ -11,13 +11,17 @@ import {
   Radio,
 } from '@patternfly/react-core';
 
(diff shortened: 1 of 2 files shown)


regexowl:
> * The inline validation logic for the date picker (`validated` + separate `FormHelperText` with the same blank-date check) duplicates the `isSnapshotValid` semantics; consider encapsulating this in a single helper or using the same validator path to avoid divergence between UI state and validation logic.
> 
> * Since `requiredDateOptions` is removed, double-check whether the `required` prop on the `DatePicker` is still needed or if it can be dropped to avoid any unintended native browser validation that might conflict with the custom wizard-progression validation behavior.

Fixed

kingsleyzissou (approved):

PR #4644, opened 2026-07-20
Wizard: Tighten row spacing and hide idle password hints (HMS-10793)

mgold1234:
Reset table row offset in Users and Groups for neater input spacing.
Only show password validation helper text when the input is focused
or has errors.
Before: 
Screenshot 2026-07-22 at 18 42 19
After:
Screenshot 2026-07-22 at 18 41 26
  

  JIRA: HMS-10793

### DIFF
--- a/src/Components/CreateImageWizard/steps/UserGroups/components/GroupRow.tsx
+++ b/src/Components/CreateImageWizard/steps/UserGroups/components/GroupRow.tsx
@@ -76,7 +76,7 @@ const GroupRow = ({ index, groupCount, group }: GroupRowProps) => {
   };
 
   return (
-    <Tr>
+    <Tr resetOffset>
       <Td>
         <ValidatedInputAndTextArea
           ariaLabel='Group name'
--- a/src/Components/CreateImageWizard/steps/Users/components/UserRow.tsx
+++ b/src/Components/CreateImageWizard/steps/Users/components/UserRow.tsx
@@ -100,7 +100,7 @@ const UserRow = ({ user, index, userCount }: UserRowProps) => {
 
   return (
     <>
-      <Tr>
+      <Tr resetOffset>
         <Td>
           <ValidatedInputAndTextArea
             ariaLabel='blueprint user name'


src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
```diff
@@ -55,19 +60,21 @@ export const PasswordValidatedInput = ({
...
+      {(isPasswordVisible || hasError) && (
+        <FormHelperText>
+          <HelperText component='ul'>
+            <HelperTextItem variant={ruleLength} component='li'>
+              Password must be at least 6 characters long
             </HelperTextItem>
-          )}
-        </HelperText>
-      </FormHelperText>
+            {environments.includes('azure') && (
```
ksiekl: I believe we have a bug here. I tried selecting Azure as the target and entering only lowercase characters, and I never got this error. I asked AI to check it out, and it seems we had a bug in here for some time, but I didn't go poking around the codebase too much: 

> There's the bug. Look at line 1101 in `useValidation.tsx`:
> 
>   isValid: isLengthValid,
> 
> The isValid result only checks length — it ignores the ruleCharacters validation entirely. Even though ruleCharacters is computed and set to 'error' when fewer than 3 character types are used, it's only used for the display (the helper text hint), not for the actual validity check.
> 
> So when you enter an all-lowercase password that's 6+ characters for Azure, ruleCharacters is 'error' but isValid is still true because the length passes. The form validation at line 930 (!isPasswordValid) won't block you.

We should definitely fix this, but it can be a follow-up.

ksiekl (commented):
Hiya! I took a look at the PR, and I'm unsure if the `compactTableStyle.ts` is the way we want to go - generally, we don't mess with the CSS properties if we can avoid it. And if I understand correctly, this new file, together with the `resetOffset,` just changes the padding of the table - but to be honest, I do not see the difference between before and after. Maybe I'm missing something, if so, feel free to correct me!

And I noticed a problem with the validation, which should definitely be fixed, but could be a follow-up.

src/Components/CreateImageWizard/utilities/PasswordValidatedInput.tsx
mgold1234: @ksiekl thank you. fixed it in a follow up pr 
https://github.com/osbuild/image-builder-frontend/pull/4729

regexowl:
I can't see much difference between the changes and the users table that's on stage. Is that intentional? (this is screencast of clicking between changes ran locally and changes on stage)

users-table

But important information for the users step - the UX team presented new mocks last week during the UX/dev sync and we'll be most probably re-revamping the user step (and firewall) to give each user more space. I don't think the mocks are final yet, but complete overhaul of the step is probably in the near future.

There are two options - either we merge this, update the follow up to fix the regression and then wait for the new mocks. Or leave the user step as is and just wait for the new mocks.

mgold1234:
@regexowl up to you, what do you prefer? merge it or wait?

mgold1234:
@regexowl can we merge it?
and update the foloow up to fix the regression ?

regexowl:
Since the only difference between the tables is the offset, let's maybe just leave the `resetOffset`. Changes to the validation logic which don't affect rendering can be removed and the tests restored. The follow up seems to be fixing an existing regression, but it's independent on these changes, is that correct?

regexowl (approved):

PR #4645, opened 2026-07-20
Use image-builder CLI for on-prem package search

kingsleyzissou:
Use image-builder CLI for on-prem package search

## Summary

Replace the HTTP-based package search in the Cockpit plugin with a direct call to the `image-builder` CLI (`image-builder pkgsearch`). The previous approach relied on a local HTTP API that is being removed; the CLI provides the same functionality with a simpler integration path.

## Changes

- Replace HTTP POST to `/search/packages` with `cockpit.spawn` call to `image-builder pkgsearch`, using `$XDG_CACHE_HOME` for the rpmmd cache directory
- Sort packages by version (descending) before deduplicating so the highest version is kept when collapsing duplicates
- Add `image-builder >= 75.0.0` as an RPM dependency in the Cockpit spec
- Remove the `typeguards` module — runtime JSON shape assertions are replaced by typed parsing via `parseJsonUnsafe`


kingsleyzissou:
Will mark this ready once we have a new image-builder release.

Fyi @croissanne

croissanne (approved):
This LGTM, I'd be ok with merging it already since it's just the feature branch.

PR #4646, opened 2026-07-20
Wizard: fix adding LabelInput values (HMS-10877)

ksiekl:
Adds the "Press Enter or Add icon" text for the user under the LabellInput fields, and adds the Add button back to keep some clickable option.

image
image
image
image
image
image

### DIFF
--- a/src/Components/CreateImageWizard/LabelInput.tsx
+++ b/src/Components/CreateImageWizard/LabelInput.tsx
@@ -1,6 +1,9 @@
 import React, { useState } from 'react';
 
 import {
+  Button,
+  Flex,
+  FlexItem,
   HelperText,
   HelperTextItem,
   Label,
@@ -9,6 +12,7 @@ import {
   TextInputGroupMain,
   Truncate,
 } from '@patternfly/react-core/dist/esm';
+import { PlusCircleIcon } from '@patternfly/react-icons';
 import { UnknownAction } from 'redux';
 
 import { StepValidation } from './utilities/useValidation';
@@ -33,6 +37,8 @@ type LabelInputProps = {
   truncateLength?: number;
   isCompact?: boolean;
   chipCollapseThreshold?: number;
+  hideAddLabel?: boolean;
+  helperText?: string;
 };
 
 const LabelInput = ({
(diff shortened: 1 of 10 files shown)


src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -70,6 +76,10 @@ const LabelInput = ({
   };
 
   const addItem = (value: string) => {
+    if (!value.trim()) {
```
regexowl: Nice, the trim makes sense here, should we trim the final value before dispatching for consistency?

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -147,62 +157,78 @@ const LabelInput = ({
...
+                  ))}
+                </LabelGroup>
+              )}
+            </TextInputGroupMain>
+          </TextInputGroup>
+        </FlexItem>
+        <FlexItem alignSelf={{ default: 'alignSelfFlexStart' }}>
+          <Button
+            variant='control'
+            aria-label='Add'
```
regexowl: Maybe we could make the aria-label a bit more specific, to make it clearer when used by screen readers. But this could also be a follow up, non blocking.

src/Components/CreateImageWizard/LabelInput.tsx
```diff
@@ -147,62 +157,78 @@ const LabelInput = ({
...
+            icon={<PlusCircleIcon />}
+            onClick={() => addItem(inputValue)}
+          >
+            {!hideAddLabel && 'Add'}
+          </Button>
+        </FlexItem>
+      </Flex>
+
+      <HelperText>
```
regexowl: This looks slightly weird for inputs that already have a helper text

Image

Did you discuss the placement with UX? If not, could you please check with them?

regexowl (commented):
Thanks, looks great! Added a few nitpicks

src/Components/CreateImageWizard/LabelInput.tsx
ksiekl: Sure, thanks!

src/Components/CreateImageWizard/LabelInput.tsx
ksiekl: Ah, yes, of course. While I was at that, I made the labels for firewall and systemd services distinct, so the screen readers don't read the same values for different fields.

src/Components/CreateImageWizard/LabelInput.tsx
ksiekl: I reached out to UX :smile:

ksiekl: Shayna recommended to merge them, so now they look like this: 
image

avitova (approved):
LGTM.

PR #4648, opened 2026-07-20
src/api: switch to image-builder for composes and their statuses (HMS-11089)

croissanne:
`systemd-run` is used to start composes, the cache directory is the default used by image-builder. The output directories are located in `/var/lib/cockpit-image-builder/`, with each build having its own uuid.

The status constructed in the following way:
1. start from pending;

2. if the unit is still active, set to building;

3. if the unit is not active and the image-builder output directory does not exist, try to find any artifacts in the osbuild-composer output directory;
3a. mark as succes if something is found, failed if not;

4. if the unit is not active and a buildlog containing an osbuild v2 result is present, try to parse the osbuild result; 4a. if osbuild succeeded, set to success and continue, 4b. if osbuild failed or if the buildlog does not contain a valid osbuild v2 result, return a failed status;

5. add progress information if it is available;'

6. add the upload result if it is available and return.


src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -29,6 +36,13 @@ import {
 import { assertComposeResponse, assertComposeStatus } from '../typeguards';
 import { type ComposerCreateBlueprintRequest } from '../types';
 
+const imagetype = (it: string) => {
+  if (it === "guest-image") {
```
kingsleyzissou: I think this might not be needed, I added all the frontend image types as aliases:
https://github.com/osbuild/image-builder/blob/b972e187a172d164a6d1db332fae6cacb3ea0db2/data/distrodefs/rhel-10/imagetypes.yaml#L1003

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -54,35 +66,44 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
             ir.upload_request.type = 'local';
           }
 
-          // this request gets saved to the local storage and needs to
-          // match the hosted format
+          const bpOnPrem = mapHostedToOnPrem(parsed as CreateBlueprintRequest)
+          const hostDistro = await getHostDistro();
+          const user = await cockpit.user();
+          const id = uuidv4();
+          const bpPath = path.join("/tmp", `cockpit-image-builder-${id}.json`);
```
croissanne: it needs to be somewhere else

croissanne: let's add a tmpfiles.d dropin to add /var/cache/cockpit-image-builder with a specific cleanup age (like 1h or smth)

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -54,35 +66,44 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
             ir.upload_request.type = 'local';
           }
 
-          // this request gets saved to the local storage and needs to
-          // match the hosted format
+          const bpOnPrem = mapHostedToOnPrem(parsed as CreateBlueprintRequest)
```
avitova, line 69:
```suggestion
          const bpOnPrem = mapHostedToOnPrem(parsed as CreateBlueprintRequest);
```

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -45,6 +52,11 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
 
         const blueprint = parsed as ComposerCreateBlueprintRequest;
         const composes: ComposeResponse[] = [];
+        const dataDir = path.join("/var/lib/cockpit-image-builder");
+        await cockpit.spawn(['mkdir', '-p', dataDir], {
+          superuser: "require",
+        });
```
avitova, line 58:
```suggestion
        const dataDir = path.join("/var/lib/cockpit-image-builder");
        const cacheDir = path.join('/var/cache/cockpit-image-builder');
        await cockpit.spawn(['mkdir', '-p', dataDir, cacheDir], {
          superuser: "require",
        });
```

avitova (commented):
I am getting
"Uncaught (in promise)" { message: 'not-found', problem: 'not-found' } when I click the `Build images` button.

My suggestion should fix it.
I also received:
```
missing artifact directory
```
error afterwards. Trying to debug.

src/store/api/backend/onprem/composerApi/composes.ts
croissanne: Ah ic, the /var/cache/cockpit-image-builder directory is actually created when installing the rpm.

croissanne: we shld definitely add this to the README at least

avitova (approved):
Okay, the next failure is not caused by your PR, I will approve these changes, if you want to change something in the README or make a change in `cockpit/devel-install` in the Makefile so that this works smootlhy for devs too, feel free to do it only in a followup.

PR #4654, opened 2026-07-21
playwright: Improve flakey behaviour (HMS-11052)

regexowl:
With the changes to the footer button behaviour in Wizard I believe there was introduced a race condition where the Wizard no longer blocks progress through the wizard when the security step is still loading.

Since playwright tests can be too fast for their own good, they get to the "Create/Save" button before the profiles are there, making the button disabled but still being furiously clicked on by the tests. This adds a simple check to wait for the security loading state to disappear and for the "Save" button to be enabled, which should hopefully help with the flakiness.

JIRA: [HMS-11052](https://issues.redhat.com/browse/HMS-11052)

[HMS-11052]: https://redhat.atlassian.net/browse/HMS-11052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/Customizations/Registration.spec.ts
+++ b/playwright/Customizations/Registration.spec.ts
@@ -409,10 +409,15 @@ registrationModes.forEach(
             ).toBeChecked();
           }
 
+          // security needs to finish loading, before trying to progress
+          await expect(frame.getByRole('progressbar')).toBeHidden();
+
           await frame.getByRole('button', { name: 'Review image' }).click();
-          await frame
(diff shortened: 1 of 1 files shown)


kingsleyzissou (approved):

regexowl:
Lol, well the good news is that the flakes this got kicked out the queue for are different than the flake this is supposed to resolve :sweat_smile:

PR #4657, opened 2026-07-21
Streamline on-prem registry auth UX and extract shared helpers

kingsleyzissou:
## Summary

Streamline the on-prem registry authentication UX by replacing the login modal with an inline form, removing outdated UI sections, and extracting shared helpers so registry auth checks can be reused across API endpoints.

## Changes

- Replace the registry login modal with an inline login form that surfaces loading state and error feedback directly in the wizard step
- Remove the pull-info expandable section that was outdated and adding visual noise
- Move the auth status indicator below the image source controls for better visual hierarchy
- Extract registry auth and host detection helpers into plain async functions to avoid circular imports when reusing them across RTK Query endpoints


src/Components/CreateImageWizard/steps/ImageOutput/components/RegistryAuthSection.tsx
regexowl, line 1: This file looks empty, but wasn't deleted.

src/Components/Blueprints/helpers/onPremToHostedBlueprintMapper.tsx
```diff
@@ -1,317 +1,17 @@
...
-  }
-
-  if (blueprint.customizations.installer) {
-    result.customizations!.installer = blueprint.customizations.installer;
-  }
-
-  return result;
-};
+// Re-export from canonical location in the store layer.
+// This file exists for backwards compatibility with Components-layer imports.
```
regexowl: Looks like this is imported only in `ImportBlueprintModal.tsx` and `tests/helpers.tsx`, I guess we could resolve those imports and get rid of this file? Not in the scope of this PR, just in general in a follow up

regexowl (commented):
Looks great! One comment and one question about possible follow up, otherwise works like magic 🧙

src/Components/Blueprints/helpers/onPremToHostedBlueprintMapper.tsx
kingsleyzissou: Ah I missed this, this was 100% bot shenanigans. I hate when it does this

kingsleyzissou: ~I'll sort it out in a follow up~ 🙏 

Changed my mind, I'll quickly do it now

kingsleyzissou: Sorry I borked the rebase, but should all be good now

regexowl: No worries, looks great now! There's still [the empty file](https://github.com/osbuild/image-builder-frontend/pull/4657#discussion_r3628083149), but we can remove it in a follow up

regexowl (approved):

src/Components/CreateImageWizard/steps/ImageOutput/components/RegistryAuthSection.tsx
kingsleyzissou: I'm half asleep lol. Ya I'm still working on this and I have some follow up in the works. Will sort it out

ochosi:
I just test-drove this locally and here's two things I noticed:
1. There's a tiny delay after logging in successfully where the status is still "not logged in" and then it switches to "logged in". I feel I was running into the same issue with my original branch, so this may just be there still.
2. Some of the errors (e.g. the registry not being reachable over the network) can get pretty long. I wonder if we can think of a better way to represent them (not blocking though)
image

kingsleyzissou:
> I just test-drove this locally and here's two things I noticed:
> 
> 1. There's a tiny delay after logging in successfully where the status is still "not logged in" and then it switches to "logged in". I feel I was running into the same issue with my original branch, so this may just be there still.
> 2. Some of the errors (e.g. the registry not being reachable over the network) can get pretty long. I wonder if we can think of a better way to represent them (not blocking though)
> 
> image

This is fair, I have a follow up that I'm working on in the meantime and I can try address those there. Shayna had some UI/UX concerns, so I will try find some time with her and iron all of this out

PR #4658, opened 2026-07-22
Split on-prem image sources into official and custom

kingsleyzissou:
# Split on-prem image sources into official and custom

## Summary

Splits the on-premises image source selection into two distinct paths — "official" (curated Red Hat images) and "custom" (user-provided container references) — so users get a guided experience for known images with pull validation, while retaining full flexibility for custom sources.

## Changes

- Split the monolithic `OnPrem` component into modular `OfficialImageSource`, `CustomImageSource`, and `ImageSelect` components under a new `OnPrem/` directory
- Add `imageSourceType` (`'official' | 'custom'`) to the wizard output slice so downstream steps can branch on the selection mode
- Populate official images from a `KNOWN_IMAGES` constant instead of querying the registry, and prefetch registry auth status on mount
- Add `pullImage` mutation and `getImageExists` query endpoints to validate that official images are actually available locally before proceeding
- Add `useImagePullValidation` hook that blocks the wizard's next step until the selected official image has been pulled
- Refetch distributions after successful registry login so the image list updates without a page reload


kingsleyzissou:
I think this is not needed, we went with another approach

PR #4660, opened 2026-07-22
Wizard: add banner with a new announce

ksiekl:
Adds banner with the single target migration announcement.
@regexowl I went for a new banner in case we want to use the "new feature" one in the meantime

### DIFF
--- a/src/Components/LandingPage/SingleTargetAlert.tsx
+++ b/src/Components/LandingPage/SingleTargetAlert.tsx
@@ -1,0 +1,67 @@
+import React, { useState } from 'react';
+
+import {
+  Alert,
+  AlertActionCloseButton,
+  AlertActionLink,
+  Content,
+  Flex,
+  FlexItem,
+} from '@patternfly/react-core';
+import { useVariant } from '@unleash/proxy-client-react';
+
+const SingleTargetAlert = () => {
+  const variant = useVariant('image-builder.single-target-migration');
+  let payload;
+  try {
+    payload = variant.payload ? JSON.parse(variant.payload.value) : undefined;
+  } catch {
+    payload = undefined;
+  }
+
+  const title = payload?.title || '';
+  const body = payload?.body || '';
+
+  const localStorageKey = 'imageBuilder.singleTargetMigration.dismissed';
+  const isAlertDismissed = window.localStorage.getItem(localStorageKey);
+
+  const [displayAlert, setDisplayAlert] = useState(!isAlertDismissed);
+  const [isTemporarilyHidden, setIsTemporarilyHidden] = useState(false);
+
+  const dismissAlert = () => {
+    setDisplayAlert(false);
+    window.localStorage.setItem(localStorageKey, 'true');
+  };
+
+  if (!variant.enabled || !displayAlert || isTemporarilyHidden || !title) {
+    return null;
+  }
+
+  return (
+    <Alert
+      data-testid='single-target-migration-banner'
+      isExpandable
+      variant='warning'
+      style={{ margin: '0 0 16px 0' }}
+      title={title}
+      actionClose={
+        <Flex>
+          <FlexItem>
+            <AlertActionLink onClick={dismissAlert}>
+              Don&apos;t show me this again
+            </AlertActionLink>
+          </FlexItem>
+          <FlexItem>
+            <AlertActionCloseButton
+              onClose={() => setIsTemporarilyHidden(true)}
+            />
+          </FlexItem>
+        </Flex>
+      }
+    >
+      <Content>{body}</Content>
+    </Alert>
+  );
+};
+
+export default SingleTargetAlert;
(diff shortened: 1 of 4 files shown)


ksiekl:
@sourcery-ai review

src/Utilities/useGetEnvironment.ts
```diff
@@ -40,6 +40,7 @@ export const useFlagWithEphemDefault = (
 const onPremFlag = (flag: string): boolean => {
   switch (flag) {
     case 'image-builder.images-table-revamp.enabled':
+    case 'image-builder.single-target-migration':
```
regexowl, line 43: I can't remember - will the migration include also on-prem or are we migrating only service in the "first wave"?

src/Components/LandingPage/SingleTargetAlert.tsx
```diff
@@ -0,0 +1,71 @@
...
+  let payload;
+  try {
+    payload = variant.payload ? JSON.parse(variant.payload.value) : undefined;
+  } catch {
+    payload = undefined;
+  }
+
+  const title = payload?.title || '';
+  const body = payload?.body || '';
+  const localStorageKey = payload?.localStorageKey || '';
```
regexowl: Since this will be probably a single purpose alert, we could simply harcode a localStorageKey. I assume the alert will get removed from the code base once the migration is finished, if we plan to repurpose it, the localStorageKey prop would make sense.

regexowl (commented):
Looks good! Added a nitpick and a question.

I assume the specific wording would be fine tuned directly in unleash before we flip the switch on?

src/Components/LandingPage/SingleTargetAlert.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import {
+  Alert,
+  AlertActionCloseButton,
+  AlertActionLink,
+  Content,
+  Flex,
+  FlexItem,
+} from '@patternfly/react-core';
+import { useVariant } from '@unleash/proxy-client-react';
```
mgold1234, line 11: useVariant imported directly from @unleash/proxy-client-react, 
The project wraps Unleash's useFlag in useGetEnvironment.ts specifically to handle Cockpit
(on-prem) builds where Unleash isn't available.
Consider adding a useVariant wrapper in useGetEnvironment.ts like the existing useFlag wrapper.

src/Components/LandingPage/SingleTargetAlert.tsx
```diff
@@ -0,0 +1,71 @@
...
+      data-testid='single-target-migration-banner'
+      isExpandable
+      variant='warning'
+      style={{ margin: '0 0 16px 0' }}
+      title={title}
+      actionClose={
+        <Flex>
+          <FlexItem>
+            <AlertActionLink onClick={dismissAlert}>
+              Don&apos;t show me this again
```
mgold1234, line 51: "Don't show me this again" button is shown even when 
  localStorageKey is missing, making the permanent dismiss a lie

  When the payload doesn't include a localStorageKey, the value defaults to ''. In that case,
  dismissAlert only calls setDisplayAlert(false) without writing to localStorage,  so the
  dismissal doesn't survive a page reload. But the button still says "Don't show me this
  again", which is misleading. Should either hide the permanent dismiss option when there's
  no localStorageKey, or treat a missing key as an error.

src/Utilities/useGetEnvironment.ts
ksiekl: Oh, well, eventually the migration will happen on both. I just started thinking only about the hosted service, but you have a good point. I'll put it outside so it shows up in the cockpit as well.

ksiekl: Hmm, well, putting the banner on prem means we would have to hardcode its content, since Unleash isn't available on-prem, right?

src/Components/LandingPage/SingleTargetAlert.tsx
ksiekl: The alert is gated by the flag in LandingPage, which returns false on-prem, so useVariant never gets called there. Maybe with showing the alert on-prem, it will have to be solved, but I'm not sure we're going to do that, because of the conversation above :arrow_up:

src/Components/LandingPage/SingleTargetAlert.tsx
ksiekl: hardcoded the key anyways, so it's not an issue - but yes, the values would need to be always in unleash, like they are with `NewAlert` now :smile:

src/Utilities/useGetEnvironment.ts
regexowl: Yes, we'd have to hardcode content for the on-prem one. Just wanted to check, keeping the flag disabled for now works perfectly.

regexowl (approved):
Looks good! 🚀

PR #4664, opened 2026-07-23
Split on-prem image source into official and custom flows (HMS-11081)

kingsleyzissou:
Split on-prem image source into official and custom flows

## Summary

Splits the monolithic on-prem image source component into distinct "Official Red Hat images" and "Custom images" flows, each backed by new API endpoints for pulling and checking image existence. This enables guided image selection with registry authentication gating for official images and a searchable typeahead for custom images.

## Changes

- Add `KNOWN_IMAGES` constant and `imageSourceType` (`official` | `custom`) to the output slice, with selectable cards to switch between flows
- Introduce `pullImage` and `getImageExists` RTK Query endpoints with per-reference cache invalidation, and prefetch registry auth status at app init
- Extract `OfficialImageSource`, `CustomImageSource`, and a reusable `ImageSelect` component from the monolithic `OnPrem.tsx`, removing the `Dropdown` passthrough layer
- Add `useImagePullValidation` hook to gate wizard progression until the official image is pulled locally
- Invalidate distributions cache on registry login so newly accessible images appear immediately

JIRA: [HMS-11081](https://issues.redhat.com/browse/HMS-11081)

[HMS-11081]: https://redhat.atlassian.net/browse/HMS-11081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kingsleyzissou:
Here are the mocks again for reference:

Screenshot 2026-07-22 at 10 35 46 AM
Screenshot 2026-07-22 at 10 36 03 AM
Screenshot 2026-07-22 at 10 36 07 AM

kingsleyzissou:
And here is a quick demo:

https://github.com/user-attachments/assets/dcf8b0ab-91d4-46d3-9e36-49245ab98cac

regexowl (approved):
Works great! The split looks really good 🚀 

Not sure if we can make the pulling a bit more transparent, that feels like the only slightly murky part of the workflow to me.

The spinner is spinning for quite some time, some added indication that things are moving would be assuring, but at the same time we probably don't have a way to check on the progress to report it to the user, do we?

Also maybe if there's the latest image pulled, the pull button could just say that the latest image is already there or at least show the "Latest image pulled" in green with a small checkmark for a bit as an added confirmation that everything went well.

But that's just thinking from the top of my head. If any of that should be considered it would be definitely in a discussion / follow up teritory and out of scope for this PR

kingsleyzissou:
> The spinner is spinning for quite some time, some added indication that things are moving would be assuring, but at the same time we probably don't have a way to check on the progress to report it to the user, do we?

I know, and this was kind of why I was pushing back on this. There isn't really a way of showing the progress at all and the images can be very large.

> Also maybe if there's the latest image pulled, the pull button could just say that the latest image is already there or at least show the "Latest image pulled" in green with a small checkmark for a bit as an added confirmation that everything went well.

Also trying to figure out if there is a newer version is a massive pain, if you have a recent enough version you won't need to pull all the layers and it will go quickly. If you have the latest version it's essentially a no-op and that should all happen very quickly.

These were all the reasons why implementing this was tricky, but it's something we really wanted for 10.3 and the discussions kept getting revived.

regexowl:
> > The spinner is spinning for quite some time, some added indication that things are moving would be assuring, but at the same time we probably don't have a way to check on the progress to report it to the user, do we?
> 
> I know, and this was kind of why I was pushing back on this. There isn't really a way of showing the progress at all and the images can be very large.
> 
> > Also maybe if there's the latest image pulled, the pull button could just say that the latest image is already there or at least show the "Latest image pulled" in green with a small checkmark for a bit as an added confirmation that everything went well.
> 
> Also trying to figure out if there is a newer version is a massive pain, if you have a recent enough version you won't need to pull all the layers and it will go quickly. If you have the latest version it's essentially a no-op and that should all happen very quickly.
> 
> These were all the reasons why implementing this was tricky, but it's something we really wanted for 10.3 and the discussions kept getting revived.

Got it, I had a feeling that if these were easy to do, they would be already a part of the PR :sweat_smile:

PR #4682, opened 2026-07-29
Wizard: add "single target migration" banner on-prem (HMS-11066)

ksiekl:
Adds the banner to on-prem to match hosted service. Since the content of the banner has to be hardcoded for on-prem, I've removed the variant and changeable content from unleash and kept it simple. It will be deleted afterwards anyways.

image

### DIFF
--- a/playwright/Customizations/Registration.spec.ts
+++ b/playwright/Customizations/Registration.spec.ts
@@ -204,7 +204,7 @@ registrationModes.forEach(
               await viewDetailsBtn.click();
               await expect(viewDetailsBtn).toBeVisible();
               await expect(frame.getByText('activation-key-')).toBeVisible();
-              await frame.getByLabel('Close').click();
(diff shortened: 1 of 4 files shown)


regexowl (dismissed):
Works nicely! 🚀

ksiekl:
@regexowl the banner on hosted service is still hidden behind an unleash flag, it's just the text on the banner that is hardcoded. But if you mean on-prem, there is no way to use Unleash, no? That's why we have the "flags" in `src/Utilities/useGetEnvironment.ts`

regexowl:
> @regexowl the banner on hosted service is still hidden behind an unleash flag, it's just the text on the banner that is hardcoded. But if you mean on-prem, there is no way to use Unleash, no? That's why we have the "flags" in `src/Utilities/useGetEnvironment.ts`

Lol, nevermind, I'm blind... sorry :sweat_smile:

regexowl (approved):

PR #4683, opened 2026-07-30
Wizard/AAP: Improve validation (HMS-10759)

regexowl:
This improves AAP validation in a few places:
- URL and CA fields trigger validation independently
- AAP validation errors block Wizard from progressing via Next
- AAP validation errors block the Create/Save blueprint button


regexowl:
> * CreateImageWizard now calls useAAPValidation directly while useIsBlueprintValid also calls useAAPValidation; if both are used in the same flow, consider reusing a single validation result (e.g., via context or props) to avoid duplicate validation work and keep the source of truth clearer.

Ha, ok, we might want look into this in a follow up as it's a pre-existing pattern that includes all other validations. This PR will at least make sure that we're handling the validators consistently.

ksiekl (approved):
Weird that we had the `useAAPValidation()` funtion written but managed to not actually call it in the step 🤦‍♀️  Thank you 💟

PR #4684, opened 2026-07-30
cockpit: support aws uploads using image-builder (HMS-11087)

croissanne:
cockpit/upload: support AWS uploads using image-builder
    
    This changes the `CloudProviderConfig` component from updatin the
    osbuild-composer worker config, to writing an upload-config file. The
    upload-config file is read to add the relevant cli args to image-builder
    when building AWS images.
    
    This also changes the region input from a dropdown into a text input,
    that way new AWS regions are automatically supported. In future
    suggestions or a typeahead could be added to ValidatedInput.
    
    As image-builder is run as root, the default location of aws credentials
    the user is ignored. In order to find the credentials file, add it to
    the upload config and set the AWS_SHARED_CREDENTIALS_FILE environment
    variable.


src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -121,7 +125,8 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
         superuser: 'try',
       });
       let composes: ComposesResponseItem[] = [];
-      const entries = Object.entries(info.entries || {});
+      // Filter to ensure the upload-config.json file is not counted as a blueprint.
+      const entries = Object.entries(info.entries || {}).filter(entry => entry[1].type === "dir");;
```
regexowl: Extra semicolon

```suggestion
      const entries = Object.entries(info.entries || {}).filter(entry => entry[1].type === "dir");
```

src/Components/CloudProviderConfig/AWSConfig.tsx
```diff
@@ -100,6 +109,113 @@ const AWSBucket = ({ value, onChange, isDisabled }: FormGroupProps<string>) => {
...
 };
 
+const ProfilePopover = () => {
+  return (
+    <Popover
+      minWidth='35rem'
+      headerContent={'What is the AWS profile?'}
+      bodyContent={
+        <Content component={ContentVariants.p}>
+          This is the name of the  profile in your AWS credentials file which contains your AWS
```
regexowl: Extra space

```suggestion
          This is the name of the profile in your AWS credentials file which contains your AWS
```

src/Components/CloudProviderConfig/validators/index.tsx
```diff
@@ -20,18 +29,30 @@ export const isAwsCredsPathValid = (credsPath?: string): boolean => {
...
+  if (!region || region === '') {
+    return false;
+  }
+
+  const regex = /^[a-z0-9](?:[a-z0-9]|[-.](?=[a-z0-9])){1,61}[a-z0-9]$/;
+  return regex.test(region);
+};
+
 export const isAwsStepValid = (
   config: AWSWorkerConfig | undefined,
```
regexowl: Should this also be changed to `AWSUploadConfig`?

regexowl (commented):
Clicking around - this seems to be working well. Added some nitpicks.

There's a bunch of checks that didn't run, is that because the changes wouldn't be going into main? For example Prettier spits out a bunch of updates that would usually fail the check.

croissanne:
> Clicking around - this seems to be working well. Added some nitpicks.
> 
> There's a bunch of checks that didn't run, is that because the changes wouldn't be going into main? For example Prettier spits out a bunch of updates that would usually fail the check.

Yea it's currently running against the feature branch. It might be useful to enable these now though, I'll do so.

edit: https://github.com/osbuild/image-builder-frontend/pull/4685

regexowl (approved):
We might want to check the placement of the AWS target description, that appears when the checkbox is selected, with the UX team. But that's in the follow up area.

Works great! 🚀

PR #4685, opened 2026-07-30
.github/workflows: enable dev-checks on image-builder-cli branch

croissanne:
This commit should be dropped when rebasing the feature branch.


croissanne:
I think it's fine if schutzbot fails for now, it's mostly about the dev checks.

regexowl (approved):
> I think it's fine if schutzbot fails for now, it's mostly about the dev checks.

Everything else looks good to me 👍

PR #4688, opened 2026-07-31
cockpit: use image-builder-cli

croissanne:
Includes:
- package search
- composes
- aws uploads


lucasgarfield:
Alright so I don't think that this works with RHEL yet, because `image-builder` gets passed `--distro rhel-10` for package search and image builds which it doesn't support. No big deal, we either need to make sure we use `--distro rhel-10.3` or add support for `--distro rhel-10` (and -9, -8, etc...) to `image-builder`. 

So for now the RHEL test is disabled, we'll need to turn that back on when we figure out how to support RHEL.

I also needed to get rid of the osbuild-composer stuff in playwright.sh and warm the cache to get the package search to not hit the timeout, which seems to be working nicely.

Also I’m not in the CODEOWNERS so I’m admin merging but consider this reviewed and approved.

kingsleyzissou:
> Alright so I don't think that this works with RHEL yet, because `image-builder` gets passed `--distro rhel-10` for package search and image builds which it doesn't support. No big deal, we either need to make sure we use `--distro rhel-10.3` or add support for `--distro rhel-10` (and -9, -8, etc...) to `image-builder`. 
> 
> 
> 
> So for now the RHEL test is disabled, we'll need to turn that back on when we figure out how to support RHEL.
> 
> 
> 
> I also needed to get rid of the osbuild-composer stuff in playwright.sh and warm the cache to get the package search to not hit the timeout, which seems to be working nicely.
> 
> 
> 
> Also I’m not in the CODEOWNERS so I’m admin merging but consider this reviewed and approved.

I'm happy to have a look at some of this this week too. I can test it out and see what needs doing

kingsleyzissou:
@lucasgarfield did you manage to get this to work at all? I'm trying on fedora and it fails to build.

I'm getting a super helpful "not-found" error message :/

Edit: I thought I had created `/var/cache/cockpit-image-builder` but it never got created properly lol

PR #4693, opened 2026-08-03
Wizard: restrict `SingleTargetAlert` banner visibility (HMS-11066)

ksiekl:
Shows the banner with the single target migration announcement only to users who have created a blueprint with multiple targets.
mt

### DIFF
--- a/src/Components/LandingPage/SingleTargetAlert.tsx
+++ b/src/Components/LandingPage/SingleTargetAlert.tsx
@@ -9,24 +9,25 @@ import {
   FlexItem,
 } from '@patternfly/react-core';
 
+import useHasMultiTargetBlueprints from './useHasMultiTargetBlueprints';
+
 const TITLE =
   'Blueprints are transitioning to single image target environments.';
 const BODY =
   'Existing multi-target blueprints will be migrated to single-target configurations.';
 
-const SingleTargetAlert = () => {
-  const localStorageKey = 'imageBuilder.singleTargetMigration.dismissed';
-  const isAlertDismissed = window.localStorage.getItem(localStorageKey);
+const localStorageKey = 'imageBuilder.singleTargetMigration.dismissed';
 
-  const [displayAlert, setDisplayAlert] = useState(!isAlertDismissed);
+const SingleTargetAlert = () => {
+  const { hasMultiTarget } = useHasMultiTargetBlueprints();
   const [isTemporarilyHidden, setIsTemporarilyHidden] = useState(false);
 
   const dismissAlert = () => {
-    setDisplayAlert(false);
+    setIsTemporarilyHidden(true);
     window.localStorage.setItem(localStorageKey, 'true');
   };
 
-  if (!displayAlert || isTemporarilyHidden) {
+  if (!hasMultiTarget || isTemporarilyHidden) {
     return null;
   }
 
@@ -66,4 +67,10 @@ const SingleTargetAlert = () => {
   );
 };
 
-export default SingleTargetAlert;
+const SingleTargetAlertWrapper = () => {
+  const isAlertDismissed = window.localStorage.getItem(localStorageKey);
+  if (isAlertDismissed) return null;
+  return <SingleTargetAlert />;
+};
+
+export default SingleTargetAlertWrapper;
--- a/src/Components/LandingPage/useHasMultiTargetBlueprints.ts
+++ b/src/Components/LandingPage/useHasMultiTargetBlueprints.ts
@@ -1,0 +1,57 @@
+import { useEffect, useState } from 'react';
+
+import {
+  useGetBlueprintsQuery,
+  useLazyGetBlueprintQuery,
+} from '@/store/api/backend';
+
+const useHasMultiTargetBlueprints = () => {
+  const { data: blueprintsData, isLoading: isListLoading } =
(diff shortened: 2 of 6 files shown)


src/Components/LandingPage/SingleTargetAlert.tsx
```diff
@@ -20,13 +22,14 @@ const SingleTargetAlert = () => {
 
   const [displayAlert, setDisplayAlert] = useState(!isAlertDismissed);
   const [isTemporarilyHidden, setIsTemporarilyHidden] = useState(false);
+  const { hasMultiTarget } = useHasMultiTargetBlueprints();
```
regexowl, line 22: This could be probably skipped when the alert was dismissed / local storage key exists. So the API call doesn't fire when the user already hid the alert in the past.

src/Components/LandingPage/useHasMultiTargetBlueprints.ts
```diff
@@ -0,0 +1,48 @@
...
+      setIsCheckingDetails(false);
+      return;
+    }
+
+    let cancelled = false;
+
+    const check = async () => {
+      const results = await Promise.all(
+        blueprintsData.data.map((bp) =>
+          triggerGetBlueprint({ id: bp.id }, true).unwrap(),
```
regexowl: ~~This might need some rejection handling in a case the `triggerGetBlueprint` throws.~~

Ha, the review is for code before the latest changes, so this comment is no longer relevant.

regexowl: Maybe we could also add an early exit here for the dismissed alert or when the first multi-target blueprint is found. We need to know if there's at least one, not the exact count.

src/Components/LandingPage/useHasMultiTargetBlueprints.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import {
+  useGetBlueprintsQuery,
+  useLazyGetBlueprintQuery,
+} from '@/store/api/backend';
+
+const useHasMultiTargetBlueprints = () => {
+  const { data: blueprintsData, isLoading: isListLoading } =
+    useGetBlueprintsQuery({ limit: 100, offset: 0 });
+  const [triggerGetBlueprint] = useLazyGetBlueprintQuery();
```
regexowl, line 11: If this returns error, the `isCheckingDetails` would get stuck as `true`, right?

regexowl (commented):
Thanks for working on this! Targeting only users with multi-target blueprints makes perfect sense.

Tested on service and it works like magic. But in cockpit the list of blueprints isn't loading for me at all. Added some comments

Image

src/Components/LandingPage/SingleTargetAlert.tsx
ksiekl: oh yeah, good point, thanks!

src/Components/LandingPage/useHasMultiTargetBlueprints.ts
ksiekl: now after that change with the finally block that the sourcery recommended, it should not

src/Components/LandingPage/useHasMultiTargetBlueprints.ts
```diff
@@ -0,0 +1,55 @@
...
+          const result = await triggerGetBlueprint(
+            { id: bp.id },
+            true,
+          ).unwrap();
+          if (result.image_requests.length > 1) {
+            setHasMultiTarget(true);
+            return;
+          }
+        }
+      } finally {
```
regexowl, line 41: Can this throw? If yes then we should probably add a catch block to swallow the error
```suggestion
        }
      } catch {
        // leaves hasMultiTarget as false
      } finally {
```

regexowl (commented):
Looks good, left one nit pick. One question - have you tried to test this on cockpit? 

I can't see the alert nor does the blueprint list load for me with the changes
Image

BUT! I believe one item on the list of 10.3 SPUR is to remove the alert before the release. So maybe we should just switch off the flag for on-prem for now. @lucasgarfield is that correct or have I hallucinated that we want to remove the alert in on-prem?

Works nicely on service though ✨

ksiekl:
Thanks @regexowl ! Yes, I can see the banner on-prem, it builds and loads for me. Could it be something else causing it for you locally?

Also, I have not heard of a list of 10.3 SPUR yet - sorry, the wisdom tooth took its toll, and I wasn't active much. Can you share it with me, please? I didn't know we want to hide the banner before release - I thought the opposite was true.

regexowl:
> Thanks @regexowl ! Yes, I can see the banner on-prem, it builds and loads for me. Could it be something else causing it for you locally?
> 
> Also, I have not heard of a list of 10.3 SPUR yet - sorry, the wisdom tooth took its toll, and I wasn't active much. Can you share it with me, please? I didn't know we want to hide the banner before release - I thought the opposite was true.

Hmm, I wonder if I'm doing something wrong...

And no worries at all, let me see if I can find any mention of the banner, it was a relatively new item I believe...

regexowl:
Aha! Found it! https://github.com/osbuild/image-builder-frontend/issues/4711 might be worth checking with @lucasgarfield as he's working on the tasks.

regexowl (approved):
Noiceeee 🚀

PR #4700, opened 2026-08-04
deps: Remove uuid (HMS-11145)

regexowl:
This replaces `uuid` dependency with crypto.randomUUID() that is available natively. Meaning we can drop another third-party package and don't have to maintain it.

JIRA: [HMS-11145](https://issues.redhat.com/browse/HMS-11145)

[HMS-11145]: https://redhat.atlassian.net/browse/HMS-11145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Hooks/Utilities/useEffectiveBlueprintId.tsx
+++ b/src/Hooks/Utilities/useEffectiveBlueprintId.tsx
@@ -2,7 +2,6 @@ import { useEffect, useRef } from 'react';
 
 import { useDispatch } from 'react-redux';
 import { useSearchParams } from 'react-router-dom';
-import { validate as uuidValidate } from 'uuid';
 
 import { useAppSelector } from '@/store/hooks';
 import {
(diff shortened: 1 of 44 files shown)


src/Hooks/Utilities/useEffectiveBlueprintId.tsx
```diff
@@ -16,8 +15,12 @@ export const useEffectiveBlueprintId = (): string | undefined => {
 
   const [searchParams] = useSearchParams();
   const blueprintIdParam = searchParams.get('blueprint_id')?.trim();
+
+  const uuidRegex =
+    /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/i;
```
regexowl, line 20: This is the only slightly more complex changes, `validate` from the `uuid` library was previously used, but it can be replaced by a UUID regex and a simple `.test` against it.

kingsleyzissou (approved):

PR #4704, opened 2026-08-05
Wizard: Hide caret for repo and pkg search inputs (HMS-10761)

regexowl:
This replaces previously used `MenuToggle` with a plain `<div>` with PF's typeahead classes for `PackageSearch` and `RepositorySearch` components.

PF6's `MenuToggle` with variant='typeahead' always renders a caret that can't be hidden/disabled. `<div>` with styling is hacky, but it's the only way to remove the dropdown caret, but still be able to pass the ref and make the component look the same.


regexowl:
Won't be addressed, we'll stick to the PatternFly pattern

PR #4716, opened 2026-08-05
Image mode subs

lucasgarfield:
While wiring up subscription support for Image Mode, I found a few bugs that were preventing Image Mode from building on RHEL and fixed them. Image Mode builds are now working on RHEL. Note that subscription support requires image-builder > v77.

The biggest change is that --distro is no longer passed to the CLI at all. This means that you can only build images equal to whatever the host is… but that’s always been the case for Cockpit.

### DIFF
--- a/src/store/api/backend/onprem/composerApi/composes.ts
+++ b/src/store/api/backend/onprem/composerApi/composes.ts
@@ -8,7 +8,6 @@ import { OnPremBuilder, onPremQueryHandler } from '@/store/api/shared';
 import {
   byCreatedAtDesc,
   getBlueprintsPath,
-  getHostDistro,
   getRegistrationArgs,
   getUploadArgs,
   imageStatusFallback,
@@ -66,7 +65,6 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
         }
 
         const bpOnPrem = mapHostedToOnPrem(parsed as CreateBlueprintRequest);
-        const hostDistro = await getHostDistro();
         const user = await cockpit.user();
         const id = crypto.randomUUID();
         const { runArgs, ibArgs } = await getUploadArgs(
@@ -86,6 +84,10 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
           id,
         );
 
+        // The distro is intentionally not passed: image-builder falls back to
+        // the host distro, including the minor version for distros that have
+        // one (e.g. rhel-10.3). Generic aliases like `rhel-10` are not
+        // supported by the CLI.
         await cockpit.spawn(
           [
             'systemd-run',
@@ -101,8 +103,6 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
             'image-builder',
             'build',
             ir.image_type,
-            '--distro',
-            bpOnPrem.distro || hostDistro,
             '--blueprint',
             bpPath,
             '--with-buildlog',
--- a/src/store/api/backend/onprem/composerApi/composes.ts
+++ b/src/store/api/backend/onprem/composerApi/composes.ts
@@ -269,6 +269,9 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
               reason:
                 'image-builder process is not running and no result was found',
             };
+            // Return before the upload result check overwrites this error
+            // with the less specific "no upload result found" one.
+            return status;
           }
         }
 
--- a/src/store/api/backend/onprem/composerApi/composes.ts
+++ b/src/store/api/backend/onprem/composerApi/composes.ts
@@ -81,12 +81,15 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
           .file(bpPath, { superuser: 'require' })
           .replace(JSON.stringify(bpOnPrem));
 
-        const registrationArgs = await getRegistrationArgs(
-          blueprint.customizations.subscription,
-          id,
-        );
+        const { args: registrationArgs, path: regsPath } =
+          await getRegistrationArgs(blueprint.customizations.subscription, id);
 
         const bootcArgs = getBootcArgs(blueprint.bootc);
+
+        // The blueprint and registrations files are only needed while the
+        // build runs, and the latter contains the activation key, so have
+        // the unit remove them when it stops (on success or failure).
+        const cacheFiles = [bpPath, ...(regsPath ? [regsPath] : [])];
 
         // The distro is intentionally not passed: image-builder falls back to
         // the host distro, including the minor version for distros that have
@@ -103,6 +106,8 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
             '--service-type=oneshot',
             '--no-block',
             '--collect',
+            '-p',
+            `ExecStopPost=/usr/bin/rm -f ${cacheFiles.join(' ')}`,
             '--',
             'image-builder',
             'build',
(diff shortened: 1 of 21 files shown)


src/store/api/backend/onprem/composerApi/helpers/registrations.ts
```diff
@@ -0,0 +1,35 @@
...
+      organization: string;
+      activation_key: string;
+      server_url: string;
+      base_url: string;
+      insights: boolean;
+      rhc: boolean;
+      proxy?: string;
+    };
+  };
+};
```
kingsleyzissou: Nitpick: this could be a follow up, but maybe we should put the types with the rest of the types?  I think there is a types/custom.ts file for some of the other types.

src/store/api/backend/onprem/composerApi/helpers/registrations.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const mapSubscriptionToRegistrations = (
```
kingsleyzissou, line 8: This could also be a follow up. The other helper functions like `getUploadArgs`  handles file writes and arg assembly, maybe this function could do the same? It would abstract more of the complexity from the `composeBlueprint` function because that is starting to grow quite significantly

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -96,6 +97,26 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
...
+            );
+          }
+          if (blueprint.bootc.iso_payload_reference) {
+            bootcArgs.push(
+              '--bootc-installer-payload-ref',
+              blueprint.bootc.iso_payload_reference,
+            );
+          }
+        }
+
```
kingsleyzissou: A `getBootcArgs` helper could own all of this. Take the `blueprint.bootc` and `ir.image_type,` return the args array and the resolved image type. Keeps the spawn call clean.

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -114,7 +135,9 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
             '--',
             'image-builder',
             'build',
-            ir.image_type,
+            // Bootc builds reject the frontend image type aliases, so
+            // translate them back to the CLI's names (e.g. qcow2).
+            blueprint.bootc ? toBackendImageType(ir.image_type) : ir.image_type,
```
kingsleyzissou: Maybe this doesn't need to be conditional? I think we should treat all image types the same. So either do this at the ibcli level or here? I don't think we have enough time for ibcli... so maybe we should just do this for everything? 

This could also be a follow up

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -96,6 +97,26 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
           registrationArgs.push('--registrations', regsPath);
         }
 
+        // Image mode: the bootc container provides the OS content. The
+        // CLI resolves the reference from root's local container storage,
+        // where the wizard has already pulled it.
+        const bootcArgs = [];
```
kingsleyzissou: The type here is implicit, but this would technically be resolved with my below suggestoin (extracting this logic to a helper).

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -79,6 +80,24 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
         await cockpit
           .file(bpPath, { superuser: 'require' })
           .replace(JSON.stringify(bpOnPrem));
+
+        // The on-prem blueprint has no subscription customization, so the
+        // registration details are passed separately via `--registrations`.
+        const registrationArgs = [];
```
kingsleyzissou: This type is implicit. We could clean this up by extracting more of this logic to a helper function. I wrote more about this in a later comment.

This could be done in a follow up though.

src/store/api/backend/onprem/composerApi/composes.ts
```diff
@@ -79,6 +80,24 @@ export const composeEndpoints = (builder: OnPremBuilder) => ({
...
+        const subscription = blueprint.customizations.subscription;
+        if (subscription) {
+          const regsPath = path.join(
+            '/var/cache/cockpit-image-builder',
+            `cockpit-image-builder-${id}-registrations.json`,
+          );
+          await cockpit
```
kingsleyzissou: This never gets cleaned up. This is not urgent, but maybe worth considering in a follow up. Or are we okay with leaving this on file? One per commit might add up

kingsleyzissou (commented):
Nice one, I just added some comments in line. Mostly small things that don't really block merging, but we should probably consider doing even as follow ups.

src/store/api/backend/onprem/composerApi/helpers/registrations.ts
lucasgarfield: Done, moved it into types/custom.ts.

src/store/api/backend/onprem/composerApi/helpers/registrations.ts
lucasgarfield: Done, getRegistrationArgs follows the getUploadArgs pattern.

src/store/api/backend/onprem/composerApi/composes.ts
lucasgarfield: Good idea, added a new commit that cleans up both the blueprint and registration files.

src/store/api/backend/onprem/composerApi/composes.ts
lucasgarfield: Good idea, added the getBootcArgs helper.

src/store/api/backend/onprem/composerApi/composes.ts
lucasgarfield: Yeah that was an AI fail I missed, sorry.

src/store/api/backend/onprem/composerApi/composes.ts
lucasgarfield: Yep, added the getBootcArgs helper.

src/store/api/backend/onprem/composerApi/composes.ts
lucasgarfield: Yep, added getRegistrationArgs helper.

src/store/api/backend/onprem/composerApi/composes.ts
lucasgarfield: It now calls toBackendImageType() unconditionally as suggested.

kingsleyzissou (commented):
Thanks for sorting out the comments. The changes look good, I just haven't had the chance to setup a VM and verify this. 

Just thought I would add my comment in case someone is able to verify and test that it works

lucasgarfield:
It will be a bit tricky to test whether it actually works right now because none of the containers are published on registry.redhat.io. At the moment only AWS is available and only on stage.

Best might be to build the containers yourself and host them on your own registry. Idk. Kind of tough to develop against a container registry that isn't serving the images yet!

regexowl (approved):

regexowl:
The code looks fine. When we're able to test this we should do that, but if there are any bugs they won't break stuff that's available to users, so yolo

PR #4719, opened 2026-08-06
Wizard/Packages: Sort packages by the newest added (HMS-10767)

regexowl:
This replaces original sorting logic, that was mostly leftover from the pre-revamp table focused workflow, with a simple sort by recency.

This will not touch sorting within the search, so all functionality as sorting active app stream to the top within the list of available options is still in place.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/components/PackagesTable.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/components/PackagesTable.tsx
@@ -33,14 +33,9 @@ import RetirementDate from './RetirementDate';
 type PackagesTableProps = {
   isSuccessEpelRepo: boolean;
   epelRepo: ApiRepositoryCollectionResponseRead | undefined;
-  activeStream: string;
 };
 
-const PackagesTable = ({
-  isSuccessEpelRepo,
-  epelRepo,
-  activeStream,
-}: PackagesTableProps) => {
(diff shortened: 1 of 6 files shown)


src/Components/CreateImageWizard/steps/Packages/components/PackagesTable.tsx
```diff
@@ -94,71 +89,11 @@ const PackagesTable = ({
     }
   };
 
-  const sortedPackages = useMemo(() => {
```
regexowl, line 97: The same sorting logic remains in `PackageSearch` where it still makes sense. In the case of packages table, there is no reason to sort the rows by stream, end date, etc. and thus adding the packages to the selected packages table in seemingly random places.

regexowl:
> * The removal of `sortedGroups` means package groups are now rendered in their insertion order instead of alphabetically; if group ordering is still expected to be deterministic for users, consider preserving a simple name-based sort or explicitly documenting the new behavior.
> 

Sorting by recency is intentional.

> * The reducer now relies on `unshift` to represent recency, but the UI doesn’t explicitly sort by a timestamp or other field; if future changes alter insertion patterns, it may be safer to encode recency in the data model rather than relying solely on array position.

Information about packages / pkg groups is stored in a flat list of names, so we need to work with that.

kingsleyzissou (approved):

PR #4723, opened 2026-08-06
Wizard: image mode official images, container installer, and optional users

lucasgarfield:
**Stacked on #4716** — this PR targets the `image-mode-subs` branch so the diff shows only the new work. Once #4716 merges, retarget this PR to `main` and delete `osbuild:image-mode-subs`.

This release scopes on-prem image mode to the official RHEL 10 images:

- **Local images fake door** (HMS-11160): the "Custom images" card becomes "Local images" and shows a note that support lands in Cockpit Image Builder 10.4, with an `image-builder` CLI example in the meantime. All official-vs-local dropdown logic is removed.
- **AWS target + env sync** (HMS-11166): the official images now include an AWS (ami) variant. Target environment radios always offer the official images' output types — the same list whether or not an image is selected — and the radios and image dropdown stay in sync both ways.
- **Container references in the dropdown**: each official image shows its registry reference as the option description.
- **Container installer** (HMS-11165): adds the bootable-container-iso official image with a disabled payload container dropdown showing the base image it deploys. Subscriptions are not supported for the iso; the existing per-type restrictions hide registration.
- **Optional users** (HMS-11164): image mode no longer requires a user (the KVM image ships cloud-init). The review step warns when a disk image has no users, and the Users step notes that on-prem passwords are stored in plain text with an `openssl passwd -6` hint.
- **Host gating**: the image mode toggle is disabled on non-RHEL-10 hosts with a "coming soon" tooltip, since only official RHEL 10 images are available.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lucasgarfield:
Hm, I was trying out Github's PR stacking but it turns out that it seems to require that we push to a branch on osbuild to work best... going to close this and re-open using our normal workflow.

PR #4729, opened 2026-08-10
Wizard: Fix password validation ignoring Azure character rules

mgold1234:
The isValid check only verified password length, ignoring the
  requirement for at least 3 character types (lowercase, uppercase,
  numbers, symbols) when Azure is selected as a target.


regexowl (approved):
Works great, thanks!

PR #4730, opened 2026-08-10
README: adjust for image-builder

ondrejbudai:
cockpit-image-builder now uses image-builder instead of osbuild-composer as the image building backend. Let's adjust the README for this change.


regexowl (approved):
Thanks!

PR #4732, opened 2026-08-10
Makefile: init cache dir for cockpit/devel-install

ondrejbudai:
The cockpit-image-builder RPM now creates /var/cache/cockpit-image-builder via a systemd-tmpfiles snippet. However, when you want to hack on cockpit-image-builder using cockpit/devel-install and you don't have the RPM installed, the directory is not created, leading to very confusing errors.

This commit adds a simple check to the cockpit/devel-install target that creates the dir if needed.


croissanne (approved):

kingsleyzissou (approved):
Sneaky approval <insert Homer shrub gif here>

Thanks, I was meaning to get to this last week but never managed

PR #4733, opened 2026-08-10
ci: skip CI on stacked pull requests

lucasgarfield:
Disable the packit and playwright jobs for now for stacked pull requests
(basically anything that isn't targeting main).

I expect we will need to tweak this in the future but for now, I just
want to establish a convention where we don't run the full CI while
using stacked pull requests - should save us time waiting for CI runs
that will just have to run again on a rebase anyway and maybe even a bit
of money.

---


croissanne (approved):
yup

PR #4736, opened 2026-08-10
Wizard: let logged-out users browse and build official images

lucasgarfield:
Registry login is no longer required to browse official images — logged-out users see the dropdown with pull disabled, and existence/auth checks stop going stale.
- Drop the login-gated empty state and "Login required" label; the image dropdown is always shown
- Logged-out users get an inline info alert offering registry login; pull buttons disabled with an explanatory tooltip
- Re-check image existence (refetchOnMountOrArgChange) on mount, so a podman rmi outside the wizard is noticed
- Re-check registry login on mount for the same staleness reason
- Scope the pull button's busy/error state to the selected image, so switching images doesn't show a stale pulling state
- Pull validation copy now tells the user how to resolve a missing image based on login state

---


lucasgarfield:
Replaced by #4756

PR #4738, opened 2026-08-10
Wizard: add the container installer official image (HMS-11165)

### LINKED ISSUE #4714: Add the container ISO installer (HMS-11165)
This might require us to show a second dropdown for the --payload-ref. The payload ref would be rhel-bootc and the bootc ref will be rhel-bootc-installer. Double check that, I always mix those two up, but those are the container names.

JIRA: [HMS-11165](https://issues.redhat.com/browse/HMS-11165)

[HMS-11165]: https://redhat.atlassian.net/browse/HMS-11165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

lucasgarfield:
Adds the RHEL 10.3 container installer ISO as a selectable official image, with a second read-only container section for its payload.
- Add the container installer (bootable-container-iso) to the official on-prem images
- Selecting its target environment shows a second read-only container section for the payload image, with its own pull button
- Default the payload reference via resolveOfficialImage; flow it to the CLI through the existing bootc argument helper
- Require both installer and payload images pulled before building
- Hide the registration step and reset registration type when the installer is selected — subscriptions aren't supported for it
- Translate the installer type to the CLI's name (bootc-generic-iso, not bootable-container-iso) so the build unit doesn't fail
- Resolves #4714

---

### DIFF
--- a/src/Components/CreateImageWizard/steps/ImageOutput/components/ImageSourceSelect/OnPrem/OfficialImageSource.tsx
+++ b/src/Components/CreateImageWizard/steps/ImageOutput/components/ImageSourceSelect/OnPrem/OfficialImageSource.tsx
@@ -5,6 +5,7 @@ import { useAppSelector } from '@/store/hooks';
 import {
   selectImageSource,
   selectImageSourceType,
+  selectIsoPayloadReference,
 } from '@/store/slices/wizard';
 
 import ContainerSection from './ContainerSection';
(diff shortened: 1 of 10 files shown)


lucasgarfield:
Replaced by #4756

PR #4739, opened 2026-08-10
Wizard: make users optional for image mode builds (HMS-11164)

### LINKED ISSUE #4713: Make adding a user optional (HMS-11164)
It's currently mandatory to add a user when building an Image Mode image. This is probably because the images built from the rhel10-bootc image don't have cloud-init so presumably we wanted to ensure users could actually get into their images. Since we're going to move the qcow2 over to the kvm bootc container in https://github.com/osbuild/image-builder-frontend/issues/4712, and the other images will also be built from bootc foundry images, we don't need to gate on users anymore.

JIRA: [HMS-11164](https://issues.redhat.com/browse/HMS-11164)

[HMS-11164]: https://redhat.atlassian.net/browse/HMS-11164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

lucasgarfield:
Users are no longer required for an on-prem build, since the KVM image ships cloud-init — the review step warns instead when a disk image has none configured.
- Drop the on-prem requirement to add a user before continuing; remove the matching note from the Users step
- Warn on Review when a disk image (qcow2/ami, not the installer ISO) has no users, pointing at cloud-init as the alternative
- Note on the Users step that on-prem passwords are stored in plain text on the host, with an openssl passwd -6 hint for a hash instead
- Resolves #4713

---

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/components/NoUsersAlert.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/components/NoUsersAlert.tsx
@@ -1,0 +1,40 @@
+import React from 'react';
+
+import { Alert, Content } from '@patternfly/react-core';
+
+import { useAppSelector } from '@/store/hooks';
+import {
+  selectImageTypes,
+  selectIsImageMode,
+  selectUsers,
+  type SupportedImageTypes,
+} from '@/store/slices/wizard';
+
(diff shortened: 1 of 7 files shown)


lucasgarfield:
Replaced by #4756

PR #4740, opened 2026-08-10
Wizard: only offer image mode on RHEL 10 hosts

lucasgarfield:
Image mode is disabled with an explanatory tooltip on non-RHEL-10 hosts, since only RHEL 10 official images ship today.
- Disable the image mode toggle on non-RHEL-10 hosts, tooltip explains RHEL-10-only support (CentOS Stream/Fedora coming soon)
- Model host distro as unknown until the check resolves — toggle starts disabled without the tooltip, so RHEL 10 users don't see a "coming soon" flash and other hosts don't get a window to enter an unsupported mode
- Hosted service is unaffected

---


lucasgarfield:
Replaced by #4756

PR #4741, opened 2026-08-10
Wizard: replace local image builds with a coming-soon note (HMS-11160)

### LINKED ISSUE #4709: Remove "Custom" images (HMS-11160)
This is too big of a can of worms to land in Cockpit 10.3. For 10.3, we agreed to only support a curated list of images that will be available on registry.redhat.io (from the bootc foundry collection).

JIRA: [HMS-11160](https://issues.redhat.com/browse/HMS-11160)

[HMS-11160]: https://redhat.atlassian.net/browse/HMS-11160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

lucasgarfield:
Local (custom) image builds aren't supported yet — on-prem image mode now shows a coming-soon empty state instead of a picker, and blocks navigation while it's selected.

- Rename "Custom images" source card to "Local images"; drop the "No login" label
- Replace the local image dropdown with a PatternFly empty state pointing at Cockpit Image Builder 10.4, with an `image-builder` CLI example for building local images in the meantime
- Remove the official-vs-local dropdown logic: the podman distributions query for local images, registry-auth filtering of known references, and the `ImageSelect` typeahead machinery only the local dropdown used
- Disable Next/Review and sidebar navigation while the local source is selected
- Resolves #4709


croissanne (dismissed):
Getting rid of the mixed bootc distributions mock confused me at first, but ig it makes sense, like this is a scenario that would never happen, right?

lucasgarfield:
> Getting rid of the mixed bootc distributions mock confused me at first, but ig it makes sense, like this is a scenario that would never happen, right?

Exactly - we're only going to be supporting RHEL, and even then we're only supporting a small set of images (the ones available on catalog.redhat.com for kvm, aws, iso, and the base container to use as the iso payload).

croissanne (approved):

croissanne (approved):

croissanne (approved):

lucasgarfield:
Replaced by #4756 - was approved but CI failing so will just let #4756 get in instead.

PR #4746, opened 2026-08-11
devDeps: Bump jsdom from 26.1.0 to 30.0.1 (HMS-11189)

regexowl:
This bumps jsdom from 26.1.0 to 30.0.1.

https-proxy-agent was added to the dev depedencies because it was previously included under jsdom, v30 dropped it, but frontend-components-config-utilities required it on the top level.

jsdom v30 also no longer separates the title and the description of a menuitem with a space.

JIRA: [HMS-11189](https://issues.redhat.com/browse/HMS-11189)

[HMS-11189]: https://redhat.atlassian.net/browse/HMS-11189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -66,7 +66,8 @@
     "eslint-plugin-react-redux": "4.2.2",
     "eslint-plugin-testing-library": "7.16.2",
     "globals": "17.9.0",
-    "jsdom": "26.1.0",
+    "https-proxy-agent": "7.0.6",
```
regexowl, line 69: Needs to be pinned to v7 for commonJS support. v9 is ESM only

regexowl:
> * The updated tests now hard-code the lack of a space between the menuitem title and description to match jsdom v30; consider adjusting the component or test queries (e.g., use a regex that allows optional whitespace or assert on separate parts) so that accessibility labels don’t become tightly coupled to a jsdom-specific concatenation quirk.

Updated

ksiekl (approved):
Image

PR #4747, opened 2026-08-11
Wizard: restrict multiple targets when creating a bp (HMS-11101)

ksiekl:
When a user creates a new blueprint, we want the target environment options to be limited to a single selection, so there will be radio buttons. However, if the user wants to edit a bp that already includes multiple target envs, we keep the checkboxes there (for now, until the migration is complete).
radiobuttons

### DIFF
--- a/playwright/Basic/imageMode.spec.ts
+++ b/playwright/Basic/imageMode.spec.ts
@@ -81,7 +81,7 @@ test('Image mode blueprint create, edit, export, import', async ({
     ).toBeHidden();
     // but the guest image is
     await expect(
-      frame.getByRole('radio', { name: 'Virtualization' }),
+      frame.getByRole('radio', { name: /^Virtualization/ }),
     ).toBeVisible();
 
     imageSourceDropdown = frame.getByRole('button', {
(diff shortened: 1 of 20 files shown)


regexowl:
The changes look great! Rebasing to get this on top of https://github.com/osbuild/image-builder-frontend/pull/4754, there were some flakes in boot tests that should hopefully not appear again (besides the openstack issue)

regexowl:
Also one minor nitpick, this alert should be hidden for radios, but that could be easily split to a follow up.

image

regexowl:
Ok, so the only weirdness is happening in the template test (ignoring the expected infra issues). But that shouldn't be connected to the target selection in any way, so I'd say we're good? Can rerun the boot tests next week for good measure.

ksiekl:
@regexowl I can fix the alert, at least we will see the boot tests again :)

regexowl:
> @regexowl I can fix the alert, at least we will see the boot tests again :)

Nice! Already re-ran the tests out of curiosity, this time it's `Repeatable build - URL source` that looks sus. But templates are good now, so on average everything seems fine! :nerd_face:

ksiekl:
I believe that was a flake, it passed before and passes now

regexowl:
> I believe that was a flake, it passed before and passes now

Agree, there's some flaking going on, but nothing jumps out as a failure caused by the changes.

regexowl (approved):
Looks great!

PR #4751, opened 2026-08-13
Wizard: handle missing `share_with_accounts` in gcp

ksiekl:
This is a rare edge case - our validation in UI requires a valid email/domain before the user can proceed past the GCP step, but the API does not. That means, if there is an image request sent with no `share_with_account` option through API, it will be an empty list, the condition `if (sharedWith)` will be true, and the UI will crash with Sentry saying that null cannot be split - as in the first item of an empty list cannot be split. This way, the UI checks if there is anything in the list.


regexowl:
I think this is it! :crossed_fingers: Super fast fix, thanks! :heart:

regexowl (approved):
Image
<br />
Image

PR #4754, opened 2026-08-13
playwright: fix race conditions causing flakiness

lucasgarfield:
I think that this might help quite a bit with the flakiness in the Playwright tests.

To simulate a degraded stage environment locally, I created (and full disclosure, everything in this PR was done with Claude) a chaos testing middleware for Playwright. It simulates API responses coming back out of order which causes failures when there are race conditions.

The first commit fixes about 90% of the flakes. The tl;dr is that it was related to validating the uniqueness of blueprint names – this had a three state (true, false, in flight) that was treated as a boolean and some crufty debounce code that resulted in a race condition on advancing the blueprint. 

The other changes all came about from other less common flakes that cropped up during the chaos testing.

I still probably didn’t manage to catch everything but the chaos testing was so useful that I went ahead and committed it so that we can use it again the next time we start to observe flakes.

I’ve watched this run a few times now in the CI on this PR and not only is Playwright passing every time, it’s passing with 0 flakes (apart from the first run where it did flake, but then I used the chaos testing tool to reproduce and fix that flake, too!).

With the flakes hopefully sorted I went ahead and also tightened up the timeouts in Playwright – over time, in response to flakes, we’d added a lot of little increased timeouts here and there and increased the global timeout from 5 seconds to 50 seconds!

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,6 +413,27 @@ PLAYWRIGHT_STATIC_PASSWORD="<your_static_user_password>"
 
 For local development purposes, you can use the same credentials as for `PLAYWRIGHT_USER` and `PLAYWRIGHT_PASSWORD` if you are using your stage account for those.
 
+### Diagnosing a failure that has no visible cause
+
+A blank page, a control that never appears, and a page that renders slowly all
+look identical from the outside: an assertion times out and says only what it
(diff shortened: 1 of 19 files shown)


regexowl (approved):
Nice! 🚀

PR #4761, opened 2026-08-17
Wizard/Packages: Refetch package info in edit mode (HMS-11206)

### LINKED ISSUE #2122: WizardV2: Selected packages have no description
Screenshot 2024-06-05 at 18 16 32

regexowl:
Since the packages are stored in a flat list of names, we need to re-fetch application stream and end date when the blueprint is opened in edit mode.

Fixes #2122

JIRA: [HMS-11206](https://issues.redhat.com/browse/HMS-11206)

[HMS-11206]: https://redhat.atlassian.net/browse/HMS-11206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
+++ b/src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
@@ -66,7 +66,6 @@ import {
   selectRecommendedRepositories,
   selectSnapshotDate,
   selectTemplate,
-  selectWizardMode,
 } from '@/store/slices/wizard';
 import { getEpelUrlForDistribution } from '@/Utilities/epel';
 import { releaseToVersion } from '@/Utilities/releaseToVersion';
@@ -109,7 +108,6 @@ const PackageSearch = ({
   const recommendedRepositories = useAppSelector(selectRecommendedRepositories);
   const template = useAppSelector(selectTemplate);
   const snapshotDate = useAppSelector(selectSnapshotDate);
-  const wizardMode = useAppSelector(selectWizardMode);
 
   const { packages: requiredPackages } = useSecuritySummary();
   const requiredSet = useMemo(
@@ -238,11 +236,6 @@ const PackageSearch = ({
       isLoading: isLoadingRecommendedGroups,
     },
   ] = useSearchPackageGroupMutation();
-
-  const [
-    searchPackageInfo,
-    { data: dataPackageInfo, isSuccess: isSuccessPackageInfo },
-  ] = useSearchRpmMutation();
 
   useEffect(() => {
     if (packageType === 'groups') {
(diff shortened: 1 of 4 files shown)


src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
```diff
@@ -388,44 +387,57 @@ const PackageSearch = ({
...
+          }),
+        );
+      } else if (rpm.package_sources?.[0]?.end_date) {
+        dispatch(
+          addPackage({
+            ...existingPackage,
+            end_date: rpm.package_sources[0].end_date,
+          }),
+        );
+      }
```
ksiekl: I'm looking at this part and getting a bit lost - when I select nginx on your branch, I see its app stream and end date:

Image

But when I open it in edit mode, I cannot see the app stream anymore:

Image

And I mean, it makes sense with the code - the `enabledModule` is false and we do have an end date, so the `else if` gets executed. I just don't think that that is the desired behavior, but maybe I'm getting this wrong?

src/Components/CreateImageWizard/steps/Packages/tests/Packages.test.tsx
```diff
@@ -992,6 +993,76 @@ describe('Packages Component', () => {
...
+          ],
+        },
+      });
+
+      await waitFor(() => {
+        expect(screen.getByText(/Jun 2030/)).toBeInTheDocument();
+      });
+    });
+
+    test('module stream and end_date are re-fetched in edit mode', async () => {
```
ksiekl, line 1025: This test is nginx when the module enablement is already in wizard state, and I think I hit a case where that doesn't happen 🤔

ksiekl (commented):
Nice! 💟 Package searches are tricky, and I got a bit bamboozled by what we want to happen - but if I understood it correctly, I think we have a bug here 🐛

src/Components/CreateImageWizard/steps/Packages/components/PackageSearch.tsx
regexowl: Forgot about this one a bit, sorry about that :sweat_smile:

This indeed looks like a bug. Do you by any chance remember how to reproduce this?

The thing is, when a package connected to a module gets added, both `addPackage` and `addModule` should be dispatched and both should be parsed into blueprint in the end. On edit - the application stream should be read from the existing `enabled_modules` array.

So either we have a bug in the blueprint mapper (tried to look, didn't find anything sus) or this might be an edge case I didn't think of :thinking:

regexowl: Just to add - if the `enabled_modules` array is indeed empty, we can't populate the app_stream column, because we would have to guess, what version of the module the user originally added.

ksiekl: On your branch, I've created a bp that only had the nginx package set, then saved it and opened it for edit. I can't remember whether `enabled_modules` was empty, I can try and reproduce this tomorrow to see.

regexowl: Lol, nevermind. Can reproduce this, will look into it. Thanks!

ksiekl: I retried again and now not even the retirement day shows: 
image

PR #4766, opened 2026-08-18
Refresh `package-lock`

regexowl:
Removed the file and node_modules and re-ran `npm i` to make the dependency definitions in sync.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -278,13 +278,13 @@
       }
     },
     "node_modules/@babel/generator": {
-      "version": "7.29.7",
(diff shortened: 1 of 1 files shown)


regexowl:
https://github.com/osbuild/image-builder-frontend/pull/4762 is failing with

```
npm error code EUSAGE
npm error
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm error
```

I assume this would repeat with every new dependabot update if we don't refresh the package-lock file.

regexowl:
Hmm, no.

PR #4772, opened 2026-08-20
Add Fedora 46 to constants

AdamWill:
Fedora 45 has branched, so Rawhide now maps to Fedora 46. This seems to be needed to prevent Rawhide/F46 updates failing their own test suite:

https://bodhi.fedoraproject.org/updates/FEDORA-2026-e82a621f3e#comment-4739429

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -63,6 +63,7 @@ export const FEDORA_42 = 'fedora-42';
 export const FEDORA_43 = 'fedora-43';
 export const FEDORA_44 = 'fedora-44';
 export const FEDORA_45 = 'fedora-45';
+export const FEDORA_46 = 'fedora-46';
 export const FEDORA_ELN = 'fedora-eln';
 export const X86_64 = 'x86_64';
 export const AARCH64 = 'aarch64';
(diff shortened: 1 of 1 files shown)


kingsleyzissou (dismissed):
Thanks 🙏

regexowl:
Hey @AdamWill, thank you for the PR! The FEDORA_46 const is not yet defined in the code base, so it will need to be added in one or two more places, you can check this Fedora 45 PR: https://github.com/osbuild/image-builder-frontend/pull/4109/changes

kingsleyzissou:
> Hey @AdamWill, thank you for the PR! The FEDORA_46 const is not yet defined in the code base, so it will need to be added in one or two more places, you can check this Fedora 45 PR: https://github.com/osbuild/image-builder-frontend/pull/4109/changes

Whoops I was too eager

AdamWill:
Oh, thanks, missed that.

AdamWill:
Updated. There's also the `RUNNER` list in `.gitlab-ci.yml` but I'm not sure whether you'd want to add this there yet (and whether it'd work).

croissanne (dismissed):

croissanne:
> Updated. There's also the `RUNNER` list in `.gitlab-ci.yml` but I'm not sure whether you'd want to add this there yet (and whether it'd work).

wouldn't work yet, we'd need fedora 46 CI images / runners first. This is fine I think , thank you!

src/store/api/backend/hosted/imageBuilderApi.ts
```diff
@@ -590,6 +590,7 @@ export type Distributions =
   | 'fedora-43'
   | 'fedora-44'
   | 'fedora-45'
+  | 'fedora-46'
```
croissanne: Oh wait, can you drop this one? This part is autogenerated. It wouldn't break anything but this is for the service (which doesn't actually expose fedora to users, we just use it internally to build CI images)

AdamWill: ah OK. will do.

croissanne: now it's perfect, thank youuuuuuuuuuuuuuuuuuuu

croissanne (approved):
perf thanks!

PR #4774, opened 2026-08-20
Wizard: hide the standalone users steps for unsupported targets

lucasgarfield:
In image mode the standalone Users and Groups steps rendered whenever isStandalone was set, ignoring shouldHide. Selecting only the container installer, which supports no customizations, still showed both steps. Guard the two consumer sites with shouldHide.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -190,7 +190,9 @@ const CreateImageWizard = () => {
     aapValidation.disabledNext ||
     snapshotValidation.disabledNext ||
     imagePullValidation.disabledNext ||
-    (restrictions.users.isStandalone && usersHaveErrors);
+    (!restrictions.users.shouldHide &&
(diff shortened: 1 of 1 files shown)


src/Components/CreateImageWizard/CreateImageWizard.tsx
```diff
@@ -528,10 +530,12 @@ const CreateImageWizard = () => {
...
-              restrictions.users.isStandalone && <UsersStep key='users' />,
-              restrictions.users.isStandalone && (
-                <UserGroupsStep key='groups' />
-              ),
+              !restrictions.users.shouldHide &&
+                restrictions.users.isStandalone && <UsersStep key='users' />,
+              !restrictions.users.shouldHide &&
+                restrictions.users.isStandalone && (
+                  <UserGroupsStep key='groups' />
+                ),
```
kingsleyzissou, line 538: Nitpick: but this block of code is getting pretty hard to read now. We should maybe come back to this at some point

kingsleyzissou (approved):

PR #4778, opened 2026-08-21
revert "cockpit/spec: bundle cockpit plugin during rpm build step"

croissanne:
Revert "cockpit/spec: bundle cockpit plugin during rpm build step"
    
    This reverts commit 4e9b5b36af37d45174006cd3bad4b328e510c2d8.
    
    The module http2-client (pulled in by codegen-openapi) has hidden
    characters in its license, this fails rpm inspect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ $(TARFILE): cockpit/download
 	npm ci --include=dev
 	touch -r package.json package-lock.json
 	touch cockpit/public/*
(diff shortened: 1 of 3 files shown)


kingsleyzissou (approved):

croissanne:
Ok so, let's waive it for the current release, since bundling on distro infra with 2 harmless control characters seems like the better choice. And then i'll keep this open, but in draft, and se if it's possible to push the fix through upstream before the next release.

PR #4797, opened 2026-08-27
Split hostname validation into granular checks with a pure validator

kingsleyzissou:
## Summary

Follow-up to the hostname validation work: splits the catch-all regex into granular, per-rule checks and replaces the validation hook with a pure `validateHostname` function, so the wizard surfaces specific, actionable errors instead of one generic message.

## Changes

- Split the hostname schema into granular Zod checks (character set, dot placement, label hyphens) so every violation is reported at once with a targeted message.
- Replace the `useHostnameValidation` hook with a pure `validateHostname` that returns structured `ValidationIssue[]`, dropping unnecessary React overhead and making the logic independently testable.
- Render every validation issue in `HostnameInput` rather than only the first failure.
- Populate each issue's `value` with the offending input.
- Add unit coverage for the granular rule messages (including multiple-error stacking) and the `validateHostname` contract.

### DIFF
--- a/src/Components/CreateImageWizard/CreateImageWizard.tsx
+++ b/src/Components/CreateImageWizard/CreateImageWizard.tsx
@@ -44,11 +44,13 @@ import {
   loadWizardState,
   parseStateFromRequest,
   selectDistribution,
+  selectHostname,
   selectImageSource,
   selectImageSourceType,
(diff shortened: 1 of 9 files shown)


ksiekl (approved):
LGTM, thanks!

PR #4807, opened 2026-08-31
Add kernel argument validation and reusable ValidatedListInput component (HMS-11247)

kingsleyzissou:
## Summary

Adds robust kernel argument validation using a Zod schema and a new reusable `ValidatedListInput` chip-input component, replacing the old `useKernelValidation` hook. Also migrates OpenSCAP-mandated kernel args to `useSecuritySummary` so required/user-added args are handled consistently.

## Changes

- Replace `useKernelValidation`/`isKernelArgumentValid` with a Zod-based `kernelArgSchema` + generic `validateList` helper, giving a single source of truth for the character allowlist and duplicate detection
- Add reusable `ValidatedListInput` component (chip input with add/remove, blueprint-restore validation, ARIA wiring, and required/locked chips) with full test coverage
- Migrate `KernelArguments` to `useSecuritySummary` and the new `ValidatedListInput`, distinguishing OSCAP-required args (non-removable) from user-added args
- Add `uniqueArray` Zod refinement factory for reusable duplicate-item validation across list schemas
- Fix duplicate/attributable error handling and add `aria-invalid` wiring based on review feedback

### DIFF
--- a/src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
+++ b/src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
@@ -1,42 +1,40 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 
 import { FormGroup } from '@patternfly/react-core';
 
-import LabelInput from '@/Components/CreateImageWizard/LabelInput';
-import { useKernelValidation } from '@/Components/CreateImageWizard/utilities/useValidation';
-import { isKernelArgumentValid } from '@/Components/CreateImageWizard/validators';
+import ValidatedListInput from '@/Components/CreateImageWizard/ValidatedListInput';
 import { useSecuritySummary } from '@/store/api/backend';
-import { useAppSelector } from '@/store/hooks';
+import { useAppDispatch, useAppSelector } from '@/store/hooks';
 import {
   addKernelArg,
   removeKernelArg,
   selectKernel,
+  validateKernelArgs,
 } from '@/store/slices/wizard';
+import {
+  type MergedListItem,
+  mergeListItems,
+} from '@/Utilities/mergeListItems';
 
 const KernelArguments = () => {
+  const dispatch = useAppDispatch();
   const kernelAppend = useAppSelector(selectKernel).append;
-
-  const stepValidation = useKernelValidation();
-
   const { kernel: requiredKernel } = useSecuritySummary();
 
-  const requiredByProfile = kernelAppend.filter((arg) =>
-    requiredKernel.append.includes(arg),
+  const items: MergedListItem[] = useMemo(
+    () => mergeListItems(requiredKernel.append, kernelAppend),
+    [kernelAppend, requiredKernel.append],
   );
 
   return (
     <FormGroup isRequired={false} label='Arguments'>
-      <LabelInput
+      <ValidatedListInput
         ariaLabel='Add kernel argument'
         placeholder='Add kernel argument'
-        validator={isKernelArgumentValid}
-        list={kernelAppend.filter((arg) => !requiredByProfile.includes(arg))}
-        requiredList={requiredByProfile}
-        item='Kernel argument'
-        addAction={addKernelArg}
-        removeAction={removeKernelArg}
-        stepValidation={stepValidation}
-        fieldName='kernelAppend'
+        validator={validateKernelArgs}
+        items={items}
+        onAdd={(value) => dispatch(addKernelArg(value))}
+        onRemove={(value) => dispatch(removeKernelArg(value))}
         helperText='Enter additional kernel boot parameters. Examples: nomodeset or console=ttyS0.'
       />
     </FormGroup>
--- a/src/Utilities/mergeListItems.ts
+++ b/src/Utilities/mergeListItems.ts
@@ -1,0 +1,11 @@
+export type MergedListItem = { value: string; required: boolean };
+
+export const mergeListItems = (
+  requiredItems: string[],
+  optionalItems: string[],
+): MergedListItem[] => [
+  ...requiredItems.map((item) => ({ value: item, required: true })),
+  ...optionalItems
+    .filter((item) => !requiredItems.includes(item))
+    .map((item) => ({ value: item, required: false })),
+];
--- a/src/store/slices/wizard/system/tests/validation/kernel.test.ts
+++ b/src/store/slices/wizard/system/tests/validation/kernel.test.ts
@@ -1,0 +1,89 @@
+import { describe, expect, it } from 'vitest';
+
+import { isKernelArgumentValid } from '@/Components/CreateImageWizard/validators';
+
+describe('kernel validation', () => {
+  describe('valid kernel arguments', () => {
+    it('accepts a simple key=value argument', () => {
+      expect(isKernelArgumentValid('quiet')).toBe(true);
+    });
+
+    it('accepts an argument with equals sign', () => {
+      expect(isKernelArgumentValid('root=/dev/sda1')).toBe(true);
+    });
+
+    it('accepts an argument with hyphens', () => {
+      expect(isKernelArgumentValid('no-scroll')).toBe(true);
+    });
+
+    it('accepts an argument with underscores', () => {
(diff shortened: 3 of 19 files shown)


src/Utilities/mergeListItems.ts
```diff
@@ -0,0 +1,11 @@
+export type MergedListItem = { value: string; required: boolean };
+
+export const mergeListItems = (
+  requiredItems: string[],
+  optionalItems: string[],
+): MergedListItem[] => [
+  ...requiredItems.map((item) => ({ value: item, required: true })),
+  ...optionalItems
+    .filter((item) => !requiredItems.includes(item))
+    .map((item) => ({ value: item, required: false })),
```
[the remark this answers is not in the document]
kingsleyzissou: This is existing behaviour for some of these components. It's very unlikely that a user would load a blueprint
from the store that already had a duplicate 🤷‍♂️

src/store/slices/wizard/validators.ts
```diff
@@ -0,0 +1,54 @@
...
+        message: i.message,
+        kind: 'duplicate',
+        value: i.params.value,
+      };
+    }
+
+    return {
+      message: i.message,
+      kind: 'format',
+      value: String(item),
```
lucasgarfield: **[suggestion]** `validateSchema` always sets format-issue `value` to `String(item)`. That is correct for hostname (a string) but wrong for `validateKernel`, which passes `{ name, append }`. Invalid append entries therefore produce `value: "[object Object]"` (confirmed against Zod 4: a failing `append[0]` issue has path `['append', 0]`, which this helper ignores). Callers currently only check `kernelErrors.length`, so Next/save still disable correctly, but the mapped issues are not reusable and will break any later UI that filters or displays `issue.value` the way `TextInputGroup` already does.

**Suggestion:** Resolve format values from the Zod issue path (same approach as `validateList`), or have `validateKernel` call `validateList(kernelSchema.shape.append, kernel.append)` until name has real constraints. Add a `validateKernel` unit test that asserts `value` is the offending argument, not `"[object Object]"`.

src/Components/CreateImageWizard/TextInputGroup.tsx
```diff
@@ -0,0 +1,199 @@
...
+    [attemptedValue, allValues, validator],
+  );
+
+  // Validate items already in the store (e.g. loaded from a blueprint).
+  const storeIssues = useMemo(
+    () => (allValues.length > 0 ? validator(allValues) : []),
+    [allValues, validator],
+  );
+
+  const allErrors = [...attemptErrors, ...storeIssues];
```
lucasgarfield: **[suggestion]** `allErrors` concatenates `attemptErrors` with `storeIssues` without deduping. If the store already has an invalid value and the user tries to add that same value, the format issue appears twice (once from the attempt, once from the store) plus a duplicate issue. The input is also marked `aria-invalid` / `validated="error"` for store-only problems even when the text field itself is empty, which is reasonable, but the duplicated helper-text lines are not.

**Suggestion:** Deduplicate by `kind`+`value`+`message` before render, or omit attempt issues that are already present in `storeIssues`.

src/Components/CreateImageWizard/TextInputGroup.tsx
```diff
@@ -0,0 +1,199 @@
...
+  isCompact = false,
+  chipCollapseThreshold = DEFAULT_CHIP_COLLAPSE_THRESHOLD,
+  hideAddLabel = false,
+  helperText,
+  addButtonAriaLabel = 'Add',
+}: TextInputGroupProps) => {
+  const [inputValue, setInputValue] = useState('');
+  // The value from the last rejected add attempt. Errors are derived from it
+  // against the current items, so removing a conflicting chip clears them.
+  const [attemptedValue, setAttemptedValue] = useState<string | null>(null);
```
lucasgarfield: **[suggestion]** `attemptedValue` is stored instead of the error list specifically so that removing a conflicting chip re-runs the validator and clears the duplicate/format message. That path is not covered: tests never add a duplicate, never add a valid item while the store already has an invalid one (the explicit `addItem` design), and never remove a chip to dismiss an attempt error.

**Suggestion:** Add tests for (1) rejecting a duplicate of an existing chip, (2) clearing that error when the existing chip is removed, and (3) successfully adding a valid value while an invalid store item is still present.

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
```diff
@@ -1,56 +1,40 @@
...
-  const { data: oscapProfileInfo } = useGetOscapCustomizationsQuery(
-    {
-      distribution: release,
-      // @ts-ignore if complianceProfileID is undefined the query is going to get skipped, so it's safe here to ignore the linter here
-      profile: complianceProfileID,
-    },
-    {
-      skip: !complianceProfileID,
-    },
-  );
```
regexowl: This should be fixed by https://github.com/osbuild/image-builder-frontend/pull/4793 and get updated after rebase.

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
```diff
@@ -1,56 +1,40 @@
...
-    oscapProfileInfo?.kernel?.append?.split(' ').includes(arg),
+  const items: MergedListItem[] = useMemo(
+    () => mergeListItems(oscapKernel.append, kernelAppend),
+    [kernelAppend, oscapKernel.append],
   );
 
   return (
     <FormGroup isRequired={false} label='Arguments'>
-      <LabelInput
+      <TextInputGroup
```
regexowl: This will be used side by side with `LabelInput` for fields with new validation and the `LabelInput` component would be deprecated at the end of migration, right?

The only nitpick from my side would be to try to find a different name to make it easier to tell the custom component and PatternFly's `TextInputGroup` apart.

src/store/slices/wizard/system/tests/validation/kernel.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      expect(isValid('a'.repeat(256))).toBe(true);
+    });
+
+    it('returns no issues for an empty list', () => {
+      expect(validateKernelArgs([])).toEqual([]);
+    });
+  });
+
+  describe('invalid kernel arguments', () => {
+    it('rejects an argument with spaces', () => {
```
regexowl, line 71: Lol, I've been fighting against adding items in `LabelInput` on Space keydown since it's valid char for kernel argument. Turns out, we're not even allowing it 🥲 

This is out of scope for this PR, just noticed it. I'll make a note to fix it some time soon 😅

regexowl (commented):
Purrfection, thanks! Happy to approve after rebase. Just one small question and a note to myself 😅

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
kingsleyzissou: Ya I dropped this commit after I saw yours got merged

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
kingsleyzissou: Ya 100%, it was getting too messy to try and use `LabelInput` for both. I'm happy to rename it so that it's a bit clearer.

I'll brainstorm a name and do that tomorrow. Small fix anyways

src/store/slices/wizard/system/tests/validation/kernel.test.ts
kingsleyzissou: Haha fair, I was trying to maintain current behaviour as much as possible. But I did have claude write the tests.

There were/are some cases where I was kind of forced to break behaviour because it didn't make sense otherwise

src/Components/CreateImageWizard/steps/Kernel/components/KernelArguments.tsx
regexowl: Yup, fair. `ValidatedListInput` sounds great, thanks!

regexowl (approved):
Looks great! 🚀

PR #4811, opened 2026-09-02
Migrate Systemd Services step to schema-based validation (HMS-11268)

kingsleyzissou:
# Migrate Systemd Services step to schema-based validation

## Summary

Migrates the Systemd Services wizard step to the shared `ValidatedListInput` component and Zod schema-based validation (building on the kernel-args groundwork). Service-name validation now lives in a single reusable schema instead of being duplicated across the component and a step-validation hook.

## Changes

- Add `serviceSchema`/`servicesSchema` Zod schemas with granular, per-rule error messages, plus `validateEnabledServices`/`validateDisabledServices`/`validateMaskedServices` and `validateServices` wrappers.
- Migrate `ServicesInput` to the shared `ValidatedListInput` component, replacing the manual `LabelInput` wiring and inline validation.
- Source OpenSCAP-required services via `useSecuritySummary` instead of manual query orchestration and inline filtering.
- Remove the legacy `useServicesValidation` hook and gate the Next button through `validateServices`, matching the existing `validateHostname`/`validateKernel` pattern (`isServiceValid` is retained for the Firewall step).
- Add schema and component test coverage for the services step.

### DIFF
--- a/playwright/Customizations/OpenSCAP.spec.ts
+++ b/playwright/Customizations/OpenSCAP.spec.ts
@@ -116,7 +116,7 @@ test('Create a blueprint with OpenSCAP customization', async ({
     await frame.getByPlaceholder('Add masked service').fill('nftables');
     await frame.getByPlaceholder('Add masked service').press('Enter');
     await expect(
-      frame.getByText('Masked service already exists'),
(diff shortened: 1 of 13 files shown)


regexowl (approved):
🚀

PR #4814, opened 2026-09-03
Show users in advanced settings consistently

kingsleyzissou:
## Summary

Move user and group customizations back into Advanced settings for all image-mode builds, removing the old on-prem standalone user model now that users are optional everywhere.

## Changes

- Render Users and Groups from Advanced settings whenever user customizations are supported.
- Remove user validation from Base settings and keep it with the Advanced settings step.
- Stop showing user/group details in the Image overview review card.
- Remove the obsolete `isStandalone` restriction metadata and related tests.

### DIFF
--- a/src/Components/CreateImageWizard/steps/Review/components/ImageOverview/index.tsx
+++ b/src/Components/CreateImageWizard/steps/Review/components/ImageOverview/index.tsx
@@ -18,12 +18,9 @@ import {
 
 import { MiscFormats, PrivateClouds, PublicClouds } from './components';
 
-import { Users } from '../AdvancedSettings/components';
-import { UserGroups } from '../AdvancedSettings/components/UserGroups';
 import { ReviewCardHeader, ReviewGroup, ReviewList } from '../shared';
-import { ReviewCardProps } from '../types';
 
-const ImageOverview = ({ restrictions }: ReviewCardProps) => {
+const ImageOverview = () => {
   const imageName = useAppSelector(selectBlueprintName);
   const description = useAppSelector(selectBlueprintDescription);
   const isOnPremise = useAppSelector(selectIsOnPremise);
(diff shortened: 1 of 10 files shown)


src/Components/CreateImageWizard/steps/Review/components/ImageOverview/index.tsx
```diff
@@ -18,12 +18,10 @@ import {
 
 import { MiscFormats, PrivateClouds, PublicClouds } from './components';
 
-import { Users } from '../AdvancedSettings/components';
-import { UserGroups } from '../AdvancedSettings/components/UserGroups';
 import { ReviewCardHeader, ReviewGroup, ReviewList } from '../shared';
 import { ReviewCardProps } from '../types';
 
-const ImageOverview = ({ restrictions }: ReviewCardProps) => {
+const ImageOverview = (_props: ReviewCardProps) => {
```
lucasgarfield, line 24: This is the first time in the codebase that we pass unused props. Feels kind of nitpicky to point out but on the other hand this is would be a new pattern so I'd at least like to be intentional about it. What do you think?

lucasgarfield (commented):
Looks great! I'm really glad to be done with isStandalone - that variable caused me some headaches fixing a bug a few weeks ago! 😄

Only found one small thing to clarify, otherwise this is a nice cleanup.

src/Components/CreateImageWizard/steps/Review/components/ImageOverview/index.tsx
kingsleyzissou: Ya sorry I missed this. We should have removed it.

kingsleyzissou: The bot was being lazy... it had a cascading effect on some of the tests and other components because we had to remove the props.

I've gone and fixed it manually

lucasgarfield (approved):

PR #4818, opened 2026-09-04
Dep updates

kingsleyzissou:
Consolidate some of the dev dependency updates in one place. This should hopefully also unblock
the larger dependabot PR


regexowl (approved):
🚀

PR #4836, opened 2026-09-10
Move timezone validation into store schemas (HMS-11315)

kingsleyzissou:
## Summary

Move timezone validation into the wizard store’s schema-based validation flow while preserving the existing supported-timezone and NTP validation behavior. Add store-level validation coverage for timezone, locale, and first-boot rules to support the remaining migrations.

## Changes

- Move the timezone list into the system slice and expose it through the store barrel as the source of truth.
- Add timezone, NTP server, duplicate-list, and optional-field schemas and store validators.
- Update timezone controls and aggregate wizard validity to use the schema-based validators.
- Remove the obsolete component-level NTP validator and redundant timezone hook validation.
- Add store-level tests covering timezone, locale, and first-boot validation behavior.


regexowl (approved):
🚀

PR #4837, opened 2026-09-11
Wizard/Locale: migrate validation to schema validators (HMS-11317)

kingsleyzissou:
## Summary

Migrate Locale validation to the shared system Zod schema and validator flow, matching the existing Timezone validation approach. This makes the schema the source of truth for supported languages and keyboards while keeping validation feedback in the wizard UI.

## Changes

- Move the language and keyboard allowlists into `store/wizard/system` constants alongside the timezone list.
- Add Locale schemas, schema-derived types, and reusable validators for languages, keyboards, and Locale objects.
- Remove the duplicated Locale validation from the wizard-level validation hooks and rely on `validateSystemSlice`.
- Add `ValidatedInputHelperText` and use it for validation messages in the Locale/Timezone controls and validated list inputs.
- Update Locale, ImportMode, and system validation tests to cover the schema-based validators.


regexowl (approved):
Nice!

PR #4847, opened 2026-09-15
Move users and groups into a dedicated wizard slice (HMS-11347)

kingsleyzissou:
## Summary

Separate user and group customizations from the system wizard state so system validation can remain focused on system settings. Unsupported network and container installer targets now clear user and group customizations that they cannot consume.

## Changes

- Add a dedicated users slice with state, actions, selectors, parsers, mappers, and tests.
- Update wizard composition, blueprint parsing, request mapping, components, and validation to use the new users slice.
- Remove user/group state and unused aggregate validators from the system slice.
- Clear users and groups when selecting `network-installer` or `bootable-container-iso` targets.
- Expand reducer, mapper, parser, listener, and component test coverage for the new state boundaries.


src/store/slices/wizard/users/parsers.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const parseGroups = ({ groups }: Customizations): UsersSlice['groups'] => {
+  if (!groups || groups.length === 0) {
+    return initialState.groups;
+  }
+
+  return groups.map((group) => ({
+    name: group.name,
+    ...(group.gid && { gid: group.gid }),
```
regexowl: I think this would drop the gid if it's 0, so maybe
```suggestion
    ...(group.gid !== undefined && { gid: group.gid }),
```

regexowl (commented):
Looks great! One minor nitpick

src/store/slices/wizard/listeners.ts
```diff
@@ -117,6 +118,25 @@ export const clearUnsupportedFilesystem: WizardListenerEffect = (
...
+const UNSUPPORTED_USERS_IMAGE_TYPES = new Set([
+  'bootable-container-iso',
+  'network-installer',
+]);
+
+export const clearUnsupportedUsersAndGroups: WizardListenerEffect = (
+  _action,
+  listenerApi,
+) => {
+  const imageType = selectImageTypes(listenerApi.getState())[0];
```
ksiekl: Why are we checking only the first image type here? If it's because of single target migration, I would still go through all of the targets for now, because the migration on cockpit still haven't started...

regexowl: Ah, good point, I was automatically like: "Single target, makes sense..." :no_mouth:

kingsleyzissou: I forgot about cockpit to be honest

kingsleyzissou: Let me re-think this because I'm doing the cockpit migration this coming sprint... the order might still be okay because all of these changes only land in the next release

kingsleyzissou: Okay I think I've added a change that future-proofs this so that it works for single + multi-target

ksiekl: Sure, I mean, it's not such a big deal since you'll be working on it in the next sprint

ksiekl: Nice! Thanks!

ksiekl (approved):

regexowl (approved):
🚀

PR #4851, opened 2026-09-18
chores(deps): fix release numbers

avitova:
After unsuccessful attempt to release v111, our release pipeline is broken, and keeps looking for v111 tarball. However, our tags now do not match the release version. This fixes it so that it looks for the correct tarball.


cockpit/cockpit-image-builder.spec
```diff
@@ -1,5 +1,5 @@
 Name:           cockpit-image-builder
-Version:        111
+Version:        113
```
kingsleyzissou, line 2: Skipping 112?

avitova: Idk what is the correct process here. We did not release tags 111 and 112. So we either stay aligned in: release-tag, or we release version 112 but with the 113 tag. Both possible, do you have some insights into this?

kingsleyzissou: Not sure, I thought the release automation should handle this 🤷 
Maybe we should merge and see what happens

kingsleyzissou (approved):

cockpit/cockpit-image-builder.spec
avitova: https://github.com/osbuild/image-builder-frontend/actions/runs/35355838613/job/105636237838 Same, but the latest release failed.

supakeen: This diff is correct though it'd be worthwhile to figure out why you didn't get a post-release version bump.

